Issue #22, opened 2026-02-25
Security: OpenClaw gateway bound to 0.0.0.0 (publicly accessible)


getmilodev:
Your openclaw.json has `"host": "0.0.0.0"` which binds the gateway to all network interfaces. If this machine has a public IP or is port-forwarded, anyone can access your OpenClaw gateway, including running commands on your machine via the agent.

**Fix:** Change to:
```json
"host": "127.0.0.1"
```

If you need remote access, use a reverse proxy with authentication or an SSH tunnel.

More details: https://getmilo.dev/blog/openclaw-security-hardening-guide

PR #70, opened 2026-03-09
feat(ape): add Agent Policy Engine extension

latentcollapse:
## Overview
APE is a lightweight policy gateway for Dream Server's autonomous agent framework. It intercepts OpenClaw tool calls before execution and evaluates them against a configurable policy.

## Motivation
OpenClaw's `exec` tool can run arbitrary commands on the host machine. Issue #22 showed the gateway was publicly accessible — fixing the binding is necessary but not sufficient. A policy layer gives operators fine-grained control over *what* an agent is allowed to do, not just *who* can reach it.

## What it provides
- **Intent classification** — infers `ExecuteCommand / WriteFile / ReadFile / NetworkFetch / SpawnAgent / Other` from tool name and args
- **Allowlist enforcement** — only approved commands execute; unsafe patterns (`rm -rf`, `curl|sh`, netcat reverse shells) blocked by regex
- **Path guards** — `WriteFile` restricted to `/workspace` and `/tmp/openclaw` by default
- **Rate limiting** — configurable RPM cap across all agent sessions
- **Audit log** — every decision appended to `audit.jsonl` (append-only, never mutated) with tool name, intent, args keys, session ID, agent ID, and timestamp
- **Hot-reload** — edit `policy.yaml`, APE picks it up within ~30s, no restart needed
- **Strict mode** — `APE_STRICT_MODE=true` makes denials return HTTP 403; default is non-blocking (log and advise)

## API
```
POST /verify    — evaluate a tool call, returns {allowed, reason, intent, decision_id}
GET  /audit     — tail the audit log (last N entries)
GET  /policy    — active policy summary
GET  /metrics   — allowed/denied/rate_limited counters
GET  /health    — liveness probe
```

## Usage
```bash
dream enable ape
```

Point OpenClaw's tool middleware at `http://ape:7890/verify`. Runs non-blocking by default so existing workflows are unaffected until you're ready to enforce.

## Footprint
- Python 3.12-slim image
- 256 MB RAM limit, 0.5 CPU
- No GPU required
- Port binds to `127.0.0.1` only

### DIFF
--- a/dream-server/config/ape/policy.yaml
+++ b/dream-server/config/ape/policy.yaml
@@ -22,8 +22,6 @@ intents:
       - head
       - tail
       - wc
-      - python3
-      - node
       - echo
       - pwd
       - env


Lightheartdevs (changes requested):
Good work — clean compose config, pinned deps, localhost binding, no-new-privileges, resource limits, hot-reload. This is solid for a v1 optional extension.

**One required change before merge:**

`path_guard` in `main.py` uses raw `str.startswith()` without path normalization:

```python
if any(path.startswith(p) for p in allowed_paths):
```

A path like `/home/node/.openclaw/workspace/../../../etc/passwd` passes this check because it starts with the allowed prefix before traversal. Fix:

```python
import posixpath
# ... in the path_guard block:
normalized = posixpath.normpath(path)
if any(normalized.startswith(p) for p in allowed_paths):
```

This is a one-line fix. Since APE is specifically a security extension, it should ship without a path traversal bypass.

Everything else (in-memory rate limiting, hardcoded CORS, no auth on /verify) is acceptable for v1. Please push the normpath fix and we'll merge.

latentcollapse:
Thank you for the feedback. I'm experiencing some network issues at the
moment, but the fixes are noted and I'll make those changes as soon as I'm
able. It's Python recreation of my Rust based APE engine.

PR #124, opened 2026-03-10
fix(features): fall back to HOST_RAM_GB for VRAM gate on Apple Silicon

yasinBursali:
## What

Fall back to `HOST_RAM_GB` for VRAM gating on Apple Silicon when `get_gpu_info_apple()` returns None.

## Why

When `HOST_CHIP` env var is empty/missing, `get_gpu_info_apple()` returns `None`, causing `gpu_vram_gb=0`. Every feature with a VRAM requirement is then marked `insufficient_vram` even on machines with 24–128 GB unified memory. Apple Silicon uses unified memory where RAM = VRAM.

## How

- In `calculate_feature_status()`: when `gpu_vram_gb == 0` and `GPU_BACKEND == "apple"`, read `HOST_RAM_GB` as the VRAM proxy.
- Same fallback applied in `api_features()` endpoint-level GPU summary, with `memory_type = "unified"`, so the dashboard GPU card is also consistent.
- Warning logged once per request (in `api_features()`) when fallback fires but `HOST_RAM_GB` is still 0, so operators can diagnose the issue.
- Import block reordered to PEP 8: stdlib → third-party → local; `logger` moved after all imports.
- New `tests/test_features.py` with 4 tests: fallback enables features when RAM ≥ requirement, blocks when RAM < requirement, does not fire on nvidia backend, and endpoint-level fallback returns correct vramGb.

## Testing

- [x] Python syntax check: PASS
- [x] pytest: 4 passed, 0 failed
- [x] Secret scan: CLEAN
- [ ] Manual: verify feature cards show correct VRAM status on macOS M4 with missing HOST_CHIP

## Platform Impact

- macOS (Apple Silicon): targeted fix
- Linux: unaffected — fallback is gated by `GPU_BACKEND == "apple"`
- Windows: not applicable

Closes Light-Heart-Labs/DreamServer#119

### DIFF
--- a/dream-server/extensions/services/dashboard-api/routers/features.py
+++ b/dream-server/extensions/services/dashboard-api/routers/features.py
@@ -1,14 +1,17 @@
 """Feature discovery endpoints."""
 
+import logging
+import os
 from typing import Optional
 
 from fastapi import APIRouter, Depends, HTTPException
 
-import os
 from config import FEATURES, GPU_BACKEND, SERVICES
 from gpu import get_gpu_info, get_gpu_tier
 from models import GPUInfo
 from security import verify_api_key
+
+logger = logging.getLogger(__name__)
 
 router = APIRouter(tags=["features"])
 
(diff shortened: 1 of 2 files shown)


Lightheartdevs (approved):
All four requested changes addressed: env var rename to HOST_RAM_GB, inline comment on zero-usage assumption, endpoint-level test added, and warning log for HOST_RAM_GB=0. Solid. ✅

PR #138, opened 2026-03-10
feat: add Workflows page, CI pipelines, and test coverage

shyboy1044:
## Summary

- **Workflows page** — new React page wired to the existing `/api/workflows` backend, with filter tabs, n8n status banner, dependency checks, and enable/disable actions. Follows the existing dashboard aesthetic and component patterns.
- **dashboard.yml CI** — frontend lint + build, dashboard-api pytest, Dockerfile linting (hadolint)
- **test-linux.yml CI** — installer contract validation (manifests, JSON configs, compose syntax), installer dry-run (bash -n), and shell unit tests
- **16 new tests** — workflow router (catalog loading, auth, path traversal, edge cases) and agents router (auth enforcement, metrics, HTML fragment)

All 123 tests pass locally (107 existing + 16 new). The one pre-existing failure (`test_preflight_required_ports_no_auth`) is unchanged.

## Test plan

- [ ] `npm run build` in `dashboard/` produces valid dist
- [ ] `pytest tests/ -v` in `dashboard-api/` passes (123 tests)
- [ ] CI workflows trigger on PR and pass
- [ ] Workflows page renders when n8n is running
- [ ] Workflows page degrades gracefully when n8n is down

### DIFF
--- a/dream-server/.github/workflows/dashboard.yml
+++ b/dream-server/.github/workflows/dashboard.yml
@@ -1,0 +1,82 @@
+name: Dashboard CI
+
+on:
+  pull_request:
+    paths:
+      - 'dream-server/extensions/services/dashboard/**'
+      - 'dream-server/extensions/services/dashboard-api/**'
+  push:
+    branches: [main]
+    paths:
+      - 'dream-server/extensions/services/dashboard/**'
+      - 'dream-server/extensions/services/dashboard-api/**'
+
+jobs:
+  dashboard-frontend:
+    name: Frontend Lint & Build
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: dream-server/extensions/services/dashboard
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: actions/setup-node@v4
+        with:
+          node-version: '20'
+          cache: 'npm'
+          cache-dependency-path: dream-server/extensions/services/dashboard/package-lock.json
+
+      - name: Install dependencies
+        run: npm ci || npm install
+
+      - name: Lint
+        run: npm run lint --if-present
+
+      - name: Build
+        run: npm run build
+
+      - name: Check build output
+        run: |
+          test -d dist || { echo "Build did not produce dist/"; exit 1; }
+          test -f dist/index.html || { echo "Missing index.html in dist/"; exit 1; }
+
+  dashboard-api:
+    name: API Lint & Test
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: dream-server/extensions/services/dashboard-api
+    steps:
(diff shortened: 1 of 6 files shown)


Lightheartdevs (changes requested):
## Request Changes

Good component structure and useful CI additions, but the frontend API calls are missing auth and won't work against the production backend.

### 🔴 Missing auth headers on all API calls

Every `fetch('/api/workflows')` in `Workflows.jsx` omits `Authorization` headers:

```javascript
// Line ~57
const resp = await fetch('/api/workflows')
// Line ~67
const resp = await fetch(url, { method })
```

The backend enforces auth via `verify_api_key` — all calls will return 401. The enable/disable actions (`enableWorkflow`, `disableWorkflow`) have the same problem. This page is non-functional as written.

Check how the existing dashboard pages handle auth (nginx injects Bearer tokens server-side for proxied `/api/` requests) and ensure the Workflows page follows the same pattern.

### 🟡 CI antipattern: `npm ci || npm install`

```yaml
- name: Install dependencies
  run: npm ci || npm install
```

`npm ci` requires a lockfile and exits non-zero when it's absent or inconsistent. Falling back to `npm install` silently masks lockfile drift and defeats the purpose of `npm ci` in CI. This should be `npm ci` only. If there's no lockfile, commit one.

### 🟡 Refresh UX — loading state flicker

```jsx
onClick={() => { setLoading(true); fetchWorkflows() }}
```

When `loading === true`, the component returns the full skeleton — filter tabs and status bar disappear during every refresh. Use a separate `refreshing` boolean or show an inline spinner on the button instead.

### 🟡 `disableWorkflow` naming

The function calls `DELETE /api/workflows/${id}`, the UI labels it "Remove," but the function is named `disableWorkflow`. Rename to `removeWorkflow` for consistency with the HTTP method and UI label.

### 🟡 `docker-compose-syntax` CI job swallows errors

```yaml
docker compose -f "$f" config --quiet 2>&1 || {
    echo "WARN: $f failed standalone validation (may need .env)"
}
```

The `||` branch always exits 0. A corrupt compose file with YAML syntax errors will not fail CI. Consider separating "needs .env for variable substitution" (acceptable) from "invalid YAML syntax" (should fail).

### 🟡 `installer-dry-run` glob edge case

If `dream-server/installers/phases/*.sh` matches zero files, bash passes the literal glob string to `bash -n`, producing a file-not-found error. Use `shopt -s nullglob` or check for file existence.

### What's good

- Component structure follows the existing dashboard aesthetic cleanly
- Path-traversal test (`../../etc/passwd` → expects 400/404/422) is a good regression guard
- CI pipeline additions (hadolint with `failure-threshold: warning`, shellcheck, contract validation) are valuable
- The `core.js` plugin registration with `order` fields is a clean extension point

PR #164, opened 2026-03-11
harden(qdrant): add API key auth, resource limits, and HTTP healthcheck

yasinBursali:
## What

Hardens the Qdrant vector DB service with API key authentication, resource limits, and a proper HTTP healthcheck.

## Why

Qdrant was running with no authentication (any local process could read/write/delete all vector embeddings), no resource limits (unbounded RAM during RAG indexing), and a raw TCP healthcheck that only verified the port was open.

## How

- **API key auth**: Auto-generate `QDRANT_API_KEY` (32-byte hex) at install time on Linux and macOS. Pass to container via `QDRANT__SERVICE__API_KEY`. Empty-string default preserves backward compatibility for existing installs. Registered in `.env.schema.json` with `secret: true` to satisfy Linux `validate-env.sh`.
- **Resource limits**: Add `deploy.resources.limits memory: 1G` — a safe default for local RAG workloads.
- **Healthcheck**: Replace `bash /dev/tcp` TCP probe with `curl -sf http://localhost:6333/healthz`, consistent with the project's established healthcheck pattern.

## Testing

- [x] `docker compose config --quiet --no-interpolate` validates successfully
- [x] No secrets in diff
- [x] QDRANT_API_KEY registered in `.env.schema.json` — `validate-env.sh` will not fail

## Review

- Critique Guardian verdict: ⚠️ APPROVED (after companion fixes: schema registration + healthcheck simplification)

## Platform Impact

- [x] macOS (Apple Silicon): QDRANT_API_KEY auto-generated in env-generator.sh
- [x] Linux: QDRANT_API_KEY auto-generated in 06-directories.sh
- [ ] Windows: out of scope for this PR

## Known Considerations

Open WebUI and n8n have optional user-configurable Qdrant integrations. Users who have manually configured those integrations will need to add their `QDRANT_API_KEY` to those services' configuration. The installer does not manage those integrations.

Closes Light-Heart-Labs/DreamServer#162

### DIFF
--- a/dream-server/extensions/services/qdrant/compose.yaml
+++ b/dream-server/extensions/services/qdrant/compose.yaml
@@ -13,7 +13,7 @@ services:
       - "127.0.0.1:${QDRANT_PORT:-6333}:6333"
       - "127.0.0.1:${QDRANT_GRPC_PORT:-6334}:6334"
     healthcheck:
-      test: ["CMD", "curl", "-sf", "http://localhost:6333/healthz"]
+      test: ["CMD-SHELL", "bash -c 'exec 3<>/dev/tcp/localhost/6333 && echo -e \"GET /healthz HTTP/1.0\\r\\n\\r\\n\" >&3 && grep -q 200 <&3'"]
       interval: 30s
       timeout: 10s
       retries: 3


Lightheartdevs (changes requested):
The two original review items are fixed (Windows installer + downstream consumer answer) — nice work.

However, the healthcheck change introduces a new issue: the official `qdrant/qdrant` Docker image **does not include `curl`** (the Qdrant team explicitly rejected adding it — see qdrant/qdrant#3491). This means:

```yaml
test: ["CMD", "curl", "-sf", "http://localhost:6333/healthz"]
```

…will always fail (`curl: not found`), marking the container as `unhealthy` after retries.

**Fix options (pick one):**
1. Revert to the original TCP check: `["CMD-SHELL", "bash -c 'echo > /dev/tcp/localhost/6333'"]`
2. Use wget if available: `["CMD-SHELL", "wget -qO- http://localhost:6333/healthz || exit 1"]`
3. Pure bash HTTP check: `["CMD-SHELL", "bash -c 'exec 3<>/dev/tcp/localhost/6333 && echo -e \"GET /healthz HTTP/1.0\r\n\r\n\" >&3 && grep -q 200 <&3'"]`

Option 1 is simplest and known to work. Everything else in this PR is good to go.

PR #170, opened 2026-03-11
Feat/preset diff command

bugman-007:
**Feature**: dream preset diff <preset1> <preset2>

## Compares two presets and shows:
* Environment variable changes (with sensitive value masking)
* Service state changes (enabled/disabled)
* Metadata differences
* Color-coded output (red=removed, green=added, yellow=changed)

## Changes:
* dream-cli: +139 lines (diff command implementation)
* tests/test-preset-diff.sh: +143 lines (6 comprehensive tests, all passing)

### DIFF
--- a/dream-server/docker-compose.base.yml
+++ b/dream-server/docker-compose.base.yml
@@ -1,0 +1,201 @@
+# Dream Server — Core Service Definitions
+# Extension services live in extensions/services/*/compose.yaml
+# GPU overlays layered on top:
+#   docker compose -f docker-compose.base.yml -f docker-compose.amd.yml up -d
+#   docker compose -f docker-compose.base.yml -f docker-compose.nvidia.yml up -d
+
+name: dream-server
+
+services:
+  # ============================================
+  # LLM Inference — llama-server (GPU-agnostic stub)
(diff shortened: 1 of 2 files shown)


Lightheartdevs (approved):
## Approve

Clean implementation of `dream preset diff <a> <b>`. Useful operator tool for comparing configurations before/after changes.

**Good details:**
- Compares metadata, service state (enabled/disabled), and environment variables
- Correctly masks sensitive values (`PASSWORD`, `SECRET`, `KEY`, `TOKEN`, `API`) — compares pre-masking, displays post-masking
- Uses associative arrays for proper set-diff logic (additions, removals, changes)
- Color-coded output: red for removals, green for additions, yellow for changes
- Updated help text with full subcommand listing

**Tests:** `test-preset-diff.sh` covers argument validation, nonexistent presets, identical presets, env changes, service state changes, and secret masking. Thorough.

Non-blocking nit: the `declare -A ext1 ext2` inside the `case` block works in bash 4+ but associative arrays aren't available in bash 3.x (macOS ships bash 3.2 by default). Since `dream-cli` already requires bash 4+ features elsewhere, this is fine — just noting it.

PR #181, opened 2026-03-12
Add extension manifest validation and versioned compatibility for Dream Server v2

fullstackdev0110:
Summary
- New script scripts/validate-manifests.sh
Validates all extensions/services/*/manifest.* files against the v1 schema and checks optional Dream Server version compatibility (compatibility.dream_min / dream_max).

- dream config validate extended
After existing .env validation, it runs scripts/validate-manifests.sh and reports success or issues. Help text updated so dream config validate is the documented way to check both .env and extension manifests.

- Installer phase 13
After preflight, the installer runs scripts/validate-manifests.sh once and prints a short “VALIDATING EXTENSIONS” block (non-blocking).

- Schema update
extensions/schema/service-manifest.v1.json now defines an optional root-level compatibility object with dream_min and dream_max (strings, semver pattern). Invalid values (e.g. compatibility: "garbage") are rejected by the schema.

- Version metadata on four extensions
compatibility.dream_min: "2.0.0" added to: llama-server, open-webui, n8n, qdrant.

- Docs
docs/EXTENSIONS.md documents the compatibility block and how to use it.

Notes
- All changes are in dream-server (Bash, JSON schema, YAML manifests, docs). No Python/JS app code changed.
- Validator uses the service registry and safe env loading; when python3 + PyYAML + jsonschema are available it runs full schema validation; otherwise it does compatibility-only checks and warns.

### DIFF
--- a/dream-server/extensions/schema/service-manifest.v1.json
+++ b/dream-server/extensions/schema/service-manifest.v1.json
@@ -8,6 +8,23 @@
     "schema_version": {
       "const": "dream.services.v1"
     },
+    "compatibility": {
+      "type": "object",
+      "description": "Dream Server version compatibility; used by validate-manifests.sh for version checks",
+      "properties": {
+        "dream_min": {
+          "type": "string",
+          "pattern": "^\\d+\\.\\d+\\.\\d+$",
+          "description": "Minimum Dream Server version this extension supports (e.g. 2.0.0)"
+        },
+        "dream_max": {
+          "type": "string",
+          "pattern": "^\\d+\\.\\d+\\.\\d+$",
+          "description": "Maximum Dream Server version this extension was tested against (optional)"
+        }
+      },
+      "additionalProperties": false
+    },
     "service": {
       "type": "object",
       "required": ["id", "name", "port", "health"],
--- a/dream-server/scripts/validate-manifests.sh
+++ b/dream-server/scripts/validate-manifests.sh
@@ -77,6 +77,7 @@ else
   warn "python3 not found — skipping schema validation (compatibility checks only)"
 fi
 
+set +e
 python3 - "$ROOT_DIR" "$EXT_DIR_REL" "$SCHEMA_PATH_REL" "$CORE_VERSION" "$PYTHON_OK" <<'PY'
 import json
 import sys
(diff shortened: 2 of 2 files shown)


Lightheartdevs (changes requested):
## Review: ❌ REQUEST CHANGES

This is a reasonable first step toward extension compatibility checking, and the non-blocking advisory approach is the right design. But the validation layer has gaps that need fixing before it becomes the foundation for a community extension ecosystem.

### Issue 1: Schema not updated for the field being validated (must fix)
The PR adds \`compatibility: dream_min: "2.0.0"\` to 4 manifests and builds a 338-line validator around it. But \`extensions/schema/service-manifest.v1.json\` was NOT modified. It works today only because the schema has \`additionalProperties: true\`. The schema cannot validate the *structure* of the compatibility block — \`compatibility: "garbage"\` would pass schema validation but crash the version-comparison code.

**Fix:** Add \`compatibility\` as an object in the schema with \`dream_min\` (string, optional) and \`dream_max\` (string, optional), with a semver pattern like \`^\d+\.\d+\.\d+$\`.

### Issue 2: Dead code after Python heredoc (must fix)
The \`if [[ $? -eq 0 ]]\` check after the Python block is unreachable. With \`set -euo pipefail\`, a non-zero Python exit terminates the script before reaching the \`if\`. The \`fail\` branch (friendly error message) can never execute — users get a raw shell error instead.

**Fix:** Either wrap the Python block in \`set +e\` / \`set -e\`, or restructure the exit handling.

### Issue 3: Duplicated Python logic (should fix)
The compatibility-checking logic (~80 lines) is copy-pasted between the \`python_ok\` and fallback branches. Refactor into shared functions.

### Issue 4: PR description doesn't match the diff
Description mentions \`dream status-json\`, \`dream mode\`, \`dream model current\` — none of which are in the diff. Appears to be a stale description.

### What works well
- Non-blocking at install time (warnings only) — correct design
- \`dream config validate\` for on-demand checking
- Graceful fallback when jsonschema module is missing
- Only 4 of 17 extensions updated — opt-in approach is fine
- EXTENSIONS.md documentation of the new field is clear

### Minor
- \`parse_version\` silently treats \`"2.0.0-beta"\` as \`(2, 0, 0)\` — pre-release semantics lost
- Inconsistent \`$INSTALL_DIR\` vs \`$SCRIPT_DIR\` between dream-cli and 13-summary.sh
- No test coverage for the validation script itself

### CI note
All failures pre-existing. Shell lint passes.

PR #199, opened 2026-03-15
bugfix: llama: unbound variable in a shell script

y-coffee-dev:
In the 4th phase of the installer, source service-registry.sh before using SERVICE_PORTS, this fixes a bug where this error appears on installs
```
line 222: llama: unbound variable in a shell script
```

### DIFF
--- a/dream-server/installers/phases/04-requirements.sh
+++ b/dream-server/installers/phases/04-requirements.sh
@@ -19,6 +19,7 @@
 chapter "REQUIREMENTS CHECK"
 
 [[ -f "${SCRIPT_DIR:-}/lib/safe-env.sh" ]] && . "${SCRIPT_DIR}/lib/safe-env.sh"
+[[ -f "$SCRIPT_DIR/lib/service-registry.sh" ]] && . "$SCRIPT_DIR/lib/service-registry.sh"
 
 REQUIREMENTS_MET=true
 TIER_RANK="$(tier_rank "$TIER")"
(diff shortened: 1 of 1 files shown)


Lightheartdevs (changes requested):
Thanks for catching this — the unbound variable under `set -u` is a real bug. However, the fix introduces a regression:

### 1. `declare -A PORTS_TO_CHECK` breaks port checking

`PORTS_TO_CHECK` is a plain string iterated with `for port in $PORTS_TO_CHECK` (line ~227). Declaring it as an associative array (`declare -A`) changes the semantics — the string assignment sets key `[0]` to the entire value, and the `for` loop gets a single element instead of individual ports. **This silently disables all port conflict detection.**

### 2. Quoting is inconsistent

Only `llama-server` and `open-webui` keys are quoted, but lines 223-225 have the same unquoted pattern for `whisper`, `tts`, `n8n`, and `qdrant`. If quoting is the fix, all 6 need it.

### 3. Root cause is likely elsewhere

The `:-` default syntax should protect against missing keys under `set -u`. The same unquoted pattern (`${SERVICE_PORTS[llama-server]:-8080}`) is used in phases 12, 13, and multiple scripts without issue. The real question is: **why is `SERVICE_PORTS` not declared/loaded when phase 04 runs?**

### Suggested fix

Instead of touching `PORTS_TO_CHECK`, guard against `SERVICE_PORTS` not being declared:

```bash
# Guard: if SERVICE_PORTS wasn't loaded, declare it empty
if ! declare -p SERVICE_PORTS &>/dev/null; then
    declare -A SERVICE_PORTS
fi
```

This protects against the unbound variable without changing the port-checking logic. Please also investigate the sourcing order — `service-registry.sh` should be loaded before phase 04.

PR #212, opened 2026-03-15
test: integration compose checks accept webui naming

boffin-dmytro:
## Summary

This PR makes the integration test suite resilient to the WebUI service naming change (`webui` → `open-webui`).

Instead of hardcoding the expected WebUI service name based on the compose overlay, the test now validates the intent of the stack contract: there must be an LLM service and a chat UI service.

## What changed

- Updated `dream-server/tests/integration-test.sh` core service checks:
  - Still requires `llama-server`
  - Accepts **either** `open-webui` **or** `webui` as the Web UI service
  - If both are present, it reports that as a pass (not a failure)

## Why

The service name has changed over time and can vary depending on compose overlays and extensions. Hardcoded expectations cause false-negative CI failures for contributors even when the stack is valid.

## Verification

- `bash -n dream-server/tests/integration-test.sh`
- `bash dream-server/tests/integration-test.sh` (suite still has known pre-existing failures unrelated to this change)

### DIFF
--- a/dream-server/tests/integration-test.sh
+++ b/dream-server/tests/integration-test.sh
@@ -125,21 +125,29 @@ else
             fi
         fi
 
-        # Verify core services are defined
+        # Verify core services are defined (behaviorally)
+        # The WebUI service name has changed over time (webui -> open-webui). We accept either.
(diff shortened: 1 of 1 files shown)


Lightheartdevs (approved):
Good fix. The integration test was hardcoding `webui` vs `open-webui` service name, causing fragile CI failures. Now accepts either name with a regex match. Clean and well-scoped.

PR #214, opened 2026-03-15
feat(ci): strengthen installer simulation summary validation

championVisionAI:


Lightheartdevs (changes requested):
The direction is right — the existing validator is brittle and shallow. But going from 60 to 357 lines (6x) without adding tests creates a new form of fragility.

1. **Needs tests proportional to complexity**: A 350-line validation framework with custom type-checking helpers (`_require_type`, `_require_key`, `_as_mapping`) needs its own test suite. Without tests, bugs in the validator will cause false CI failures.

2. **Consider jsonschema instead**: The hand-rolled type validators duplicate what `jsonschema` does. A JSON Schema definition + `jsonschema.validate()` would be ~50 lines and self-documenting.

3. **Python version**: Uses `from __future__ import annotations` for recursive type aliases — confirm this works on Python 3.8+ (the lowest version some target systems might have).

Add tests for the validation functions and this becomes mergeable.

PR #215, opened 2026-03-15
feat(healthcheck): add retries, status/body matching, and JSON output

championVisionAI:


Lightheartdevs (changes requested):
The retry logic and HEAD-to-GET fallback are genuinely useful for container healthchecks. Backward compatible, well-scoped to one file.

Two things needed:

1. **Unit tests**: A 400-line script with retry logic, regex matching, status code parsing, and TCP mode needs at least basic tests for the parsing functions (`_parse_expected_status`, `_parse_target`, `_parse_host_port`) and the retry behavior.

2. **Document the 1 MiB body read**: For a healthcheck running every 30s in a Docker HEALTHCHECK, reading up to 1 MiB per check could add up. A comment noting why the cap exists and that it's intentional would help.

Add tests and this is ready.

PR #233, opened 2026-03-16
feat(backup/restore): disk space preflight

bugman-007:
## Summary
Adds a lightweight disk space preflight to backup and restore flows to prevent failures mid-operation due to insufficient free space.

### Backup (`dream-backup.sh`)
- Estimates required space for the requested backup type (best-effort).
- Checks available disk space on the filesystem containing `BACKUP_ROOT`.
- Fails fast with a clear message when space is insufficient and suggests using `--output` to another disk.

### Restore (`dream-restore.sh`)
- Estimates restore size from:
  - extracted backup directory (via `du -sb`), or
  - compressed `.tar.gz` contents (sum of file sizes from `tar -tvzf`).
- Checks available disk space on the filesystem containing `DREAM_DIR`.
- Fails fast with a clear message when space is insufficient.

## Tests
- Adds `tests/test-disk-space-preflight.sh` (stubs `df` to simulate low disk space and verifies that backup/restore are blocked).

## Notes
- Best-effort estimation: designed to be simple and safe, not perfectly exact.
- No changes to service registry, presets, or restore semantics beyond preflight.

### DIFF
--- a/dream-server/dream-backup.sh
+++ b/dream-server/dream-backup.sh
@@ -68,7 +68,7 @@ estimate_backup_bytes() {
         for p in "${user_data_paths[@]}"; do
             if [[ -d "$DREAM_DIR/$p" ]]; then
                 local b
-                b=$(du -sb "$DREAM_DIR/$p" 2>/dev/null | awk '{print $1}')
+                b=$(du -sk "$DREAM_DIR/$p" 2>/dev/null | awk '{print $1 * 1024}')
                 total=$(( total + ${b:-0} ))
             fi
         done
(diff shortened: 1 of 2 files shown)


bugman-007:
##  Fixed: macOS/BSD Compatibility

I've addressed the reviewer's concern about `du -sb` being GNU-specific. The fix replaces all instances with `du -sk` and converts kilobytes to bytes for consistent byte-level calculations.

**Changes made:**
- `dream-backup.sh`: Updated all `du -sb` calls in `estimate_backup_bytes()` to use `du -sk` with conversion
- `dream-restore.sh`: Updated `estimate_restore_bytes_dir()` to use `du -sk` with conversion

This ensures the disk space preflight works correctly on macOS, BSD, and Linux systems. The logic remains the same—just using a POSIX-compatible flag that's available everywhere.

Ready for review!

PR #242, opened 2026-03-16
fix: missing variable initialization and enable LAN access

onyxhat:
- remove `127.0.0.1` port binding as it disables LAN access without including a reverse proxy in the solution
- add proper handling for missing `jq` as it fails the install process (is **not** optional)
- add dot-sourcing of `service-registry` in `04-requirements` as variables aren't initialized and fail the installation on Ubuntu-based distros (ex. Linux Mint 24.04)


Lightheartdevs (changes requested):
The variable initialization fixes across installer phases are fine.

**Security blocker:** Removing `127.0.0.1:` from ALL compose port mappings exposes every service (llama-server, dashboard-api, qdrant, token-spy, etc.) to the entire LAN by default. This is a security regression — these services should only be accessible from localhost unless the user explicitly opts in.

**Recommendation:** Split this PR:
1. Variable initialization fixes → separate PR, ready to merge
2. LAN access → make it opt-in via a `BIND_ADDRESS` env var that defaults to `127.0.0.1`. Users who want LAN access can set `BIND_ADDRESS=0.0.0.0`.

PR #300, opened 2026-03-17
feat: add progress indicators to backup/restore operations

boffin-dmytro:
## Summary
Adds progress indicators to backup and restore operations so users can see transfer progress instead of wondering if the process is hung during large operations.

## Changes
- Add `rsync_with_progress()` function with `--info=progress2` support
- Replace silent rsync calls with progress-enabled versions in:
  - `backup_user_data()`
  - `backup_config()`
  - `backup_cache()`
  - `restore_user_data()`
- Fallback to `--progress` for older rsync versions
- Fallback to basic rsync if progress not supported

## Testing
- Bash syntax validated
- Compatible with rsync 3.0+ (progress2) and older versions (progress)
- No behavioral changes, only adds visibility

## Impact
- Significantly improves UX for large backup/restore operations (multi-GB data)
- Users can see real-time progress instead of staring at a blank screen
- Especially helpful for model backups (10GB+) and full system restores

Total LOC: ~41 lines (18 new function + 23 call site updates)

### DIFF
--- a/dream-server/dream-backup.sh
+++ b/dream-server/dream-backup.sh
@@ -28,6 +28,9 @@ log_success() { echo -e "${GREEN}[SUCCESS]${NC} $*"; }
 log_warn() { echo -e "${YELLOW}[WARN]${NC} $*"; }
 log_error() { echo -e "${RED}[ERROR]${NC} $*"; }
 
+# Source shared rsync utilities
+. "$DREAM_DIR/lib/rsync.sh"
+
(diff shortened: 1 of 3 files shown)


Lightheartdevs (changes requested):
rsync_with_progress is copy-pasted identically into both dream-backup.sh and dream-restore.sh. Extract it to a shared lib (e.g. lib/rsync.sh) to avoid drift.

PR #304, opened 2026-03-17
feat: add health checks to APE and token-spy Dockerfiles

reo0603:
## Summary
- Adds HEALTHCHECK directives to APE and token-spy services
- Improves container reliability and orchestration support

## Changes
- **APE**: Added health check on port 7890 using /health endpoint
- **token-spy**: Added health check on port 8080 (respects UVICORN_PORT env) using /health endpoint

Both services already expose /health endpoints (defined in manifest.yaml), this change adds Docker-native health monitoring.

## Benefits
- Docker/Compose can detect unhealthy containers and restart them automatically
- Better integration with orchestration tools (Kubernetes, Swarm, etc.)
- Improved observability and debugging capabilities
- Consistent with other services (dashboard-api, privacy-shield, comfyui)

## Test Plan
- Verified both services already have /health endpoints in their manifests
- Health check uses Python urllib (already available in base images)
- Follows same pattern as existing health checks in the codebase

PR #309, opened 2026-03-17
fix: add USER directive to token-spy Dockerfile for security hardening

reo0603:
## Summary
- Adds non-root user execution to token-spy service
- Addresses security gap where token-spy was running as root

## Changes
- Creates 'dreamer' user (UID 1000) before switching context
- Sets proper ownership of /app directory
- Runs uvicorn as non-root user

## Context
Token-spy was the only Python service running as root. Other services (dashboard-api, privacy-shield, ape) already run as non-root users. This PR brings token-spy in line with the rest of the stack.

## Benefits
- Reduces attack surface by limiting container privileges
- Follows principle of least privilege
- Consistent with other services in the stack
- Aligns with container security best practices

## Test Plan
- Verified user creation and ownership changes
- Confirmed uvicorn can run as non-root on port 8080 (non-privileged)
- Follows same pattern as dashboard-api and other Python services

### DIFF
--- a/dream-server/extensions/services/token-spy/Dockerfile
+++ b/dream-server/extensions/services/token-spy/Dockerfile
@@ -9,6 +9,11 @@ COPY . .
 
 RUN mkdir -p /app/data
 
+# Create non-root user
+RUN useradd -m -u 1000 dreamer && chown -R dreamer:dreamer /app
+
+USER dreamer
+
 EXPOSE 8080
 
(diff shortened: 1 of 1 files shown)


Lightheartdevs (changes requested):
Conflicts with PR #304 which also modifies the token-spy Dockerfile. Please coordinate — one should incorporate the other. Also useradd -m creates an unnecessary home directory; use --system --no-create-home to match the pattern in the APE Dockerfile.

PR #317, opened 2026-03-17
unify port contract across installer, schema, compose, manifests and …

eva57gr:
…dashboard


Lightheartdevs (changes requested):
Changing OLLAMA_PORT default from 11434 to 8080 is a breaking change for every existing install. Any external tool, bookmark, or script that uses 11434 will break silently. This needs a migration note in the docs, a migrate-config.sh rule, and a version bump. Please add a migration path.

PR #361, opened 2026-03-17
fix: agent_monitor logger, workflow tests, update endpoint, and Ruff so API and lint pass

fullstackdev0110:
- agent_monitor.py: add logger = logging.getLogger(__name__) (fixes NameError in test_agent_monitor).
- Workflow tests: align with API — use GET /api/workflows for categories and n8n info, DELETE /api/workflows/{id} for disable auth check.
- Updates router: validate action before script path check so unknown action returns 400 (fixes test_trigger_update_unknown_action).
- Ruff: remove unused imports in dashboard-api tests (test_updates, test_workflows), scripts/healthcheck.py, scripts/validate-sim-summary.py, tests/contracts/test-port-contracts.py.

### DIFF
--- a/dream-server/extensions/services/dashboard-api/agent_monitor.py
+++ b/dream-server/extensions/services/dashboard-api/agent_monitor.py
@@ -5,11 +5,14 @@ Collects real-time metrics on agent swarms, sessions, and throughput.
 
 import asyncio
 import json
+import logging
+import os
 from datetime import datetime, timedelta
 from typing import List
-import os
 
 import aiohttp
+
+logger = logging.getLogger(__name__)
 
 TOKEN_SPY_URL = os.environ.get("TOKEN_SPY_URL", "http://token-spy:8080")
 TOKEN_SPY_API_KEY = os.environ.get("TOKEN_SPY_API_KEY", "")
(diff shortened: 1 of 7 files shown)


Lightheartdevs (changes requested):
## Review: REQUEST CHANGES

The logger fix and workflow test realignment are correct. One design concern and one minor issue.

### Issue: Fragile `else` fallthrough in `routers/updates.py`

The PR moves the unknown-action guard to the top of `trigger_update`, then changes the final `elif action.action == "update":` to a bare `else` with a comment. This is fragile — if someone later adds a new valid action to the top guard but forgets to add a branch, they silently fall into the update path.

**Fix (preferred):** Change `UpdateAction.action` to `action: Literal["check", "backup", "update"]` in `models.py`. Pydantic returns 422 automatically for invalid values, eliminating the manual guard entirely.

**Fix (alternative):** Keep explicit `elif action.action == "update":` with a final `else: raise HTTPException(400)` as a defensive fallback.

### Minor: Duplicate workflow tests

`test_n8n_status_requires_auth` and `test_workflow_categories_requires_auth` now test the same endpoint (`GET /api/workflows`) with the same assertion (401). Consider consolidating or adding distinct assertions (e.g., checking response body keys).

### What's good:
- `agent_monitor.py` logger fix is correct — was a `NameError` at runtime
- Workflow tests properly realigned to actual API surface (old tests hit endpoints that don't exist)
- Ruff cleanup is straightforward

🤖 Reviewed with [Claude Code](https://claude.com/claude-code)

PR #411, opened 2026-03-19
feat(update): add --dry-run flag showing version, images, and .env diff without applying

buddy0323:
## Summary

- `dream update --dry-run` (alias `-n`) previews what an update would change without pulling images or restarting any containers
- New `GET /api/update/dry-run` endpoint in `routers/updates.py` returns version comparison, configured image tags, and update-relevant `.env` keys as structured JSON
- CLI output is divided into four sections: **Version** (installed vs. latest, changelog URL), **Image tags** (what `docker compose pull` would fetch), **Model configuration** (TIER, LLM_MODEL, GGUF_FILE, CTX_SIZE, GPU_BACKEND, N_GPU_LAYERS), and **.env keys** the update path reads/writes
- Dashboard API is queried first to avoid a duplicate GitHub call; GitHub releases API is used as a direct fallback when the API is unreachable
- `cmd_update()` dispatcher now passes `"$@"` so all flags reach the function

## Files changed

- `routers/updates.py` — added `GET /api/update/dry-run` endpoint and `_UPDATE_ENV_KEYS` constant
- `dream-cli` — added `cmd_dry_run()`, wired `--dry-run/-n` into `cmd_update()`, fixed dispatcher arg forwarding, added help example

## Test plan

- [ ] `dream update --dry-run` on an up-to-date install → shows "up to date", lists current images and .env keys, exits 0
- [ ] `dream update --dry-run` when a newer GitHub release exists → shows "update available" with version diff and changelog URL
- [ ] `dream update -n` (short flag) works identically
- [ ] `dream update` (no flag) still performs the real update unchanged
- [ ] `GET /api/update/dry-run` via dashboard API returns `dry_run: true`, correct `current_version`, `images` list, and `env_keys`
- [ ] With dashboard API unreachable, CLI falls back to GitHub API and local compose parsing without error
- [ ] No containers are started, stopped, or re-created during dry run (verify with `docker events`)

### DIFF
--- a/dream-server/dream-cli
+++ b/dream-server/dream-cli
@@ -650,19 +650,26 @@ cmd_dry_run() {
     read -ra flags <<< "$flags_str"
 
     # Prefer API response; fall back to parsing compose files directly.
+    # Use process substitution so flags like images_shown update in this shell (not a pipe subshell).
     local images_shown=false
     if [[ -n "$api_json" ]] && echo "$api_json" | jq -e '.images | length > 0' >/dev/null 2>&1; then
-        echo "$api_json" | jq -r '.images[]' | sort -u | while read -r img; do
-            echo "  ${img}"
-        done
-        images_shown=true
+        local _img_line _any_api=false
+        while IFS= read -r _img_line || [[ -n "$_img_line" ]]; do
+            [[ -z "$_img_line" ]] && continue
+            echo "  ${_img_line}"
+            _any_api=true
+        done < <(echo "$api_json" | jq -r '.images[]' | sort -u)
+        [[ "$_any_api" == "true" ]] && images_shown=true
     fi
     if [[ "$images_shown" == "false" ]]; then
-        if docker compose "${flags[@]}" config 2>/dev/null \
-                | grep -E '^\s+image:' | sed 's/.*image:\s*//' | sort -u \
-                | while read -r img; do echo "  ${img}"; done; then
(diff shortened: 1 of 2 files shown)


Lightheartdevs (changes requested):
## Review: Needs Work

The dry-run concept is solid but there are a couple bugs:

### 1. `--force` flag silently removed (breaking change)
The new argument parsing in `cmd_update()` replaces the old `case` block and drops `--force|-f` handling entirely. Unrecognized flags are now silently ignored via `*) shift`. Users relying on `dream update --force` will get no error — it'll just be ignored. Please restore the `--force` flag alongside `--dry-run`.

### 2. Bash subshell variable scoping bug
`images_shown` is set to `false`, then updated to `"true"` inside a pipe (`| while read ...`). Due to bash subshell scoping, the update never propagates to the parent shell. The fallback image display will always run regardless. Fix: use a process substitution or temp file instead of a pipe.

### 3. Ruff lint failure
CI shows a Python lint failure — likely the `str | None` union syntax which requires Python 3.10+. Use `Optional[str]` for broader compatibility.

### What's good
- The `cmd_dry_run()` function is strictly read-only — verified no writes anywhere
- Good coverage: version, images, model config, .env keys
- API endpoint properly auth-gated

Co-Authored-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>

PR #538, opened 2026-03-21
fix(extensions-library): unify immich DB password env var to IMMICH_DB_PASSWORD

yasinBursali:
## What
Unify both immich-server and postgres on `IMMICH_DB_PASSWORD` with backward-compatible fallback to `IMMICH_DB_PASS`.

## Why
immich-server used `IMMICH_DB_PASS` while postgres used `IMMICH_DB_PASSWORD` — two different env var names for the same secret. Both defaulted to "postgres" masking the mismatch. Users could not rotate the password with a single variable.

## How
Both services now use: `${IMMICH_DB_PASSWORD:-${IMMICH_DB_PASS:-postgres}}`

This means:
- `IMMICH_DB_PASSWORD` set → both use it (new canonical name)
- Only `IMMICH_DB_PASS` set → both use it (backward compat)
- Neither set → both default to "postgres"

## Scope
All changes within `resources/dev/extensions-library/services/immich/compose.yaml`.

## Testing
- YAML validation: passed
- Env var consistency check: passed (both services use identical expression)
- Docker Compose nested variable interpolation: supported in Compose v2+
- Critique Guardian: APPROVED

## Review
Critique Guardian verdict: APPROVED — nested fallback correct, all scenarios verified, zero regression risk.

Originally reported in yasinBursali/DreamServer#90

## Merge Order
- No conflicts with any other open PRs — can merge independently in any order.

### DIFF
--- a/resources/dev/extensions-library/services/immich/compose.yaml
+++ b/resources/dev/extensions-library/services/immich/compose.yaml
@@ -13,7 +13,7 @@ services:
       - LLM_API_URL=${LLM_API_URL:-http://localhost:8000}
       - DB_HOSTNAME=${IMMICH_DB_HOST:-postgres}
       - DB_USERNAME=${IMMICH_DB_USER:-postgres}
-      - DB_PASSWORD=${IMMICH_DB_PASS:-postgres}
+      - DB_PASSWORD=${IMMICH_DB_PASSWORD:-${IMMICH_DB_PASS:-postgres}}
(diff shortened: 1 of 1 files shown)


Lightheartdevs:
Approved but has merge conflicts from other PRs that just landed. Please rebase against main and we'll merge. Thanks for the solid work on these\! 🙏

PR #670, opened 2026-03-29
feat(extensions-portal): add installer support for extensions library

yasinBursali:
## Summary
- Installer copies extensions-library to data dir at install time
- resolve-compose-stack.sh scans data/user-extensions for enabled compose files
- Adds DREAM_EXTENSIONS_LIBRARY_DIR and DREAM_AGENT_PORT env vars to dashboard-api

## Test plan
- [ ] Linux install copies library to DATA_DIR/extensions-library
- [ ] macOS install does the same
- [ ] resolve-compose-stack picks up user-extension compose files
- [ ] Disabled extensions (.disabled suffix) correctly skipped

Part 2 of 5 for the Extensions Portal feature.

### DIFF
--- a/.github/workflows/validate-env.yml
+++ b/.github/workflows/validate-env.yml
@@ -55,6 +55,10 @@ jobs:
             # Generate minimal .env with all expected keys
             cat > .env << ENVEOF
           # Test .env for tier ${{ matrix.tier }}
+          WEBUI_SECRET=test-secret-changeme
+          N8N_USER=admin
+          N8N_PASS=test-pass-changeme
+          OPENCLAW_TOKEN=sk-test-openclaw-token
           DREAM_VERSION=2.1.0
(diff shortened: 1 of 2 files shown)


Lightheartdevs (approved):
## Audit Review

Clean and correct. Small scope (89 additions, 4 files).

- Extensions library copy at install time with proper guard (`[[ -d "$_ext_lib_src" ]]`)
- `resolve-compose-stack.sh` correctly discovers user-installed extensions and their GPU overlays
- macOS LaunchAgent uses correct restart semantics (`KeepAlive.SuccessfulExit: false`)
- Both `install.sh` and `install-macos.sh` paths covered

Minor note: Python path is baked into the macOS plist at install time — if the user changes their Python install, the plist would point to a stale path. Standard practice, not blocking.

LGTM.

PR #672, opened 2026-03-29
feat(extensions-portal): add extensions API with host agent integration

yasinBursali:
## Summary
Full extensions REST API for the dashboard:
- GET catalog with status enrichment, source field (core/user/library), filters
- GET detail with setup instructions and credential info
- POST install/enable/disable, DELETE uninstall with filesystem + agent calls
- POST logs proxied to host agent for live container logs
- Graceful fallback when agent is offline (restart_required flag)

## Security (16-check compose scan)
Blocks: privileged, docker socket, host paths, extra_hosts, sysctls, devices, dangerous caps (10 types), pid:host, network_mode:host, root user, build contexts, 0.0.0.0 ports, core service name collisions. Re-scans at enable time to prevent post-install modification.

## Test plan
- [ ] 53 unit tests pass
- [ ] Install extension via API, container starts
- [ ] Disable stops container before renaming compose file
- [ ] Agent offline: returns restart_required: true

Part 4 of 5 for the Extensions Portal feature.

### DIFF
--- a/.github/workflows/validate-env.yml
+++ b/.github/workflows/validate-env.yml
@@ -55,6 +55,10 @@ jobs:
             # Generate minimal .env with all expected keys
             cat > .env << ENVEOF
           # Test .env for tier ${{ matrix.tier }}
+          WEBUI_SECRET=test-secret-changeme
+          N8N_USER=admin
+          N8N_PASS=test-pass-changeme
+          OPENCLAW_TOKEN=sk-test-openclaw-token
           DREAM_VERSION=2.1.0
           DREAM_TIER=${{ matrix.tier }}
           GPU_BACKEND=cpu
--- a/dream-server/.env.schema.json
+++ b/dream-server/.env.schema.json
@@ -17,11 +17,12 @@
     },
     "DREAM_MODE": {
       "type": "string",
-      "description": "LLM backend mode: local, cloud, or hybrid",
+      "description": "LLM backend mode: local, cloud, hybrid, or lemonade (AMD)",
       "enum": [
         "local",
         "cloud",
-        "hybrid"
+        "hybrid",
+        "lemonade"
       ],
       "default": "local"
     },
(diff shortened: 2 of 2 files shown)


Lightheartdevs (changes requested):
## Audit Review — Security-Critical PR

This is the strongest PR in the chain. The filesystem safety (regex + resolve + is_relative_to + lock + staged scan + symlink rejection) and the 16-check compose scanner are well-designed. The test suite (955 lines, covers install/enable/disable/uninstall/path traversal/scan edge cases) is thorough. A few scanner gaps need closing:

### High: Missing compose scanner checks

Three namespace isolation bypasses are not caught:

1. **`ipc: host`** — gives container access to host IPC namespace (shared memory of other processes)
2. **`userns_mode: host`** — disables user namespace isolation, maps container root to host root
3. **Named volumes with `driver_opts`** — a top-level volume definition with `driver_opts: {type: none, o: bind, device: /etc}` mounts arbitrary host paths. The service-level reference is just a named volume, so the absolute path check never fires.

**Fix:** Add checks for `ipc` and `userns_mode` on each service def. Scan top-level `volumes:` definitions for `driver_opts` containing bind mount patterns.

### Medium: Case-sensitive capability check

The blocklist uses exact string matching (`cap in {"SYS_ADMIN", ...}`). Docker is case-insensitive for capabilities, so `cap_add: ["sys_admin"]` passes the scanner but Docker grants it.

**Fix:** `cap.upper() in DANGEROUS_CAPS`

### Medium: Dict-form ports bypass 127.0.0.1 enforcement

Docker Compose supports ports as dicts: `{target: 80, published: 8080}`. The scanner does `str(port)` which produces a dict repr that doesn't parse correctly with `split(":")`. The port binding check fails to enforce localhost.

**Fix:** Handle both string and dict port formats.

### Medium: Enable TOCTOU

The compose scan runs outside the lock. Between scan and `os.rename` inside the lock, file contents could be modified. Move `_scan_compose_content` inside `_extensions_lock()`.

### Bug: Reverse dependency warning never fires

The disable endpoint reads `peer_data.get("depends_on", [])` at the manifest root, but manifests store deps under `service.depends_on`. Should be `peer_data.get("service", {}).get("depends_on", [])`. The test passes by accident because it also writes deps at the wrong level.

### Good
- Atomic install via temp dir + `os.rename`
- `_copytree_safe` skips symlinks, FIFOs, block/char devices, sockets
- `fcntl.flock(LOCK_EX)` for concurrent mutation safety
- 50MB size limit on extensions
- Path traversal: regex + `.resolve()` + `.is_relative_to()` — belt and suspenders
- Symlink rejection under lock (TOCTOU defense)
- Re-scan at enable time (defense against post-install modification)
- Dependency checking on enable, reverse-dep warnings on disable
- 955-line test file covering all critical paths

PR #736, opened 2026-04-02
fix: replace localhost with 127.0.0.1 in all healthcheck definitions

yasinBursali:
## What
Replace all `localhost` references in Docker healthcheck `test:` lines with `127.0.0.1` to prevent IPv6 resolution failures.

## Why
On dual-stack hosts, `localhost` may resolve to `::1` (IPv6). Services listening only on IPv4 become unreachable, causing healthcheck failures that cascade through `depends_on: condition: service_healthy` — potentially bringing down the entire stack.

## How
19 replacements across 17 files: `docker-compose.base.yml`, `docker-compose.amd.yml`, 12 extension compose files, and 2 extension templates. Only healthcheck `test:` lines changed — environment variables left untouched.

## Testing
- Grep confirms zero remaining `localhost` in active healthcheck lines ✅
- YAML syntax validated ✅

## Platform Impact
All platforms (especially dual-stack Linux hosts with IPv6 enabled)

### DIFF
--- a/dream-server/docker-compose.amd.yml
+++ b/dream-server/docker-compose.amd.yml
@@ -44,7 +44,7 @@ services:
       - LEMONADE_LLAMACPP_BACKEND=rocm
       - LEMONADE_CTX_SIZE=${CTX_SIZE:-131072}
     healthcheck:
-      test: ["CMD", "curl", "-sf", "http://localhost:8080/api/v1/health"]
+      test: ["CMD", "curl", "-sf", "http://127.0.0.1:8080/api/v1/health"]
       interval: 15s
       timeout: 10s
       retries: 5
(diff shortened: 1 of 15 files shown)


Lightheartdevs (changes requested):
## Audit Review

The change itself is correct and follows Docker best practices — `127.0.0.1` is more reliable than `localhost` in containers that may lack `/etc/hosts` or resolve `localhost` to IPv6.

### Blocking: Docker Compose validation CI failure

"Validate Docker Compose files" is failing. Investigate whether the changes introduced a syntax issue, or if this is the pre-existing llama-server image failure. If it's the pre-existing one (llama-server has no image in base.yml), then this is not blocking.

PR #742, opened 2026-04-03
fix: mark Fooocus as unsupported locally, add uninstall volume warning

yasinBursali:
## What
Mark Fooocus extension as non-functional for local deployment, and add Docker volume cleanup hints to the extension uninstall response.

## Why
The Fooocus extension uses `runpod/fooocus:2.5.3`, a RunPod-specific Docker image that requires `RUNPOD_POD_ID` and `RUNPOD_API_KEY`. On local Docker, the container exits or hangs in a restart loop. Separately, uninstalling extensions leaves orphaned Docker volumes with no warning.

## How
- Cleared `compose_file` in Fooocus manifest so it won't be included in the compose stack
- Renamed `compose.yaml` → `compose.yaml.reference` and `compose.nvidia.yaml` → `compose.nvidia.yaml.reference` to prevent filesystem-based discovery during user-install (the resolver's user-extensions scanner checks for `compose.yaml` by filename)
- Updated manifest description to note RunPod requirement
- Added `message` and `cleanup_hint` fields to uninstall API response (no auto-deletion — just informs the user)

## Testing
- Python syntax verified
- Verified resolver skips empty `compose_file` and `.reference` suffix files
- Verified `service_id` in cleanup_hint is regex-validated (no injection risk)

## Review
Critique Guardian: REJECTED → fixed file renames + cleanup_hint → APPROVED

## Platform Impact
- **macOS:** No impact (Fooocus is nvidia-only, uninstall response is pure Python)
- **Linux:** Fooocus no longer causes restart loops when installed
- **Windows/WSL2:** Same as Linux

> **Note:** This PR touches `extensions.py` which is also modified by #733. Rebase needed after #733 merges.

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/dream-server/config/extensions-catalog.json
+++ b/dream-server/config/extensions-catalog.json
@@ -1,5 +1,5 @@
 {
-  "generated_at": "2026-04-03T09:13:20Z",
+  "generated_at": "2026-04-03T09:14:27Z",
   "schema_version": "1.0.0",
   "extensions": [
     {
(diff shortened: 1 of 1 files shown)


Lightheartdevs:
#735 just merged and caused a conflict on the Fooocus manifest. Needs a rebase, then it's good to go.

PR #750, opened 2026-04-03
feat: AMD Multi-GPU Support

y-coffee-dev:
# feat: AMD Multi-GPU Support

End-to-end multi-GPU support for AMD GPUs, matching the existing NVIDIA multi-GPU feature set.

Previously, AMD support was limited to single-GPU. This branch implements end to end support with hardware discovery, topology analysis, GPU assignment, Docker Compose isolation, CLI management, and monitoring, to handle multiple AMD GPUs.

### Hardware Detection
- Multi-GPU AMD detection via sysfs (counts all `vendor=0x1002` cards)
- Handles XCP virtual cards on MI300X (filters by non-empty vendor file)
- Total VRAM aggregation across all detected AMD GPUs
- Mixed APU + discrete GPU classification

### Topology Detection
- Full AMD topology library with three detection backends: amd-smi JSON, rocm-smi text, sysfs NUMA/IOMMU fallback
- Inter-GPU link classification: XGMI, PCIe-SameSwitch, PCIe-HostBridge, PCIe-CrossNUMA
- Link ranking system (0–100) for topology-aware GPU assignment
- Per-GPU metadata: render node, GFX version, PCI BDF, VRAM, memory type (unified/discrete)
- GPU identification with three fallback methods: amd-smi UUID, sysfs unique_id, composite PCI BDF

### Installer Integration
- AMD topology detection phase (runs when GPU_COUNT > 1 and backend is AMD)
- Vendor-aware GPU assignment extraction
- AMD multi-GPU env vars written to `.env` only when applicable
- Render node verification in AMD tuning phase

### Docker Compose Overlays
- AMD multi-GPU overlay for llama-server with Lemonade passthrough
- `--split-mode` passed via `--llamacpp-args` (Lemonade's official mechanism)
- ROCm backend selected via `LEMONADE_LLAMACPP` env var (compatible with both Python and C++ Lemonade builds)
- Per-service GPU isolation via `ROCR_VISIBLE_DEVICES`
- Renamed existing NVIDIA overlays from generic `multigpu` to `multigpu-nvidia`

### CLI (`dream gpu` commands)
- All five GPU commands (`status`, `topology`, `assign`, `reassign`, `monitor`) are AMD-aware
- AMD GPU status table with VRAM, utilization, temperature, power via amd-smi
- AMD topology display with GFX versions, memory types, and render nodes
- AMD GPU reassignment writes `LLAMA_SERVER_GPU_INDICES` and per-service `*_GPU_INDEX`

### Dashboard API
- AMD GPU monitoring via amd-smi and sysfs hwmon
- Per-GPU metrics: utilization, VRAM, temperature, power, fan speed
- GPU assignment decoding from `GPU_ASSIGNMENT_JSON_B64`

### Environment & Schema
- `LLAMA_SERVER_GPU_INDICES` — comma-separated GPU indices for `ROCR_VISIBLE_DEVICES`
- `COMFYUI_GPU_INDEX`, `WHISPER_GPU_INDEX`, `EMBEDDINGS_GPU_INDEX` — per-service GPU index
- `LLAMA_ARG_SPLIT_MODE`, `LLAMA_ARG_TENSOR_SPLIT` — llama.cpp multi-GPU parameters
- All new env vars added to `.env.schema.json` and `.env.example`

### Tests
- 25 BATS unit tests for amd-topo.sh (render node, GFX version, GPU name, GPU ID, topology parsing)
- 23 shell integration tests with fixture files (4-GPU XGMI, 2-GPU PCIe, field-based select, cross-tool agreement)
- 16 pytest tests for dashboard-api AMD GPU monitoring
- Fixture files extracted from real hardware for amd-smi JSON and rocm-smi text output
- Real hardware tests on 4x AMD Instinct MI300X

## Files Changed

| Area | Files |
|------|-------|
| Topology detection | `installers/lib/amd-topo.sh` (new) |
| Hardware detection | `installers/lib/detection.sh`, `installers/phases/02-detection.sh` |
| Installer | `installers/phases/03-features.sh`, `installers/phases/06-directories.sh`, `installers/phases/10-amd-tuning.sh` |
| Compose overlays | `docker-compose.multigpu-amd.yml` (new), `docker-compose.multigpu-nvidia.yml` (renamed) |
| Service overlays | `extensions/services/{comfyui,whisper,embeddings}/compose.multigpu-amd.yaml` (new), NVIDIA renamed |
| CLI | `dream-cli` |
| Assignment | `scripts/assign_gpus.py`, `scripts/resolve-compose-stack.sh` |
| Dashboard API | `extensions/services/dashboard-api/gpu.py` |
| Config | `config/gpu-database.json`, `.env.schema.json`, `.env.example` |
| Tests | `tests/bats-tests/amd-topo.bats`, `tests/test-amd-topo.sh`, `tests/fixtures/amd/*`, `extensions/services/dashboard-api/tests/test_gpu_amd.py` |


Lightheartdevs (changes requested):
## Audit Review — AMD Multi-GPU Support

Strong PR. The architecture mirrors the NVIDIA multi-GPU system faithfully, the code is well-structured, and the test coverage is excellent (64+ tests across BATS, shell, and pytest with real MI300X hardware fixtures). No security concerns — sysfs reads are guarded, PCI BDF strings are regex-filtered, jq uses numeric variables, no shell injection vectors.

Two bugs to fix, two things to verify:

### Bug 1: Bash evaluation order in _gpu_status power reading

```bash
if [[ "$pw_uw" -eq 0 || ! "$pw_uw" =~ ^[0-9]+$ ]]
```

The `-eq` runs before the regex check. If `pw_uw` is non-numeric (e.g., sysfs returns an error string), bash throws an integer comparison error. Swap the order:

```bash
if [[ ! "$pw_uw" =~ ^[0-9]+$ || "$pw_uw" -eq 0 ]]
```

### Bug 2: Dead jq expression in _gpu_reassign auto-mode

The `llama_indices` assignment uses a jq filter that always returns empty string:
```bash
llama_indices=$(echo "$assignment_json" | jq -r '
    [.gpu_assignment.services.llama_server.gpus[]] as $uuids |
    "" ')
```
The actual index extraction happens in the while loop below. Remove the dead code.

### Verify: compute_subset rank_matrix indexing

In the APU+dGPU hybrid path (`assign_gpus.py` ~line 488):
```python
discrete_gpus = [g for g in llama_subset.gpus if g.memory_type == "discrete"]
discrete_subset = compute_subset(discrete_gpus, rank_matrix)
```
Confirm `compute_subset` uses `gpu.index` (original topology index) for rank_matrix lookups, not list position. If it uses list position, the filtered subset will produce wrong link rankings.

### Verify: Old multigpu.yml filename references

The rename from `docker-compose.multigpu.yml` to `docker-compose.multigpu-nvidia.yml` is clean, but grep the full codebase for any hardcoded references to the old filename in dream-cli, docs, or tests.

### Non-blocking notes (follow-up material)
- `gpu-database.json`: RX 7900 XTX/XT/GRE share device_id `0x744c` — matcher must prioritize `name_patterns` over `device_ids`
- Missing schema entries for `ROCR_VISIBLE_DEVICES`, `VIDEO_GID`, `RENDER_GID`, `HSA_OVERRIDE_GFX_VERSION`
- `_gpu_reassign` interactive prompts have no `--yes` flag — will hang in non-interactive pipelines
- No tests for 0-GPU and 1-GPU edge cases, `_detect_topo_sysfs` fallback, or mixed APU+dGPU in the API detailed view

### Good
- 3-backend fallback chain (amd-smi -> rocm-smi -> sysfs) with graceful degradation
- Hybrid APU+dGPU strategy correctly routes lightweight services to APU, freeing discrete VRAM for LLM
- All 5 CLI GPU commands are now vendor-aware
- Tests use real hardware fixtures (4x MI300X XGMI, 2-GPU PCIe) — no flaky mocks
- Docker Compose overlays follow existing patterns (`ROCR_VISIBLE_DEVICES` parallels `NVIDIA_VISIBLE_DEVICES`)
- `resolve-compose-stack.sh` correctly constructs vendor-specific overlay filenames
- All CI failures are pre-existing on main, none caused by this PR

Lightheartdevs (commented):
**Audit: NEEDS DISCUSSION — scope and existing review**

This is a substantial PR (+2,866 lines) touching installer phases, CLI, dashboard API, compose overlays, and tests. The architecture looks sound:
- Multi-backend topology detection (amd-smi, rocm-smi, sysfs fallback) with proper fallback chains
- `ROCR_VISIBLE_DEVICES` for GPU isolation in compose overlays
- `GROUP_ADD` for video/render GIDs is correct for AMD GPU access
- Impressive test coverage (25 BATS + 23 shell + 16 pytest)
- No `shell=True` in subprocess calls

**Concerns:**
- The `multigpu` → `multigpu-nvidia` rename is a breaking change if any automation references the old filename — verify `resolve-compose-stack.sh` handles it
- Several AMD GPUs share `device_ids` (e.g., RX 7900 XTX and XT both have `0x744c`) — the consumer code must use `name_patterns` to disambiguate. Verify matching priority.
- `amd-topo.sh` parses sysfs and amd-smi output — needs review for command injection in the parsing paths
- XCP virtual card filtering (MI300X) relies on empty vendor file — hardware-specific, hard to test without real MI300X

**Status:** Already has changes requested. The outstanding review comments should be addressed before further deep audit. The core architecture is solid but the blast radius warrants careful incremental review.

y-coffee-dev:
Thanks for the thorough review! I did some adjustments, and here's where each concern lands after verification, with fixes in the latest commit.

### `multigpu` to `multigpu-nvidia` rename

Verified. The old files existed on `main` and were renamed (identical content). `resolve-compose-stack.sh` was updated to construct filenames dynamically per backend (`f"docker-compose.multigpu-{gpu_backend}.yml"`). `dream-cli` has zero references to any multigpu filename. Updates go through `git pull` which handles renames transparently.

### Shared `device_id` `0x744c` (XTX / XT / GRE) - fixed

Valid concern, and it was introduced by this PR's new `gpu-database.json` entries. I traced through the matching logic in `classify-hardware.sh` and found two issues:

**Substring collision:** `"RX 7900 XT"` is a substring of `"RX 7900 XTX"`. The old matcher broke on the first dual match, making JSON entry ordering load-bearing. Fixed by tracking the longest matched pattern. `"RX 7900 XTX"` (13 chars) now wins over `"RX 7900 XT"` (10 chars) regardless of entry order.

**Empty name fallback:** When sysfs `product_name` is unavailable, the old code fell back to the first device_id match (always XTX). Fixed by using VRAM proximity as a tiebreaker. The classifier already receives `--vram-mb`, so a 20 GB card now resolves to XT and a 16 GB card to GRE. When VRAM is also 0, it picks the smallest matching card because under-provisioning is safer (smaller model runs fine), over-provisioning may crash the model loader.

Added 35 test assertions to `test-installer-contracts.sh` covering the happy path as well as substring safety (XT must not match XTX and vice versa), tier/bandwidth correctness, VRAM tiebreaker with exact and approximate values, zero-VRAM fallback, the `0x7480` pair, name-only matches, and unknown GPU heuristic fallback.

### Command injection in parsing paths

Walked every external data ingestion point in `amd-topo.sh`. No injection vector exists. All sysfs reads use kernel-controlled paths, `jq` calls use integer loop counters, `awk` calls use `-v` parameter passing, PCI BDFs are regex-filtered, zero uses of `eval`. Python side: zero `shell=True` in entire `dashboard-api/`.

Found one inconsistency: `$vram_bytes` was interpolated directly into an awk program string while the rest of the file used `-v`. Fixed it in this commit. Also added a missing `grep -oP '^\d+'` filter on `pcie_width` to match `pcie_gen`'s existing filter. MI300X fixture data confirmed `current_link_width` can be `"0"` and `current_link_speed` can be `"Unknown"`.

### XCP virtual card filtering (MI300X)

Confirmed using real MI300X fixture data. The 4x MI300X system exposes 32 card entries: 4 real GPUs (card0, card8, card16, card24) with `vendor=0x1002`, and 28 XCP virtual partitions with empty vendor files. The standard `vendor == "0x1002"` check naturally filters them. All three code paths (`detection.sh`, `amd-topo.sh`, `gpu.py`) use the same mechanism. Python path is tested in `test_gpu_amd.py`, bash paths use identical logic. Already validated on real MI300X hardware.

Thanks again for the review!

Lightheartdevs (changes requested):
Appreciate the scope — end-to-end AMD multi-GPU is a real gap and the work is substantial: topology library, GPU assignment, compose overlays, CLI commands, dashboard monitoring, plus real-hardware testing on 4x MI300X. The architecture (three detection backends with fallback chain, link-ranking 0-100, per-service GPU isolation via `ROCR_VISIBLE_DEVICES`) is thoughtful.

**Blocking issues:**

1. **Merge state is CONFLICTING.** Needs rebase on main before any meaningful review. The file list shows overlap with multiple recent PRs (`.env.schema.json`, `detection.sh`, `resolve-compose-stack.sh`, `dream-cli`, `routers/gpu.py`) — conflict likely non-trivial.

2. **Breaking rename: `docker-compose.multigpu.yml` → `docker-compose.multigpu-nvidia.yml`.** This affects:
   - Any existing install's `.compose-flags` cache that lists the old filename
   - [`docker-compose.multigpu.yml` exists at root and is referenced by `resolve-compose-stack.sh:138`](dream-server/scripts/resolve-compose-stack.sh) via `(script_dir / "docker-compose.multigpu.yml").exists()`
   - Documentation that references the old name
   
   Need either: (a) a migration step in `scripts/` that rewrites `.compose-flags` on first run, or (b) keep the old name as a symlink/alias for a release cycle. Right now this will break any existing multi-GPU NVIDIA install on update.

3. **No CI coverage for AMD multi-GPU.** The BATS/pytest tests are excellent for the pure-Python/shell pieces, but there's no smoke test that the overlay resolution actually produces a working compose stack. The `test-installer-contracts.sh` additions (+87 lines) help but don't validate runtime. Consider a dry-run smoke job in `.github/workflows/` that asserts the resolved compose is valid `docker compose config` output.

4. **`LEMONADE_LLAMACPP` env var** — please add to `.env.schema.json` with a description and allowed values. I see it's used in the compose overlay but don't see a schema entry.

5. **First-time contributor, large feature.** I don't know if that's a blocker for this repo, but please confirm there's a maintainer committed to ongoing AMD multi-GPU support. The maintenance surface (three detection backends, fixture refresh on new ROCm versions, MI300X XCP virtual card handling) is real.

**Non-blocking notes:**

- Link ranking system is smart — XGMI > same-switch PCIe > cross-NUMA. Consider documenting the ranking numbers in `dream-server/docs/` so future changes don't silently invert them.
- Fixture files from real hardware (`tests/fixtures/amd/*`) are excellent. Please note which ROCm/amd-smi version they were captured from, so they can be refreshed when the tools evolve.
- `ROCR_VISIBLE_DEVICES` per-service isolation is the right call over relying on the llama.cpp `--split-mode` alone.

Happy to re-review once the rebase + the rename migration + the LEMONADE_LLAMACPP schema are addressed. The core work is solid — don't want it to bitrot. Thanks for the contribution.

y-coffee-dev:
Thanks for the detailed review, I appreciate it!

1. **Merge state:** I rebased, PR has no conflicts at the moment.
2. **Rename migration:** Went with the migration step approach. `dream-cli` now validates every path in `.compose-flags` before trusting it, and silently re-resolves when any overlay is missing. Existing multi-GPU NVIDIA installs will pick up the rename on their next `dream-cli` invocation without any user action.
3. **CI coverage:** Three new jobs/steps in `validate-compose.yml`, base+backend overlay validation, multi-GPU overlay validation (both AMD and NVIDIA), and a resolver assertion that `docker-compose.multigpu-amd.yml` shows up for `--gpu-backend amd --gpu-count 2`. Everything goes through `docker compose config --quiet` so we catch schema drift before runtime.
4. **`LEMONADE_LLAMACPP`:** Added to `.env.schema.json` with the `rocm`/`vulkan`/`cpu` enum and `rocm` default. The compose overlay now uses `${LEMONADE_LLAMACPP:-rocm}` so schema and overlay agree.

On the non-blocking notes:
- Link-ranking numbers are documented inline at the top of `amd-topo.sh`. Happy to move them to `dream-server/docs/` as a standalone reference if you'd prefer that layout.

Ready for re-review whenever you have a moment. Thanks again!

Lightheartdevs:
Audit follow-up: promising AMD work, but needs revision before merge.

The multi-GPU architecture is directionally right, and the synthetic topology/dashboard checks looked good. The blocker is that several compose refresh/cache call sites omit `--gpu-count`, so a `GPU_COUNT=2` AMD stack can later resolve/cache only `docker-compose.base.yml + docker-compose.amd.yml` and drop `docker-compose.multigpu-amd.yml`. Please pass the GPU count through every resolver refresh path, then rerun the AMD multi-GPU regression/hardware validation notes.

y-coffee-dev:
Good catch, I fixed it, and it turns out it's broader than just AMD. Only installers/lib/compose-select.sh was passing --gpu-count; every other caller (03-features.sh, 11-services.sh, dream-cli, dream-update.sh, dream-host-agent.py, dream-preflight.sh, validate.sh, dream-macos.sh) silently defaulted to 1 and dropped both the top-level multigpu-{backend}.yml overlay and the per-extension compose.multigpu-{backend}.yaml overlays. The damage was persistent because 11-services.sh wrote the broken result into .compose-flags, locking the bug into every subsequent dream-cli/dream-update/host-agent invocation. GPU_COUNT is already loaded from .env by load_env/load_env_file in every Bash caller; host agent now reads it at startup alongside GPU_BACKEND/TIER. Pinned with a regression contract in test-installer-contracts.sh that walks every shipped resolver call site and fails if any invocation lacks --gpu-count.

Lightheartdevs:
Re-audit update: this branch has made useful progress on earlier AMD multi-GPU findings.

What looks improved:

- The previous `--gpu-count` refresh gap is fixed in the installer and dream-cli regeneration paths.
- `LEMONADE_LLAMACPP` is now present in `.env.schema.json`.
- The stale `.compose-flags` cache problem from the multigpu filename rename is at least partly addressed by validating cached `-f` paths before reuse.

Still not merge-ready:

- GitHub marks the branch `DIRTY`, and a local merge simulation into current `origin/main` conflicts in `.github/workflows/validate-compose.yml`, `dream-server/installers/lib/detection.sh`, `dream-server/tests/bats-tests/docker-phase.bats`, and `dream-server/tests/contracts/test-installer-contracts.sh`.
- `git diff --check` fails on trailing whitespace in `dream-server/tests/fixtures/amd/rocm_smi_showtopo_4gpu_xgmi.txt`.
- I could syntax-check the touched shell scripts, but `tests/test-amd-topo.sh` could not run in this local environment because `jq` is missing. I did not run real AMD hardware verification in this pass.

This is still valuable AMD-partnership work, but it needs a rebase/conflict pass and whitespace cleanup before it should come back to the merge queue.

y-coffee-dev:
Thanks! I addressed each blocker:

**1. Conflicts in 4 files.** Fixed. Branch rebased onto main and conflicts resolved.

**2. `git diff --check` whitespace failure on `rocm_smi_showtopo_4gpu_xgmi.txt`.** Worth a quick discussion before stripping it, the trailing whitespace is real `rocm-smi --showtopo` column padding, and fixtures in `tests/fixtures/amd/` exist to capture what the tool actually emits. The amd-topo.sh awk parser is whitespace-tolerant, so it's harmless at runtime.

Took a slightly different route: a `.gitattributes` rule scoped to the fixture directory that exempts captured tool output from the whitespace check:

```
dream-server/tests/fixtures/amd/*.txt -whitespace
```

Verified: `git diff --check` is now clean and `git check-attr whitespace` returns `unset`. 

**3. `test-amd-topo.sh` requires jq.** I think this one's an environmental thing rather than a code issue. The test already exits with a clear `ERROR: jq is required` message at `test-amd-topo.sh:25-28`, and amd-topo.sh itself hard-depends on jq for parsing `amd-smi --json` (its header notes `Expects: jq`), so a no-jq environment can't validly exercise either the production code or its test. `make test` doesn't invoke this script, it's opt-in.

**4. CI failure: `detect_gpu: does not misidentify discrete AMD GPU as APU`.** The test was written for main's pre-multi-GPU `detect_gpu`, where discrete AMD cards fell through to CPU-only mode and the test asserted that fallthrough. The branch's loop refactor now classifies a 32 GB VRAM / 16 GB GTT card as `amd` / `discrete` that's the *intended* multi-GPU contract, but the test body still asserted the old `assert_failure`. Test name is unchanged and still accurate; body now asserts the correct positive contract:

```bash
assert_equal "$GPU_BACKEND"     "amd"
assert_equal "$GPU_MEMORY_TYPE" "discrete"
```

Ready for re-review, thank you so much!

Lightheartdevs (changes requested):
Re-audit of current head `3aea5eb`: still not merge-ready, but still valuable.

What looks better now:

- `git diff --check origin/main...HEAD` is clean; the fixture whitespace exemption works.
- Shell syntax check over the touched installer/CLI scripts passed locally.
- The prior `--gpu-count` refresh gap appears addressed across the resolver call sites.
- CI was green on the PR head when it last ran.

Remaining blockers:

1. The branch is still not mergeable into current `origin/main`. A local merge simulation conflicts in:
   - `dream-server/.env.schema.json`
   - `dream-server/dream-cli`
   - `dream-server/installers/phases/03-features.sh`

2. I could not run `dream-server/tests/test-amd-topo.sh` in this local environment because `jq` is not installed (`ERROR: jq is required but not installed`). That is an environment limitation, but it means I did not locally re-exercise the AMD topology regression suite in this pass.

3. I did not run live AMD hardware validation. Given the scope (installer detection, compose overlays, CLI reassignment, dashboard metrics, and MI300X/XGMI topology), this needs a final post-rebase run on AMD hardware before merge.

Recommendation: do not merge as-is. This PR is worth preserving, but it needs a current-main rebase/conflict pass plus a real AMD validation pass. I am not directly fixing and merging it in this queue because the conflicts touch active installer/CLI/schema surfaces and the correctness claim depends on hardware I cannot exercise here.

y-coffee-dev:
Thanks for the re-audit. Addressed each blocker, plus four issues that surfaced when testing on AMD hardware after the rebase.

**1. Not mergeable - conflicts in 3 files.** Rebased onto `main` and resolved conflicts
**2. `test-amd-topo.sh` requires jq.** Environmental, not a code issue. `jq` is a hard project dep, `installers/lib/amd-topo.sh` (the production code under test) invokes it around 10 times, including to *construct* its JSON output. `installers/phases/01-preflight.sh:47-62` auto-installs `jq` on every supported distro.

I softened the test's failure path so the next reviewer without `jq` isn't stuck on a bare error, it now prints an install command suggestion:

```
Install it: apt-get install -y jq | dnf install -y jq | brew install jq
```

`sudo apt-get install -y jq` and re-run gets you to 23/23.

**3. Live AMD hardware validation.** Ran it on an MI300X box. Four issues surfaced, all on-branch and pushed:

- **`24b67118`** — Lemonade now rejects `llamacpp.backend: rocm`:

  ```
  Error 500: 'llamacpp.backend' must be one of: auto, vulkan, cpu
  ```

  Lemonade upstream tightened the schema. Swapped the CLI arg and env-var defaults to `auto` in `docker-compose.amd.yml`.

- **`96f65e41`** — Our schema entry for `LEMONADE_LLAMACPP` had `enum: ["rocm", "vulkan", "cpu"]` / `default: "rocm"`. Updated to `["auto", "vulkan", "cpu"]` / default `auto` so `validate-env.sh` doesn't reject working configs or seed a fresh `.env` with the now-invalid `rocm`.

- **`bdff0fa9`** — The multi-GPU compose overlay fully *replaces* the base command and had been omitting `--llamacpp` on the assumption that the `LEMONADE_LLAMACPP` env var was equivalent. Field-testing showed it isn't — only the CLI arg overrides Lemonade's internal default. Added back as `--llamacpp ${LEMONADE_LLAMACPP:-auto}` so multi-GPU AMD matches single-GPU.

- **`c2eb97e1`** — The real branch defect. Phase 06 was emitting `HSA_OVERRIDE_GFX_VERSION=11.5.1` and `AMDGPU_TARGET=gfx1151` *unconditionally* for every AMD install. On MI300X (real ISA = gfx942) the runtime then tries to launch gfx1151 kernels on gfx942 silicon -> `HSA_STATUS_ERROR_INVALID_ISA (0x100f)` at model-load.

  The branch already detects per-GPU gfx via `amd-topo.sh:amd_gfx_version()` and stores it in `GPU_TOPOLOGY_JSON.gpus[].gfx_version` — that data just wasn't wired into Phase 06's `.env` generation. Phase 06 now:

  - sets `AMDGPU_TARGET` to the detected target,
  - emits `HSA_OVERRIDE_GFX_VERSION=11.5.1` **only** when the target is `gfx1151` (the override is Strix-Halo-specific; gfx942/gfx90a/gfx1100 are in ROCm's native support list and must *not* have it),
  - emits `LEMONADE_LLAMACPP_ROCM_BIN=/opt/llama-custom/llama-server` **only** when the target is `gfx1151` (the custom binary is built with Strix-Halo MMQ patches + `AMDGPU_TARGET=gfx1151`; pointing Lemonade at it from any other architecture ISA-faults or runs perf-regressed).

  **Strix Halo is preserved bit-for-bit.** On real gfx1151 hardware the `case` arm produces the same env block as before; on empty / `null` / `unknown` detection the fallback chain coerces to `gfx1151` so prior installs without a refreshed topology are untouched. Verified by simulating 5 scenarios (MI300X gfx942, Strix Halo gfx1151, RX 7900 gfx1100, empty topology, unknown detection). AMD/Lemonade contracts (17) and installer contracts both still green.

Ready for re-review on the rebase + jq tweak + four AMD fixes above. Thank you!

Lightheartdevs:
I audited this and could not safely update the contributor branch directly because maintainerCanModify=false. I opened a repaired replacement PR at #1261 that carries this AMD multi-GPU work onto current main, resolves the merge conflicts, fixes the Strix-only HSA/custom-binary handling, fixes AMD topology memory classification/card ordering, and has green CI. I’d use #1261 as the merge path for this work.

Lightheartdevs:
Merged this work via the repaired replacement PR #1261 because this contributor branch could not be updated directly (maintainerCanModify=false) and remained unsafe/conflicting against current main. Thank you again for the AMD multi-GPU contribution — the repaired path is now landed in main as d8549218.

PR #804, opened 2026-04-04
feat(host-agent): add per-container stats endpoint and extract dir_size_gb

yasinBursali:
## What
- Add `GET /v1/service/stats` to the host agent returning per-container CPU, memory, and PID metrics
- Extract `dir_size_gb()` from nested function in `main.py` to module-level in `helpers.py`
- Refactor `_compute_storage()` to use the imported version

## Why
- Dashboard needs per-service resource metrics for the upcoming resources page
- `dir_size_gb()` is needed by both the resources endpoint and data lifecycle features — extraction enables reuse
- The nested function prevented importing from other modules

## How
- **Host agent:** `docker stats --no-stream --format '...'` returns one JSON object per line. Parse CPU%, memory usage/limit/percent, PIDs. Filter to `dream-*` containers only. Auth required via bearer token
- **Memory parsing:** `_parse_mem_value()` handles Docker's IEC units (B, KiB, MiB, GiB, TiB) with longest-suffix-first matching to avoid false `endswith("B")` matches
- **dir_size_gb extraction:** Identical logic plus symlink-skip improvement (`not f.is_symlink()`) to prevent double-counting and avoid following links outside DATA_DIR

## Three Pillars Impact
- **Install Reliability:** No installer code touched — runtime API only
- **Broad Compatibility:** `docker stats --format` available since Docker 1.13+. Python 3.10+ `removeprefix` safe. IEC units only (Docker standard)
- **Extension Coherence:** No extension system changes

## Modified Files
- `dream-server/bin/dream-host-agent.py` — `_parse_mem_value()`, `_iso_now()`, `_handle_service_stats()`, route in `do_GET`
- `dream-server/extensions/services/dashboard-api/helpers.py` — `dir_size_gb()` module-level function
- `dream-server/extensions/services/dashboard-api/main.py` — import change, nested function removed

### Automated
- Python compile: PASS (all 3 files)
- Existing tests: 173 pass, 1 pre-existing failure (Python 3.14 asyncio — not introduced by this PR)

### Manual
- [ ] `curl -H "Authorization: Bearer $KEY" http://localhost:$PORT/v1/service/stats` — verify containers array with CPU/memory data
- [ ] Verify `memory_used_mb > 0` for running services
- [ ] Verify only `dream-*` containers in response
- [ ] `/api/storage` endpoint returns same values as before (regression test for `_compute_storage` refactor)
- [ ] No auth → 401

## Review
- CG: ⚠️ APPROVED WITH WARNINGS (4 non-blocking advisories: lazy import in `_iso_now`, returncode unguarded, PIDs int cast, no unit tests for `_parse_mem_value`)
- Compatibility: PASS
- Security: PASS (auth enforced, read-only, dream-* filter, no shell=True)

## Platform Impact
- **macOS:** Supported (Docker Desktop VM-scoped metrics)
- **Linux:** Supported (native Docker, accurate metrics)
- **Windows (WSL2):** No host agent — consumers degrade gracefully with empty stats

## Sequence
PR 3 of 7 (Phase 2 — shared infrastructure)

### DIFF
--- a/dream-server/bin/dream-host-agent.py
+++ b/dream-server/bin/dream-host-agent.py
@@ -251,6 +251,8 @@ class AgentHandler(BaseHTTPRequestHandler):
                  "--format", '{"name":"{{.Name}}","cpu":"{{.CPUPerc}}","mem_usage":"{{.MemUsage}}","mem_percent":"{{.MemPerc}}","pids":"{{.PIDs}}"}'],
                 capture_output=True, text=True, timeout=10,
             )
+            if result.returncode != 0:
+                logger.warning("docker stats returned non-zero: %s", result.stderr[:200] if result.stderr else "")
 
             containers = []
             for line in result.stdout.strip().splitlines():
@@ -283,6 +285,11 @@ class AgentHandler(BaseHTTPRequestHandler):
 
                 service_id = name.removeprefix("dream-")
 
+                try:
+                    pids = int(raw.get("pids", "0") or "0")
+                except (ValueError, TypeError):
+                    pids = 0
+
                 containers.append({
                     "service_id": service_id,
                     "container_name": name,
@@ -290,7 +297,7 @@ class AgentHandler(BaseHTTPRequestHandler):
                     "memory_used_mb": round(mem_used_mb),
                     "memory_limit_mb": round(mem_limit_mb),
                     "memory_percent": round(mem_percent, 1),
-                    "pids": int(raw.get("pids", "0") or "0"),
+                    "pids": pids,
                 })
 
             json_response(self, 200, {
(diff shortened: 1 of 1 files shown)


yasinBursali:
## Addressing review feedback

All 3 bugs fixed:

**Bug 1 (`int("--")` ValueError) — Fixed:**
```python
try:
    pids = int(raw.get("pids", "0") or "0")
    pids = 0
```
Docker returns `"--"` for PIDs when containers are in Created/Restarting/Paused state. Now handled gracefully.

**Bug 2 (Missing broad exception handler) — Fixed:**
Added `except Exception as exc:` after `TimeoutExpired`, matching the existing `_handle_logs()` pattern. Returns JSON 500 instead of uncontrolled HTML from BaseHTTPRequestHandler.

**Bug 3 (No returncode check) — Fixed:**
Added `if result.returncode != 0:` warning log after `docker stats` call. Logs stderr (truncated to 200 chars) but continues processing — `docker stats` may produce partial valid output even on non-zero exit.

PR #808, opened 2026-04-04
feat(dashboard-api): route logs through read-only service endpoint

yasinBursali:
## What
- Update `extension_logs()` to call `/v1/service/logs` instead of `/v1/extension/logs`, enabling log viewing for core services
- Add `container_name` to SERVICES dict for accurate container name resolution
- Replace bare `except Exception:` with specific `urllib.error.HTTPError` and `urllib.error.URLError` catches
- Add Docker Compose label spoofing prevention in `_scan_compose_content()`

## Why
- Core services (llama-server, open-webui, dashboard-api) return 403 when viewing logs because `validate_service_id()` blocks them
- The dashboard Terminal button already renders for core services but the fetch fails silently
- Container name hardcoding (`dream-{service_id}`) fails for open-webui (container is `dream-webui`)
- A malicious extension could spoof Docker Compose labels to masquerade as a core service

## How
- **URL change:** `/v1/extension/logs` → `/v1/service/logs` (the new read-only endpoint that allows all services)
- **Error propagation:** `HTTPError` handler forwards the actual status code from host agent (not always 503). `URLError`/`OSError` returns 503 with f-string fallback: "Host agent unavailable. Use 'docker logs dream-{service_id}' in terminal."
- **Label security:** Rejects any extension compose YAML that sets `com.docker.compose.*` labels (handles both dict and list label formats)
- **container_name:** Read from manifest, falls back to `dream-{service_id}`

## Three Pillars Impact
- **Install Reliability:** No installer changes
- **Broad Compatibility:** Standard urllib, all Python versions
- **Extension Coherence:** Existing `/v1/extension/logs` unchanged; label spoofing prevention strengthens extension security

## Modified Files
- `dashboard-api/config.py` — `container_name` field in SERVICES dict
- `dashboard-api/routers/extensions.py` — `extension_logs()` rewrite + label spoofing check

### Automated
- Python compile: PASS
- Existing tests: 173 pass (1 pre-existing failure unrelated)

### Manual
- [ ] View logs for core service (llama-server) via dashboard Terminal → should work (was 403 before)
- [ ] View logs for user extension → still works
- [ ] Host agent down → 503 with helpful fallback message
- [ ] Extension with `com.docker.compose.service` label → rejected at install time

## Review
- CG: ⚠️ APPROVED WITH WARNINGS (D must merge first — operational dependency)
- Security: PASS (label spoofing covers dict+list formats)

## Platform Impact
- All platforms supported. WSL2 (no host agent) gets helpful 503 fallback

## Sequence
PR 5 of 7 (Phase 3). Depends on PR 4 (#806 — host agent logs endpoint)

### DIFF
--- a/dream-server/extensions/services/dashboard-api/routers/extensions.py
+++ b/dream-server/extensions/services/dashboard-api/routers/extensions.py
@@ -490,7 +490,7 @@ async def extension_logs(
             detail = err_body.get("error", f"Host agent error: HTTP {exc.code}")
         except (json.JSONDecodeError, OSError):
             detail = f"Host agent returned HTTP {exc.code}"
-        raise HTTPException(status_code=exc.code, detail=detail)
+        raise HTTPException(status_code=502, detail=detail)
     except (urllib.error.URLError, OSError):
         raise HTTPException(
             status_code=503,


yasinBursali:
## Addressing review feedback

**Bug 1 (Raw status codes forwarded) — Fixed:**
Changed `status_code=exc.code` to `status_code=502` in the `HTTPError` handler. All host-agent errors now return 502 (Bad Gateway) to the dashboard client, preventing:
- 401/403 from host agent confusing the dashboard into thinking its own auth failed
- 500 from host agent leaking internal paths or Docker error details

Error detail string is still forwarded (sanitized via `err_body.get("error", ...)`) for debugging, but the status code is always 502.

**Re: XSS via Docker log output** — The dashboard frontend renders logs using text content (React's default JSX escaping), not `innerHTML` or `dangerouslySetInnerHTML`. No XSS vector.

**Re: `container_name` in SERVICES dict** — This field is consumed by PR #810 (`resources.py`) for the container-to-service reverse mapping. Required merge order: #806 → #808 → #810.

Lightheartdevs (approved):
**Re-audit: APPROVE — fix verified**

HTTPError handler now returns 502 (Bad Gateway) instead of forwarding raw host-agent status codes. Prevents internal error leakage and auth confusion. Label spoofing prevention is solid.

CI all green. Depends on #806.

PR #716, opened 2026-04-01
fix(extensions-library): add sensible defaults for required env vars

Arifuzzamanjoy:
8 compose files were failing validate-compose.sh because they required env vars to be set before compose config could run. Added sensible defaults so services can start out of the box.

**Changes:**
- anythingllm: JWT_SECRET, AUTH_TOKEN
- flowise: USERNAME, PASSWORD (empty = no auth for local dev)
- frigate: RTSP_PASSWORD
- jupyter: TOKEN
- librechat: CREDS_KEY, CREDS_IV, JWT secrets, MONGO_PASSWORD, MEILI_KEY
- open-interpreter: API_KEY (optional)
- paperless-ngx: SECRET_KEY
- weaviate: API_KEY

All defaults are obvious placeholders meant for local dev. Users should override them for production.

PR #815, opened 2026-04-04
fix(extensions-library): add default values to manifest env_vars

Arifuzzamanjoy:
5 services had optional env_vars without default values in their manifest.yaml files. This makes the CLI and dashboard unable to determine sensible defaults for these settings.

Changes:
- ollama: OLLAMA_MODEL defaults to 'llama3'
- piper-audio: PIPER_VOICE defaults to 'en_US-lessac-medium'
- dify: 4 env vars now have explicit defaults
- librechat: CREDS_KEY/CREDS_IV default to empty (setup.sh generates)
- aider: API keys default to empty (use local LLM)

Note: This complements PR #716 which adds defaults to compose.yaml files for Docker validation. This PR adds defaults to manifest.yaml for CLI/dashboard discovery.


Lightheartdevs (changes requested):
**Audit: REQUEST CHANGES — sk-dreamserver default**

**Blocker: `DIFY_OPENAI_API_KEY` default is `sk-dreamserver`**
- The `sk-` prefix mimics real OpenAI API key format
- This field is marked `secret: true` in the manifest
- Will trigger gitleaks and other secret scanners
- Could be sent as an auth header to external endpoints if Dify is misconfigured
- The previous implied default was `dummy-key` (from the description text)

**Fix:** Change to `dummy-key` or `local-only-placeholder`.

**Everything else looks good:**
- `OPENAI_API_KEY: ""` and `ANTHROPIC_API_KEY: ""` for aider — correct for local LLM usage ✓
- `DIFY_EXTERNAL_URL: "http://localhost:8002"` — sensible local default ✓
- `DIFY_OPENAI_API_BASE: "http://llama-server:8080/v1"` — correct Docker internal URL ✓
- `CREDS_KEY: ""` and `CREDS_IV: ""` for librechat — empty, setup.sh generates them ✓
- `OLLAMA_MODEL: "llama3"` — sensible default ✓
- `PIPER_VOICE: "en_US-lessac-medium"` — sensible default ✓
- These are in `resources/dev/extensions-library/`, not deployed manifests, so existing installs are unaffected ✓

PR #818, opened 2026-04-05
fix(windows): avoid CPU fallback on NVIDIA GPU passthrough smoke test

gabsprogrammer:
Two bugs caused Windows users with NVIDIA GPUs to run inference on CPU instead of GPU after a fresh install:

**Bug 1 — Smoke test pulls a 150 MB image on cold install (05-docker.ps1)**

The installer verified GPU passthrough by running:
  docker run --rm --gpus all nvidia/cuda:12.0-base-ubuntu22.04 nvidia-smi

On a fresh Docker Desktop install, this image is not cached and must be downloaded (~150 MB). On slow connections or when Docker Hub rate-limits, this pull fails silently in under a second, setting gpuPassthroughFailed=true and permanently downgrading the stack to docker-compose.cpu.yml with --n-gpu-layers 0 (pure CPU inference).

Fix: Replace the image-pull approach with a two-stage zero-download check:
  1. Inspect \docker info --format '{{json .Runtimes}}'\ for the nvidia/cdi runtime entry -- zero network I/O, instant result.
  2. If the runtime is present, attempt \docker run --gpus all --pull never hello-world\ using only locally-cached images.

This means GPU passthrough is confirmed without any image download.

**Bug 2 — Hardcoded cpus: '16.0' crashes on machines with fewer cores (docker-compose.nvidia.yml)**

The NVIDIA overlay set a hard CPU limit of 16.0 for llama-server. On machines with fewer than 16 logical CPUs (e.g. a 12-thread i5), \docker compose up\ fails immediately:
  Error response from daemon: Range of CPUs is from 0.01 to 12.00, as
  there are only 12 CPUs available.

Fix: Replace the hardcoded value with \\\, matching the env-var pattern already used in docker-compose.cpu.yml. Users who want more threads can set LLAMA_CPU_LIMIT in their .env. Also lowered the default memory cap from 64G to 16G (cpu.yml default is 6G; 16G is a safe NVIDIA entry-tier default that does not OOM smaller systems).

Reproducer (both bugs hit simultaneously on a 12-thread machine with RTX 4060):
  1. Fresh Windows install, Docker Hub slow / rate-limited
  2. Installer silently falls back to CPU (Bug 1)
  3. Manual re-run with nvidia overlay crashes on cpus limit (Bug 2) Result: user stuck on CPU inference with no clear error message

Fixes: #<issue>

### DIFF
--- a/dream-server/docker-compose.base.yml
+++ b/dream-server/docker-compose.base.yml
@@ -22,6 +22,7 @@ services:
   # Serves OpenAI-compatible API on port 8080
   # ============================================
   llama-server:
+    image: ${LLAMA_SERVER_IMAGE:-ghcr.io/ggml-org/llama.cpp:server}
     container_name: dream-llama-server
     restart: unless-stopped
     volumes:
--- a/dream-server/docker-compose.nvidia.yml
+++ b/dream-server/docker-compose.nvidia.yml
@@ -14,8 +14,8 @@ services:
               count: all
               capabilities: [gpu]
         limits:
-          cpus: '16.0'
-          memory: ${LLAMA_SERVER_MEMORY_LIMIT:-64G}
+          cpus: '${LLAMA_CPU_LIMIT:-8.0}'
+          memory: ${LLAMA_SERVER_MEMORY_LIMIT:-16G}
 
   dashboard-api:
     deploy:
--- a/dream-server/installers/windows/phases/05-docker.ps1
+++ b/dream-server/installers/windows/phases/05-docker.ps1
@@ -97,44 +97,53 @@ if ($dryRun) {
     }
 
     # ── NVIDIA GPU passthrough smoke test ─────────────────────────────────────
-    # Only run if NVIDIA GPU detected AND WSL2 backend is confirmed.
-    # This test starts a minimal container with --gpus all and checks that
-    # nvidia-smi is accessible. If it fails, Phase 08 falls back to CPU-only
-    # inference (docker-compose.cpu.yml) instead of crashing docker compose up.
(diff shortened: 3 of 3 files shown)


gabsprogrammer:
Done <3

Lightheartdevs (changes requested):
**Audit: REQUEST CHANGES — auto-remediation removal breaks one-command install**

The GPU detection improvements are solid — the three-state model (CONFIRMED/INCONCLUSIVE/FAILED) is well-designed, eliminating the 150MB CUDA image download that caused false CPU fallbacks. The CPU limit fix (`cpus: '16.0'` → `${LLAMA_CPU_LIMIT:-8.0}`) and memory default change (64G → 16G) are both correct and needed. The `try/finally` error handling is strictly better than the old save/restore pattern.

**However, the auto-installation of the NVIDIA Container Toolkit is completely removed.**

DreamServer's core selling point is one-command deployment. The old code auto-installed the Container Toolkit when it was missing — it was fragile (assumed Debian WSL, ran `sudo apt-get`, could prompt for password), but it kept the "one install" promise. The new code removes that fallback entirely. A fresh Windows user with an NVIDIA GPU but no Container Toolkit now gets FAILED status and has to install it manually.

This is a UX regression for first-time users on Windows.

**What I'd like to see:**
- Keep the improved three-state detection ✓
- Keep the CPU limit and memory fixes ✓
- **Restore or replace the auto-install fallback** for the Container Toolkit. If the old Debian-WSL approach was too fragile, replace it with something better — a clearer guided step within the installer flow, or a more robust detection of the WSL distro before running apt-get, or a fallback that at least attempts the install before giving up.
- The installer should exhaust automatic options before asking the user to do something manually.

PR #823, opened 2026-04-06
feat(polish dashboard): frontend layout and status flows

gabsprogrammer:
## Summary
- builds on top of the accessible splash-screen branch so the full frontend refresh can be reviewed together
- redesigns the dashboard surface with a denser, more professional information layout while preserving existing routes and data sources
- keeps the splash-screen accessibility and session behavior from the current frontend base intact

## What Changed
- tightens the main dashboard layout and applies a darker market-style shell to the primary content area
- refreshes the sidebar with stronger DreamServer typography, refined quick links, and a cleaner collapsed state
- reorganizes feature

### DIFF
--- a/dream-server/installers/lib/logging.sh
+++ b/dream-server/installers/lib/logging.sh
@@ -13,7 +13,8 @@
 # ============================================================================
 
 install_elapsed() {
-  local secs=$(( $(date +%s) - INSTALL_START_EPOCH ))
+  local now_epoch="${INSTALL_NOW_EPOCH:-$(date +%s)}"
+  local secs=$(( now_epoch - INSTALL_START_EPOCH ))
   local m=$(( secs / 60 ))
   local s=$(( secs % 60 ))
   printf '%dm %02ds' "$m" "$s"
--- a/dream-server/installers/macos/lib/ui.sh
+++ b/dream-server/installers/macos/lib/ui.sh
@@ -12,7 +12,8 @@ DIVIDER="───────────────────────�
 
 # Elapsed time since install start
 install_elapsed() {
-    local secs=$(( $(date +%s) - INSTALL_START_EPOCH ))
+    local now_epoch="${INSTALL_NOW_EPOCH:-$(date +%s)}"
+    local secs=$(( now_epoch - INSTALL_START_EPOCH ))
     local m=$(( secs / 60 ))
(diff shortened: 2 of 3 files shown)


gabsprogrammer:
> **udit: REQUEST CHANGES — don't lose monitoring data**
> 
> The visual upgrade is great — grouped service layout, token signal charts, liquid metal design system, collapsible sidebar, and the splash screen now has proper accessibility, prefers-reduced-motion support, session persistence, and skip mechanism. The Google Fonts CDN issue from #819 is properly fixed. Clean PR from a security standpoint — no external resources, no XSS vectors, no data leaks.
> 
> **However, this removes monitoring data that users need:**
> 
> * **Mem Free** metric card is removed — users need this to know if they can load a bigger model or if the system is under memory pressure during inference
> * **CPU Temp** metric card is removed — matters for sustained inference on consumer hardware (NUCs, gaming PCs, laptops) where thermal throttling is a real concern
> 
> For a self-hosted AI stack where users are technical and actively monitoring their system, losing these readings is a functional downgrade even though the UI looks better.
> 
> **Request:** Restore Mem Free and CPU Temp into the new layout. They don't need to be the old-style metric cards — they'd fit naturally into the system status panel or as compact readings alongside the existing metrics. The goal is: keep all the visual improvements, but don't lose data that was previously visible.
> 
> **One other note:** `package-lock.json` has 77 packages resolved from `registry.npmmirror.com` instead of `registry.npmjs.org`. Integrity hashes are present so no security risk, but recommend regenerating the lockfile from the official npm registry for supply chain hygiene.

Thanks for the review. I addressed both points in the latest update to this PR.

Resolved:
- Restored `Mem Free` in the new `System Status` layout as a compact reading inside the `Inference telemetry` panel.
- Restored `CPU Temp` in the same panel alongside `GPU Temp`, so the monitoring data remains visible without reverting to the old card layout.
- Kept the visual refresh intact while bringing back the previously visible monitoring signals you called out.

Lockfile:
- Regenerated/normalized the dashboard lockfile so resolved package URLs now point to the official `registry.npmjs.org` instead of `registry.npmmirror.com`.

Validation:
- `npm run build` passed
- `vitest` passed (`32` tests)
- `eslint` still reports the same pre-existing warnings outside the scope of this change

PR #848, opened 2026-04-08
fix: add log rotation to all extension compose files

yasinBursali:
## What
Add explicit `logging:` sections (json-file driver, max-size 10m, max-file 3) to all extension service definitions across 21 compose files.

## Why
Core services in `docker-compose.base.yml` use the `x-logging` YAML anchor for log rotation. Extension compose files are separate YAML documents and cannot reference base anchors. All extension containers ran with Docker's default unlimited logging, risking disk exhaustion on long-running installs.

## How
Added identical logging block to every service definition in 21 files:
- 19 extension compose files (ape, comfyui GPU overlays, embeddings, litellm, n8n, openclaw, perplexica, privacy-shield, qdrant, searxng, token-spy, tts, whisper + GPU overlays)
- langfuse disabled file (8 sub-services)
- dreamforge disabled file

Skipped: `comfyui/compose.yaml` (stub), `.local.yaml` dev files, core services (already in base.yml)

## Testing
- All 21 YAML files validated
- Live verified: all 12 running containers confirmed with `{json-file map[max-file:3 max-size:10m]}`

## Platform Impact
- **All platforms:** Fixed — 30MB max log per service (10m × 3 files)

### DIFF
--- a/dream-server/extensions/services/ape/compose.yaml
+++ b/dream-server/extensions/services/ape/compose.yaml
@@ -32,3 +32,8 @@ services:
       timeout: 5s
       retries: 3
       start_period: 10s
+    logging:
+      driver: "json-file"
+      options:
+        max-size: "10m"
+        max-file: "3"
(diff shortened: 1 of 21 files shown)


Lightheartdevs (approved):
**APPROVE.** Rebased cleanly, CI all green (23 passing). Log rotation config is correct — json-file driver, 10MB max-size, 3 files = 30MB cap per service across all 21 extension compose files.

PR #874, opened 2026-04-09
fix: use $DOCKER_CMD instead of bare docker in installer phases 11-13

yasinBursali:
## What
Replace 5 bare `docker` calls with `$DOCKER_CMD` in installer phases 11–13.

## Why
Phase 05 sets `DOCKER_CMD="sudo docker"` when the user is not in the docker group. Five locations in later phases bypass this by calling `docker` directly, causing "permission denied" on the Docker socket during container recovery (phase 11) and health checks (phases 12/13).

Reported by a user on a fresh Ubuntu 24.04 install (Dell Pro Max GB10) where the installer correctly detected `sudo docker` was needed but later phases ignored it.

## How
Substitute `docker` → `$DOCKER_CMD` in all 5 call sites:

- `11-services.sh:306` — `docker image inspect` (DreamForge image check)
- `11-services.sh:349` — `docker start $(docker ps -a ...)` (container recovery step 1)
- `11-services.sh:353` — `docker start $(docker ps -a ...)` (container recovery step 3)
- `12-health.sh:89` — `docker inspect` (Perplexica auto-config gate)
- `13-summary.sh:291` — `docker inspect` (Perplexica post-install validation)

The inner `$()` subshells on lines 349/353 also get `2>/dev/null` to suppress stderr from the subshell when permissions fail.

## Testing
- `bash -n`: all 3 modified files pass syntax check
- `shellcheck`: no new warnings introduced
- `$DOCKER_CMD` scope verified: Phase 05 defines it before Phases 11/12/13
- Manual: requires Linux VM where user is NOT in docker group

## Review
Critique Guardian: ✅ APPROVED (all five pillars clean)

## Platform Impact
- **Linux:** Fix applies — resolves permission denied for non-docker-group users
- **WSL2:** Fix applies — same code path as Linux
- **macOS:** Not affected — uses separate installer (`installers/macos/`)

Closes #871

### DIFF
--- a/dream-server/installers/phases/11-services.sh
+++ b/dream-server/installers/phases/11-services.sh
@@ -303,7 +303,7 @@ MODELS_INI_EOF
     [[ "$ENABLE_COMFYUI" == "true" ]] && _build_services+=(comfyui)
     if [[ "${ENABLE_DREAMFORGE:-}" == "true" ]]; then
         _dreamforge_image="${DREAMFORGE_IMAGE:-ghcr.io/light-heart-labs/dreamforge:latest}"
-        if ! docker image inspect "$_dreamforge_image" &>/dev/null; then
+        if ! $DOCKER_CMD image inspect "$_dreamforge_image" &>/dev/null; then
             _build_services+=(dreamforge)
         else
             log "DreamForge image found locally — skipping source build"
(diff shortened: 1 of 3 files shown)


Lightheartdevs:
Closing — superseded by #882 which applies the same $DOCKER_CMD fix to phases 11-13, rebased on top of the Phase 11 retry loop changes from #878. Thanks for identifying the issue, Yasin.

PR #883, opened 2026-04-10
fix(models): platform-aware activation + real download progress + cancel

yasinBursali:
## ⚠️ Merge Order — READ FIRST

This PR **targets `main`** but its code depends on PR #870 (`feat/model-library-actions`) which is currently stranded on the `feat/model-library` branch (not yet merged to main).

**Required merge order:**
1. **Merge `feat/model-library` → `main` first** (contains the `_handle_model_*` host agent handlers from PR #870 + the activation lock from commit a8e9727)
2. **Then merge this PR** — it applies cleanly on top

Until step 1 is done, this PR will show conflicts / missing context on `main`. That is expected and will resolve once `feat/model-library` lands.

Rebased on `upstream/feat/model-library` which already includes the activation concurrency lock the maintainer added at `a8e9727`. Our fix no longer duplicates `#293`.

No conflicts with PR #881 (litellm lemonade wildcard + bootstrap-upgrade.sh) — completely separate files.

## Summary
Fixes five bugs in the model management system introduced in PR #870 and adds a download cancel feature.

## What — The 5 bugs fixed

| # | Severity | Platform | Bug |
|---|----------|----------|-----|
| 1 | Critical | macOS | Activation tried Docker restart for native llama-server → false success, old model kept running |
| 2 | High | All | Download progress stuck at 0% during entire download; frontend field contract mismatch |
| 3 | High | Linux NVIDIA | Activation ignored `llama_server_image` → Gemma4 models failed to load |
| 4 | Medium | Linux IPv6 | Health check used `localhost` → resolved to `::1` → 5-minute false rollback |
| 5 | Medium | All | Used stale `.compose-flags` file → `docker stop/start` fallback didn't re-read `.env` |

Plus a **new cancel feature** — users can now abort in-progress downloads.

### Bug 1 — macOS native llama-server
`docker-compose.macos.yml` sets `replicas: 0` for llama-server because the process runs natively via Metal on the host. The activate handler had only `amd` and `else` (NVIDIA Docker) branches. On macOS, docker commands silently failed and the native process was untouched, but `.env` was already updated — leaving the install with a new config pointing at a model that was never loaded.

### Bug 2 — Download progress
Two sub-bugs:
- `_download()` ran curl as a blocking `subprocess.run()`, so the status JSON was only updated before and after the download — never during.
- Backend wrote `bytesDownloaded/bytesTotal/status:"failed"` but frontend expected `percent/speedBytesPerSec/eta/status:"error"`. Also `data.error` vs `data.message` mismatch.

### Bug 3 — LLAMA_SERVER_IMAGE
`model-library.json` Gemma4 entries have `llama_server_image: "ghcr.io/ggml-org/llama.cpp:server-cuda-b8648"` but the activation handler only wrote `GGUF_FILE/LLM_MODEL/CTX_SIZE/MAX_CONTEXT` to `.env`. Switching from Qwen to Gemma4 on NVIDIA left the container running the older Qwen-compatible llama.cpp build.

### Bug 4 — localhost → ::1
`docker-compose.base.yml` binds to `127.0.0.1:${OLLAMA_PORT:-8080}` explicitly. On IPv6-enabled Linux hosts, `localhost` resolves to `::1` first, so the health check hit a port that wasn't bound and timed out after 5 minutes, triggering an unnecessary rollback even when activation succeeded.

### Bug 5 — `.compose-flags`
The activate handler read `INSTALL_DIR/.compose-flags` which is never created by the installer. When absent, it fell back to raw `docker stop dream-llama-server` + `docker start dream-llama-server`. `docker start` restores the previous configuration — it does NOT re-read `.env`. So `.env` was updated but the container kept running the old model.

Meanwhile, `resolve_compose_flags()` already existed in the same file and was used by every other Docker operation. This PR unifies on that.

### `dream-server/bin/dream-host-agent.py`

**Activation fixes (in `_do_model_activate()`):**
- New `if gpu_backend == "apple":` branch — stops native process via PID file (SIGTERM → 10s wait → SIGKILL), re-launches via `subprocess.Popen` with Metal args, mirroring `dream-macos.sh:126-185`
- Reads `llama_server_image` from library entry; writes to `.env` when set and not apple; runs `docker compose pull llama-server` before restart if image changed
- Health check URL changed from `localhost` to `127.0.0.1`
- Removed `.compose-flags` reading + bare `docker stop/start` fallbacks; uses `resolve_compose_flags()` for all Docker paths
- Apple rollback path: stops new process, re-launches with old params

**Download fixes (in `_handle_model_download()`):**
- Replaced `subprocess.run()` with `subprocess.Popen()` + polling loop that reads `part_file.stat().st_size` every 2 seconds
- Takes the **last** `Content-Length` header from `curl -sI -L` (HuggingFace redirects — the first header is the redirect page ~1KB)
- Added `started_at` param to `_write_model_status()`

**New cancel endpoint:**
- `POST /v1/model/download/cancel` handler
- `_model_download_cancel: threading.Event()` + `_model_download_proc: subprocess.Popen`
- Cancel flag checked in retry loop and poll loop
- On cancel: kills curl, unlinks `.part` file, writes `cancelled` status
- Cancel handler captures local reference to avoid TOCTOU race

### `dream-server/extensions/services/dashboard-api/routers/models.py`
- New `POST /api/models/download/cancel` route proxying to host agent

### `dream-server/extensions/services/dashboard/src/hooks/useDownloadProgress.js`
- Handles `verifying` as downloading state (SHA256 phase)
- Handles `failed` and `cancelled` as error states
- Reads `data.error` (was `data.message`)
- Percent computed client-side with `Math.min(..., 100)` safety cap
- Removed `speedMbps`, `eta`, `speedBytesPerSec` fields (backend never sent them)
- Exposes `cancelDownload()` function

### `dream-server/extensions/services/dashboard/src/pages/Models.jsx`
- Added red Cancel button in the `DownloadProgressBar` component

## Companion hardening (from Critique Guardian review)

- **Path traversal validation** on download + activate handlers via `.resolve()` + `.is_relative_to()`, matching the existing delete handler pattern
- **PID verification** via `ps -p {pid} -o comm=` before SIGTERM on apple forward AND rollback paths (prevents PID reuse accidents), matching `bootstrap-upgrade.sh:327` pattern
- **Missing binary rollback** — when `bin/llama-server` is missing on apple, rolls back `.env`/`models.ini` before returning 500
- **AMD stop+up -d** — changed from `docker compose restart` (which doesn't re-read `.env`) to `stop` + `up -d`, matching NVIDIA path
- **curl stderr → DEVNULL** — was `PIPE` with no reader, could block on >64KB stderr output during long downloads

## Testing

**Automated (all PASS):**
- `py_compile` — clean
- `ruff check` — clean (all checks passed)
- Mechanical diff review — exactly 4 expected files modified

**Manual test plan:**
- [ ] macOS Apple Silicon: activate a different downloaded model → verify `.llama-server.pid` updated, new process running correct model
- [ ] macOS rollback: deliberately fail health check → verify old model restored and running
- [ ] Linux NVIDIA with Gemma4: activate Gemma4 model → verify `docker compose pull` runs and new image deploys
- [ ] Linux AMD: activate model → verify `docker compose stop` + `up -d` picks up new env vars
- [ ] Concurrent activation: fire two requests → verify one gets 409
- [ ] Download progress: start a download → verify bar advances past 0% within 4 seconds
- [ ] Download cancel mid-download: click Cancel around 30% → verify curl killed, `.part` removed, status shows cancelled
- [ ] Delete active model: try to delete loaded model → verify 409

## Review

Critique Guardian: **APPROVED WITH WARNINGS** (all three warnings addressed in this PR):
- W1 (TOCTOU on `_model_download_proc`) — fixed with local reference capture + `AttributeError` catch
- W2 (apple rollback ps-verify inconsistency) — fixed, rollback now matches forward path
- W3 (stderr PIPE buffering) — changed to DEVNULL

## Platform Impact

| Platform | Impact |
|----------|--------|
| **macOS** (Apple Silicon) | Activation now works. Native process managed via PID file. Rollback path validated. |
| **Linux NVIDIA** | Health check fixed (IPv6). LLAMA_SERVER_IMAGE now updates on model swap. `stop`+`up -d` re-reads .env. |
| **Linux AMD** (Lemonade) | `stop`+`up -d` instead of `restart` — now re-reads .env. Health endpoint `/api/v1/health` preserved. |
| **Windows/WSL2** | Inherits Linux Docker path. No WSL-specific changes. |

## Known Considerations

- `resolve_compose_flags()` may raise `RuntimeError` if `scripts/resolve-compose-stack.sh` is missing. Propagates to outer handler → 500 with clear message. Fail-loud is intentional.
- On apple, if the maintainer restarts the native llama-server out-of-band (via `dream-macos.sh restart llama-server`), the PID file is updated by that script. Our ps-verify will still correctly identify it.
- Download cancel has a narrow race if cancel arrives exactly as the download completes naturally — the `.part` file (already renamed to the final file) may briefly exist before being unlinked, then marked cancelled. Acceptable: user intent wins.

### DIFF
--- a/dream-server/bin/dream-host-agent.py
+++ b/dream-server/bin/dream-host-agent.py
@@ -50,6 +50,8 @@ _service_locks: dict[str, threading.Lock] = collections.defaultdict(threading.Lo
 # Model download state — only one download at a time
 _model_download_lock = threading.Lock()
 _model_download_thread: threading.Thread | None = None
+# Model activation lock — prevent concurrent .env writes and Docker restarts
+_model_activate_lock = threading.Lock()
 
 
 def load_env(env_path: Path) -> dict:
@@ -714,7 +716,7 @@ class AgentHandler(BaseHTTPRequestHandler):
                         import hashlib
                         sha = hashlib.sha256()
                         with open(target, "rb") as f:
-                            for chunk in iter(lambda: f.read(8192), b""):
+                            for chunk in iter(lambda: f.read(1048576), b""):  # 1MB chunks
                                 sha.update(chunk)
                         actual = sha.hexdigest()
                         if actual != gguf_sha256:
@@ -746,6 +748,17 @@ class AgentHandler(BaseHTTPRequestHandler):
             json_response(self, 400, {"error": "model_id is required"})
             return
 
+        if not _model_activate_lock.acquire(blocking=False):
+            json_response(self, 409, {"error": "Another model activation is in progress"})
+            return
+
+        try:
+            self._do_model_activate(model_id)
+        finally:
+            _model_activate_lock.release()
+
+    def _do_model_activate(self, model_id: str):
+        """Inner activate logic — called with _model_activate_lock held."""
         # Look up model in library
         library_path = INSTALL_DIR / "config" / "model-library.json"
         model = None
--- a/dream-server/bin/dream-host-agent.py
+++ b/dream-server/bin/dream-host-agent.py
@@ -50,6 +50,8 @@ _service_locks: dict[str, threading.Lock] = collections.defaultdict(threading.Lo
 # Model download state — only one download at a time
 _model_download_lock = threading.Lock()
 _model_download_thread: threading.Thread | None = None
+_model_download_proc: subprocess.Popen | None = None
+_model_download_cancel = threading.Event()
 # Model activation lock — prevent concurrent .env writes and Docker restarts
 _model_activate_lock = threading.Lock()
 
@@ -353,6 +355,8 @@ class AgentHandler(BaseHTTPRequestHandler):
             self._handle_service_logs()
         elif self.path == "/v1/model/download":
             self._handle_model_download()
+        elif self.path == "/v1/model/download/cancel":
+            self._handle_model_download_cancel()
         elif self.path == "/v1/model/activate":
             self._handle_model_activate()
         elif self.path == "/v1/model/delete":
@@ -654,7 +658,10 @@ class AgentHandler(BaseHTTPRequestHandler):
             return
 
         models_dir = INSTALL_DIR / "data" / "models"
-        target = models_dir / gguf_file
+        target = (models_dir / gguf_file).resolve()
+        if not target.is_relative_to(models_dir.resolve()):
+            json_response(self, 400, {"error": "Invalid file path"})
+            return
         if target.exists():
             json_response(self, 200, {"status": "already_downloaded"})
             return
@@ -665,15 +672,20 @@ class AgentHandler(BaseHTTPRequestHandler):
                 json_response(self, 409, {"error": "Another download is in progress"})
                 return
 
+            _model_download_cancel.clear()
+
             def _download():
+                global _model_download_proc
                 status_path = INSTALL_DIR / "data" / "model-download-status.json"
                 part_file = models_dir / f"{gguf_file}.part"
+                started_at = ""
                 try:
                     models_dir.mkdir(parents=True, exist_ok=True)
-                    # Write initial status
-                    _write_model_status(status_path, "downloading", gguf_file, 0, 0)
+                    # Write initial status with startedAt
+                    started_at = _iso_now()
+                    _write_model_status(status_path, "downloading", gguf_file, 0, 0, started_at=started_at)
 
-                    # Get total size
+                    # Get total size — take the LAST Content-Length header (HuggingFace redirects)
                     total_bytes = 0
                     try:
                         head_result = subprocess.run(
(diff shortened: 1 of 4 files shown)


yasinBursali:
Closing this draft — the base branch (feat/model-library) was superseded by PR #886 which merged the model action infrastructure directly to main with a different implementation. Opening a fresh PR on top of main that applies only the remaining unique fixes: macOS native restart, IPv6 health check, compose-flags fix, LLAMA_SERVER_IMAGE apple guard, path traversal hardening, and download cancel.

PR #893, opened 2026-04-11
fix(models): platform-aware activation + download cancel

yasinBursali:
> **Merge order:** Merge last — depends on #906, #905, #900, #908, and #902.

## Summary

Addresses five bugs in the model management system and adds a download cancel feature. Builds on the model action infrastructure already in main.

## What — Bugs fixed

| # | Severity | Platform | Bug |
|---|----------|----------|-----|
| 1 | Critical | macOS | Activation ran `_compose_restart_llama_server` which silently succeeded (replicas: 0 in docker-compose.macos.yml) while the native llama-server process kept the old model loaded |
| 2 | High | Linux NVIDIA | `LLAMA_SERVER_IMAGE` written to `.env` even on `apple` backend — meaningless write that could corrupt the env on macOS |
| 3 | High | Linux IPv6 | Health check used `localhost` → resolved to `::1` on IPv6-enabled hosts — 5-minute false rollback even when model loaded successfully |
| 4 | Medium | All | `_compose_restart_llama_server` read `.compose-flags` directly (file may not exist) and used `docker compose restart` for AMD — restart does not re-read `.env` |
| 5 | Medium | All | No path traversal protection in activate handler — inconsistent with the delete handler which already had it |

Plus: **new download cancel feature** — users can abort in-progress downloads.

### `dream-server/bin/dream-host-agent.py`

**Bug 1 — macOS native restart:**
Added `if gpu_backend == "apple":` branch in `_do_model_activate` before the `_in_container` / `_compose_restart` paths. Stops the existing native process via PID file (ps-verify to avoid PID reuse accidents → SIGTERM → 10s wait → SIGKILL), then re-launches via `subprocess.Popen` with Metal args matching `scripts/bootstrap-upgrade.sh:431-438`. Rollback path mirrors the forward path. Extracted shared `_launch_native_llama_server()` helper to avoid duplicating the ~20-line launch block in both directions.

**Bug 2 — apple guard on LLAMA_SERVER_IMAGE:**
Reads `gpu_backend` from `.env` before the update block and skips `LLAMA_SERVER_IMAGE` when `gpu_backend == "apple"`.

**Bug 3 — 127.0.0.1 vs localhost:**
Changed `llama_host = "localhost"` to `llama_host = "127.0.0.1"` on the host-native health check path. Docker binds to `127.0.0.1` explicitly; on IPv6-enabled Linux hosts, `localhost` resolves to `::1` first.

**Bug 4 — `_compose_restart_llama_server` rework:**
- Replaced inline `.compose-flags` read with `resolve_compose_flags()`, which already falls back to running `resolve-compose-stack.sh` when the cache file is absent
- Changed all Docker restart paths to `stop + up -d` (was: `restart` for AMD, `docker start` for no-compose-flags NVIDIA) — `up -d` re-reads `.env`; `restart` and `start` do not
- Named volumes (`lemonade-cache`, `lemonade-llama`, `lemonade-recipe`) survive `stop + up -d` so there is no Lemonade binary re-cache penalty

**Bug 5 — path traversal:**
Added `.resolve() + .is_relative_to()` check on the GGUF target path in `_do_model_activate`, matching the existing delete handler pattern.

**Cancel feature:**
- Added `_model_download_proc: subprocess.Popen | None` and `_model_download_cancel: threading.Event` globals
- Switched `subprocess.run()` to `subprocess.Popen()` + `proc.wait()` in the download loop so the process can be killed from outside the thread
- The `_poll_progress` thread checks `_model_download_cancel` and kills the active curl proc
- `_model_download_cancel.wait(5)` replaces `time.sleep(5)` in retry delay so cancel is immediate
- Added `POST /v1/model/download/cancel` host agent endpoint
- Cancel handler captures a local reference to `_model_download_proc` to avoid TOCTOU race

### `dream-server/extensions/services/dashboard-api/routers/models.py`
- Added `POST /api/models/download/cancel` proxying to host agent via existing `_call_agent_model` helper

### `dream-server/extensions/services/dashboard/src/hooks/useDownloadProgress.js`
- Added `cancelDownload()` async function exposed from the hook
- Handles `cancelled` status (alongside existing `failed`/`error`)
- Logs cancel errors with `console.error` instead of swallowing silently

### `dream-server/extensions/services/dashboard/src/pages/Models.jsx`
- Added red Cancel button inside `DownloadProgressBar` component, rendered only when `helpers.cancelDownload` is available

## Testing

**Automated (all pass):**
- `py_compile` — clean
- `ruff check` — clean
- Critique Guardian — APPROVED (all CG observations addressed)

**Manual (Apple Silicon, local install):**
- Activated Qwen3.5-2B while Qwen3.5-9B was running → old PID killed, new process launched, health check passed in 1 attempt, `.llama-server.pid` updated
- Attempted activate on non-downloaded model → 400, running process untouched
- Cancel with no active download → `{"status": "no_download"}`

## Platform impact

| Platform | Impact |
|----------|--------|
| **macOS Apple Silicon** | Activation now works — native process managed via PID file |
| **Linux NVIDIA** | `127.0.0.1` health check (IPv6 fix), `stop+up -d` re-reads `.env` |
| **Linux AMD (Lemonade)** | `stop+up -d` re-reads `.env`; named volumes preserve cached binary |
| **Docker Desktop / WSL2** | `_recreate_llama_server` path unchanged |

### DIFF
--- a/dream-server/bin/dream-host-agent.py
+++ b/dream-server/bin/dream-host-agent.py
@@ -80,6 +80,8 @@ def _to_bash_path(path: Path) -> str:
 # Model download state — only one download at a time
 _model_download_lock = threading.Lock()
 _model_download_thread: threading.Thread | None = None
+_model_download_proc: subprocess.Popen | None = None
+_model_download_cancel = threading.Event()
 # Model activation lock — prevent concurrent .env writes and Docker restarts
 _model_activate_lock = threading.Lock()
 
@@ -597,6 +599,8 @@ class AgentHandler(BaseHTTPRequestHandler):
             self._handle_service_logs()
         elif self.path == "/v1/model/download":
             self._handle_model_download()
+        elif self.path == "/v1/model/download/cancel":
+            self._handle_model_download_cancel()
         elif self.path == "/v1/model/activate":
             self._handle_model_activate()
         elif self.path == "/v1/model/delete":
@@ -1312,7 +1316,10 @@ class AgentHandler(BaseHTTPRequestHandler):
                 json_response(self, 409, {"error": "Another download is in progress"})
                 return
 
+            _model_download_cancel.clear()
+
             def _download():
+                global _model_download_proc
                 import time as _time
                 status_path = INSTALL_DIR / "data" / "model-download-status.json"
                 try:
@@ -1321,6 +1328,8 @@ class AgentHandler(BaseHTTPRequestHandler):
                     _write_model_status(status_path, "downloading", label, 0, 0)
 
                     for part_idx, (part_file_name, part_url) in enumerate(download_plan, 1):
+                        if _model_download_cancel.is_set():
+                            break
                         part_target = models_dir / part_file_name
                         part_tmp = models_dir / f"{part_file_name}.part"
                         part_label = part_file_name if len(download_plan) == 1 else f"{part_file_name} (part {part_idx}/{len(download_plan)})"
@@ -1343,11 +1352,19 @@ class AgentHandler(BaseHTTPRequestHandler):
 
                         _write_model_status(status_path, "downloading", part_label, 0, part_total)
 
-                        # Progress polling: update status by checking .part file size
+                        # Progress polling: update status by checking .part file size.
+                        # Also kills the active curl process when cancel is requested.
                         _stop_progress = threading.Event()
 
                         def _poll_progress():
                             while not _stop_progress.is_set():
+                                if _model_download_cancel.is_set():
+                                    proc_ref = _model_download_proc
+                                    if proc_ref is not None:
+                                        try:
+                                            proc_ref.kill()
+                                        except (OSError, AttributeError):
+                                            pass
                                 try:
                                     if part_tmp.exists():
                                         current = part_tmp.stat().st_size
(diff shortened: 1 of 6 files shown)


Lightheartdevs (dismissed):
Five bugs + a feature in one PR is larger than I'd normally like, but every piece is tightly coupled to model activation/download so I'll let it stand. Each bug is legitimate:

**Bug 1 (Critical, macOS):** the silent-success activation bug is particularly nasty — `_compose_restart_llama_server` "succeeded" because replicas:0 in `docker-compose.macos.yml` made the compose call a no-op, while the native llama-server process kept the old model loaded. User sees "success," asks a question, gets the old model's response. Worst kind of bug.

**Fix is correct:** the `if gpu_backend == "apple":` branch before the container/compose paths stops the native process via PID file (with ps-verify to avoid PID reuse — good paranoia), launches fresh via `subprocess.Popen` with Metal args matching `bootstrap-upgrade.sh:431-438`. Shared `_launch_native_llama_server()` helper is the right extraction.

**Bug 2:** writing `LLAMA_SERVER_IMAGE` to `.env` on the apple backend is meaningless and could corrupt the env. Skip is correct.

**Bug 3:** `localhost` → `127.0.0.1` on the host-native health check. Consistent with #977 (dreamforge/perplexica healthcheck) and #975 (native llama-server binds). Good.

**Bug 4:** `stop + up -d` over `restart` — same pattern as #935. `restart` doesn't re-read `.env`; `up -d` does. The named-volume analysis (lemonade-cache, lemonade-llama, lemonade-recipe survive) correctly addresses the "does this nuke the cached binary" concern.

**Bug 5:** path traversal protection on activate matches the existing delete handler pattern — good consistency.

**Cancel feature:** the `subprocess.Popen` + `threading.Event` + TOCTOU-safe local-ref pattern is correct. `_model_download_cancel.wait(5)` replacing `time.sleep(5)` for fast cancel is a nice touch. Dashboard UI button is scoped correctly (only renders when `helpers.cancelDownload` is available).

**Merge order (per author):** last, after #906, #905, #900, #908, #902. That's a 5-deep dependency chain — please bundle these into a merge train or the leaf PRs will keep needing rebases. Ship.

Lightheartdevs:
Cross-PR coordination note from a deeper re-read.

The new `_launch_native_llama_server` helper binds to `--host 0.0.0.0` — this **preserves** the existing behavior in main (both `bootstrap-upgrade.sh` and `install-macos.sh` currently bind to `0.0.0.0` on macOS native launches), so it's not a regression vs the current state.

**However, it conflicts with the direction of draft PR #975**, which changes all native launches (macOS installer, CLI, host agent, bootstrap-upgrade, Windows) from `--host 0.0.0.0` to `--host 127.0.0.1`. If #975's security fix is split out and lands first (as I've recommended in #975's review), this new helper will need to be updated to match — otherwise merging this PR would silently re-introduce the `0.0.0.0` bind for the macOS activation path that #975 just removed.

Options:
1. **If this PR merges first**: #975 needs to add `_launch_native_llama_server` to its list of native-launch fixes (currently it only covers the 7 call sites that existed when #975 was branched).
2. **If #975 merges first**: update this PR to use `127.0.0.1` in the new helper.

Either works; just needs coordination between you and yasin so one doesn't silently undo the other. Not blocking — still approving.

PR #895, opened 2026-04-11
fix(dashboard): use theme token for info tooltip background

yasinBursali:
## What
`Dashboard.jsx:518` swaps a hardcoded `bg-[#0d0b12]/95` literal on the FeatureCard info tooltip for the `bg-theme-card/95` theme token.

## Why
A prior commit migrated `Extensions.jsx` to theme tokens but was scoped exclusively to that file. `Dashboard.jsx` retained one residual hardcoded literal — the info tooltip — which rendered as a near-black rectangle against the page background on every non-dark theme (`light`, `lemonade`, `arctic`).

## How
One-line substitution. Matches the migration pattern applied earlier to the equivalent `Extensions.jsx` tooltip.

## Testing
- `grep '#0d0b12' dream-server/extensions/services/dashboard/src/` → empty (no residuals)
- `grep 'bg-theme-card/95' Dashboard.jsx` → single match at L518

## Platform Impact
- **macOS / Linux / Windows:** Identical. Static JSX class string. Users on `light`, `lemonade`, or `arctic` themes now see the info tooltip blend with the page surface instead of rendering as a dark rectangle.

### DIFF
--- a/dream-server/extensions/services/dashboard/src/pages/Dashboard.jsx
+++ b/dream-server/extensions/services/dashboard/src/pages/Dashboard.jsx
@@ -515,7 +515,7 @@ const FeatureCard = memo(function FeatureCard({ icon: Icon, title, description, 
           <CircleHelp size={13} />
         </div>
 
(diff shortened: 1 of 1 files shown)


yasinBursali:
Replaced by combined PR

PR #898, opened 2026-04-11
fix(bootstrap-upgrade): filename arg, docker perms, and JSON escape

yasinBursali:
## What
Harden `bootstrap-upgrade.sh` Phase 4b and the Lemonade warm-up request path.

## Why
Three bugs in the Linux bootstrap-upgrade flow:

1. Phase 4b hardcoded `BOOTSTRAP_GGUF="Qwen3.5-2B-Q4_K_M.gguf"` instead of
   reading the authoritative name from `installers/lib/bootstrap-model.sh`.
   Any rename of the bootstrap model would leave stale files behind that
   Lemonade's `--extra-models-dir` could pick up instead of the full model.
2. The script ran bare `docker` and `docker compose` commands, so any user
   not in the docker group (fresh install, no relog) hit a silent failure
   during model hot-swap — the bootstrap stays resident and the upgrade
   status panel never updates.
3. The Lemonade warm-up curl interpolated `$FULL_GGUF_FILE` straight into
   the JSON body. A filename containing `"` produced malformed JSON — a
   latent bug that would silently corrupt any future catalog entry with a
   quote in its name.

## How
- Accept the bootstrap GGUF filename as a 7th positional argument
  (`${7:-Qwen3.5-2B-Q4_K_M.gguf}`). Phase 11 passes `$BOOTSTRAP_GGUF_FILE`
  which it already sources from `bootstrap-model.sh`, so the two stay in
  sync without duplication.
- Add a self-contained Docker permission detection block: probe
  `docker info`, then `sudo -n docker info`, then fail loudly with a
  diagnostic pointing at `sudo`/docker-group remediation. Mirrors the
  pattern Phase 05 uses. Replaces seven bare `docker` / `docker compose`
  call sites with `$DOCKER_CMD` / `$DOCKER_COMPOSE_CMD`. The macOS
  native llama-server fallback (which checks for `.llama-server.pid`) is
  preserved — `DOCKER_CMD` stays empty and the existing macOS branch is
  taken.
- Escape embedded `"` in `FULL_GGUF_FILE` when building the warm-up JSON
  body, mirroring the existing `_safe_model` pattern at the top of the
  file.

## Testing
- `bash -n` and `shellcheck` clean on both files (zero new warnings vs
  `upstream/main` baseline).
- Manual repro of the JSON escape path against `python3 -c "import json;
  json.loads(...)"`.
- Helper-sourcing smoke test confirms `BOOTSTRAP_GGUF_FILE` propagates.

### Manual test plan
- **Linux AMD / in docker group:** `dream model swap phi4-mini-q4` → AMD
  Lemonade hot-swap path runs, bootstrap file is removed after full model
  arrives.
- **Linux / not in docker group + passwordless sudo:** same command, log
  should show "Detected docker requires sudo"; swap still succeeds.
- **Linux / not in docker group, no NOPASSWD sudo:** swap fails loudly
  with a "docker is installed but not accessible" message and writes
  `failed` to the upgrade status file (previously silent).
- **Linux NVIDIA:** stop + `up -d` recreates the container with the new
  `--model` flag.
- **macOS native (no docker):** path falls through to `.llama-server.pid`
  branch unchanged.

## Platform Impact
- **macOS:** unchanged. Script falls through its existing native-llama branch.
- **Linux (Docker):** primary target of the fix. Users outside the docker group now get a clear failure instead of a silent one, and users who are in the group still hit the happy path.
- **Windows (WSL2):** inherits Linux behavior. Same improvements.

### DIFF
--- a/dream-server/installers/phases/11-services.sh
+++ b/dream-server/installers/phases/11-services.sh
@@ -380,6 +380,7 @@ MODELS_INI_EOF
         nohup bash "$SCRIPT_DIR/scripts/bootstrap-upgrade.sh" \
             "$INSTALL_DIR" "$FULL_GGUF_FILE" "$FULL_GGUF_URL" \
             "$FULL_GGUF_SHA256" "$FULL_LLM_MODEL" "$FULL_MAX_CONTEXT" \
+            "$BOOTSTRAP_GGUF_FILE" \
             > "$INSTALL_DIR/logs/model-upgrade.log" 2>&1 &
         _upgrade_pid=$!
 
--- a/dream-server/scripts/bootstrap-upgrade.sh
+++ b/dream-server/scripts/bootstrap-upgrade.sh
@@ -9,8 +9,13 @@
 # Usage (called by phase 11, not directly by users):
 #   nohup bash bootstrap-upgrade.sh \
 #       <install_dir> <gguf_file> <gguf_url> <gguf_sha256> \
-#       <llm_model> <max_context> \
+#       <llm_model> <max_context> [<bootstrap_gguf_file>] \
 #       > logs/model-upgrade.log 2>&1 &
+#
+# Arg 7 (bootstrap_gguf_file) is optional and defaults to the historical
+# Qwen3.5-2B-Q4_K_M.gguf for backwards compatibility. Phase 11 must pass the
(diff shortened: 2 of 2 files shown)


yasinBursali:
Replaced by combined PR

PR #899, opened 2026-04-11
fix(macos-installer): dynamic launchd PATH + extensions-lib source

yasinBursali:
> **Merge order:** Merge before #940 and #920 — all three modify `install-macos.sh`.

## What
Three related macOS installer fixes:

1. Compute the launchd PATH for both the dream host-agent and OpenCode
   plists from the user's actual `docker` and `brew` prefixes instead
   of hardcoding `/opt/homebrew/bin:/usr/local/bin:/usr/bin:/bin`.
2. Surface `launchctl bootstrap` errors instead of silently redirecting
   stderr to `/dev/null`. Emit a concrete recovery hint when the
   launchd throttle EIO state is detected.
3. Correct the extensions-library copy source path so a fresh macOS
   install actually populates `data/extensions-library/`.

## Why
1. A user running DreamServer on a non-default Homebrew prefix (anything
   other than `/opt/homebrew` or `/usr/local`) or a non-default Docker
   install (e.g. Docker.app on an external volume) crashes the host
   agent at startup: `shutil.which("docker")` returns None, the agent
   `sys.exit(1)`, launchd retries until it gives up and moves the
   service to on-demand with `LastExitStatus = 19968`.
2. Because the bootstrap call redirects `stderr` to `/dev/null`, the
   installer always prints `[OK] Dream host agent installed` even when
   the service is already throttled. The user gets no feedback until
   they hit the dashboard and see a disconnected host-agent.
3. `SOURCE_ROOT` resolves to `dream-server/`, so
   `${SOURCE_ROOT}/resources/dev/extensions-library/services` points at
   a directory that does not exist. The copy is silently skipped,
   `data/extensions-library/` is never created, and the Extensions page
   returns HTTP 503. The Linux installer gets this right via
   `$SCRIPT_DIR/..`; the macOS installer was missing the extra `..`.

## How
- New helper `_compute_launchd_path()` computes a deduped, ordered PATH
  containing (optional prefix) + `dirname "$(command -v docker)"` +
  `$(brew --prefix)/bin` + the four legacy paths. Used by the OpenCode
  plist (with `${HOME}/.opencode/bin` as the prefix arg) and the dream
  host-agent plist (no prefix). `command -v` existence checks are used
  as structured booleans rather than `|| true`.
- Warn up front at install time if `docker` is not on PATH — the user
  learns about the missing binary immediately instead of debugging a
  dead host-agent later.
- Capture `launchctl bootstrap`'s stderr + exit code into a local var
  and surface it via `ai_warn`. When stderr contains `Input/output
  error` (the launchd throttle sentinel), emit a recovery hint showing
  the exact `launchctl bootout; sleep 5;` sequence. Done for both the
  host-agent and OpenCode bootstrap calls.
- `bootout` is left with `>/dev/null 2>&1 || true` — it legitimately
  errors when nothing is loaded.
- Add `..` to `_ext_lib_src` so it climbs from `dream-server/` to repo
  root, and replace the silent-skip `if [[ -d ]]` with an `else`
  branch that emits a warning pointing at what was expected vs. found.

## Testing
- `bash -n` clean.
- `shellcheck` zero-delta vs. baseline.
- Extracted both plist heredocs with substituted vars and ran
  `plutil -lint` → both OK.
- Direct invocation of `_compute_launchd_path` on a machine with Docker
  at `/Volumes/X/applications/Docker.app/...` and Homebrew at
  `/Volumes/X/homebrew` produces the expected deduped order.

### Manual test plan
- **macOS with non-default Homebrew prefix:** fresh install, inspect
  `~/Library/LaunchAgents/com.dreamserver.host-agent.plist`, confirm
  `<key>PATH</key>` contains the actual `dirname $(command -v docker)`.
  `launchctl list com.dreamserver.host-agent` should show
  `LastExitStatus = 0`, `OnDemand = false`, no throttle.
- **macOS with docker briefly off PATH:** re-run installer, expect
  `ai_warn "docker not found on PATH at install time..."`.
- **Throttle recovery:** force a bad-PATH plist, bootout/bootstrap until
  EIO, confirm the new "launchd is throttled. Recover with..." hint.
- **Extensions library:** after install,
  `ls $INSTALL_DIR/data/extensions-library/` should contain aider,
  anythingllm, audiocraft, baserow, ... etc.

## Platform Impact
- **macOS:** primary target of all three fixes.
- **Linux:** untouched. `installers/phases/06-directories.sh:109` is already correct.
- **Windows (WSL2):** untouched. Uses the Linux installer path.

### DIFF
--- a/dream-server/installers/macos/install-macos.sh
+++ b/dream-server/installers/macos/install-macos.sh
@@ -863,8 +863,12 @@ OPENCODE_EOF
             ai_ok "OpenCode config already exists"
         fi
 
-        # Install as macOS LaunchAgent (auto-start on login)
-        mkdir -p "$HOME/Library/LaunchAgents"
+        # Install as macOS LaunchAgent (auto-start on login).
+        # Log path is intentionally decoupled from INSTALL_DIR: xpcproxy denies
+        # file-write-create on non-$HOME volumes, which causes the launchd spawn
+        # to exit 78 before the target process ever runs. $HOME/Library/Logs is
+        # always inside xpcproxy's sandbox writable set, so use that instead.
+        mkdir -p "$HOME/Library/LaunchAgents" "$HOME/Library/Logs/DreamServer"
         OPENCODE_LAUNCHD_PATH="$(_compute_launchd_path "${HOME}/.opencode/bin")"
         cat > "$OPENCODE_PLIST" <<PLIST_EOF
 <?xml version="1.0" encoding="UTF-8"?>
@@ -899,9 +903,9 @@ OPENCODE_EOF
         <false/>
     </dict>
     <key>StandardOutPath</key>
-    <string>${INSTALL_DIR}/data/opencode-web.log</string>
+    <string>${HOME}/Library/Logs/DreamServer/opencode-web.log</string>
     <key>StandardErrorPath</key>
-    <string>${INSTALL_DIR}/data/opencode-web.log</string>
+    <string>${HOME}/Library/Logs/DreamServer/opencode-web.log</string>
 </dict>
 </plist>
 PLIST_EOF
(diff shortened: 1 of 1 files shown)


yasinBursali:
Added a second commit that redirects the launchd `StandardOutPath` / `StandardErrorPath` of both `com.dreamserver.opencode-web` and `com.dreamserver.host-agent` to `$HOME/Library/Logs/DreamServer/` instead of `${INSTALL_DIR}/data/`.

The previous path triggered `xpcproxy` sandbox denial when `INSTALL_DIR` lives outside `$HOME` (non-`$HOME` APFS volume, external drive, etc.), causing the launchd spawn to exit 78 before the target process ever ran. Verified with a scratch plist:

- **Buggy path** → `state = not running, runs = 1, last exit code = 78: EX_CONFIG`, no log file, kernel log shows `Sandbox ... deny(1) file-write-create /Volumes/.../data/...log`.
- **Fixed path** → target process spawned cleanly, its own stdout hits the `$HOME`-rooted log file.

`WorkingDirectory` stays on `INSTALL_DIR` — `chdir` is not sandbox-gated, only log-file creation. The log dir is `mkdir -p`'d alongside the existing `LaunchAgents` mkdir in both plist blocks.

Validation on the new commit: `bash -n` PASS, `make lint` PASS, `make test` PASS (113 tests), `shellcheck` clean (one preexisting info-level SC2153 on the `OLLAMA_PID` line that's already on `main`), `pre-commit` hooks PASS. Diff is 13 insertions / 7 deletions in `installers/macos/install-macos.sh`, no other files touched.

PR #900, opened 2026-04-11
fix(models): populate SHA256 catalog + verify every download part

yasinBursali:
> **Merge order:** Merge after #906 and #905 — touches `dream-host-agent.py` and `model-library.json`.

## What
Populate missing SHA256 checksums in the model catalog and harden the
host-agent download verifier to check every part of every model.

## Why
`config/model-library.json` shipped with `gguf_sha256` populated for
only 5 of 19 entries. For the other 14 single-file models, the host-
agent's verifier hit `if gguf_sha256 and len(download_plan) == 1:` and
silently skipped the check. Split-file models (two of them) also hit
that gate unconditionally because `len(download_plan) > 1`, so every
multi-part download was never verified. Result: a corrupted or
tampered GGUF could slip through and only surface as a runtime crash
in llama-server, with no signal to the user or operator.

The host-agent also read `gguf_sha256` from the client request body
and used the client-supplied value as ground truth. That added an
unnecessary trust boundary — a caller supplying their own matching-but-
bogus hash could fake a "verified" download.

## How
- Populate `gguf_sha256` for all 12 single-file models that shipped
  with an empty string. For the 2 split-file models, add a `sha256`
  field to each entry in `gguf_parts`. Top-level `gguf_sha256` on
  split-file models is intentionally left empty — for those, the
  per-part sha is authoritative. All SHAs were fetched from
  HuggingFace LFS `x-linked-etag` response headers, cross-validated
  against the 5 pre-existing catalog SHAs as a sanity check.
- In `dream-host-agent.py`, drop the `len(download_plan) == 1` gate.
  Build an `expected_sha_by_file: dict[str, str]` map inside the
  existing library-validation block (keyed by filename, single-file
  and split-file both supported) and iterate over every entry in
  `download_plan`. For each part:
  - Empty catalog sha → `logger.warning(...)` and continue (no silent
    skip; the warning drives the operator to populate the catalog).
  - Non-empty sha → stream-hash in 1 MiB chunks and compare. On
    mismatch, `unlink` the bad part, write `failed` status, return.
- Drop the client-supplied `gguf_sha256` body field. The catalog is
  the sole trust root. Dashboard-api continues to accept the field in
  the request body for backwards compatibility but the host-agent
  ignores it.
- Verifier status label distinguishes `"foo.gguf (part 2/3)"` so the
  dashboard can show meaningful per-part progress during multi-part
  verification.

## Testing
- `python3 -m py_compile dream-host-agent.py`, `ruff check` clean
  (the body-field drop also fixed a pre-existing F841 dead-variable).
- `python3 -c "import json; json.load(...)"` confirms the catalog
  stays well-formed; 19 models preserved; key order and 4-space
  indentation untouched.
- Existing dashboard-api test suite green (22 host_agent tests, 13
  models tests).
- Chunked-hash path validated against a 2.3 MB synthetic file with a
  known SHA, plus a deliberate mismatch confirming the failure path.

### Manual test plan
- **Any platform:** trigger a single-file model download, watch
  host-agent log for `verifying` status, expect `complete` with no
  warning.
- **Split-file:** trigger llama4-scout-q4 or qwen3.5-122b-a10b-q4,
  status should pass through `verifying ... (part N/M)` for each
  part.
- **Tamper test:** temporarily edit the catalog to a bogus sha for one
  entry, retry, expect the download to land in `failed` with a clear
  mismatch message in the status file.
- **Empty-sha warning:** blank one entry's sha, retry, expect a
  `WARNING: SHA256 verification skipped for ...` in stderr but still
  `complete`.

## Platform Impact
- **macOS / Linux / Windows:** identical. Pure Python stdlib (`hashlib`, `pathlib`, `json`), no platform-specific behavior.

### DIFF
--- a/dream-server/bin/dream-host-agent.py
+++ b/dream-server/bin/dream-host-agent.py
@@ -1063,7 +1063,6 @@ class AgentHandler(BaseHTTPRequestHandler):
 
         gguf_file = body.get("gguf_file", "")
         gguf_url = body.get("gguf_url", "")
-        gguf_sha256 = body.get("gguf_sha256", "")
         gguf_parts = body.get("gguf_parts", [])
 
         if not gguf_file or (not gguf_url and not gguf_parts):
@@ -1079,9 +1078,12 @@ class AgentHandler(BaseHTTPRequestHandler):
         else:
             download_plan = [(gguf_file, gguf_url)]
 
-        # Validate against library (prevent arbitrary URL downloads)
+        # Validate against library (prevent arbitrary URL downloads).
+        # Also harvest expected SHA256s keyed by filename so verification can
+        # cover every part of split-file downloads, not just single-file models.
         library_path = INSTALL_DIR / "config" / "model-library.json"
         allowed = False
+        expected_sha_by_file: dict = {}
         if library_path.exists():
             try:
                 lib = json.loads(library_path.read_text(encoding="utf-8"))
@@ -1090,12 +1092,21 @@ class AgentHandler(BaseHTTPRequestHandler):
                         continue
                     if gguf_parts:
                         # Verify every (file, url) in the request matches the library
-                        lib_parts = {(p["file"], p["url"]) for p in m.get("gguf_parts", [])}
+                        lib_parts_meta = {
+                            (p["file"], p["url"]): p.get("sha256", "")
+                            for p in m.get("gguf_parts", [])
+                            if p.get("file") and p.get("url")
+                        }
                         req_parts = set(download_plan)
-                        if req_parts and req_parts <= lib_parts:
+                        if req_parts and req_parts <= set(lib_parts_meta.keys()):
                             allowed = True
+                            expected_sha_by_file = {
+                                file: lib_parts_meta[(file, url)]
+                                for file, url in download_plan
+                            }
                     elif m.get("gguf_url") == gguf_url:
                         allowed = True
+                        expected_sha_by_file = {gguf_file: m.get("gguf_sha256", "")}
                     break
             except (json.JSONDecodeError, OSError):
                 pass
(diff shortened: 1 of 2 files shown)


Lightheartdevs (approved):
**This is a real supply-chain fix.** Previously:

- 14 of 19 single-file models shipped with empty `gguf_sha256` — the host-agent verifier silently skipped them.
- Both split-file models hit `if gguf_sha256 and len(download_plan) == 1:` and were *never* verified.
- Client-supplied `gguf_sha256` in request body was treated as ground truth — a caller could supply a matching-but-bogus hash and get a "verified" download.

Net effect: a corrupted or tampered GGUF could land on disk and only surface as a crash in llama-server, with no signal. That's a legitimate security hole, and given this project's whole pitch is "sovereign local AI you can trust," having silent no-op verification is a problem.

**Fix is correct:**
- Populates `gguf_sha256` from HuggingFace LFS `x-linked-etag` headers for all 14 missing entries.
- Adds per-part `sha256` to split-file `gguf_parts`; leaves top-level `gguf_sha256` empty for split-file (per-part is authoritative).
- Drops the `len(download_plan) == 1` gate; iterates over every entry.
- Empty-sha path now logs a `WARNING` instead of silently continuing — operator gets a signal to populate the catalog.
- Drops the client-supplied `gguf_sha256` — catalog is sole trust root.

**Nice touches:**
- Per-part status label (`"foo.gguf (part 2/3)"`) for dashboard progress.
- Chunked 1 MiB hash reads for memory efficiency.
- Tamper-test in manual test plan is exactly the test I'd ask for.

**Nitpick:** cross-validating new SHAs against 5 pre-existing catalog SHAs is good but not enough — consider adding a CI job that re-fetches all 19 HF etags and compares against the catalog monthly to catch upstream model updates. Follow-up, not blocking.

Merge order (per author): after #906 and #905. Agreed. Ship.

Lightheartdevs (approved):
Re-checked against the live branch state. The SHA catalog + full-part verification change still looks correct and the CI surface is green. Approved.

PR #905, opened 2026-04-11
fix(host-agent): use new hooks schema + sandboxed env for install hooks

yasinBursali:
> **Merge order:** Merge after #906, before #900 and #908 — all modify `dream-host-agent.py`.

## What
`_handle_install`'s inner `_run_install` worker now uses the new `hooks.post_install` manifest schema and passes a minimal allowlist environment to the subprocess, matching the pattern already used by `_execute_hook` elsewhere in the host agent.

## Why
Two bugs, one function:

1. **Legacy resolver** — `_run_install` was the last call site still using `_resolve_setup_hook()`, which only reads the old `service.setup_hook` manifest field. Any extension declaring its setup via `hooks.post_install` was silently skipped during install.
2. **Secret leak via subprocess env inheritance** — `subprocess.run(...)` had no `env=` kwarg, so the extension's setup.sh inherited the full host-agent environment including `AGENT_API_KEY`, `DREAM_AGENT_KEY`, `DASHBOARD_API_KEY`. A malicious extension could exfiltrate host-agent credentials during install.

## How
- Replace `_resolve_setup_hook(ext_dir)` with `_resolve_hook(ext_dir, \"post_install\")` (the schema-aware resolver that `_handle_setup_hook` and `_handle_hook` already use).
- Build the same minimal `hook_env` dict `_execute_hook` uses (8 keys: PATH, HOME, SERVICE_ID, SERVICE_PORT, SERVICE_DATA_DIR, DREAM_VERSION, GPU_BACKEND, HOOK_NAME). Pass via `env=hook_env`.
- Delete `_resolve_setup_hook` (now unreferenced repo-wide).
- Add `TestInstallHookEnvAllowlist` — 4 source-level assertions that fail loudly if anyone reverts the kwarg, re-adds a secret, or removes a required allowlist key.

## Testing
- \`pytest dashboard-api/tests/test_host_agent.py\` → 26 passed (22 pre-existing + 4 new)
- \`python3 -m py_compile dream-server/bin/dream-host-agent.py\` → clean
- Repo-wide grep for \`_resolve_setup_hook\` → empty (dead code removed)
- **Manual test needed:** install an extension whose \`manifest.yaml\` declares \`hooks.post_install: setup.sh\` where setup.sh writes \`env > /tmp/leak.txt\`. Verify leak.txt does NOT contain any of \`AGENT_API_KEY\`, \`DREAM_AGENT_KEY\`, \`DASHBOARD_API_KEY\`.

## Platform Impact
[...]

PR #906, opened 2026-04-11
fix(host-agent): surface docker failures in _compose_restart_llama_server

yasinBursali:
> **Merge order:** Merge before #905, #900, #908, #902, and #893 — all modify `dream-host-agent.py`.

## What
`_compose_restart_llama_server` now checks return codes on every `subprocess.run` for docker commands and raises `RuntimeError` with the captured stderr on non-zero exit.

## Why
All six docker calls in the function (`compose stop`, `compose up`, `compose restart`, `docker restart`, `docker stop`, `docker start`) previously ran without inspecting returncode. If any failed — daemon hiccup, permission error, missing compose file, disk full — the function returned silently. `_do_model_activate` then proceeded into a 5-minute health-check polling loop and eventually reported the generic "Health check failed — rolled back" with no indication of the real cause. Operators saw a slow failure with no actionable error.

## How
Introduce a nested `_run(argv, timeout)` helper inside the function that calls `subprocess.run` with `text=True` and raises `RuntimeError(f"{cmd} failed (exit N): {stderr[:300]}")` on non-zero. Route all six docker calls through it. Update the docstring to document the raising contract.

The caller `_do_model_activate` already wraps its body in `except Exception as exc` and returns a 500 with the error message, so the real docker failure now surfaces immediately instead of the 5-minute health-check stall.

## Testing
- \`python3 -m py_compile dream-server/bin/dream-host-agent.py\` → clean
- **Manual test needed:** trigger a model activate on a native-host install (Linux/macOS) while \`docker compose up\` is guaranteed to fail — e.g., temporarily \`chmod 000 .env\` or stop the docker daemon. Expected: immediate 500 with docker error in response, not a 5-minute wait before "Health check failed".
- Rollback path (L1382) also calls this function — if rollback fails, the outer \`except Exception\` still catches the RuntimeError and returns 500. Acceptable per project's "Let It Crash" philosophy.

## Platform Impact
- **macOS:** affected — this is the native-host code path (launchd agent).
- **Linux:** affected — native-host code path (systemd agent).
- **Windows/WSL2:** **not affected** — WSL2 uses \`_recreate_llama_server\` which is gated on \`_in_container\` at \`_do_model_activate:1322\` and has its own returncode handling.

## Review notes
- Only the function body changes. No callers modified, no imports added.
- No test coverage pre-existed for this function; a manual test is the only validation for the non-zero-return path.

### DIFF
--- a/dream-server/bin/dream-host-agent.py
+++ b/dream-server/bin/dream-host-agent.py
@@ -1599,9 +1599,11 @@ def _write_lemonade_config(install_dir: Path, gguf_file: str):
 def _compose_restart_llama_server(env: dict):
     """Restart llama-server via docker compose (host-native path).
 
-    This is the primary restart strategy for Linux (systemd) and macOS
-    (launchd) where the agent runs natively on the host.  It mirrors the
-    proven pattern from bootstrap-upgrade.sh lines 289-304.
+    Primary restart strategy for Linux (systemd) and macOS (launchd) where the
+    agent runs natively on the host. Mirrors bootstrap-upgrade.sh lines 289-304.
+
+    Raises RuntimeError on any docker-layer failure so _do_model_activate can
+    surface the error immediately instead of waiting for the health-check loop.
     """
     gpu_backend = env.get("GPU_BACKEND", "nvidia")
     compose_flags = []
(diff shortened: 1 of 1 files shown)


Lightheartdevs (approved):
Aligns with CLAUDE.md's Let-It-Crash rule. Previously `_compose_restart_llama_server` used `subprocess.run(...)` with `capture_output=True` but discarded the result — a docker-layer failure (daemon down, image missing, compose flags stale) would silently succeed, and the caller only found out via the health-check timeout 30+ seconds later.

The `_run()` helper:
- Checks `returncode != 0` and raises `RuntimeError` with the first 300 chars of stderr.
- Truncates stderr to a sane size so log lines stay readable.
- Preserves `timeout` per subprocess call.

Downstream `_do_model_activate` now surfaces the error immediately instead of waiting for health checks to time out. Good UX improvement.

**Minor:** the `' '.join(argv[:3])` in the error message only shows the first three args (`docker compose -f`), which is fine for identification. Could include the operation (`restart`/`up`/`stop`) for clearer messages, but not blocking.

Ship.

Lightheartdevs (approved):

PR #907, opened 2026-04-11
fix(host-agent,dashboard-api): allow API management of built-in extensions

yasinBursali:
> **Merge order:** Merge after #909 and #926, before #935 — all modify `extensions.py`.

## What
The host agent and dashboard-api blocklist now correctly identifies the 4 always-on base-compose services (llama-server, open-webui, dashboard, dashboard-api) as the only services that cannot be managed via API. The 16 other built-in extensions (n8n, tts, whisper, comfyui, litellm, langfuse, openclaw, perplexica, searxng, token-spy, qdrant, embeddings, ape, dreamforge, opencode, privacy-shield) are now first-class manageable extensions.

## Why
`CORE_SERVICE_IDS` contains all 20 known service IDs. `validate_service_id` (host agent) and `_assert_not_core` (dashboard-api) were using it as a blocklist, rejecting all 20 with 403 "Cannot manage core service" — except for `privacy-shield` which was whitelisted via `TOGGLABLE_CORE_SERVICES`. This broke every dashboard UI action (start/stop/restart/purge) for 15 optional extensions that users are supposed to be able to toggle, blocked multiple in-flight features (template apply, deps+hooks, UI extension cards), and made the dashboard API lie about what it could actually manage.

Root cause: `core-service-ids.json` was originally a whitelist of known service IDs (for input validation), but got repurposed as a blocklist of "services the API is not allowed to touch" — those are two different concepts and the file contents only match the whitelist meaning.

## How
- Introduce `ALWAYS_ON_SERVICES = frozenset({"llama-server", "open-webui", "dashboard", "dashboard-api"})` in `dream-host-agent.py` and `dashboard-api/config.py` — the real 4-service blocklist.
- Update `validate_service_id` to check `ALWAYS_ON_SERVICES` instead of the `CORE_SERVICE_IDS`-minus-`TOGGLABLE_CORE_SERVICES` pattern, and make the extension-dir fallback (`EXTENSIONS_DIR / sid`) unconditional so all built-ins resolve.
- Update `_assert_not_core`, `_is_dep_satisfied`, and the purge handler in `dashboard-api/routers/extensions.py` to check `ALWAYS_ON_SERVICES`.
- **Keep** `CORE_SERVICE_IDS` and `_scan_compose_content`'s use of it — that's a legitimate name-collision check preventing user extensions from shadowing built-in service names.
- **Delete** the now-unused `TOGGLABLE_CORE_SERVICES` constant from the host agent.
- Add `TestAssertNotCoreAllowsBuiltins` — 20 parametrized test cases covering all 16 built-ins as "allow" and all 4 always-on as "block with 403".
- Update `test_purge_403_core_service` assertion substring to match the renamed error message ("always-on service" instead of "core service").

## Testing
- `pytest dashboard-api/tests/test_extensions.py -k assert_not_core` → 20 passed
- Full `test_extensions.py` → 124 passed, 0 failed (104 pre-existing + 20 new)
- `python3 -m py_compile` on both modified .py files → clean
- Pre-existing 12 pytest failures (test_workflows, test_gpu_detailed, test_main GPU cached) are unrelated — verified by reverting this diff and re-running.
- **Manual test needed:** install fresh, disable tts via CLI (`dream disable tts`), open dashboard → click Enable on tts card → expect success (was 403 before this fix). Also verify that `dream stop llama-server` via API still 403s.

## Platform Impact
- **macOS:** affected — Python-only change, identical behavior on all three platforms.
- **Linux:** affected — same.
- **Windows/WSL2:** affected — same.

## Merge order
**This PR should merge AFTER PR #909 (`fix/apply-template-sync-builtin-extensions`).**

Reason: this PR's change to `_is_dep_satisfied` makes disabled built-in extensions fall through to `_activate_service`, which on upstream/main still only checks `USER_EXTENSIONS_DIR` and would return 404 for built-ins. PR #909 fixes `_activate_service` to check both dirs. If this lands first, there's a brief window where a user extension declaring a disabled built-in as a dep would fail to auto-resolve.

The two PRs touch `dashboard-api/routers/extensions.py` in disjoint functions (`_assert_not_core`/`_is_dep_satisfied`/purge handler here vs `_activate_service` in #909), so the merge is a trivial fast-forward in either order.

## Known considerations
- Follow-up issue filed at https://github.com/yasinBursali/DreamServer/issues/327 for the architecture ambiguity between "core / built-in / user" extension lanes. Not a blocker for this fix.
- `core-service-ids.json` is NOT renamed — intentional, avoids config migration churn. A cleaner name like `known-service-ids.json` is a follow-up.

### DIFF
--- a/dream-server/bin/dream-host-agent.py
+++ b/dream-server/bin/dream-host-agent.py
@@ -49,8 +49,9 @@ AGENT_API_KEY: str = ""
 GPU_BACKEND: str = "nvidia"
 TIER: str = "1"
 CORE_SERVICE_IDS: set = set()
-# Core services that can be toggled via the extension API (e.g., privacy shield)
-TOGGLABLE_CORE_SERVICES: set = {"privacy-shield"}
+# Always-on services defined in docker-compose.base.yml — never stoppable via API.
+# Distinct from CORE_SERVICE_IDS (which is the allowlist of known service IDs).
+ALWAYS_ON_SERVICES: frozenset = frozenset({"llama-server", "open-webui", "dashboard", "dashboard-api"})
 USER_EXTENSIONS_DIR: Path = Path()
 EXTENSIONS_DIR: Path = Path()
 
@@ -358,14 +359,14 @@ def validate_service_id(handler, body: dict) -> str | None:
     if not isinstance(sid, str) or not SERVICE_ID_RE.match(sid):
         json_response(handler, 400, {"error": "Invalid service_id"})
         return None
-    if sid in CORE_SERVICE_IDS and sid not in TOGGLABLE_CORE_SERVICES:
-        json_response(handler, 403, {"error": f"Cannot manage core service: {sid}"})
+    if sid in ALWAYS_ON_SERVICES:
+        json_response(handler, 403, {"error": f"Cannot manage always-on service: {sid}"})
         return None
     # Verify the service_id maps to an actual installed extension.
-    # Check user-extensions first, then core extensions for togglable services.
+    # Check user-extensions first, then built-in extensions.
     ext_dir = USER_EXTENSIONS_DIR / sid
-    if not ext_dir.is_dir() and sid in TOGGLABLE_CORE_SERVICES:
-        ext_dir = INSTALL_DIR / "extensions" / "services" / sid
+    if not ext_dir.is_dir():
+        ext_dir = EXTENSIONS_DIR / sid
     manifest_exists = any((ext_dir / n).exists() for n in ("manifest.yaml", "manifest.yml", "manifest.json"))
     if not ext_dir.is_dir() or not manifest_exists:
         json_response(handler, 404, {"error": f"Extension not found: {sid}"})
--- a/dream-server/bin/dream-host-agent.py
+++ b/dream-server/bin/dream-host-agent.py
@@ -589,6 +589,10 @@ class AgentHandler(BaseHTTPRequestHandler):
             self._handle_setup_hook()
         elif self.path == "/v1/extension/hooks":
             self._handle_hook()
+        elif self.path == "/v1/extension/activate":
+            self._handle_extension_compose_toggle(activate=True)
+        elif self.path == "/v1/extension/deactivate":
+            self._handle_extension_compose_toggle(activate=False)
         elif self.path == "/v1/service/logs":
             self._handle_service_logs()
         elif self.path == "/v1/model/download":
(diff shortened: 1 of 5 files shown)


Lightheartdevs (dismissed):
Good feature split. Previously `_assert_not_core` blocked **all** `CORE_SERVICE_IDS` from API management, which meant built-in extensions (n8n, tts, whisper, etc.) couldn't be enabled/disabled via the dashboard. This PR correctly separates two concepts:

- **`ALWAYS_ON_SERVICES`** (frozenset of 4 base-compose services): never stoppable via API. Enforced in both host agent (`validate_service_id`) and dashboard-api (`_assert_not_core`).
- **`CORE_SERVICE_IDS`** (all built-in extensions): can be managed via API.

**New host agent endpoints `/v1/extension/activate` and `/v1/extension/deactivate`**: necessary because the extensions dir is mounted `:ro` in the dashboard-api container. Agent performs the compose.yaml ↔ compose.yaml.disabled rename on the host filesystem. Correctly gated by `check_auth`, `SERVICE_ID_RE`, `ALWAYS_ON_SERVICES` check, `_service_locks` for concurrency, and TOCTOU check inside the lock. Clean.

**`_resolve_extension_dir` helper**: centralizes the user-first, built-in-fallback resolution. Correctly rejects path traversal via `is_relative_to`. This is a better version of the inline logic in PR #909.

**Cross-PR dependency chain:**
- Depends on #926 (`skip_name_collision` parameter) — must merge after #926.
- Conflicts with #909 (both modify `_activate_service`) — #909 must rebase after this lands.

Recommended merge order: #926 → #907 → rebase #909 on top.

Approving on condition of the merge order. Ship.

Lightheartdevs:
Two follow-up observations from a deeper re-read — neither blocks merge, but both worth addressing:

**1. Attack-surface shift worth documenting.** The old host-agent logic at `dream-host-agent.py:395` was:
```python
if sid in CORE_SERVICE_IDS and sid not in TOGGLABLE_CORE_SERVICES:
    # block
```
where `TOGGLABLE_CORE_SERVICES = {"privacy-shield"}`. Net: blocked everything except privacy-shield from being toggled via API.

The new logic blocks only the 4 `ALWAYS_ON_SERVICES`. Net: n8n, tts, whisper, searxng, qdrant, privacy-shield, and any future built-in become toggleable via API.

This is the **intended behavior** — without it, template apply for built-in extensions doesn't work. But it's a real attack-surface expansion worth noting in the PR description / CHANGELOG: an attacker who compromises `dashboard-api` (and obtains `DREAM_AGENT_KEY`) can now stop a broader set of built-in services, not just privacy-shield. Nothing catastrophic — local services, user experience impact only — but the threat model shifted.

**2. Inconsistent gate on `purge_extension_data`.** At [`extensions.py:1340`](dream-server/extensions/services/dashboard-api/routers/extensions.py#L1340):
```python
if service_id in CORE_SERVICE_IDS:
    raise HTTPException(status_code=403, detail="Cannot purge core service data")
```

After this PR, a user can toggle n8n **off** via the dashboard but **can't purge its data** — because the purge check still uses `CORE_SERVICE_IDS`, not `ALWAYS_ON_SERVICES`. Inconsistent UX.

Either:
- Change this to `ALWAYS_ON_SERVICES` (so purge matches the toggle permissions), or
- Explicitly document that purge is stricter than toggle (reasonable: disable is reversible, purge isn't)

Suggest either is fine, but pick one and be explicit.

Still approving.

PR #909, opened 2026-04-11
fix(dashboard-api): non-blocking apply_template + resolve built-in extensions

yasinBursali:
> **Merge order:** Merge before #926, #907, and #935 — all modify `extensions.py`.

## What
Two related dashboard-api bugs:

1. `apply_template` was `async def` but called blocking urllib-based helpers with 300s timeouts, stalling the FastAPI event loop and freezing every other request for potentially minutes at a time.
2. `_activate_service` only resolved extension directories under `USER_EXTENSIONS_DIR`, silently 404'ing on the 11 built-in extensions that live in `EXTENSIONS_DIR`. All 11 shipped templates (chat-playground, voice-assistant, creative-studio, llm-platform, ...) reference at least one built-in, so template apply silently "succeeded" with `enabled_count: 0` and the UI showed "Template applied" but nothing actually started.

## Why
**Event-loop stall (async handler + blocking I/O):** `apply_template` at `templates.py:98` is `async def`, so FastAPI runs it directly on the asyncio event loop. Inside, it calls `_call_agent_hook`, `_call_agent`, `_call_agent_invalidate_compose_cache`, and `_install_from_library` — all of which use `urllib.request.urlopen` with blocking 300s timeouts. A template with 5 services could freeze the entire dashboard-api for several minutes while health checks, log streams, and GPU metric polls all time out.

**Silent skip of built-ins:** `_activate_service` at `extensions.py:977` resolves `ext_dir = (USER_EXTENSIONS_DIR / service_id).resolve()`. Built-in extensions are installed by the product at `EXTENSIONS_DIR / <name>/`, not `USER_EXTENSIONS_DIR`. `ext_dir.is_dir()` returned False → `HTTPException(404, "Extension not installed: n8n")`. `apply_template` caught the HTTPException and recorded `"skipped: Extension not installed: n8n"` — no user-visible error. The in-code comment at `templates.py:163` said "_activate_service checks both user-installed and built-in extension dirs", which was false.

### Event-loop stall fix
Keep `apply_template` as `async def` and wrap every blocking helper call in `await asyncio.to_thread(...)` — same pattern `main.py:api_settings_env_save` already uses. Introduce two nested helpers (`_install_with_lock`, `_activate_with_lock`) to keep `_extensions_lock` acquisition inside a single off-loop call, because `fcntl` file locks cannot cross thread boundaries in Python.

### Built-in resolution fix
Rewrite `_activate_service` to check both `USER_EXTENSIONS_DIR` and `EXTENSIONS_DIR`, with `is_relative_to` path-traversal validation on each. User dir takes precedence when both exist (matches existing precedence in `_is_dep_satisfied`). The rest of the function (`compose.yaml.disabled` → `compose.yaml` rename logic) operates on the resolved `ext_dir` unchanged.

### Tests
- `TestActivateServiceBuiltinBranch` — 3 tests covering the disabled-compose rename path, the already-enabled idempotent path, and the user-dir-takes-precedence path (all via `tmp_path` + `monkeypatch`).
- `test_templates.py` — structural AST regression test that walks `apply_template`'s source and asserts every blocking helper (`_call_agent_hook`, `_call_agent`, `_get_missing_deps_transitive`, `_install_with_lock`, `_activate_with_lock`) is invoked as the first positional arg of an `await asyncio.to_thread(...)` Call. Any revert to a direct blocking call fails the test with a clear regression message.

## Testing
- \`pytest dashboard-api/tests/test_extensions.py tests/test_templates.py\` → 126 passed (123 pre-existing + 3 new + structural template tests)
- \`python3 -m py_compile\` on both modified .py files → clean
- **Manual test needed:**
  1. Install cleanly, disable a built-in (\`dream disable n8n\`).
  2. In dashboard, apply a template that references n8n (e.g. \`no-code-ai-builder\`). Expected: template apply succeeds, n8n container shows up in \`docker ps\`, UI shows n8n as healthy. Before fix: "Template applied" but n8n never started.
  3. During a template apply, open the dashboard in another tab and hit \`/health\`. Expected: immediate response. Before fix: hung until the apply finished.

## Platform Impact
- **macOS:** affected — Python-only change, identical behavior on all three platforms.
- **Linux:** affected — same.
- **Windows/WSL2:** affected — same.

## Merge order
**This PR should merge BEFORE PR #907 (\`fix/validate-service-id-always-on\`).**

Reason: PR #907's change to \`_is_dep_satisfied\` makes disabled built-ins fall through to \`_activate_service\`. If #907 lands first, \`_activate_service\` still only checks \`USER_EXTENSIONS_DIR\` on upstream/main and would 404 for built-ins. This PR fixes \`_activate_service\` to check both dirs. Same file, disjoint functions — merge is clean in either order but the sequence affects behavior during the brief window between the two landings.

## Known considerations
- **Langfuse edge case:** of the 15 built-in extensions, 14 ship with \`compose.yaml\` and only \`langfuse\` ships with \`compose.yaml.disabled\`. Enabling langfuse via a template renames the disabled compose inside the shipped extensions tree (same pattern the installer's feature-selection already uses at install time — see \`installers/phases/03-features.sh:78-96\`). The installer's \`dream update\` flow preserves file-level state, so this is behavior-consistent with how comfyui/dreamforge are already handled. A follow-up issue at https://github.com/yasinBursali/DreamServer/issues/327 proposes adding langfuse to the install-time feature menu for full parity with the other optional extensions.
- **\`_is_dep_satisfied\` correctness:** the pre-existing function treats any \`CORE_SERVICE_IDS\` member as "always satisfied", which is wrong for built-in extensions that haven't actually been enabled yet. This is fixed in the companion PR #907 (which should merge after this one).

### DIFF
--- a/dream-server/extensions/services/dashboard-api/routers/extensions.py
+++ b/dream-server/extensions/services/dashboard-api/routers/extensions.py
@@ -1036,15 +1036,27 @@ def _get_missing_deps_transitive(
 def _activate_service(service_id: str) -> dict:
     """Core enable logic — NO lock acquisition. Called inside _extensions_lock.
 
+    Checks both USER_EXTENSIONS_DIR (user-installed) and EXTENSIONS_DIR
+    (built-in) so templates can enable built-in extensions like n8n, tts, etc.
+
     Returns a result dict for the service. Cycle detection is handled
     upstream by _get_missing_deps_transitive.
     """
-    ext_dir = (USER_EXTENSIONS_DIR / service_id).resolve()
-    if not ext_dir.is_relative_to(USER_EXTENSIONS_DIR.resolve()):
+    user_dir = (USER_EXTENSIONS_DIR / service_id).resolve()
+    builtin_dir = (EXTENSIONS_DIR / service_id).resolve()
+
+    in_user = user_dir.is_relative_to(USER_EXTENSIONS_DIR.resolve())
+    in_builtin = builtin_dir.is_relative_to(EXTENSIONS_DIR.resolve())
+    if not in_user and not in_builtin:
         raise HTTPException(
             status_code=404, detail=f"Extension not found: {service_id}",
         )
-    if not ext_dir.is_dir():
+
+    if in_user and user_dir.is_dir():
+        ext_dir = user_dir
+    elif in_builtin and builtin_dir.is_dir():
+        ext_dir = builtin_dir
+    else:
         raise HTTPException(
             status_code=404, detail=f"Extension not installed: {service_id}",
         )
--- a/dream-server/extensions/services/dashboard-api/routers/templates.py
+++ b/dream-server/extensions/services/dashboard-api/routers/templates.py
@@ -1,5 +1,6 @@
 """Service template endpoints."""
 
+import asyncio
 import logging
 
 from fastapi import APIRouter, Depends, HTTPException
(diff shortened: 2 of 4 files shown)


Lightheartdevs (approved):
Two good changes bundled:

1. **`asyncio.to_thread` wrapping** in `apply_template`: urllib installs with 300s timeouts and host-agent network calls were blocking the event loop. The `_install_with_lock` and `_activate_with_lock` wrappers are correctly scoped — `_extensions_lock` is a thread-context manager and can't cross await boundaries, so re-entering the lock inside `to_thread` is the right call.

2. **`_activate_service` checks both USER_EXTENSIONS_DIR and EXTENSIONS_DIR**: lets templates enable built-in extensions like n8n, tts. Tests cover all three branches (built-in disabled → enabled, built-in already enabled, user takes precedence over built-in).

The AST-based test (`test_apply_template_blocking_calls_run_in_to_thread`) is a really nice pattern — replaces a flaky concurrent-request timing test with a deterministic structural proof that blocking helpers are always wrapped. Worth copying for other async routers.

**Cross-PR conflict (IMPORTANT):** #907 modifies the same `_activate_service` function to use a `_resolve_extension_dir()` helper (cleaner extraction of the same logic). These two PRs cannot both merge as-is.

Recommended order:
1. Merge #926 (adds `skip_name_collision` parameter)
2. Merge #907 (adds `_resolve_extension_dir` helper + built-in extension management)
3. Rebase this PR on top — the `_activate_service` changes here become a no-op after #907, and only the `asyncio.to_thread` wrapping + tests remain.

Approving the concept and the async wrapping; the `_activate_service` diff needs to drop out after #907 lands.

PR #911, opened 2026-04-11
fix(07-devtools): recover from corrupt opencode.json when jq parse fails (#332)

yasinBursali:
## Summary

- Adds a deterministic recovery path when `07-devtools.sh`'s jq rewrite finds an existing `opencode.json` it cannot parse (issue #332).
- Extracts the first-install HEREDOC template into a small shell helper `_opencode_write_fresh` and calls it from both the fresh-install branch and the recovery branch.
- When jq parse fails, warns the operator (so the corruption stays visible) and regenerates the file from the template instead of leaving the broken file in place.

## Context

This **stacks on #901** (`fix/opencode-json-sed`). #901 fixes the WRITE path so new installs don't corrupt `opencode.json` by eating the trailing comma. #332 is the RECOVERY gap it leaves behind: users who upgrade from a pre-#901 broken state had no way out of the bad state because the jq path bailed out instead of falling through to any repair logic.

Closes #332. Cross-ref: parent #309 (closed by #901).

## Runtime validation (WSL2 / Ubuntu 24.04 / jq 1.7)

Reproduced against a real pre-existing corrupted file on the running integration install (missing comma between `baseURL` and `apiKey` siblings, the exact #309/#332 symptom):

```
$ jq -e . ~/.config/opencode/opencode.json
jq: parse error: Expected separator between values at line 11, column 16
$ echo $?
5
```

Rsync'd the patched `07-devtools.sh` into the running install and re-ran the phase:

```
⚠ OpenCode config jq rewrite failed (existing file unparseable) — regenerating from template
✓ OpenCode config regenerated from template (recovered from corruption)
```

Post-fix:

```
$ jq -e . ~/.config/opencode/opencode.json >/dev/null && echo "valid"
valid
$ jq -r '.provider["llama-server"].options.baseURL, .provider["llama-server"].options.apiKey' ~/.config/opencode/opencode.json
http://127.0.0.1:11434/v1
no-key
```

Install health unchanged: 17 containers healthy, `dream-host-agent.service` active, dashboard-api `/health` 200.

## Test plan

- [x] Reproduces the #332 symptom on a corrupt pre-existing `opencode.json`
- [x] Recovery path regenerates a valid JSON file
- [x] Fresh-install branch still produces the same file via the shared helper
- [x] jq-absent sed fallback branch is unchanged in behavior
- [x] Warns before regenerating so the operator sees the corruption was detected

### DIFF
--- a/dream-server/installers/phases/07-devtools.sh
+++ b/dream-server/installers/phases/07-devtools.sh
@@ -123,8 +123,11 @@ else
             _opencode_key="no-key"
         fi
 
-        if [[ ! -f "$OPENCODE_CONFIG_DIR/opencode.json" ]]; then
-            cat > "$OPENCODE_CONFIG_DIR/opencode.json" <<OPENCODE_EOF
+        # Writes a fresh opencode.json from the template. Used for first-install
+        # and as deterministic recovery when the jq rewrite path finds an
+        # existing malformed file it cannot parse (issue #332).
+        _opencode_write_fresh() {
+            cat > "$1" <<OPENCODE_EOF
 {
   "\$schema": "https://opencode.ai/config.json",
   "model": "llama-server/${LLM_MODEL}",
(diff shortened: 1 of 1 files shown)


yasinBursali:
Replaced by combined PR #932

PR #926, opened 2026-04-12
fix(extensions): trust built-in composes in _scan_compose_content name-collision check

### LINKED ISSUE #338: [#907/#909 follow-up] _scan_compose_content rejects all built-ins except comfyui — second gate blocks EXTENSIONS_DIR activation even after #299 fix
## Summary

PR #907 (origin: #299) correctly relaxed the top-level `CORE_SERVICE_IDS` blocklist down to the 4 `ALWAYS_ON_SERVICES`, and PR #909 correctly added the `EXTENSIONS_DIR` branch to `_activate_service`. Together they *should* make built-in extensions activatable via the internal template-apply path. But there's a **second, downstream gate** in the same code that still rejects 19 of 20 built-ins by name — `_scan_compose_content` at `routers/extensions.py:187`.

When `_activate_service` tries to activate a built-in in its `compose.yaml.disabled` state (line 1023), it calls `_scan_compose_content(disabled_compose)`. That function walks the compose file's internal `services:` map and rejects any entry whose key matches `CORE_SERVICE_IDS` with:

```python
for svc_name in services:
    if svc_name in CORE_SERVICE_IDS:
        raise HTTPException(
            status_code=400,
            detail=f"Extension rejected: service name '{svc_name}' conflicts with core service",
        )
```

This is a legitimate **name-collision defense** against a user extension trying to shadow a built-in's service name. PR #907 explicitly kept it, correctly — this check has real security value for the `/install` path (line 856). But when the caller is `_activate_service` activating a **known built-in whose compose file LEGITIMATELY declares its own built-in service name**, the check misfires: the compose IS declaring `n8n: ...` precisely because the extension is named `n8n`.

Result: every built-in whose `compose.yaml` contains its own service id in the `services:` map gets rejected at this gate *before* reaching the rename at line 1032. The only built-in that reliably slips past is **comfyui**, because its `compose.yaml` is a stub with `services: {}` (the real service definitions live in `compose.{nvidia,amd,multigpu}.yaml` GPU overlays).

So the full state of built-in activatability post-PR-907+909 is:

| Built-in | internal `services:` map | Activatable via `_activate_service`? |
|---|---|---|
| `comfyui` | `{}` (empty stub) | ✅ reaches `os.rename` — but then hits #331 EROFS |
| `n8n` | `{n8n: ...}` | ❌ blocked at `_scan_compose_content:204` with 400 |
| `tts` | `{tts: ...}` | ❌ blocked |
| `whisper` | `{whisper: ...}` | ❌ blocked |
| `langfuse` | `{langfuse-*: ...}` | ❌ blocked (its sub-services include `langfuse`) |
| `openclaw` | `{openclaw: ...}` | ❌ blocked |
| `searxng` | `{searxng: ...}` | ❌ blocked |
| `perplexica` | `{perplexica: ...}` | ❌ blocked |
| `token-spy` | `{token-spy: ...}` | ❌ blocked (verified empirically below) |
| `qdrant` | `{qdrant: ...}` | ❌ blocked |
| `embeddings` | `{embeddings: ...}` | ❌ blocked |
| `ape` | `{ape: ...}` | ❌ blocked |
| `dreamforge` | `{dreamforge: ...}` | ❌ blocked |
| `litellm` | `{litellm: ...}` | ❌ blocked |
| `privacy-shield` | `{privacy-shield: ...}` | ❌ blocked |
| `opencode` | host-only, no compose | N/A |
| `llama-server`, `open-webui`, `dashboard`, `dashboard-api` | always-on | blocked at `_assert_not_core` (correct) |

**Net: 1 of 16 togglable built-ins is activatable past the second gate — and that 1 then immediately hits #331's `:ro`-mount EROFS.** No built-in can actually be activated via the integration branch's template-apply path without fixing BOTH this gate and #331.

## Reproduction

On macOS Apple Silicon, integration branch (`upstream/main` + 17 open PRs):

```bash
# 1. Put token-spy into disabled state
$ docker stop dream-token-spy
$ mv /Volumes/X/dream-server-test/extensions/services/token-spy/compose.yaml \\
     /Volumes/X/dream-server-test/extensions/services/token-spy/compose.yaml.disabled

# 2. Apply a template that references token-spy
$ curl -s -X POST -H \"Authorization: Bearer \$KEY\" \\
    http://127.0.0.1:3002/api/templates/llm-platform/apply

{
  \"template_id\": \"llm-platform\",
  \"results\": {
    \"litellm\": \"already_enabled\",
    \"langfuse\": \"skipped: Extension rejected: service name 'langfuse' conflicts with core service\",
    \"token-spy\": \"skipped: Extension rejected: service name 'token-spy' conflicts with core service\",
    ...
  }
}
```

token-spy and langfuse both get skipped at the `_scan_compose_content` gate, **not** at `_assert_not_core` (which they correctly pass now). The skip is silent at the HTTP layer (200 OK with per-service error strings), so a user running a template apply just sees \"some built-ins skipped with opaque conflicts-with-core-service errors\" and no actionable feedback.

## Why existing tests don't catch this

- **`TestActivateServiceBuiltinBranch`** (added by #909) uses `monkeypatch` to set up test compose files in `tmp_path` with generic service names like `fakesvc`. Since `fakesvc` isn't in `CORE_SERVICE_IDS`, it slips past `_scan_compose_content` and the test correctly verifies the `EXTENSIONS_DIR` branch. But the test **doesn't exercise a real built-in name**, so it doesn't hit the second gate.

- **`TestAssertNotCoreAllowsBuiltins`** (added by #907) tests `_assert_not_core` **in isolation** and doesn't call `_activate_service` or run a compose file through `_scan_compose_content`. It verifies the name gets past the *first* gate; it doesn't verify it gets past the *second* gate.

Either test would have caught this if it had driven `_activate_service` against a real built-in's real `compose.yaml.disabled`.

## Relationship to adjacent issues

- **#299** — Reports the blocklist-misuse at the `validate_service_id()` level. PR #907 fixes that layer. This issue is downstream of #299 — a second check nobody realized existed.
- **#331** — `:ro` mount + `os.rename` crash. Only reachable via **comfyui** on the integration branch because comfyui is the only built-in that gets past the `_scan_compose_content` gate. Fixing just #331 without this gate still leaves 15 built-ins non-activatable.
- **#333** — `/enable`+`/disable` REST handlers miss `EXTENSIONS_DIR` fallback. Orthogonal — #333 is about the PUBLIC endpoints not finding built-ins in the first place; this issue is about even the INTERNAL template-apply path being unable to activate them.

## Suggested fix

Add a `trusted=True` branch to the `_scan_compose_content` call at line 1023 when the compose being scanned is under `EXTENSIONS_DIR` (built-in trust root), same as the `_install_from_library` call at line 856 already does:

```diff
     # Re-scan compose content (TOCTOU prevention)
-    _scan_compose_content(disabled_compose)
+    # Skip the name-collision check for built-in extensions — their compose
+    # files LEGITIMATELY declare their own service id, which is in CORE_SERVICE_IDS
+    # by definition. The other checks (privileged, label spoofing, unsafe volumes)
+    # still run.
+    is_builtin = ext_dir.resolve().is_relative_to(EXTENSIONS_DIR.resolve())
+    _scan_compose_content(disabled_compose, trusted=is_builtin)
```

Then update `_scan_compose_content(trusted=True)` semantics: skip the `CORE_SERVICE_IDS` collision check, keep the `privileged: true` + label-spoofing + unsafe-mount checks (which are still valid defenses for built-ins).

Alternatively: gate the collision check on `ext_dir` being under `USER_EXTENSIONS_DIR` only — built-ins get scanned by the other defenses but not the name check.

Either approach preserves the security intent of PR #907's \"keep `CORE_SERVICE_IDS`\" for name-shadowing defense against user extensions, while unblocking the legitimate template-apply path for built-ins.

## Severity

Medium-high. #907's promised behavior (\"16 other built-in extensions are now first-class manageable\") doesn't ship at the internal activation layer *or* the public REST layer (see #333). Template apply is the only documented user-facing path, and it's broken for 15 of 16 toggleable built-ins. Combined with #331, the only one that slips through the name gate then crashes on EROFS. Net: every template in the catalog that contains any toggleable built-in either silently skips it with an opaque error, or crashes with HTTP 500.

Not a regression in the direct sense — on `upstream/main` the first gate (`validate_service_id`) blocked everything, so nothing reached the second gate. PR #907 opened the first gate and exposed that the second gate was always there.

## Environment

- macOS Darwin 25.2.0, Apple Silicon M4, 24 GB unified memory
- Integration branch: `local/integration-test` = `upstream/main` + all 17 open yasinBursali PRs (#893–909)
- Install at `/Volumes/X/dream-server-test`
- Reproduced while cross-verifying #331 on Mac (which only triggered on comfyui because every other built-in gets blocked by this earlier gate).

yasinBursali:
> **Merge order:** Merge after #909, before #907 and #935 — all modify `extensions.py`.

## What
Allow built-in extensions (n8n, tts, whisper, langfuse, token-spy, etc.) to be activated via templates and the enable API without being rejected by the CORE_SERVICE_IDS name-collision check.

## Why
`_scan_compose_content` blocks any compose file whose `services:` map declares a name found in `CORE_SERVICE_IDS`. This is a legitimate anti-shadowing defense for user extensions — it prevents a user extension from defining `services: {n8n: ...}` and hijacking the built-in n8n container. But every built-in extension naturally declares its own service name in its compose file. The same check therefore blocks `_activate_service` from re-enabling any built-in except comfyui (which happens to have an empty `services: {}` stub because its real definitions live in GPU-specific overlays).

Templates that call `_activate_service` for built-ins — e.g. the `llm-platform` template enabling n8n, tts, whisper — all fail silently with `"Extension rejected: service name 'X' conflicts with core service"` and never reach the compose rename.

### 1. New `skip_name_collision` parameter on `_scan_compose_content`
A keyword-only `skip_name_collision: bool = False` parameter, separate from the existing `trusted` (which controls the `build:` directive check). When `True`, only the `CORE_SERVICE_IDS` loop is skipped. All other security checks remain active:
- Privileged mode
- Docker socket mounts
- Absolute host path mounts
- Dangerous capabilities (SYS_ADMIN, NET_ADMIN, etc.)
- Docker Compose label spoofing
- Host PID/network/IPC/userns namespaces
- Root user
- Security option relaxation

The separation from `trusted` is important: library installs (`_install_from_library`) pass `trusted=True` to allow `build:` directives but MUST still check core-service names. An initial implementation that overloaded `trusted` for both broke the existing `test_scan_rejects_core_service_name` test — a library extension shadowing `open-webui` was silently allowed.

### 2. Dual-directory lookup in `_activate_service`
`_activate_service` now checks both `USER_EXTENSIONS_DIR` (user-installed) and `EXTENSIONS_DIR` (built-in), user-dir first. Path containment is verified via `resolve()` + `is_relative_to()` so a user extension cannot be misclassified as built-in. Then: `is_builtin = ext_dir.is_relative_to(EXTENSIONS_DIR.resolve())` → `_scan_compose_content(disabled_compose, skip_name_collision=is_builtin)`.

### Automated
- `pytest tests/test_extensions.py`: **108/108 PASS** (104 existing + 4 new)
- `make lint`: PASS
- `make test`: PASS
- `pre-commit`: PASS

### New unit tests (`TestScanComposeSkipNameCollision`)
| Test | Scenario | Asserts |
|---|---|---|
| `test_rejects_core_name_by_default` | `skip_name_collision=False` + core service name | HTTPException 400, "conflicts with core service" |
| `test_allows_core_name_when_skipped` | `skip_name_collision=True` + core service name | No exception |
| `test_privileged_still_blocked_when_skipped` | `skip_name_collision=True` + `privileged: true` | HTTPException 400, "privileged" |
| `test_docker_socket_still_blocked_when_skipped` | `skip_name_collision=True` + docker socket mount | HTTPException 400, "Docker socket" |

### Runtime (against live macOS install)
- Template apply for `llm-platform` reached `os.rename` for built-in langfuse (which then hit EROFS from the `:ro` bind-mount — a separate known issue). The error is "Read-only file system", NOT "conflicts with core service" — proving the scan gate was passed.
- Security verification inside the container via `docker exec`:
  - Untrusted user extension shadowing core name → correctly REJECTED
  - Trusted built-in with its own name → correctly ALLOWED
  - Trusted but privileged → correctly REJECTED

- **All platforms**: disable a built-in extension's compose (`mv compose.yaml compose.yaml.disabled`), apply a template that includes it, verify it passes the scan gate (reaches either `os.rename` success or the `:ro` mount error, but NOT "conflicts with core service"). Then verify a user extension with a core service name is still blocked via the install endpoint.

## Platform Impact
- **macOS**: affected — built-in activation is the primary use case for templates on Mac installs.
- **Linux**: affected — same code path runs in the dashboard-api container.
- **Windows/WSL2**: affected — same.

## Known Considerations
- **Relationship to PRs #907 and #909.** PR #907 relaxes `_assert_not_core` from `CORE_SERVICE_IDS` to `ALWAYS_ON_SERVICES`. PR #909 adds the dual-dir lookup to `_activate_service`. This PR adds the `skip_name_collision` parameter to `_scan_compose_content` and reproduces #909's dual-dir lookup so the fix is standalone against upstream/main. Once all three merge, the full "make built-ins manageable" surface is covered.
- **The `:ro` mount issue (#331).** Even with this fix, `os.rename` still fails on `:ro` bind-mounts. That's a separate orthogonal fix (delegate rename to host-agent). Both fixes are needed for full built-in activation support.

## Fork issue
Closes yasinBursali/DreamServer#338

### DIFF
--- a/dream-server/extensions/services/dashboard-api/routers/extensions.py
+++ b/dream-server/extensions/services/dashboard-api/routers/extensions.py
@@ -229,7 +229,12 @@ def _assert_not_core(service_id: str) -> None:
         )
 
 
-def _scan_compose_content(compose_path: Path, *, trusted: bool = False) -> None:
+def _scan_compose_content(
+    compose_path: Path,
+    *,
+    trusted: bool = False,
+    skip_name_collision: bool = False,
+) -> None:
     """Reject compose files containing dangerous directives."""
     try:
         data = yaml.safe_load(compose_path.read_text(encoding="utf-8"))
@@ -246,7 +251,7 @@ def _scan_compose_content(compose_path: Path, *, trusted: bool = False) -> None:
         return
 
     for svc_name in services:
-        if svc_name in CORE_SERVICE_IDS:
+        if not skip_name_collision and svc_name in CORE_SERVICE_IDS:
             raise HTTPException(
                 status_code=400,
                 detail=f"Extension rejected: service name '{svc_name}' conflicts with core service",
@@ -1073,8 +1078,14 @@ def _activate_service(service_id: str) -> dict:
             status_code=404, detail=f"Extension has no compose file: {service_id}",
         )
 
-    # Re-scan compose content (TOCTOU prevention)
-    _scan_compose_content(disabled_compose)
+    # Re-scan compose content (TOCTOU prevention). Built-in extensions
+    # legitimately declare their own service name in their compose file, so
+    # skip the CORE_SERVICE_IDS name-collision check for them. User extensions
+    # still get the full anti-shadowing scan. The `trusted` flag is separate
+    # and controls whether `build:` directives are allowed (library installs
+    # need it, built-in activations do not).
+    is_builtin = ext_dir.is_relative_to(EXTENSIONS_DIR.resolve())
+    _scan_compose_content(disabled_compose, skip_name_collision=is_builtin)
 
     # Reject symlinks
     st = os.lstat(disabled_compose)
--- a/dream-server/extensions/services/dashboard-api/tests/test_extensions.py
+++ b/dream-server/extensions/services/dashboard-api/tests/test_extensions.py
@@ -5,7 +5,9 @@ import json
 from pathlib import Path
 from unittest.mock import AsyncMock, patch
 
+import pytest
 import yaml
+from fastapi import HTTPException
 from models import ServiceStatus
 
 
@@ -1185,6 +1187,45 @@ class TestComposeScanEdgeCases:
         )
         assert resp.status_code == 400
         assert "dangerous security_opt" in resp.json()["detail"]
+
+
+# --- skip_name_collision flag isolation ---
+
+
+class TestScanComposeSkipNameCollision:
+    """Direct unit tests for the skip_name_collision parameter added for
+    built-in activation (fork issue #338)."""
+
+    def test_rejects_core_name_by_default(self, tmp_path):
+        from routers.extensions import _scan_compose_content
+        compose = tmp_path / "compose.yaml"
+        compose.write_text("services:\n  open-webui:\n    image: test\n")
+        with pytest.raises(HTTPException) as exc:
+            _scan_compose_content(compose, skip_name_collision=False)
+        assert exc.value.status_code == 400
+        assert "conflicts with core service" in exc.value.detail
+
+    def test_allows_core_name_when_skipped(self, tmp_path):
+        from routers.extensions import _scan_compose_content
+        compose = tmp_path / "compose.yaml"
+        compose.write_text("services:\n  open-webui:\n    image: test\n")
+        _scan_compose_content(compose, skip_name_collision=True)
+
+    def test_privileged_still_blocked_when_skipped(self, tmp_path):
+        from routers.extensions import _scan_compose_content
+        compose = tmp_path / "compose.yaml"
+        compose.write_text("services:\n  svc:\n    image: test\n    privileged: true\n")
+        with pytest.raises(HTTPException) as exc:
+            _scan_compose_content(compose, skip_name_collision=True)
+        assert "privileged" in exc.value.detail
+
+    def test_docker_socket_still_blocked_when_skipped(self, tmp_path):
+        from routers.extensions import _scan_compose_content
+        compose = tmp_path / "compose.yaml"
+        compose.write_text("services:\n  svc:\n    image: test\n    volumes:\n      - /var/run/docker.sock:/var/run/docker.sock\n")
+        with pytest.raises(HTTPException) as exc:
+            _scan_compose_content(compose, skip_name_collision=True)
+        assert "Docker socket" in exc.value.detail
 
 
 # --- Size quota enforcement ---


Lightheartdevs (approved):
Surgical fix. `_scan_compose_content` rightfully rejects extensions whose service name collides with a core service — prevents name-shadowing attacks. But built-in extensions (n8n, tts, whisper) legitimately declare their own service name, so the check false-positives when the extension-enable flow re-scans their compose for TOCTOU.

The `skip_name_collision` parameter threads cleanly: defaults to `False` (safe for user extensions), set to `True` only for built-in activation paths. All other scans (privileged, Docker socket, host networking, etc.) still run — verified in the test suite you added.

Tests are thorough: rejects-by-default, allows-when-skipped, and critically **still blocks privileged/socket even when skipped**. Good separation of concerns.

**Cross-PR dependency note:** #907 uses this new parameter (`_scan_compose_content(disabled_compose, skip_name_collision=is_builtin)`). #907 must merge after this one or be rebased. Ship this first.

PR #928, opened 2026-04-12
fix(dashboard): make theme tokens alpha-aware in Tailwind config

### LINKED ISSUE #336: PR #895 regression: bg-theme-card/95 never compiles — feature-card info tooltip ends up with transparent background on ALL themes
## Summary

PR #895 (`fix(dashboard): use theme token for info tooltip background`, merged as part of the integration-test branch of 17 open PRs) replaces a working hardcoded arbitrary-value background class with a theme-token class that **Tailwind cannot compile** given the current `tailwind.config.js`. The result is a transparent background on the FeatureCard info tooltip on every theme (dream, light, lemonade, arctic), not just light — objectively worse than the pre-PR behavior.

This was not caught by static review because the class name is correctly present in the JSX source. It only surfaces when you build the CSS and check whether a matching rule exists.

## The change

From `dream-server/extensions/services/dashboard/src/pages/Dashboard.jsx:518` (FeatureCard info tooltip):

```diff
-        <div className="pointer-events-none absolute bottom-[calc(100%+0.45rem)] right-0 z-20 w-52 rounded-lg border border-white/10 bg-[#0d0b12]/95 px-3 py-2 ...">
+        <div className="pointer-events-none absolute bottom-[calc(100%+0.45rem)] right-0 z-20 w-52 rounded-lg border border-white/10 bg-theme-card/95 px-3 py-2 ...">
```

## Why the old class worked and the new one doesn't

Tailwind's arbitrary-value syntax `bg-[#0d0b12]/95` is parsed by Tailwind itself: it reads the hex, converts to RGB, and emits `rgba(13, 11, 18, 0.95)` in the generated CSS:

```css
.bg-\[\#0d0b12\]\/95 { background-color: rgba(13, 11, 18, 0.95) }
```

The new class `bg-theme-card/95` uses a **named theme color** that's declared in `tailwind.config.js` as a raw CSS custom property reference:

```js
// dream-server/extensions/services/dashboard/tailwind.config.js
theme: {
  card: 'var(--theme-card)',  // ← just a raw var, no <alpha-value> placeholder
  bg: 'var(--theme-bg)',
  ...
}
```

Because the color value is a raw `var(--theme-card)` string (with no `<alpha-value>` placeholder), Tailwind has **no way to inject alpha** and therefore **does not generate the `/95` variant rule at all**. The class name ends up in the DOM, but no matching CSS rule exists.

## Evidence

On the integration branch, after `npm run build` and rebuilding/recreating the dashboard container (`index-B5NJ5w4H.js`):

**1. The class is present in the rendered JSX** (confirmed via Chrome DOM inspection):
```
<div class="pointer-events-none absolute bottom-[calc(100%+0.45rem)] right-0 z-20 w-52 rounded-lg border border-white/10 bg-theme-card/95 px-3 py-2 ...">
```
`hasThemeCardClass: true`, `hasHardcodedHex: false`.

**2. The CSS contains the base rule but NOT the `/95` variant:**
```bash
$ curl -s http://localhost:3001/assets/index-ZKvbaKnW.css | grep -oE '\.bg-theme-card[^{]*\{[^}]*'
.bg-theme-card{background-color:var(--theme-card)  ← base rule, no opacity variant
```
```bash
$ curl -s http://localhost:3001/assets/index-ZKvbaKnW.css | grep -c 'bg-theme-card\\/95'
0   ← no rule at all for the /95 variant
```

**3. At runtime on a light theme, the tooltip computed background-color is fully transparent:**
```js
getComputedStyle(tooltip).backgroundColor
// → 'rgba(0, 0, 0, 0)'
```
`--theme-card` is correctly set to `#FFFFFF` for the light theme, but it isn't being applied to this element because no rule matches the class.

**4. Same issue affects several other pre-existing classes** (not caused by this PR, but caused by the same tailwind config gap):
- `bg-theme-bg/50` — 0 rules in CSS
- `bg-theme-bg/80` — 0 rules in CSS
- `bg-theme-text-muted/45` — 0 rules in CSS

These were already broken on `Light-Heart-Labs/DreamServer main` before this PR. PR #895 is the first PR where silently removing the working hardcoded tooltip background makes the bug *user-visible*.

## Pre-PR vs post-PR comparison

| Theme | Before PR #895 | After PR #895 |
|---|---|---|
| `dream` (dark) | `rgba(13, 11, 18, 0.95)` — near-black panel, correct | transparent — hover shows through to card below |
| `light` | `rgba(13, 11, 18, 0.95)` — ugly near-black rectangle on white cards (this is what the PR was trying to fix) | transparent — text floats over the card below |
| `lemonade` | `rgba(13, 11, 18, 0.95)` — ugly near-black rectangle | transparent |
| `arctic` | `rgba(13, 11, 18, 0.95)` — ugly near-black rectangle | transparent |

So the PR trades \"ugly\" for \"invisible\" on light themes, and downgrades \"correct\" to \"invisible\" on the dark theme. On the dark theme this is a **clear net regression**.

### Option A — fix `tailwind.config.js` so theme tokens support alpha modifiers (preferred)

Requires every `--theme-*` CSS var to be in `R G B` space-separated format instead of `#rrggbb`, then declare the Tailwind colors with `<alpha-value>`:

```js
// tailwind.config.js
theme: {
  bg:  'rgb(var(--theme-bg) / <alpha-value>)',
  card: 'rgb(var(--theme-card) / <alpha-value>)',
  border: 'rgb(var(--theme-border) / <alpha-value>)',
  ...
}
```

And update `index.css`:
```diff
 :root {
-  --theme-bg: #0f0f13;
-  --theme-card: #18181b;
+  --theme-bg: 15 15 19;
+  --theme-card: 24 24 27;
 }
 :root[data-theme=light] {
-  --theme-bg: #F4F4F5;
-  --theme-card: #FFFFFF;
+  --theme-bg: 244 244 245;
+  --theme-card: 255 255 255;
 }
 ... (same for lemonade, arctic)
```

This is the Tailwind-idiomatic pattern and fixes all 4 currently-broken classes (`bg-theme-card/95`, `bg-theme-bg/50`, `bg-theme-bg/80`, `bg-theme-text-muted/45`) in one change.

### Option B — use modern CSS `color-mix` directly via arbitrary value

Keep the CSS vars as hex, but rewrite the tooltip class:
```diff
- bg-theme-card/95
+ bg-[color-mix(in_oklab,var(--theme-card)_95%,transparent)]
```
Verbose but works without touching `tailwind.config.js`. Browser support is modern-enough (Chrome 111+, Safari 16.2+, Firefox 113+) for a dev tool.

### Option C — revert PR #895

Restore `bg-[#0d0b12]/95`. Accepts \"ugly on light themes\" as the lesser evil until Option A is done.

**Recommendation:** Option A. It's a one-time config change that fixes PR #895 *and* the three pre-existing broken classes, and it unblocks future use of `/opacity` modifiers on theme tokens.

## Reproduction (integration branch)

1. `git fetch origin fix/dashboard-tooltip-theme-token && git checkout origin/fix/dashboard-tooltip-theme-token` (or use the integration-test branch)
2. `cd dream-server/extensions/services/dashboard && npm run build`
3. Rebuild + recreate the dashboard container so it serves the new `index-*.js`
4. In Chrome DevTools at http://localhost:3001/, hover any feature card info icon
5. Inspect the tooltip `div.absolute` — `getComputedStyle(el).backgroundColor === 'rgba(0, 0, 0, 0)'` regardless of theme
6. `curl -s http://localhost:3001/assets/index-*.css | grep 'bg-theme-card'` — only the base rule exists, no `/95` variant

## Severity

- Net-negative on the `dream` (default) theme: loses a working dark panel.
- Net-neutral on light/lemonade/arctic: swaps an ugly panel for an invisible one.
- Visual-only, no functional break — users can still read the tooltip text over whatever's behind it.

Medium-low severity overall, but it's worth filing because:
- It's the first runtime regression I've observed in any of the 17 open PRs during integration testing.
- The root cause (Tailwind config missing `<alpha-value>`) silently breaks 3 pre-existing classes too and will break any future attempt to use `/opacity` on theme colors — so fixing this unlocks a whole class of future styling work.

## Found during

Integration-test branch (`upstream/main + 17 yasinBursali PRs`) runtime validation on macOS. Rebuilt dashboard image from integration source, recreated container, verified via Chrome DevTools + direct CSS inspection via curl. 5-minute investigation from \"PR #895 marked PASS in static triage\" to \"PR #895 is actually a regression in runtime\".

yasinBursali:
## What
Make the dashboard's theme color tokens compatible with Tailwind's opacity-modifier syntax (e.g. `bg-theme-card/95`), fixing invisible tooltips introduced by PR #895 and 3 pre-existing broken classes.

## Why
`tailwind.config.js` declared theme colors as bare `var(--theme-card)` references. Tailwind's `/95` opacity modifier requires an `<alpha-value>` placeholder in the color definition to inject the alpha channel. Without it, Tailwind silently skips CSS rule generation — the class appears in the DOM but has no matching rule, so the element renders transparent.

PR #895 replaced a working `bg-[#0d0b12]/95` with `bg-theme-card/95` (correct intent — respect the active theme) but the config gap meant the new class had no CSS. Three older classes — `bg-theme-bg/50`, `bg-theme-bg/80`, `bg-theme-text-muted/45` — had the same latent issue.

## How
**`tailwind.config.js`**: rewrite all 11 theme color keys from `'var(--theme-X)'` to `'rgb(var(--theme-X) / <alpha-value>)'`.

**`src/index.css`**: convert 44 `--theme-*` color vars (11 tokens × 4 themes: dream, light, lemonade, arctic) from hex (`#rrggbb`) to space-separated RGB triplets (`R G B`). Wrap 12 raw CSS consumers (linear-gradient fallbacks, color-mix border, dashboard-shell background) with `rgb()` so they produce valid CSS from the new triplet format. Non-color vars and scrollbar/focus vars not in the Tailwind config are left as hex.

### Automated
- ESLint: 0 errors, 5 pre-existing warnings
- Vite build: `✓ built in 1.25s`
- CSS output grep: 46 opacity-modified rules now compile (was 0 before), including `bg-theme-card/95`, `bg-theme-bg/50`, `bg-theme-bg/80`, `bg-theme-text-muted/45`

### Runtime (Chrome)
Dashboard rebuilt + recreated, healthy.

**Dream theme** tooltip: `bgColor: "rgba(24, 24, 27, 0.95)"` (was `rgba(0, 0, 0, 0)` = transparent)
**Light theme** tooltip: `bgColor: "rgba(255, 255, 255, 0.95)"`

Theme restored to `dream` after testing.

Open the dashboard, hover over a feature card's info icon. The tooltip should have a visible semi-transparent background matching the active theme's card color at 95% opacity. Test on all 4 themes (dream, light, lemonade, arctic).

## Platform Impact
- **macOS / Linux / Windows**: all affected equally — CSS runs in the browser, same on all platforms.

## Relationship to PR #895
PR #895 made the class-name change (`bg-[#0d0b12]/95` → `bg-theme-card/95`). This PR is the companion config fix that makes Tailwind actually emit CSS for that class. Both are needed together for correct behavior.

## Fork issue
Closes yasinBursali/DreamServer#336

### DIFF
--- a/dream-server/extensions/services/dashboard/src/index.css
+++ b/dream-server/extensions/services/dashboard/src/index.css
@@ -6,16 +6,16 @@
 
 /* Dream — default dark theme with indigo accents */
 :root, :root[data-theme="dream"] {
-  --theme-bg: #0f0f13;
-  --theme-card: #18181b;
-  --theme-border: #27272a;
-  --theme-text: #e4e4e7;
-  --theme-text-secondary: #a1a1aa;
-  --theme-text-muted: #71717a;
-  --theme-accent: #9d00ff;
-  --theme-accent-hover: #8900df;
-  --theme-accent-light: #d7a4ff;
-  --theme-surface-hover: #27272a;
+  --theme-bg: 15 15 19;
+  --theme-card: 24 24 27;
+  --theme-border: 39 39 42;
+  --theme-text: 228 228 231;
+  --theme-text-secondary: 161 161 170;
+  --theme-text-muted: 113 113 122;
+  --theme-accent: 157 0 255;
+  --theme-accent-hover: 137 0 223;
+  --theme-accent-light: 215 164 255;
+  --theme-surface-hover: 39 39 42;
   --theme-scrollbar-track: #18181b;
   --theme-scrollbar-thumb: #3f3f46;
   --theme-scrollbar-hover: #52525b;
@@ -23,7 +23,7 @@
   --theme-loading-logo: #d7a4ff;
   --theme-gradient-from: rgba(157,0,255,0.42);
   --theme-gradient-to: rgba(215,164,255,0.26);
-  --theme-sidebar: #18181b;
+  --theme-sidebar: 24 24 27;
   --dashboard-shell-base: #050507;
   --dashboard-shell-grid: rgba(255, 255, 255, 0.028);
   --dashboard-shell-depth: rgba(255, 255, 255, 0.014);
@@ -32,8 +32,8 @@
   --theme-card-shadow: none;
   --theme-card-radius: 12px;
   --liquid-border-size: 1px;
-  --liquid-card-fill: linear-gradient(var(--theme-card), var(--theme-card));
-  --liquid-soft-fill: linear-gradient(var(--theme-card), var(--theme-card));
+  --liquid-card-fill: linear-gradient(rgb(var(--theme-card)), rgb(var(--theme-card)));
+  --liquid-soft-fill: linear-gradient(rgb(var(--theme-card)), rgb(var(--theme-card)));
   --liquid-pressed-shadow:
     inset 0 1px 0 rgba(255,255,255,0.12),
     inset 0 -18px 28px rgba(0,0,0,0.18);
@@ -67,16 +67,16 @@
 
 /* Lemonade — warm parchment theme matching lemonade-server.ai branding */
 :root[data-theme="lemonade"] {
-  --theme-bg: #F0E8CF;
-  --theme-card: #FDFBF3;
-  --theme-border: #E4DABB;
-  --theme-text: #262626;
-  --theme-text-secondary: #6B665C;
-  --theme-text-muted: #8A8478;
-  --theme-accent: #F5C518;
-  --theme-accent-hover: #E2B20F;
-  --theme-accent-light: #F2DE88;
-  --theme-surface-hover: #F3E9CF;
+  --theme-bg: 240 232 207;
+  --theme-card: 253 251 243;
+  --theme-border: 228 218 187;
+  --theme-text: 38 38 38;
+  --theme-text-secondary: 107 102 92;
+  --theme-text-muted: 138 132 120;
+  --theme-accent: 245 197 24;
+  --theme-accent-hover: 226 178 15;
+  --theme-accent-light: 242 222 136;
+  --theme-surface-hover: 243 233 207;
   --theme-scrollbar-track: #F6EFD9;
   --theme-scrollbar-thumb: #E1D4B0;
   --theme-scrollbar-hover: #D0C4A0;
@@ -84,7 +84,7 @@
   --theme-loading-logo: #E2B20F;
   --theme-gradient-from: rgba(245,197,24,0.15);
   --theme-gradient-to: rgba(226,178,15,0.15);
-  --theme-sidebar: #EFE4C6;
+  --theme-sidebar: 239 228 198;
   --dashboard-shell-base: #f4ecd5;
   --dashboard-shell-grid: rgba(88, 72, 32, 0.05);
   --dashboard-shell-depth: rgba(88, 72, 32, 0.03);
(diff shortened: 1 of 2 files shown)


yasinBursali:
Replaced by combined PR

PR #935, opened 2026-04-12
fix(dream-cli): template apply resilience + restart cache invalidation

yasinBursali:
> **Merge order:** Merge after #909, #926, and #907 — all modify `extensions.py` and/or `dream-cli`.

Combined PR replacing #903 and #904. See commit message.

Replaces: #903, #904

### DIFF
--- a/dream-server/dream-cli
+++ b/dream-server/dream-cli
@@ -760,12 +760,12 @@ cmd_restart() {
 
     if [[ -z "$service" ]]; then
         log "Restarting all services..."
-        docker compose "${flags[@]}" restart
+        docker compose "${flags[@]}" up -d
         success "All services restarted"
     else
         service=$(resolve_service "$service")
         log "Restarting $service..."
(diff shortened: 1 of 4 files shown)


Lightheartdevs (approved):
Three reasonable fixes bundled:

1. **`docker compose restart` → `docker compose up -d`**: correct. `restart` reuses the existing container (baked env); `up -d` recreates with fresh env from `.env`. This is the canonical way to pick up model/config changes.

2. **`( cmd_enable "$svc" )` subshell in template apply**: isolates `set -euo pipefail` so one service failure doesn't kill the loop. The `|| warn` already handles the failure message — the subshell is needed to make it actually work. Subtle but correct.

3. **Fallback to `resolve-compose-stack.sh` in `dream-macos.sh`**: good — when `.compose-flags` cache is missing, falling through to static overlay list would miss user-installed extensions. Delegating to the resolver fixes that.

All three are small, correct, and well-motivated. Ship.

PR #945, opened 2026-04-12
fix(bootstrap): replace docker start with docker rm in fallback path

yasinBursali:
> **Merge order:** Merge after #934 — both modify `bootstrap-upgrade.sh`.

## Summary
- The fallback path in Phase 5 of `bootstrap-upgrade.sh` called `docker start` which reuses the baked container environment (`GGUF_FILE=2B model`)
- If the bootstrap model was already deleted in Phase 4b, this puts the container into a crash loop
- Replaced with `docker rm` + status write so `dream restart` launches a fresh container with the correct model

## Test plan
- [ ] Run bootstrap upgrade on a system with no `.compose-flags` file — should rm the container and write "failed" status
- [ ] Verify `dream restart` afterward creates a fresh container with the upgraded model

### DIFF
--- a/dream-server/scripts/bootstrap-upgrade.sh
+++ b/dream-server/scripts/bootstrap-upgrade.sh
@@ -359,8 +359,14 @@ if [[ -n "$DOCKER_CMD" ]] && $DOCKER_CMD ps --filter name=dream-llama-server --f
             $DOCKER_COMPOSE_CMD "${COMPOSE_ARGS[@]}" stop llama-server 2>&1 || true
             $DOCKER_COMPOSE_CMD "${COMPOSE_ARGS[@]}" up -d llama-server 2>&1 || true
         else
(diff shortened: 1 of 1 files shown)


Lightheartdevs (approved):
Correct fix. When `.compose-flags` is absent, `docker start` reuses the baked-in container environment — which points at the bootstrap model that Phase 4b just deleted. Result: crash-loop. Switching to `docker rm` forces a fresh container that re-reads `GGUF_FILE` from `.env`.

Also correctly adds:
- User-visible `WARNING` log so they know to run `dream restart`
- `write_status "failed"` so upstream monitoring picks up the problem instead of hanging on "success"

**Cross-PR note:** draft PR #975 contains this same change as part of its much larger scope. If this merges first (recommended — it's a focused fix), #975 needs to drop the corresponding lines. I've already flagged the split in my #975 review. Ship this one now.

PR #950, opened 2026-04-13
fix(helpers): add TTL cache to dir_size_gb to prevent blocking rglob walks

boffin-dmytro:
## Summary

- `dir_size_gb()` uses `path.rglob("*")` which recursively walks every file in a directory tree — on large model directories (tens of GB, thousands of files) this blocks the request thread on every call
- Added a per-path TTL cache (60s) so repeated calls from the resources, extensions, and storage endpoints return cached values instead of re-walking the filesystem
- Added `invalidate_dir_size_cache()` and wired it into `purge_extension_data` so the cache stays accurate after mutations

## Test plan

- [ ] Hit `/api/services/resources` multiple times within 60s — second call should return instantly without re-walking disk
- [ ] Purge an extension via `/api/extensions/{id}/data` — subsequent resource calls should reflect the freed space
- [ ] Verify existing tests still pass (all tests mock `dir_size_gb`, so no test changes needed)

### DIFF
--- a/dream-server/extensions/services/dashboard-api/helpers.py
+++ b/dream-server/extensions/services/dashboard-api/helpers.py
@@ -17,6 +17,31 @@ import httpx
 from config import SERVICES, INSTALL_DIR, DATA_DIR, LLM_BACKEND
 from models import ServiceStatus, DiskUsage, ModelInfo, BootstrapStatus
 
+
+class _DirSizeCache:
+    """Per-path TTL cache for dir_size_gb to avoid repeated rglob walks."""
+
+    def __init__(self, ttl: float = 60.0):
+        self._ttl = ttl
+        self._store: dict[str, tuple[float, float]] = {}
+
+    def get(self, path: Path) -> float | None:
+        key = str(path.resolve())
+        entry = self._store.get(key)
+        if entry is None:
+            return None
+        expires_at, value = entry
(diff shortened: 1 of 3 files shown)


Lightheartdevs (dismissed):
TTL cache is the right shape for `dir_size_gb`. Reviewed the implementation:

- 60s TTL is reasonable — dashboard polls at ~5s intervals so this cuts rglob walks by an order of magnitude.
- `invalidate_dir_size_cache()` correctly called after `shutil.rmtree` in `purge_extension_data` so the next poll doesn't return stale size.
- `clear_dir_size_cache()` is a nice escape hatch but I didn't see it called anywhere — consider adding it to `_clear_settings_caches` if settings changes can move data dirs.

**Minor note:** the cache uses `path.resolve()` as key — good because it canonicalizes symlinks, but beware if callers pass relative paths from different cwds during a single process lifetime. Current callers all pass absolute paths, so fine for now.

The unrelated update-script-path resolution in `routers/updates.py` is a small bonus cleanup (collapses a confusing nested conditional). Ship.

Lightheartdevs (dismissed):
Re-submitting — prior approval review was dismissed (likely by an automation). TTL cache + invalidation pattern is correct, tests look good, 60s TTL is sensible for dashboard polling cadence. Details in the dismissed review.

PR #952, opened 2026-04-13
fix(installer): replace silent || true masks with warnings in phase scripts

boffin-dmytro:
## Summary

- **Phase 06 (directories)**: source copy fallback now warns when `cp` fails instead of silently proceeding with missing files; chmod failures warn that scripts may not be executable; OpenClaw default config copy failure now warns the user
- **Phase 11 (services)**: `.env` patching loop now checks each awk/cat/rm chain individually and warns on failure instead of always printing success
- Errors are logged to the install log file instead of discarded to `/dev/null`

## Test plan

- [ ] Run installer with `--dry-run` to verify no syntax errors in modified phase scripts
- [ ] Verify warnings appear in `/tmp/dream-server-install.log` when copy/patch operations fail
- [ ] Run `bash -n` on both modified phase scripts to confirm no syntax errors

### DIFF
--- a/dream-server/extensions/services/dashboard-api/routers/updates.py
+++ b/dream-server/extensions/services/dashboard-api/routers/updates.py
@@ -227,6 +227,7 @@ async def get_update_dry_run():
     latest: Optional[str] = None
     changelog_url: Optional[str] = None
     update_available = False
+    version_check_error: Optional[str] = None
 
     try:
         req = urllib.request.Request(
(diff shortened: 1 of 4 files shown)


Lightheartdevs (approved):
Exactly the right fix — directly addresses the CLAUDE.md rule (*"Never `|| true` or `2>/dev/null`"*). Converts silent failure masks into named `warn` calls with specific messages ("Source copy incomplete — some files may be missing", etc.) and routes stderr to `LOG_FILE` for diagnostics.

The awk pipeline refactor in 11-services.sh is a nice bonus: loop-based approach is more readable and lets each key patch report failure independently.

**Flag:** #953 (BATS tests) appears to carry these same changes. Land this one first, then #953 can rebase cleanly.

Lightheartdevs (approved):
Focused change, good direction, and current checks are green. Replacing silent masks with visible warnings improves operator feedback without changing the happy path. Approved.

PR #965, opened 2026-04-15
fix(token-spy): bundle dashboard charts for offline installs

boffin-dmytro:
## Summary

This removes the shipped Token Spy dashboard's dependency on jsDelivr so the dashboard keeps working in DreamServer offline installs.

Before this change, the dashboard loaded Chart.js and the date adapter from `cdn.jsdelivr.net`. In offline or air-gapped environments, the page would still load but all chart visualizations would fail.

## What changed

- remove external Chart.js and date-adapter script tags from the shipped dashboard HTML
- add a bundled local chart helper at `extensions/services/token-spy/dashboard_charts.js`
- serve that asset from Token Spy at `/dashboard-assets/charts.js`
- replace the Chart.js-based render calls with local canvas rendering for line, stacked bar, and doughnut charts
- add a contract check to prevent jsDelivr chart dependencies from returning to the shipped Token Spy dashboard

## Validation

- `python3 -m py_compile dream-server/extensions/services/token-spy/main.py`
- `node --check dream-server/extensions/services/token-spy/dashboard_charts.js`
- `bash -n dream-server/tests/contracts/test-installer-contracts.sh`
- targeted offline asset check:
  - local `dashboard_charts.js` exists
  - `main.py` references `/dashboard-assets/charts.js`
  - `main.py` no longer references jsDelivr Chart.js assets

## Notes

- this PR is intentionally scoped to the shipped Token Spy dashboard only
- it does not change the separate dashboard service, which still has its own external asset usage
- full `tests/contracts/test-installer-contracts.sh` could not be run locally because `jq` is not installed in this environment

### DIFF
--- a/dream-server/extensions/services/token-spy/dashboard_charts.js
+++ b/dream-server/extensions/services/token-spy/dashboard_charts.js
@@ -1,0 +1,318 @@
+(function () {
+  const COLORS = {
+    text: '#e6edf3',
+    muted: '#8b949e',
+    grid: '#21262d',
+  };
+
+  function getContext(canvas) {
+    const dpr = window.devicePixelRatio || 1;
+    const width = Math.max(canvas.clientWidth || 320, 320);
+    const height = Math.max(parseInt(canvas.dataset.height || '', 10) || canvas.clientHeight || 280, 220);
+    canvas.width = Math.floor(width * dpr);
+    canvas.height = Math.floor(height * dpr);
+    canvas.style.width = width + 'px';
+    canvas.style.height = height + 'px';
+    const ctx = canvas.getContext('2d');
+    ctx.setTransform(dpr, 0, 0, dpr, 0, 0);
+    ctx.clearRect(0, 0, width, height);
+    return { ctx, width, height };
+  }
+
+  function legend(target, items) {
+    if (!target) return;
+    target.innerHTML = items.map(item =>
+      '<span class="chart-legend-item"><span class="chart-legend-swatch" style="background:' + item.color +
+      '"></span><span>' + item.label + '</span></span>'
+    ).join('');
+  }
+
+  function noData(ctx, width, height) {
+    ctx.fillStyle = COLORS.muted;
+    ctx.font = '13px system-ui, sans-serif';
+    ctx.textAlign = 'center';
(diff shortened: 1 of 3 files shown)


Lightheartdevs (approved):
Good offline-install fix. Replaces jsDelivr Chart.js CDN deps with a 318-line bundled canvas renderer ([`dashboard_charts.js`](dream-server/extensions/services/token-spy/dashboard_charts.js)) served from `/dashboard-assets/charts.js`.

Spot-checked:
- Canvas-based line/stacked-bar/doughnut implementations look correct (DPR handling, tick generation, no-data fallbacks).
- Contract test addition in `test-installer-contracts.sh` locks in the no-CDN invariant.
- `main.py` references `/dashboard-assets/charts.js` and no longer references jsDelivr.

**Minor observation:** 318 lines of hand-rolled charting is real maintenance. If you ever need richer chart types (scatter, heatmap, stacked-area), the tradeoff reverses and a bundled Chart.js in the container image becomes cheaper. For the current 3 chart types, this is fine.

**Follow-up noted by author:** the separate dashboard service still has external asset usage. Worth tracking as a separate issue if offline parity is a goal across both. Ship.

Lightheartdevs:
Two minor follow-up observations from a deep re-read of `dashboard_charts.js` — non-blocking, worth a one-line fix when convenient:

1. **`line()` doesn't call `noData()` on empty input.** `stackedBar()` and `donut()` both do:
   ```javascript
   if (!labels.length || !datasets.length) {
     legend(legendEl, []);
     noData(ctx, width, height);
     return;
   }
   ```
   But `line()`'s empty-data branch returns without drawing the "No chart data for this time range" text, so an empty line chart renders as bare axes instead of the same friendly message. Cosmetic, not correctness.

2. **NaN y-values aren't filtered.** `(s.points || []).filter(p => Number.isFinite(+new Date(p.x)))` strips bad timestamps but a NaN `y` would produce a jump in the path. Consider:
   ```javascript
   const points = (s.points || []).filter(p =>
     Number.isFinite(+new Date(p.x)) && Number.isFinite(Number(p.y))
   );
   ```

Neither blocks merge — the PR is a clear net improvement over the CDN dependency. Happy to ship as-is.

PR #977, opened 2026-04-17
fix(healthcheck): use 127.0.0.1 instead of localhost/hostname

yasinBursali:
## What
Replace `localhost` and `os.hostname()` with `127.0.0.1` in DreamForge and Perplexica healthchecks.

## Why
DreamForge used `localhost` which can resolve to `::1` (IPv6) on dual-stack kernels — the server binds to `0.0.0.0` (IPv4 only), so the IPv6 connection fails. Perplexica used `require('os').hostname()` which relies on Docker DNS, adding fragility. Both violated the project convention used by all other 17 services.

## How
Three one-line substitutions:
- `dreamforge/compose.yaml` healthcheck: `localhost` → `127.0.0.1`
- `dreamforge/Dockerfile.rust` HEALTHCHECK: `localhost` → `127.0.0.1`
- `perplexica/compose.yaml` healthcheck: `require('os').hostname()` → `127.0.0.1`

## Testing
- YAML valid, `127.0.0.1` confirmed in all 3 files
- Live test: both containers report `healthy` with new healthchecks
- All 19/19 services now consistently use `127.0.0.1`

## Platform Impact
- **macOS:** Fixed (Docker VM dual-stack)
- **Linux:** Fixed (kernel IPv6 loopback)
- **Windows/WSL2:** Fixed (Docker Desktop)

### DIFF
--- a/dream-server/extensions/services/dreamforge/Dockerfile.rust
+++ b/dream-server/extensions/services/dreamforge/Dockerfile.rust
@@ -43,6 +43,6 @@ EXPOSE 3010
 VOLUME ["/workspace", "/data/dreamforge"]
 
 HEALTHCHECK --interval=30s --timeout=5s --retries=3 \
-    CMD curl -f http://localhost:3010/health || exit 1
+    CMD curl -f http://127.0.0.1:3010/health || exit 1
 
 ENTRYPOINT ["dreamforge-server"]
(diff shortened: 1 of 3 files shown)


Lightheartdevs (approved):
Verified: dreamforge Dockerfile.rust sets `DREAMFORGE_HOST=0.0.0.0`, so container-internal `127.0.0.1` resolves to the listener. Reasoning about IPv6 `::1` ambiguity on dual-stack kernels is accurate. All 3 healthchecks now consistent with the 17 other services in the stack. Ship.

PR #978, opened 2026-04-17
fix(shell): replace find|head with find -print -quit

yasinBursali:
## What
Replace `find ... | head -1` with `find ... -print -quit` in two installer scripts.

## Why
`find | head -1` under `set -euo pipefail` can produce SIGPIPE (exit 141) when find returns multiple results. `head` closes the pipe after reading one line, sending SIGPIPE to `find`, which `pipefail` propagates as a fatal error.

## How
Two one-line substitutions using `-print -quit` which stops `find` after the first match without any pipe:
- `installers/lib/detection.sh` — Linux secure boot sign-file detection
- `installers/macos/install-macos.sh` — llama-server binary extraction

`-print -quit` is portable across both GNU find (Linux) and BSD find (macOS 10.11+).

## Testing
- shellcheck clean (pre-existing SC2034 warnings only)
- `-print -quit` verified portable across GNU/BSD find

## Platform Impact
- **macOS:** Fixed (install-macos.sh)
- **Linux:** Fixed (detection.sh)
- **Windows/WSL2:** Fixed (shares Linux detection.sh)

### DIFF
--- a/dream-server/installers/lib/detection.sh
+++ b/dream-server/installers/lib/detection.sh
@@ -390,7 +390,7 @@ fix_nvidia_secure_boot() {
         fi
     done
     if [[ -z "$sign_file" ]]; then
-        sign_file=$(find /usr/src /usr/lib -name sign-file -executable 2>/dev/null | head -1)
+        sign_file=$(find /usr/src /usr/lib -name sign-file -executable -print -quit 2>/dev/null)
     fi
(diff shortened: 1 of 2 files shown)


Lightheartdevs (approved):
Semantically equivalent to `| head -1` but avoids SIGPIPE on the `find` process. `-quit` is supported by both GNU and BSD `find` (covers Linux + macOS). Ship.

PR #979, opened 2026-04-17
fix(security): generate dedicated DREAM_AGENT_KEY in installers

yasinBursali:
## What
Generate a dedicated `DREAM_AGENT_KEY` in both macOS and Linux installers.

## Why
`DREAM_AGENT_KEY` was registered in `.env.schema.json` and preferred by both the host agent and dashboard-api, but never generated by either installer. All installations fell back to `DASHBOARD_API_KEY`, meaning the browser-facing dashboard key doubled as the privileged host agent credential (container management, `.env` writes, extension hooks).

## How
Added 32-byte hex key generation to both Bash-based installers:
- **macOS** (`env-generator.sh`): `new_secure_hex 32` — matches `dashboard_api_key` pattern
- **Linux** (`06-directories.sh`): `_env_get DREAM_AGENT_KEY` with `openssl rand` — uses re-install merge logic to preserve existing keys

No changes needed to host agent, dashboard-api, compose, or schema — wiring already exists.

## Testing
- shellcheck clean
- Live test: `.env` contains `DREAM_AGENT_KEY=<64-char hex>`
- Schema validation passes (key already registered in `.env.schema.json`)
- Existing installs retain `DASHBOARD_API_KEY` fallback (graceful degradation)

## Known Considerations
- Windows installer (`env-generator.ps1`) not updated — tracked separately. DASHBOARD_API_KEY fallback prevents runtime failure.
- Existing macOS installs without `--force` won't get the key (early-return on existing `.env`). Fallback covers this.

## Platform Impact
- **macOS:** Fixed — new installs get dedicated key
- **Linux:** Fixed — new installs get key, re-installs preserve existing
- **Windows/WSL2:** Fallback covers — follow-up planned

### DIFF
--- a/dream-server/installers/macos/lib/env-generator.sh
+++ b/dream-server/installers/macos/lib/env-generator.sh
@@ -132,6 +132,8 @@ generate_dream_env() {
     livekit_api_key=$(new_secure_hex 16)
     local dashboard_api_key
     dashboard_api_key=$(new_secure_hex 32)
+    local dream_agent_key
+    dream_agent_key=$(new_secure_hex 32)
     local openclaw_token
     openclaw_token=$(new_secure_hex 24)
     local qdrant_api_key
@@ -221,6 +223,7 @@ LANGFUSE_PORT=3006
 #=== Security (auto-generated, keep secret!) ===
 WEBUI_SECRET=${webui_secret}
 DASHBOARD_API_KEY=${dashboard_api_key}
+DREAM_AGENT_KEY=${dream_agent_key}
 N8N_USER=admin@dreamserver.local
 N8N_PASS=${n8n_pass}
 LITELLM_KEY=${litellm_key}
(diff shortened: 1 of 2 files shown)


Lightheartdevs (approved):
The problem is real — `DREAM_AGENT_KEY` is read in 10+ places (config.py:303, extensions.py, privacy.py, models.py, resources.py, main.py, host agent) via `DREAM_AGENT_KEY or DASHBOARD_API_KEY`. Until this PR, the browser-facing dashboard key doubled as the privileged host-agent credential used for container management, `.env` writes, and extension hooks. Good catch.

Fix is correct:
- macOS `env-generator.sh`: `new_secure_hex 32` matches the dashboard_api_key pattern.
- Linux `06-directories.sh`: `_env_get` preserves existing keys on re-install.

**Known gaps (acknowledged by author):**
- Windows `env-generator.ps1` not updated — fallback covers, but should be tracked as a follow-up.
- Existing macOS installs without `--force` won't get the key. Fallback covers, but new installs and re-runs should eventually converge.

**Note on CI:** `integration-smoke` is failing here but it's also failing on other recent PRs — looks unrelated to this change.

Ship.

PR #980, opened 2026-04-17
fix: portable ANSI stripping and narrowed GPU exception

yasinBursali:
## What
Two targeted error handling fixes: portable sed ANSI stripping and narrowed GPU detection exception.

## Why
1. `dream-preflight.sh` used `\x1b` hex escape in sed regex — older BSD sed (macOS) treats this literally, leaving raw ANSI codes in log files.
2. `routers/models.py` caught bare `Exception` in GPU VRAM detection, silently hiding all errors (including programming bugs) per a blanket `return None`.

## How
1. **sed fix:** Changed `\x1b` to `$'\033'` (Bash ANSI-C quoting). The shell resolves the escape before sed sees it — works on all platforms.
2. **Python fix:** Narrowed `except Exception` to `(ImportError, FileNotFoundError, OSError, KeyError, AttributeError)` with `logger.warning()`. Covers I/O-boundary failures while letting programming errors propagate visibly.

The `fitsVram=True` fallback when GPU is unknown is deliberately unchanged — it's the optimistic path that lets users browse and download models even when GPU detection fails.

## Testing
- shellcheck clean, Python compile OK
- pytest: 172/173 pass (1 pre-existing failure in unrelated test)
- `logger` already defined in models.py (line 16)

## Platform Impact
- **macOS:** Fixed (BSD sed portability for older versions)
- **Linux:** Fixed (narrowed exception + logging visibility)
- **Windows/WSL2:** Fixed (same Python code)

### DIFF
--- a/dream-server/dream-preflight.sh
+++ b/dream-server/dream-preflight.sh
@@ -80,7 +80,7 @@ WARN=0
 
 log() {
     echo -e "$1"
-    echo -e "$1" | sed 's/\x1b\[[0-9;]*m//g' >> "$LOG_FILE"
+    echo -e "$1" | sed $'s/\033\\[[0-9;]*m//g' >> "$LOG_FILE"
 }
 pass() { log "${GREEN}✓${NC} $1"; PASS=$((PASS+1)); }
 fail() { log "${RED}✗${NC} $1"; FAIL=$((FAIL+1)); }
--- a/dream-server/extensions/services/dashboard-api/routers/models.py
+++ b/dream-server/extensions/services/dashboard-api/routers/models.py
@@ -79,7 +79,8 @@ def _get_gpu_vram() -> Optional[ModelLibraryGpu]:
             vramUsed=round(used_gb, 1),
             vramFree=round(total_gb - used_gb, 1),
         )
-    except Exception:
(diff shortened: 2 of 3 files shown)


Lightheartdevs (changes requested):
The ANSI stripping fix is correct — BSD `sed` on macOS doesn't reliably accept `\x1b`, and `$'\033...'` ANSI-C quoting is the portable form. Happy to ship that half.

**The narrowed exception list is incomplete.** In `routers/models.py`:

```python
except (ImportError, FileNotFoundError, OSError, KeyError, AttributeError) as exc:
```

`pynvml.NVMLError` is the most common failure mode here (driver not loaded, NVML shared library not found, permission denied). It inherits from `Exception` directly, not from any of the listed types — so this change will crash uncaught where the previous `except Exception` silently returned `None`.

Please add `pynvml.NVMLError` (or `RuntimeError` as a broader net) to the tuple. Otherwise this regresses `_get_gpu_vram()` on any non-NVIDIA or misconfigured host.

Suggestion:

```python
except (ImportError, FileNotFoundError, OSError, KeyError, AttributeError, pynvml.NVMLError) as exc:
```

(And a test asserting `_get_gpu_vram()` returns `None` when pynvml raises `NVMLError_DriverNotLoaded` would lock this down permanently.)

PR #981, opened 2026-04-17
fix(dreamforge): pin image to v0.1.0 for reproducible deployments

yasinBursali:
## What
Pin DreamForge image from `:latest` to `:v0.1.0` across all 4 reference locations.

## Why
DreamForge was the only service using `:latest`. All other 9+ services use pinned version tags. Installs at different times could pull different DreamForge binaries, making regression bugs impossible to bisect.

## How
Changed `:latest` to `:v0.1.0` (from `Cargo.toml`) in all 4 locations:
- `extensions/services/dreamforge/compose.yaml` — runtime image
- `installers/phases/08-images.sh` — pre-pull list
- `installers/phases/11-services.sh` — build-skip check
- `.env.example` — documented override example

The `DREAMFORGE_IMAGE` env var override remains available for user customization.

## Testing
- YAML valid
- grep confirms zero `:latest` references remain for dreamforge
- Install flow coherent: pre-pull → build-skip → compose all use same tag

## Platform Impact
- **macOS:** Compose tag updated
- **Linux:** Full coverage (installer phases + compose)
- **Windows/WSL2:** Compose tag updated

### DIFF
--- a/dream-server/.env.example
+++ b/dream-server/.env.example
@@ -256,7 +256,7 @@ LANGFUSE_INIT_USER_PASSWORD=   # auto-generated during install
 # LLAMA_SERVER_MEMORY_LIMIT=64G
 
 #=== DreamForge (Local Agentic Coding) ===
-# DREAMFORGE_IMAGE=ghcr.io/light-heart-labs/dreamforge:latest
+# DREAMFORGE_IMAGE=ghcr.io/light-heart-labs/dreamforge:v0.1.0
 # DREAMFORGE_PORT=3010
 # DREAMFORGE_PERMISSION_MODE=accept_edits
 # DREAMFORGE_MAX_TURNS=25
--- a/dream-server/extensions/services/dreamforge/compose.yaml
+++ b/dream-server/extensions/services/dreamforge/compose.yaml
@@ -1,6 +1,6 @@
 services:
   dreamforge:
-    image: ${DREAMFORGE_IMAGE:-ghcr.io/light-heart-labs/dreamforge:latest}
+    image: ${DREAMFORGE_IMAGE:-ghcr.io/light-heart-labs/dreamforge:v0.1.0}
     build:
       context: ./extensions/services/dreamforge
       dockerfile: Dockerfile.rust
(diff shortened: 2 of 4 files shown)


Lightheartdevs (dismissed):
All 3 sites updated consistently (.env.example, compose.yaml, 11-services.sh fallback). If the ghcr.io tag doesn't exist, install fails loudly at phase 08 — desired Let-It-Crash behavior rather than silently pulling a moving `latest`. Ship.

Lightheartdevs (changes requested):
**Dismissing my prior approval — deep audit caught a missed site.** Apologies for the churn; this is the value of reviewing twice.

The PR description says "All 3 sites updated consistently (.env.example, compose.yaml, 11-services.sh fallback)." There are actually **four** sites in the tree that reference `ghcr.io/light-heart-labs/dreamforge:latest`:

1. ✅ `dream-server/.env.example:251` — updated
2. ❌ [`dream-server/installers/phases/08-images.sh:50`](dream-server/installers/phases/08-images.sh#L50) — **NOT UPDATED**, still `:latest`
3. ✅ `dream-server/installers/phases/11-services.sh:305` — updated
4. ✅ `dream-server/extensions/services/dreamforge/compose.yaml:3` — updated

```bash
$ grep -rn 'dreamforge:latest\|dreamforge:v0.1.0' .
dream-server/.env.example:251:# DREAMFORGE_IMAGE=ghcr.io/light-heart-labs/dreamforge:latest
dream-server/installers/phases/08-images.sh:50:[[ "${ENABLE_DREAMFORGE:-}" == "true" ]] && PULL_LIST+=("ghcr.io/light-heart-labs/dreamforge:latest|DREAMFORGE — agent system")
dream-server/installers/phases/11-services.sh:305:        _dreamforge_image="${DREAMFORGE_IMAGE:-ghcr.io/light-heart-labs/dreamforge:latest}"
dream-server/extensions/services/dreamforge/compose.yaml:3:    image: ${DREAMFORGE_IMAGE:-ghcr.io/light-heart-labs/dreamforge:latest}
```

**Consequence if this merges as-is:**

- Phase 08 pulls `ghcr.io/light-heart-labs/dreamforge:latest` (whatever's latest at install time).
- Phase 11 runs `$DOCKER_CMD image inspect "ghcr.io/light-heart-labs/dreamforge:v0.1.0"` — the pulled `:latest` doesn't match this tag.
- Phase 11 adds `dreamforge` to `_build_services` and builds from source anyway.

Net result: every install wastes bandwidth pulling an unused `:latest` image, then forces a local build from source. **The PR's stated goal ("reproducible deployments") is not achieved** — builds are not reproducible because `:latest` is a moving target and then the build path is hit on every install.

**Fix:** update line 50 of 08-images.sh to use `:v0.1.0`. One-line change. After that, I'll re-approve.

Thanks for the PR direction — pinning is the right call, it just needs to be consistent across all 4 sites.

PR #998, opened 2026-04-23
fix(dream-cli): pipefail + surface LLM failures + exit-code contract

yasinBursali:
## What

Three independent error-discipline improvements to `dream-cli`:

1. Promote shell options from `set -e` to `set -eo pipefail`, aligning with CLAUDE.md's project standard. Audit `| head -1` sites (converted to `| sed -n 1p` for SIGPIPE-safety).
2. `dream chat` and `dream benchmark` now actually detect LLM failures instead of silently reporting "success" against a dead backend. Uses `curl --fail --show-error --max-time` + `jq -er` strict parsing, plus a `/v1/models` preflight probe.
3. Three commands (`dream disable`, `dream agent logs`, `dream gpu reassign`) now exit non-zero on failure/skip conditions, fixing the exit-code contract for scripted callers.

## Why

Before this PR: `dream chat` against a dead backend printed `[dream] Sending to ...`, emitted an empty line, and exited 0. `dream benchmark` propagated that silent failure and happily reported sub-second "Excellent" performance against a non-running LLM. Root cause: curl was invoked with `-s` (silent, no error-show, no fail-on-HTTP-error) and responses were parsed with `jq -r .choices[0].message.content // .error.message // "Error: no response"` — every failure path masked behind a human-friendly string on stdout.

`dream disable bogus-service` used to silently stop a non-existent container (`|| true`) and print a success-shaped message. Several other commands exited 0 after warning about skip conditions, so automation couldn't distinguish "success" from "skipped."

## How

Three commits, each self-contained:

- `dfaa4c8f` — adopt `set -eo pipefail`; convert two `| head -1` sites to `| sed -n 1p`.
- `385372ec` — rewrite chat/benchmark with preflight probe + strict `curl --fail` + `jq -er`. `cmd_benchmark` uses `if ! response=$(cmd_chat ...)` so chat failure aborts the benchmark. On the main chat call, the flag is bare `--show-error` (no `--fail*`) so HTTP 4xx/5xx responses flow through to the existing `jq` fallback which extracts `.error.message` cleanly — also keeps compat with `curl < 7.76` (Ubuntu 20.04, Debian 11, RHEL 8).
- `10588c2a` — exit-code contract for `cmd_disable`, `cmd_agent logs`, `_gpu_reassign`.

## Testing

- `dream chat "hi"` against a stopped llama-server: now exits 1 with "llama-server not reachable..." (previously silent empty output, exit 0).
- `dream benchmark` against a stopped backend: now aborts with "Benchmark failed: LLM unreachable..." (previously false-positive sub-second "Excellent").
- `dream disable bogus-service`: exits 1 with "Unknown service" (previously exit 0 after silent stop).
- Round-1 review: Critique Guardian approved. Round-2 adversarial audit: verified across 5 failure paths including the `--fail` / no-flag asymmetry on the two curl sites.

## Known Considerations

Nounset (`set -u`) is **not** added in this PR — it requires a full bare-variable audit, which lands as the sibling `refactor(dream-cli): enable set -u` PR on top of this one.

## Platform Impact

- **macOS:** chat/benchmark runs locally against llama-server; curl 8.x tested.
- **Linux Ubuntu 20.04 / Debian 11 / RHEL 8:** the `--fail-with-body` flag (curl ≥ 7.76, Mar 2021) was deliberately avoided on the main chat call so these older-curl distros work too.
- **Windows (WSL2):** same behavior as Linux.

## ⚠️ Merge order — must land #1008 first

Adopting `set -eo pipefail` here turns 7 existing `grep "^KEY=" .env | cut | tr` pipelines in `dream-cli` into hard exits when the key is absent. PR **#1008** (`refactor(dream-cli): guard .env grep pipelines against pipefail kill on missing keys`) appends `|| true` to those 7 sites so the surrounding defensive `[[ -n ... ]]` / `${var:-default}` checks keep working.

If this PR merges alone, `dream update`, `dream rollback`, `dream enable`, `dream preset`, and `dream dry-run` can abort mid-flow on installs that predate (or have manually-edited-out) one of those .env keys. **#1008 is preventive on `main` today and load-bearing the moment this PR lands.** Marked as Draft until #1008 is in to make this gate mechanical.

---

## 📋 Chain status (operator's reminder — apr-25)

This PR sits in the middle of a `dream-cli` strict-mode chain:

```
#1008 (READY) ──▶ #998 (this — DRAFT) ──▶ #1002 (DRAFT)
                                     └──▶ #1018 (DRAFT)
```

**Before this PR can be promoted:**
- #1008 must merge first (the .env grep-pipeline guards above are load-bearing the moment `set -eo pipefail` lands here).

**After this PR merges, the next operator step is:**
1. Rebase #1002 onto post-#998 main. The 6+ shared hunks (line 6 set-line, identical sub-hunks in `cmd_chat`/`cmd_benchmark`/`cmd_disable`/`cmd_agent`/`_gpu_reassign`, `| head -1` → `sed -n '1p'` swaps) drop out automatically. What remains is #1002's actual unique work: the `-u` flag and `:-` empty-string default guards.
2. Promote #1002 from DRAFT → ready (`gh pr ready 1002`).
3. Same pattern for #1018 once its other deps (#994, #1003, #1016) have landed.

#1002 and #1018 are kept in DRAFT as a mechanical safeguard so they cannot merge before this one.

### DIFF
--- a/dream-server/Makefile
+++ b/dream-server/Makefile
@@ -37,6 +37,9 @@ test: ## Run unit and contract tests
 	@echo ""
 	@echo "=== Bind address sweep ==="
 	@bash tests/test-bind-address-sweep.sh
+	@echo ""
+	@echo "=== dream-cli pipefail tolerance ==="
+	@bash tests/test-dream-cli-pipefail-tolerance.sh
 
 bats: ## Run BATS unit tests for shell libraries
 	@echo "=== BATS unit tests ==="
--- a/dream-server/dream-cli
+++ b/dream-server/dream-cli
@@ -1547,6 +1547,7 @@ cmd_disable() {
     # Check built-in extensions first, then dashboard-installed user extensions
     local ext_dir="$INSTALL_DIR/extensions/services/$service_id"
     [[ ! -d "$ext_dir" ]] && ext_dir="$INSTALL_DIR/data/user-extensions/$service_id"
+    [[ -d "$ext_dir" ]] || error "Unknown service: $input"
     local cf="$ext_dir/compose.yaml"
 
     # Check it's not a core service
@@ -2488,6 +2489,7 @@ cmd_agent() {
                 tail -f "$log_file"
             else
                 warn "No log file at $log_file"
+                return 1
             fi
             ;;
         *)
@@ -2849,19 +2851,19 @@ _gpu_reassign() {
 
     if ! command -v nvidia-smi &>/dev/null; then
         warn "GPU reassign is only supported for NVIDIA. Use 'dream gpu status' for AMD."
-        return
+        return 1
     fi
 
     if ! command -v jq &>/dev/null; then
         warn "jq not found — required for GPU reassignment"
-        return
+        return 1
     fi
 
     local topo_lib="$SCRIPT_DIR/installers/lib/nvidia-topo.sh"
     local assign_script="$SCRIPT_DIR/scripts/assign_gpus.py"
 
-    [[ -f "$topo_lib" ]] || { warn "Topology library not found: $topo_lib"; return; }
-    [[ -f "$assign_script" ]] || { warn "Assignment script not found: $assign_script"; return; }
+    [[ -f "$topo_lib" ]] || { warn "Topology library not found: $topo_lib"; return 1; }
+    [[ -f "$assign_script" ]] || { warn "Assignment script not found: $assign_script"; return 1; }
 
     # warn and err are already defined; source is safe
     . "$topo_lib"
--- a/dream-server/dream-cli
+++ b/dream-server/dream-cli
@@ -3,7 +3,7 @@
 # Mission: M5 (Clonable Dream Setup Server)
 # Version: 2.0.0 — Registry-driven service resolution
 
-set -e
+set -eo pipefail
 
 # Require Bash 4+ (associative arrays used by service registry and dream-cli)
 if (( BASH_VERSINFO[0] < 4 )); then
@@ -251,7 +251,7 @@ _check_version_compat() {
 
     # 1. Installed version
     _COMPAT_INSTALLED_VER=$(grep '^DREAM_VERSION=' "$INSTALL_DIR/.env" 2>/dev/null \
-        | head -1 | cut -d= -f2 | tr -d '[:space:]')
+        | sed -n '1p' | cut -d= -f2 | tr -d '[:space:]')
     # Fall back to .version file (written by dream-update.sh / PR #349 convention)
     if [[ -z "$_COMPAT_INSTALLED_VER" && -f "$INSTALL_DIR/.version" ]]; then
         _COMPAT_INSTALLED_VER=$(jq -r '.version // empty' "$INSTALL_DIR/.version" 2>/dev/null \
@@ -1892,7 +1892,7 @@ META
 
             # Validate archive structure
             local archive_name
-            archive_name=$(tar tzf "$archive" 2>/dev/null | head -1 | cut -d/ -f1)
+            archive_name=$(tar tzf "$archive" 2>/dev/null | sed -n '1p' | cut -d/ -f1)
             [[ -z "$archive_name" ]] && error "Invalid archive structure"
 
             # Check if preset already exists
--- a/dream-server/dream-cli
+++ b/dream-server/dream-cli
@@ -1148,20 +1148,44 @@ cmd_chat() {
     local _llm_port="${SERVICE_PORTS[llama-server]:-11434}"
     _llm_port="${OLLAMA_PORT:-${LLAMA_SERVER_PORT:-$_llm_port}}"
 
-    # Get model from llama-server if not specified
+    # Pre-flight reachability check.
+    local _probe
+    _probe=$(curl --silent --show-error --fail --max-time 3 \
+        "http://localhost:${_llm_port}/v1/models" 2>&1) || {
+        error "llama-server not reachable at http://localhost:${_llm_port} — is 'dream status' showing it healthy?"
+    }
+
     if [[ -z "$model" ]]; then
-        model=$(curl -s "http://localhost:${_llm_port}/v1/models" | jq -r '.data[0].id // "local"')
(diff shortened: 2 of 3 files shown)


Lightheartdevs:
Audit follow-up: keep draft / rebase on current `main`.

The focused pipefail guard from #1008 is now merged. This draft still needs to absorb or preserve that fix: `_check_version_compat` must tolerate a `.env` without `DREAM_VERSION` and fall back to `.version`/`manifest.json` instead of exiting under pipefail.

yasinBursali:
Pushed audit follow-up — rebased onto current `upstream/main`, conflict resolved, audit-required `|| true` preserved, 4-case regression test added, plus 3 latent split-local pipefail bombs fixed (CG follow-up).

**Audit resolution:**
The conflict in `_check_version_compat` was exactly the audit's target — branch's commit `dfaa4c8f` had stripped the `|| true` from the `DREAM_VERSION` grep pipeline; current `upstream/main` (post-#1008) carries the audit-required tolerance. Conflict resolution kept upstream's `|| true` form with an inline rationale comment citing #1008 + this audit so a future "cleanup" reviewer cannot remove it without context.

```bash
_COMPAT_INSTALLED_VER=$(grep '^DREAM_VERSION=' "$INSTALL_DIR/.env" 2>/dev/null \
    | sed -n '1p' | cut -d= -f2 | tr -d '[:space:]' || true)
```

The `.version` and `manifest.json` fallback branches that follow are intact and reachable.

**Regression test (`9d7aa45e`):** `tests/test-dream-cli-version-compat-pipefail.sh` (4 cases, wired into `make test`):
- dream-cli runs under strict mode (precondition).
- DREAM_VERSION grep pipeline ends with `|| true` (active code; comments stripped before grep so the rationale block alone can't satisfy the assertion).
- `.version` and `manifest.json` fallback branches present.
- Anti-regression: bare-close form (no trailing `|| true`) does NOT appear.

Sabotage-validated: stripping `|| true` from the pipeline makes cases 2 and 4 both fail.

**CG follow-up (`3ecbfe13`):** CG returned APPROVED WITH WARNINGS and flagged 3 latent split-local pipelines now fragile under the pipefail flip:
- `cmd_status` reading `bootstrap-status.json` (lines 647/650/651) — could abort `dream status` mid-output on a partial/mid-write JSON.
- `cmd_preset list` reading `meta.txt` (lines 1954/1955) — a single malformed preset would abort the listing of all subsequent ones.
- `cmd_chat` jq-fallback (line 1279, newly added in `61be91b0`) — unparseable LLM JSON would silently abort instead of reaching the friendly `LLM error: ${_err:-unparseable response}`.

All 3 sites now have `|| true` with inline rationale comments. Sites #1 and #2 are pre-existing code newly exposed by this PR's pipefail flip; site #3 is new code from this PR.

`make lint` green, `make test` green (including the new pipefail-tolerance regression).

Promoting from draft.

Lightheartdevs:
Audit pass (2026-05-02): the prerequisite #1008 is now merged, so the original `set -eo pipefail` blocker is resolved at the project level. Re-reviewed the intended diff (3 stated commits: `dfaa4c8f`, `385372ec`, `10588c2a`) — adding pipefail, the `| head -1` → `| sed -n 1p` SIGPIPE-safety swap, the `curl --fail` + `jq -er` chat/benchmark hardening, and the exit-code contract for cmd_disable / cmd_agent logs / _gpu_reassign all look correct.

**Cannot review or merge while CONFLICTING.** `git diff main HEAD` against current main shows ~4000 lines of unrelated deletions (test files, host-agent.py, support-bundle scripts) that aren't part of this PR's intent — that's the rebase delta from being branched off old main. Please rebase onto current main; the actual unique diff should drop to the ~190-line shape the PR body describes. Will re-review after rebase.

Lightheartdevs (changes requested):
This fixes the original pipefail/version-fallback finding: `_check_version_compat` now keeps `|| true` on the `DREAM_VERSION` lookup, and the new focused regression passed locally.

Still not merge-ready because the branch is stale/dirty against current `main`:

- `tests/test-dream-cli-version-compat-pipefail.sh` passed locally (4/4), and `bash -n` passed for the touched shell files.
- A local merge simulation into current `origin/main` conflicts in both `dream-server/Makefile` and `dream-server/dream-cli`.

Please rebase onto current `main` and resolve those conflicts. This one looks valuable after that rebase; the blocker is merge hygiene rather than the original correctness issue.

yasinBursali:

Four of the original five commits dropped as empty during rebase — their content already landed in upstream via other merged PRs:
- `3ecbfe13 fix(dream-cli): tolerate malformed JSON/meta input in 3 split-local pipelines under pipefail`
- `9d7aa45e test(dream-cli): regression for _check_version_compat pipefail tolerance`
- `04a43e9e fix(dream-cli): exit-code contract for disable / agent logs / gpu reassign`
- `61be91b0 fix(dream-cli): surface LLM failures in 'dream chat' and 'dream benchmark'`

Surviving commit:
- `a5f435d5 fix(dream-cli): add pipefail and audit pipe/|| true sites`

Net diff: **1 file, 7+/2−** in `dream-server/dream-cli`. The load-bearing pipefail change. Mergeable. Ready for re-review.

Lightheartdevs (approved):
Deep-audited the refreshed pipefail PR. The surviving diff was too small for a safe global set -eo pipefail flip: several tolerant read paths still aborted on missing/partial optional data. I fixed those on the PR branch in ca843b4c and added 	ests/test-dream-cli-pipefail-tolerance.sh, wired into make test, covering empty .env, missing mode/model keys, malformed bootstrap status, malformed preset metadata, dead chat backend, and benchmark propagation.\n\nLocal checks run:\n- ash -n dream-cli tests/test-dream-cli-pipefail-tolerance.sh\n- ash tests/test-dream-cli-pipefail-tolerance.sh with a local python3 shim to exercise PyYAML-backed registry cases: 8 passed\n- native Git Bash run without PyYAML: 6 passed, 1 skipped\n- git diff --check\n\nGitHub checks are green on head ca843b4c. Ready from my side.

PR #1000, opened 2026-04-23
feat(dream-cli): --json flag on list/status and document doctor --json

yasinBursali:
## What

`dream list --json` and `dream status --json` now work as advertised. Previously the `--json` flag was silently dropped because the main case dispatch forwarded no arguments to the subcommands. Also documents the previously-undocumented `dream doctor --json`, keeps `dream status-json` as a back-compat hyphenated alias.

Plus an audit follow-up: a regression test proving `dream list --json` stays valid JSON on stdout even when registry/schema diagnostics fire on stderr.

## Why

Scripts piping `dream list` / `dream status` output to `jq` got parse errors because the subcommands emitted human-readable output regardless of any flag passed. The human-readable hyphenated alias `dream status-json` was the only JSON-emitting path; `--json` appeared to work (no error) but emitted the wrong format.

Once the `--json` plumbing is fixed, the next concern (raised by maintainer audit) is that the JSON contract must hold even when registry loading emits diagnostics. After PR #1006 moved `log()`/`warn()` to stderr, the producer/consumer separation is in place — the regression locks it in.

### Functional fix (commit `af9acf0f`)
- Main case dispatch now forwards `"$@"` to both `cmd_list` and `cmd_status`.
- `cmd_list --json` emits a JSON array of `{id, category, status}`.
- `cmd_status --json` delegates to the existing `cmd_status_json`. The delegation is wrapped in a subshell (`( cmd_status_json )`) to isolate the RETURN trap set inside `cmd_status_json` — without the subshell, the trap would leak into `cmd_status`'s frame and crash with an unbound-variable error when the sibling nounset PR lands on top.
- Unknown flags on these subcommands now error cleanly (exit 1).
- `cmd_help` advertises the new flags and the existing `doctor --json`.

### Audit follow-up (commits `6203c819` + `8d2b5493`)
New `dream-server/tests/test-dream-list-json-clean-stdout.sh` (5 cases, wired into `make test`):

1. Hermetic scaffold — copies `dream-cli` + minimal `lib/` (`service-registry.sh` required, plus `safe-env.sh`/`python-cmd.sh` if present) into a tempdir. Because `dream-cli` derives `SCRIPT_DIR` from `BASH_SOURCE[0]`, and `lib/service-registry.sh` derives `EXTENSIONS_DIR="${SCRIPT_DIR:-$(pwd)}/extensions/services"`, the test is fully isolated from the live repo's extensions tree.
2. Plants two manifests inside the scaffold:
   - `extensions/services/valid-svc/manifest.yaml` (well-formed).
   - `extensions/services/broken-svc/manifest.yaml` with `schema_version: dream.services.v1` but missing `service.id` — the canonical trigger for `lib/service-registry.sh:127` to emit `# SKIP: <path>: missing required "id" field` to stderr.
3. Runs `NO_COLOR=1 DREAM_HOME=$tempdir $tempdir/dream-cli list --json` with stdout and stderr captured separately.
4. Five assertions:
   - exit code 0.
   - stdout parses as JSON via `python3 json.loads`.
   - JSON contains `valid-svc` and excludes `broken-svc`.
   - stderr contains the literal `# SKIP:` diagnostic — proves the warning fired during the same invocation that produced the JSON.
   - stdout has no `# SKIP:` / `⚠` / `[dream]` / ANSI escape (literal `0x1B`) bytes — the jq-pipeline contract.
5. Skips gracefully if `python3` or PyYAML are unavailable on the host.

CG-follow-up tightenings in `8d2b5493`: `# SKIP:` literal prefix (was bare `SKIP`), real `$'\x1b['` ANSI ESC byte (was `\\033\[` literal-string match), and `[[ -f X ]] && cp X` (was `cp X 2>/dev/null || true` — violated project CLAUDE.md `|| true` ban).

## Testing

- `make lint` PASS.
- `make test` PASS — new test runs at the end of the suite.
- `shellcheck` clean on the new test file.
- 5/5 cases pass; CG sabotage probe (replacing the SKIP token with a guaranteed-absent string) produces 4/1 fail — confirming the test is non-tautological.

## Review

Local CG **APPROVED WITH WARNINGS** on the regression. All three labelling-strength warnings were folded into commit `8d2b5493` before push:

- `# SKIP:` literal prefix instead of bare `SKIP` substring.
- Real ANSI ESC byte test instead of literal four-char `\033[` string match.
- `[[ -f ]] && cp` instead of `cp ... 2>/dev/null || true`.

CG flagged two coverage gaps as non-blocking follow-ups (not addressed here):
- `cmd_status --json` regression — same shape but exercises the subshell-isolated `cmd_status_json` path. Worth a sibling test in a follow-up.
- PyYAML *parse* error coverage — current scaffold triggers a registry-validation warning (`# SKIP: missing id`); a malformed-YAML manifest would trigger a different warn shape (`warn "Service registry: manifest parser failed"`). Assertion (e) would still catch leakage from either.

## Known Considerations

`cmd_status_json`'s internal RETURN trap (single-quoted, `$tmp` resolved at fire-time) is the latent root cause and exists pre-PR on upstream `main` too — dormant there because upstream currently has only `set -e`. The subshell wrap in this PR defensively hardens the new caller path without modifying the pre-existing function (the wider trap fix belongs with the nounset PR that adds `set -u`).

`dream status-json` hyphenated alias is kept for one release for back-compat.

## Platform Impact

- **macOS / Linux / Windows (WSL2):** identical. No platform branching.
- The hermetic test scaffold uses `mktemp -d`, `cp`, `python3`, `bash` — portable across the supported platforms.

### DIFF
--- a/dream-server/dream-cli
+++ b/dream-server/dream-cli
@@ -515,6 +515,21 @@ get_compose_flags() {
 #=============================================================================
 
 cmd_status() {
+    local json_mode="false"
+    while [[ $# -gt 0 ]]; do
+        case "$1" in
+            --json) json_mode="true"; shift ;;
+            *) error "Unknown argument to 'dream status': $1" ;;
+        esac
+    done
+
+    if [[ "$json_mode" == "true" ]]; then
+        # Subshell isolates cmd_status_json's RETURN trap so it can't leak
+        # into this caller's frame and re-fire with an unbound $tmp under set -u.
+        ( cmd_status_json )
+        return $?
+    fi
+
     check_install
     cd "$INSTALL_DIR"
     load_env
@@ -1646,10 +1661,42 @@ cmd_purge() {
 }
 
 cmd_list() {
+    local json_mode="false"
+    while [[ $# -gt 0 ]]; do
+        case "$1" in
+            --json) json_mode="true"; shift ;;
+            *) error "Unknown argument to 'dream list': $1" ;;
+        esac
+    done
+
     sr_load
     load_env 2>/dev/null || true
     local active_flags
     active_flags=$(get_compose_flags)
+
+    if [[ "$json_mode" == "true" ]]; then
+        # JSON output: array of {id, category, status}
+        local _first=1
+        printf '['
+        for sid in "${SERVICE_IDS[@]}"; do
+            local cat="${SERVICE_CATEGORIES[$sid]}"
+            local cf="${SERVICE_COMPOSE[$sid]}"
+            local status
+            if [[ "$cat" == "core" ]]; then
+                status="always-on"
+            elif [[ -n "$cf" && -f "$cf" && "$active_flags" == *"${cf#"$INSTALL_DIR/"}"* ]]; then
+                status="enabled"
+            else
+                status="disabled"
+            fi
+            (( _first == 1 )) || printf ','
+            printf '{"id":"%s","category":"%s","status":"%s"}' "$sid" "$cat" "$status"
+            _first=0
+        done
+        printf ']\n'
+        return 0
+    fi
+
     echo -e "${BLUE}━━━ Available Services ━━━${NC}"
     printf "%-20s %-12s %-10s\n" "SERVICE" "CATEGORY" "STATUS"
     printf "%-20s %-12s %-10s\n" "───────" "────────" "──────"
@@ -3260,9 +3307,9 @@ Usage: dream <command> [options]
 ${CYAN}Commands:${NC}
   gpu [status|topology|assignment|validate|reassign]
                       Inspect and manage multi-GPU configuration
-  status              Show service health and GPU status
-  status-json         Machine-readable status (JSON) with mode/tier/model
-  list                List all services and their status
+  status [--json]     Show service health and GPU status (--json = machine-readable)
+  status-json         Alias for 'status --json' (kept for back-compat)
+  list [--json]       List all services and their status (--json = machine-readable)
   enable <service>    Enable an extension service
   disable <service>   Disable an extension service
   purge <service>     Permanently delete service data
@@ -3287,7 +3334,7 @@ ${CYAN}Commands:${NC}
                       View, edit, or validate configuration
   chat "<message>"    Quick chat with the LLM
   benchmark           Run a quick performance test
-  doctor [report]     Run diagnostics and write JSON report
+  doctor [report|--json]   Run diagnostics (--json writes JSON to stdout)
   repair|fix          Run basic repairs (currently redirects to doctor)
   template [action]   Apply pre-built service templates (list|preview|apply)
   audit [extensions]  Audit extension manifests and compose contracts
@@ -3385,9 +3432,9 @@ EOF
 #=============================================================================
 case "${1:-help}" in
     gpu|g)       shift; cmd_gpu "$@" ;;
-    status|s)    cmd_status ;;
+    status|s)    shift; cmd_status "$@" ;;
     status-json) cmd_status_json ;;
-    list|ls)     cmd_list ;;
+    list|ls)     shift; cmd_list "$@" ;;
     enable)      shift; cmd_enable "$@" ;;
     disable)     shift; cmd_disable "$@" ;;
     purge)       shift; cmd_purge "$@" ;;
--- a/dream-server/dream-cli
+++ b/dream-server/dream-cli
@@ -658,6 +658,21 @@ _dream_cli_rebuild_images() {
 #=============================================================================
 
 cmd_status() {
+    local json_mode="false"
+    while [[ $# -gt 0 ]]; do
+        case "$1" in
+            --json) json_mode="true"; shift ;;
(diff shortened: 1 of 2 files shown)


Lightheartdevs:
Audit follow-up: needs a clean-JSON regression before merge.

The `--json` feature is useful, but machine-readable stdout must stay parseable even when registry/schema loading emits diagnostics. Now that #1006 has moved `log()`/`warn()` to stderr on `main`, please rebase and add a regression proving `dream list --json` remains valid JSON when PyYAML/registry loading warnings occur.

yasinBursali:
Pushed audit follow-up (`6203c819` + `8d2b5493`).

Rebased onto current `upstream/main`. The post-#1006 stderr split is now in scope.

New `tests/test-dream-list-json-clean-stdout.sh` (5 cases, wired into `make test`):

1. Hermetic tempdir with copied `dream-cli` + minimal `lib/` so `EXTENSIONS_DIR` resolves inside the scaffold (no live-repo pollution).
2. Plants `extensions/services/valid-svc/manifest.yaml` (well-formed) and `extensions/services/broken-svc/manifest.yaml` (missing `service.id` — triggers `# SKIP: <path>: missing required "id" field` from `lib/service-registry.sh:127`).
3. Runs `NO_COLOR=1 DREAM_HOME=$tempdir $tempdir/dream-cli list --json` with stdout/stderr captured separately.
4. Asserts:
   - exit 0
   - stdout parses as JSON via `python3 json.loads`
   - JSON contains `valid-svc`, excludes `broken-svc`
   - stderr contains the literal `# SKIP:` diagnostic — proves the warning fired in the same invocation that produced the JSON
   - stdout has no `# SKIP:` / `⚠` / `[dream]` / ANSI ESC byte (`0x1B`) — the jq-pipeline contract.
5. Skips gracefully if `python3` or PyYAML are unavailable.

Local CG returned APPROVED WITH WARNINGS. All three labelling-strength warnings folded into `8d2b5493` before push: `# SKIP:` literal prefix, real `$'\x1b['` ANSI byte test, `[[ -f ]] && cp` instead of `cp ... || true` (project CLAUDE.md style).

CG flagged two coverage gaps as non-blocking follow-ups: a `cmd_status --json` regression (same shape but exercises the subshell-isolated `cmd_status_json` path) and a PyYAML-parse-error case (current scaffold triggers a registry-validation warning; a malformed-YAML manifest would trigger `warn "Service registry: manifest parser failed"` — assertion (e) would still catch leakage from either).

Lightheartdevs:
Audit pass (2026-05-02): `--json` on `dream list` and `dream status` is the right shape. Manual run confirms clean stdout (`[]`) with PyYAML warnings on stderr, exit code 0. Cannot review while CONFLICTING — please rebase. Codex's note about adding the focused test to CI with PyYAML installed (so it exercises the broken-manifest path instead of skipping) stands.

Lightheartdevs (changes requested):
This is closer than the first audit: the JSON flag is useful, `warn()` now writes to stderr via current main, and the branch adds a regression for `dream list --json` keeping stdout parseable while registry diagnostics go to stderr.

I can't merge it yet because the branch is stale/dirty against current `main`:

- GitHub marks it `DIRTY`.
- A local merge simulation conflicts in `dream-server/Makefile`.
- The added regression script syntax-checks, but in this Windows/Git Bash environment it skipped because the `python3` command available to the script did not have PyYAML. Please make sure the CI path actually runs the new test rather than skipping it silently.

Please rebase/resolve the Makefile conflict and verify the JSON stdout regression runs in CI. After that, this should be a straightforward re-review.

yasinBursali:

Two of the original three commits dropped as empty during rebase (already in upstream):
- `8d2b5493 test(dream-cli): tighten stdout-leakage regex + drop `|| true` from optional cp`
- `6203c819 test(dream-cli): regression — `dream list --json` keeps stdout valid JSON when registry warnings emit`

Surviving commit:
- `8e66036b feat(dream-cli): --json flag on list/status and document doctor --json`

Net diff: **1 file, 53+/6−** in `dream-server/dream-cli`. Mergeable. Ready for re-review.

Lightheartdevs (approved):
Audited the refreshed dream-cli JSON flag changes in depth. I found one JSON escaping gap in dream list --json for user-extension registry strings, fixed it on the PR branch, and added 	ests/test-dream-list-json-escaping.sh to cover quoted/backslash category values.\n\nLocal checks run:\n- ash -n dream-cli tests/test-dream-list-json-escaping.sh\n- ash tests/test-dream-list-json-escaping.sh with a local python3 shim for this Windows Git Bash environment\n- DREAM_HOME=C:\Users\conta\Documents\Codex\2026-05-10\clone-and-audit-this-repo-and\DreamServer-pr-audit-worktree\dream-server NO_COLOR=1 ./dream-cli list --json | python3 -m json.tool with the same shim\n- git diff --cached --check before commit\n\nGitHub checks are green on head 8c650a46. Looks ready to merge from my side.

PR #1001, opened 2026-04-23
fix(dream-cli): compose summary wrapper for restart/start/stop/update

yasinBursali:
## What

Wrap `docker compose` calls from `dream restart`, `start`, `stop`, and `update` with a summary function that surfaces `"<verb> — done"` on success or a red `"<verb> failed:"` banner with up-to-20 error-keyword lines + full log path on failure. Also teaches `sr_resolve` to accept container names (e.g. `dream-token-spy` as copied from `docker ps`) by stripping the `dream-` prefix when the literal input doesn't match any alias.

## Why

Previously, a compose failure like "dream-llama-ready is unhealthy" was one line buried in 40+ lines of state transitions. Users couldn't distinguish "all up" from "one unhealthy" without re-running and squinting. Separately, users copying container names from `docker ps` output would get "no such service" errors because `sr_resolve` only matched registered aliases.

## How

Single commit `431e83cf`:

- `_compose_run_with_summary <verb> <docker-compose-args...>`: runs compose with `--progress quiet`, captures stdout+stderr to `mktemp`, prints either a success line or a red failure banner with up to 20 `grep -iE 'error|unhealthy|failed|dependency'` matches plus the preserved log path.
- All four callers (`cmd_restart`, `cmd_start`, `cmd_stop`, `cmd_update`) wrap through the helper. For `cmd_update`, the existing `if ! …; error` guard on both pull and up-recreate calls is preserved so the rollback hint remains on top of the surfaced error context.
- The summary grep pipeline ends with `|| warn "(no error keywords matched in compose log)"` — `|| warn` is CLAUDE.md's project-blessed form for "tolerate this specific non-match and log why the summary is empty." Upstream `main` today runs under `set -e` only (no `pipefail`), so the guard is defensive for the forthcoming pipefail adoption; it costs nothing today and keeps the function correct under future pipefail.
- `sr_resolve` now strips a leading `dream-` prefix if the literal input has no alias and the stripped form does — recovers the service ID for container names. All registry-loader-generated container names follow `dream-<sid>` by convention.

## Testing

- `dream restart dream-token-spy` (container name): resolves to `token-spy`, restart succeeds (previously "no such service").
- Synthetic compose failure: error banner + log path + correct compose exit code.
- Round-1 review: Critique Guardian approved. Round-2 adversarial audit by shell-verifier: flagged an earlier `|| true` as a CLAUDE.md silent-swallow violation; replaced with `|| warn` + corrected block comment. Re-verified.

## Platform Impact

- **macOS / Linux / Windows (WSL2):** identical. No platform branching. `docker compose --progress quiet`, `grep -iE`, `sed`, `head`, `mktemp` are all portable across the three.

### DIFF
--- a/dream-server/dream-cli
+++ b/dream-server/dream-cli
@@ -510,6 +510,60 @@ get_compose_flags() {
     fi
 }
 
+# Run `docker compose <args...>` with a compact summary on success and a
+# surfaced error banner on failure.
+#
+# Usage:  _compose_run_with_summary <verb> <compose_args...>
+#   <verb>           Human-readable gerund phrase, e.g. "Restarting all services"
+#   <compose_args>   Everything that goes after `docker compose`, including
+#                    any `-f` flags resolved by get_compose_flags.
+#
+# Behavior:
+#   - Runs `docker compose --progress quiet <compose_args>`, capturing
+#     stdout+stderr to a mktemp log file.
+#   - On success: prints "<verb> — done" and removes the log file.
+#   - On failure: prints an error banner, surfaces up to 20 lines matching
+#                 /error|unhealthy|failed|dependency/, preserves the full
+#                 log file for inspection, and returns the compose exit
+#                 code so the caller (under `set -e`) aborts with it.
+#
+# The summary grep pipeline can legitimately produce zero matches if the
+# failing compose output has no error-keyword hits. `upstream/main` today
+# runs under `set -e` only (no `pipefail`), so grep's exit 1 is absorbed
+# by the pipeline's final-stage exit and the function continues to the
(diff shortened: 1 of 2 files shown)


yasinBursali:
Superseded by #1016, which is a superset of this PR.

#1016 contains all of this PR's `_compose_run_with_summary` wrapper changes in `dream-cli` and `lib/service-registry.sh`, and additionally adds:
- `trap '...' INT TERM` to fix a `/tmp/tmp.*` leak when the user hits Ctrl-C mid-`docker compose`
- A zero-match refactor that's robust under `pipefail`
- Apple GPU output polish + dream-doctor macOS fixes

Merging this PR alone would ship a known resource leak. Closing in favor of #1016.

PR #1002, opened 2026-04-23
refactor(dream-cli): enable set -u and add guards for conditional variables

yasinBursali:
## What

Enable `set -u` (nounset) in `dream-cli` on top of the existing `set -eo pipefail` foundation. Audit every variable read; add `${VAR:-}` guards to six legitimately-conditional sites (missing positional args for `preset diff`, sparse associative-array lookups when comparing preset ext/env maps, registry lookups for service dirs without valid manifests). All other variables in the file are unconditionally set before use; no guard added elsewhere.

## Why

Nounset catches typos in variable names at runtime instead of silently expanding to empty strings. The audit also surfaced three pre-existing latent bugs that `set -u` would have caught immediately:

- `preset diff`: bare `$2` / `$3` crashed with no args.
- `preset diff`: `ext1[$key]` / `env1[$key]` associative-array reads crashed partway through comparison when a key exists only in the other preset.
- `enable <svc>`: bare `SERVICE_DEPENDS[$service_id]` crashed for service directories that exist on disk but have no valid manifest (edge case: user-extensions with malformed manifest).

## How

Single commit (tip `bea96fe2`; the original `a7bd6cac` was rebased onto this PR's parent branch `fix/dream-cli-pipefail-exit-codes` after that branch's commits were adjusted during review — PR-11's nounset audit content is unchanged, only the base commits it sits on have updated SHAs).

Audited by running every read-only subcommand end-to-end in a fake `INSTALL_DIR` and against a live install: `help`, `version`, `list`, `status`, `status-json`, `config show`, `gpu {status|topology|validate|assignment}`, `model {current|list}`, `mode`, `stt {current|status}`, `preset {list|save|load|delete|export|import|diff}`, `agent status`, `doctor --json`, `template {list|preview}`, `audit`, `chat` (against down backend), `benchmark`, `enable`/`disable`/`purge` (against both valid-service and unknown paths). Also verified with minimal environment (`env -i PATH HOME BASH INSTALL_DIR dream-cli ...`) to catch implicit `TERM`/`USER`/`LANG` assumptions.

## Testing

- End-to-end coverage of ~25 read-only dream-cli subcommands on the operator's integration branch.
- Minimal-environment verification via `env -i`.
- Operator's integration-branch test battery: green.

## Scope Note

Depends on the `set -eo pipefail` foundation from `fix/dream-cli-pipefail-exit-codes` — this branch's base commits are that PR's three commits (rebased onto the review-adjusted tips). Intended to merge after that PR lands.

## Platform Impact

- **macOS / Linux / Windows (WSL2):** identical. Portable Bash 4+ parameter expansion. No platform branching. No variables that are set on one OS but not another were found during the audit — all guarded reads guard against missing registry entries or unprovided positional args, neither of which is platform-specific.

---

## 📋 Chain status (operator's reminder — apr-25)

This PR is the **last link in the `dream-cli` strict-mode chain**:

```
#1008 (READY) ──▶ #998 (DRAFT) ──▶ #1002 (this — DRAFT)
```

**Before this PR can be promoted:**
- #1008 must merge first (.env grep-pipeline guards).
- #998 must merge next (`set -eo pipefail` + exit-code contract — the foundation `-u` builds on).

**This PR currently shares 6+ identical hunks with #998** (line 6 set-line, identical `cmd_chat`/`cmd_benchmark`/`cmd_disable`/`cmd_agent`/`_gpu_reassign` rewrites, `sed -n '1p'` swaps). Merging in parallel = textual conflict.

**After #998 merges, the operator must:**
1. Rebase this branch onto post-#998 main. The shared hunks drop out automatically (already in main).
2. What remains is the unique `-u` work: the extra flag at line 6, plus `:-` empty-string default guards in `cmd_preset`, `cmd_enable`, etc.
3. Promote this PR from DRAFT → ready (`gh pr ready 1002`).

Kept DRAFT until then as a mechanical safeguard — cannot merge before #998.

### DIFF
--- a/dream-server/dream-cli
+++ b/dream-server/dream-cli
@@ -1547,6 +1547,7 @@ cmd_disable() {
     # Check built-in extensions first, then dashboard-installed user extensions
     local ext_dir="$INSTALL_DIR/extensions/services/$service_id"
     [[ ! -d "$ext_dir" ]] && ext_dir="$INSTALL_DIR/data/user-extensions/$service_id"
+    [[ -d "$ext_dir" ]] || error "Unknown service: $input"
     local cf="$ext_dir/compose.yaml"
 
     # Check it's not a core service
@@ -2488,6 +2489,7 @@ cmd_agent() {
                 tail -f "$log_file"
             else
                 warn "No log file at $log_file"
+                return 1
             fi
             ;;
         *)
@@ -2849,19 +2851,19 @@ _gpu_reassign() {
 
     if ! command -v nvidia-smi &>/dev/null; then
         warn "GPU reassign is only supported for NVIDIA. Use 'dream gpu status' for AMD."
-        return
+        return 1
     fi
 
     if ! command -v jq &>/dev/null; then
         warn "jq not found — required for GPU reassignment"
-        return
+        return 1
     fi
 
     local topo_lib="$SCRIPT_DIR/installers/lib/nvidia-topo.sh"
     local assign_script="$SCRIPT_DIR/scripts/assign_gpus.py"
 
-    [[ -f "$topo_lib" ]] || { warn "Topology library not found: $topo_lib"; return; }
-    [[ -f "$assign_script" ]] || { warn "Assignment script not found: $assign_script"; return; }
+    [[ -f "$topo_lib" ]] || { warn "Topology library not found: $topo_lib"; return 1; }
+    [[ -f "$assign_script" ]] || { warn "Assignment script not found: $assign_script"; return 1; }
 
     # warn and err are already defined; source is safe
     . "$topo_lib"
--- a/dream-server/dream-cli
+++ b/dream-server/dream-cli
@@ -3,7 +3,7 @@
 # Mission: M5 (Clonable Dream Setup Server)
 # Version: 2.0.0 — Registry-driven service resolution
 
-set -e
+set -eo pipefail
 
 # Require Bash 4+ (associative arrays used by service registry and dream-cli)
 if (( BASH_VERSINFO[0] < 4 )); then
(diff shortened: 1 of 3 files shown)


yasinBursali:
**Marked as draft pending merge of #998.**

Per the PR description, this branch's base commits are the three commits of #998 (`fix/dream-cli-pipefail-exit-codes`). Merging this before #998 would pull #998's changes into this PR's diff out of order.

Will un-draft and request review once #998 lands.

Lightheartdevs:
Audit follow-up: keep draft / rebase on current `main`.

The broad nounset/pipefail direction overlaps with focused fixes that have now landed, especially #1008. Please rebase and ensure `_check_version_compat` does not abort when `.env` lacks `DREAM_VERSION`; it should still fall back to `.version` or `manifest.json` under `set -euo pipefail`.

Lightheartdevs:
Holding this out of the merge queue because it is still a draft and conflicts with current `main`.

Re-audit notes:

- The branch still enables `set -euo pipefail` while `_check_version_compat` reads `DREAM_VERSION` through a bare `grep ... | sed ... | cut ...` pipeline without `|| true`.
- I reproduced the failure shape directly: with `set -euo pipefail` and an `.env` that lacks `DREAM_VERSION`, the assignment exits before the intended `.version` / `manifest.json` fallback can run.
- A local merge simulation into current `origin/main` conflicts in `dream-server/dream-cli`.

Suggested path: absorb the focused #1008/#998-style `|| true` guard into this broader nounset audit, rebase onto current `main`, and undraft when the branch is intentionally ready.

Lightheartdevs:
Audit update: ready after rebase and direct corrections.

What I changed:
- Rebased the branch onto current `main`, dropping the already-merged pipefail stack so this PR is only the nounset work plus tests.
- Kept the intended `set -euo pipefail` change and guarded the conditional reads in `cmd_enable` and `preset diff`.
- Strengthened `tests/test-dream-cli-pipefail-tolerance.sh` to assert nounset and cover one-sided preset env/service keys plus missing `preset diff` args.
- Fixed the Git Bash secret-mask test harness to use wrappers instead of temp symlinks/copies, so the jq/python PATH matrix works on Windows too.

Local verification:
- `bash -n dream-cli tests/test-dream-cli-pipefail-tolerance.sh tests/test-preset-diff.sh tests/test-dream-config-secret-mask.sh`
- `tests/test-dream-cli-pipefail-tolerance.sh` -> 9 passed, 0 failed, 1 local PyYAML skip
- `tests/test-preset-diff.sh` -> 6 passed
- `tests/test-dream-config-secret-mask.sh` -> 9 passed
- minimal `env -i` command sweep for help/version/config/mode/model/preset/list/chat: no unbound/syntax errors
- `git diff --check`

GitHub CI is green on `6acc0c24f15f143aaa779dc47d9020f7ac264fa4`.

Lightheartdevs (approved):
Approved after rebase, nounset regression coverage, Git Bash test-harness fix, and green CI.

PR #1011, opened 2026-04-23
chore(bash32): guard declare -A callers + route dream-cli validate through $BASH

yasinBursali:
> **Merge after: #994.** This branch overlaps with #994 on `dream-server/dream-cli` and `scripts/validate-env.sh` (the schema-driven secret masking + Bash 4 `validate-env.sh` invocation). Merging this first would conflict with #994 / re-add reverted hunks. Once #994 lands, rebasing this drops the duplicated `dream-cli` / `validate-env.sh` hunks and the unique `installers/phases/03-features.sh`, `lib/progress.sh`, `scripts/pre-download.sh`, `scripts/dream-test-functional.sh` Bash-3.2 guards remain.

## What
Adds Bash 4+ guards to five scripts that use `declare -A` without one, and routes two `dream config validate` subprocesses through `"$BASH"` so they inherit dream-cli's modern bash interpreter.

Files touched (6):
- `dream-server/scripts/pre-download.sh` — Pattern A guard (bare `exit 1`)
- `dream-server/scripts/dream-test-functional.sh` — Pattern A guard + drop dead `2>/dev/null || true` fallback on `declare -A SERVICE_PORTS`
- `dream-server/scripts/validate-env.sh` — Pattern A guard
- `dream-server/lib/progress.sh` — Pattern B guard (`return 1 2>/dev/null || exit 1`)
- `dream-server/installers/phases/03-features.sh` — Pattern B guard
- `dream-server/dream-cli` — prefix validate-env.sh and validate-manifests.sh subprocess execs with `"$BASH"`

Guard style mirrors `dream-cli:8-14` (Pattern A, standalone scripts) and `lib/service-registry.sh:18-24` (Pattern B, sourced libraries).

## Why
On macOS, system `/bin/bash` is 3.2.57, which does not support associative arrays. Two user-facing problems:

1. Invoking any of these scripts directly under system bash (e.g. `./scripts/pre-download.sh --tier 3` from FAQ.md) produced a cryptic `declare -A: invalid option` crash with no actionable hint.
2. Even when dream-cli itself was launched under a modern Homebrew bash (its own Bash 4+ guard passes), it invoked `scripts/validate-env.sh` via shebang — which re-resolves to `/bin/bash` 3.2 regardless of the parent interpreter. `dream config validate` was silently non-functional on macOS for users without bash in PATH.

Guards turn (1) into a friendly error with a `brew install bash` hint. The `"$BASH"` prefix makes (2) work by inheriting the parent interpreter.

## How
- Guard block placed after the file header/purpose comments but before `set -euo pipefail` and before the first `declare -A`. Pattern B uses `return 1 2>/dev/null || exit 1` so sourced-vs-exec'd-direct both behave correctly.
- `dream-cli:1117` and `:1125`: `"$INSTALL_DIR/scripts/<script>.sh" ...` → `"$BASH" "$INSTALL_DIR/scripts/<script>.sh" ...`. `$BASH` is bash's own path to its interpreter; since dream-cli guards at line 9 require Bash 4+, `$BASH` is guaranteed to point at a modern bash when this code runs.

## Testing
- [x] `bash -n` syntax check — all 6 files pass
- [x] `shellcheck` — per-file warning counts identical between upstream/main and PR (no new diagnostics). SC2317 on Pattern B `exit 1` is accepted baseline, identical to `service-registry.sh:23`.
- [x] `make lint` — clean
- [x] Pre-commit (gitleaks / private-key / large-file) — clean
- [x] **Live macOS test on Bash 3.2.57**: each script exits with "ERROR: ... requires Bash 4.0+ (you have 3.2.57)" when run under `/bin/bash`. Pattern B sourced-path correctly propagates non-zero `$?` to the caller without aborting the parent shell.
- [x] Pattern B error-text wording aligned to `lib/service-registry.sh:20` ("Install a modern version") for consistency.

## Platform Impact
- **macOS** (primary target): `dream config validate` now works without requiring the user to manually invoke via `/opt/homebrew/bin/bash`. Direct invocations of the 5 scripts under system bash produce a clean error message instead of a cryptic crash.
- **Linux**: no impact (Bash 5+ everywhere; guards never fire).
- **Windows WSL2**: no impact (Bash 5+ in WSL2; guards never fire).

### DIFF
--- a/dream-server/Makefile
+++ b/dream-server/Makefile
@@ -41,6 +41,9 @@ test: ## Run unit and contract tests
 	@echo "=== dream-cli pipefail tolerance ==="
 	@bash tests/test-dream-cli-pipefail-tolerance.sh
 	@echo ""
+	@echo "=== Bash 3.2 guard contracts ==="
+	@bash tests/test-bash32-guards.sh
+	@echo ""
 	@echo "=== dream config show secret-mask matrix ==="
 	@bash tests/test-dream-config-secret-mask.sh
 	@echo ""
--- a/dream-server/dream-cli
+++ b/dream-server/dream-cli
@@ -1114,7 +1114,7 @@ cmd_config() {
         validate)
             cd "$INSTALL_DIR"
             if [[ -x "$INSTALL_DIR/scripts/validate-env.sh" ]]; then
-                "$INSTALL_DIR/scripts/validate-env.sh" "$INSTALL_DIR/.env" "$INSTALL_DIR/.env.schema.json"
+                "$BASH" "$INSTALL_DIR/scripts/validate-env.sh" "$INSTALL_DIR/.env" "$INSTALL_DIR/.env.schema.json"
             else
                 warn "validate-env.sh not found at $INSTALL_DIR/scripts/validate-env.sh"
                 warn "Make sure you're on a recent Dream Server release."
@@ -1122,7 +1122,7 @@ cmd_config() {
             echo ""
             if [[ -x "$INSTALL_DIR/scripts/validate-manifests.sh" ]]; then
                 log "Validating extension manifests and compatibility..."
-                if "$INSTALL_DIR/scripts/validate-manifests.sh"; then
+                if "$BASH" "$INSTALL_DIR/scripts/validate-manifests.sh"; then
                     success "Extension manifests validated"
                 else
                     warn "Extension manifest validation reported issues. See output above."
--- a/dream-server/dream-cli
+++ b/dream-server/dream-cli
@@ -1470,7 +1470,7 @@ cmd_config() {
             echo ""
             if [[ -x "$INSTALL_DIR/scripts/validate-manifests.sh" ]]; then
                 log "Validating extension manifests and compatibility..."
-                if "$INSTALL_DIR/scripts/validate-manifests.sh"; then
+                if "$BASH" "$INSTALL_DIR/scripts/validate-manifests.sh"; then
                     success "Extension manifests validated"
                 else
                     warn "Extension manifest validation reported issues. See output above."
--- a/dream-server/installers/phases/03-features.sh
+++ b/dream-server/installers/phases/03-features.sh
@@ -20,6 +20,14 @@
 #   Add new optional features to the Custom menu here.
 # ============================================================================
 
+# Require Bash 4+ (associative arrays used for GPU topology/link maps)
+if (( BASH_VERSINFO[0] < 4 )); then
+    echo "ERROR: $(basename "${BASH_SOURCE[0]}") requires Bash 4.0+ (current: $BASH_VERSION)" >&2
+    echo "  macOS ships Bash 3.2 due to licensing. Install a modern version:" >&2
+    echo "    brew install bash" >&2
+    return 1 2>/dev/null || exit 1
+fi
+
 dream_progress 18 "features" "Selecting features"
 if $INTERACTIVE && ! $DRY_RUN; then
     show_phase 2 6 "Feature Selection" "~1 minute"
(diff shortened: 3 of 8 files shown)


yasinBursali:
**Marked as draft pending merge of #994.**

This PR overlaps with open PR #994 (`fix/dream-cli-config-security-macos`):

- #994 already adds `"$BASH"` invocation of `validate-env.sh` from `dream-cli` → duplicates this PR's `dream-cli:1117` edit
- #994 already adds a Bash 4+ guard to `scripts/validate-env.sh` → duplicates this PR's guard on that file

Once #994 lands, I'll rebase this branch; the duplicate changes will drop automatically and the remaining scope is:

- Guards on `scripts/pre-download.sh`, `scripts/dream-test-functional.sh`, `lib/progress.sh`, `installers/phases/03-features.sh` (4 scripts #994 does not touch)
- `"$BASH"` routing for `validate-manifests.sh` subprocess call (`dream-cli:1125` — #994 does not cover this)
- Pattern B error-text wording alignment in `lib/progress.sh` and `installers/phases/03-features.sh` to match `lib/service-registry.sh:20`

Will un-draft and request review after the post-#994 rebase is clean.

Lightheartdevs:
Audit follow-up: keep draft and rebase after the strict-mode foundations.

The Bash 4 compatibility guard direction is reasonable, but this branch overlaps with the strict-mode/pipefail stack that has now partially landed. Please rebase on current `main`, keep only the unique Bash 3.2 compatibility changes, and rerun the relevant shell/BATS coverage before marking ready.

Lightheartdevs:
Holding this out of the merge queue because it is still a draft.

Re-audit notes:

- The patch is small and the intent is reasonable: guard Bash-4-only scripts and route validation helpers through `$BASH`.
- `bash -n` passed for the touched shell files.
- A local merge simulation into current `origin/main` applied cleanly.

Before this should merge, please undraft it and clarify the user-facing Bash contract: on macOS, `dream-cli config validate` now depends on the interpreter used for `dream-cli` itself. If the installed `dream-cli` is launched under `/bin/bash` 3.2, routing helpers through `$BASH` still points at Bash 3.2 and the new guard exits. That may be the intended fail-fast behavior, but the PR should state it plainly and include/update a test that proves the intended message/path.

Lightheartdevs:
Audit update: ready after rebase and direct corrections.

What I changed:
- Rebased onto current `main`, dropping the duplicate `validate-env.sh` guard/routing work that already landed.
- Kept the unique `validate-manifests.sh` `$BASH` routing in `dream-cli config validate`.
- Kept Bash 4 fail-fast guards for `lib/progress.sh`, `installers/phases/03-features.sh`, `scripts/pre-download.sh`, and `scripts/dream-test-functional.sh` before their `declare -A` usage.
- Added `tests/test-bash32-guards.sh` and wired it into `make test` so the Bash 3.2 compatibility contract is explicit.

Local verification:
- `bash -n` on touched shell files plus the new test
- `tests/test-bash32-guards.sh` -> 10 passed
- `tests/test-dream-cli-pipefail-tolerance.sh` -> 9 passed, 0 failed, 1 local PyYAML skip
- `tests/test-dream-config-secret-mask.sh` -> 9 passed
- `git diff --check`

Local limitation:
- `tests/test-validate-env.sh` and `tests/test-validate-manifests.sh` are blocked on this Windows checkout because local Git Bash lacks `jq`; CI has jq and is green.

GitHub CI is green on `d51237a66fbae931ec13755146b674cd60f6d12e`.

Lightheartdevs (approved):
Approved after rebase, Bash 3.2 guard contract coverage, and green CI.

PR #1015, opened 2026-04-23
fix(dashboard): template picker defensive fixes (handleApply + vacuous-truth)

yasinBursali:
## ✅ Rebased on current `main` 2026-04-28

#1003 has merged. The PR diff is now exactly the 2-line "Our delta" view below: `lib/templates.js` (+1 line) and `TemplatePicker.jsx` (1 line modified).

## What
Two small defensive fixes in template-related dashboard code:

- **`lib/templates.js` `getTemplateStatus`**: add early-return `if (services.length === 0) return 'available'` guarding against `[].every(s => s === 'enabled')` returning true (vacuous truth). Without this, templates with `services: []` get misclassified as `'applied'` — silently hidden from the Extensions page (filter drops 'applied') and rendered as a disabled green "Applied" card in the setup wizard added by #1003.
- **`TemplatePicker.jsx` `handleApply`**: replace `const data = await res.json().catch(() => ({}))` with `const data = await res.json()`. The silent `.catch(() => ({}))` swallowed malformed server responses and fabricated an empty data object → UI showed a fake "already active" state. The outer `try/catch` at line 159 already sets `setError('Failed to apply template')` — letting parse errors bubble there gives users a visible error state.

## Our delta
Two files, two lines:
- `dashboard/src/lib/templates.js`: `+ if (services.length === 0) return 'available'`
- `dashboard/src/components/TemplatePicker.jsx`: `- const data = await res.json().catch(() => ({}))` → `+ const data = await res.json()`

## Testing
- [x] Verified outer `catch (err) { setError(...) }` at TemplatePicker.jsx:159 shows a user-visible error path
- [x] Vacuous-truth early-return placed BEFORE the `.every()` check — empty-services templates now correctly return `'available'`
- [x] Pre-commit hooks clean
- [x] ESLint clean on touched files
- [ ] Unit tests for these two code paths — filing as a fork follow-up issue (not blocking)

## Platform Impact
- **All three platforms**: browser-side React/JS; dashboard is served via nginx, same behavior on macOS/Linux/Windows (WSL2).

### DIFF
--- a/dream-server/CONTRIBUTING.md
+++ b/dream-server/CONTRIBUTING.md
@@ -15,6 +15,26 @@ No CLA. No committee. No waiting for permission. If it makes Dream Server better
 If you're adding or extending services, read these first:
 - [docs/EXTENSIONS.md](docs/EXTENSIONS.md) — how to add a new service in 30 minutes
 - [docs/INSTALLER-ARCHITECTURE.md](docs/INSTALLER-ARCHITECTURE.md) — how the installer works under the hood
+- [docs/DASHBOARD-API-DEVELOPMENT.md](docs/DASHBOARD-API-DEVELOPMENT.md) — how to actually iterate on the dashboard-api FastAPI backend without your changes silently no-op'ing
+
+## Dashboard API development
+
+If you're touching `extensions/services/dashboard-api/`, read [docs/DASHBOARD-API-DEVELOPMENT.md](docs/DASHBOARD-API-DEVELOPMENT.md) **before** you start editing.
+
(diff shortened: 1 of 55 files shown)


Lightheartdevs:
Audit follow-up: keep draft and rebase after #1003.

The defensive TemplatePicker/error handling ideas are useful, but #1003 is now on `main`, and this draft needs to be restacked against that sentinel-based setup flow. Please rebase, keep only the still-needed UI defensive fixes, and rerun the dashboard tests before marking ready.

yasinBursali:
Rebased on current `main` (#1003 has merged). Diff collapsed to exactly the 2-line delta the description always promised — `lib/templates.js` early-return + `TemplatePicker.jsx` `.catch(() => ({}))` removal. ESLint clean on touched files. Description updated.

PR #973, opened 2026-04-16
docs: sync documentation with codebase after 50+ merged PRs

yasinBursali:
## Summary

Systematic audit of the last 50 merged PRs against all existing documentation revealed stale, contradictory, and missing docs. This PR fixes them:

- **WINDOWS-QUICKSTART.md**: Remove "Coming Soon — Preflight Checks Only" language (Windows has been fully supported since March 2026). Add real install flow, verified `dream.ps1` commands, and accurate installer flags
- **MODE-SWITCH.md**: Add lemonade mode section — auto-configured on AMD hardware, not user-switchable via `dream mode`
- **QUICKSTART.md / README.md**: Update all stale Qwen2.5 model names to Qwen3.5/Qwen3 matching current `tier-map.sh`
- **POST-INSTALL-CHECKLIST.md**: Rewrite 22-line skeleton (unchecked boxes, no commands) with real verification steps
- **SECURITY.md**: Add `DREAM_AGENT_BIND` / LAN access documentation with platform defaults
- **FAQ.md**: Add backup/restore, service templates, bootstrap fast-start, and expanded update/rollback docs
- **HOST-AGENT-API.md**: Add Windows platform note (host agent not yet available)
- **Langfuse README.md**: New service README (ports, env vars, volumes, troubleshooting)
- **CATALOG.md**: Add missing Langfuse entry
- **.env.example**: Document `LLAMA_CPU_LIMIT` for macOS/CPU-only mode
- **SUPPORT-MATRIX.md**: Link to AMD system-tuning guide
- **Root README.md**: Fix Apple Silicon "4B" → "9B" for 16-24GB tier

Every claim was verified against source code by an Opus-level critique pass (tier-map.sh, dream-cli, install-windows.ps1, dream.ps1, dream-host-agent.py, compose files, manifests).

## Test plan

- [ ] Verify markdown renders correctly on GitHub (tables, code blocks, links)
- [ ] Spot-check model names in QUICKSTART.md tier tables against `installers/lib/tier-map.sh`
- [ ] Verify `docs/MODE-SWITCH.md` lemonade section matches `installers/phases/06-directories.sh` behavior
- [ ] Verify `WINDOWS-QUICKSTART.md` installer flags match `install-windows.ps1` param block
- [ ] Confirm `extensions/services/langfuse/README.md` ports match `manifest.yaml`
- [ ] Confirm no broken relative links in changed files

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1017, opened 2026-04-23
docs(security): Linux host-agent fallback is 127.0.0.1 post-#988

yasinBursali:
## ⚠️ Draft — depends on #988 AND #973 merging first

- **#988** (`fix/security-loopback`) changes the Linux Docker-bridge-gateway detection fallback in `bin/dream-host-agent.py` from `0.0.0.0` to `127.0.0.1`.
- **#973** (`docs/sync-documentation-with-codebase`) introduces the "Host Agent Network Binding" table in `SECURITY.md` whose Linux row documents pre-#988 behavior.

Once both merge, I'll rebase and the PR diff will show exactly the 1-line change below.

## What
One-cell docs edit in the "Host Agent Network Binding" table in `dream-server/SECURITY.md`, Linux row:

```diff
- | Linux | auto-detected | Detects the Docker bridge gateway IP (e.g. `172.17.0.1`) so containers can reach the agent; LAN devices cannot. Falls back to `0.0.0.0` if detection fails. |
+ | Linux | auto-detected | Detects the Docker bridge gateway IP (e.g. `172.17.0.1`) so containers can reach the agent; LAN devices cannot. Falls back to `127.0.0.1` if detection fails. |
```

## Why
#973's Linux row was accurate when written but becomes stale the moment #988 merges. Actual runtime behavior per `bin/dream-host-agent.py:2241`: `bind_addr = _detect_docker_bridge_gateway() or "127.0.0.1"`. Docs should match.

## Testing
- [x] Verified against #988's code change in `bin/dream-host-agent.py`
- [x] No other "Falls back to 0.0.0.0" phrase remains in SECURITY.md (the remaining `0.0.0.0` mentions are legitimate user-opt-in override examples)
- [x] Pre-commit hooks clean

## Platform Impact
- **All three**: docs read identically. Linux row describes Linux-specific behavior.


yasinBursali:
Re-marking as draft as a mechanical merge-order safeguard.

This branch contains commits `ba556a9f` and `a6350e45` from #973 (docs-sync). Merging this PR before #973 lands would bring #973's content in as a side-effect rather than on its own merit.

**Merge order:** #973 first, then this PR rebases clean and gets re-marked ready.

Lightheartdevs:
Audit follow-up: keep draft and rebase to a docs-only delta.

#988 has now landed, so the security-doc direction is right, but this branch is currently draft/conflicting and still carries part of the old code/docs stack. Please rebase on current `main` and keep only the documentation changes that remain necessary after #988.

yasinBursali:
Pushed audit follow-up — rebased onto current `upstream/main`.

The audit ("rebase to a docs-only delta after #988") needed clarification because of an interaction with the still-open #973: SECURITY.md on current `upstream/main` does not yet contain the "Host Agent Network Binding" table at all (#973 introduces it). So a "docs-only delta" rebase resolves to one of two paths:

- **Path A (current):** this PR introduces the corrected table directly (with `127.0.0.1` in the Linux row, the post-#988 fallback). 3 commits, all yasinBursali, all docs-only after rebase. Diff: ~24 lines added to SECURITY.md.
- **Path B:** wait for #973 to merge first; rebase reduces this PR to a 1-line cell correction.

If #973 merges before this PR, the rebase will trivially collapse the first two commits (their SECURITY.md additions are already on main) and leave just the 1-line correction on the third commit (`ce676b99`). If this PR merges first, #973's later rebase will need a small SECURITY.md hunk reconciliation (it would otherwise re-add the table with the stale `0.0.0.0` value).

Suggested merge order to minimise rebase friction: **#973 first, then #1017.** I will update #973 in this same session to use `127.0.0.1` directly, so the `0.0.0.0`-stale-cell mismatch never reaches main. After that, #1017 may simply be closed as redundant — happy to close it with a one-line note pointing at the #973 update if you'd prefer that path.

Lightheartdevs:
Holding this out of the merge queue because it is still a draft and the scope has drifted beyond the title.

The security-doc fallback note itself looks directionally correct after the Linux bind fallback moved to `127.0.0.1`, and a local merge simulation into current `origin/main` applied cleanly. But this branch also changes multiple quickstarts, FAQ/mode-switch docs, `.env.example`, extension catalog docs, and adds a Langfuse README. That is too broad for a PR titled as a narrow security doc update.

Suggested path: either split the host-agent bind-address security doc into a small doc-only PR, or retitle/undraft this as a broader documentation refresh and call out the additional Langfuse/quickstart changes explicitly.

yasinBursali:
Closing — verified the PR's intent is already in upstream/main @ `0c6d5f05`. After rebase, 0 net diff remains (all 3 commits became empty):

- `f8b56a17 docs: sync documentation with codebase after 50+ merged PRs` — already merged as #973
- `9c0a6881 docs(host-agent): correct stale Windows entries — host agent now runs on Windows` — upstream now documents Windows host-agent support directly
- `ce676b99 docs(security): Linux host-agent fallback is 127.0.0.1 post-#988` — upstream/main `SECURITY.md` already reflects the post-#988 `127.0.0.1` fallback

No remaining work in this branch.

PR #1019, opened 2026-04-23
test+fix(setup): complete __DREAM_RESULT__ sentinel contract (exception path + tests)

yasinBursali:
## ✅ Rebased on current `main` 2026-04-28 + audit feedback addressed

#1003 has merged. The PR diff reduced to: `routers/setup.py` exception-path sentinel + 3 new test files (484 lines total). Two test fixes added on top per the audit:

- **SetupWizard.test.jsx — mock by URL**: replaced `fetch.mockImplementationOnce` with a URL-dispatched fetch helper (`makeFetchMock` + `stubFetchWithSetupStream`). Step 2 of the wizard fetches `/api/templates` and `/api/extensions/catalog` before the diagnostic POST, so the one-shot mock was being consumed by the wrong call. Routing by URL is the only stable pattern.
- **TemplatePicker.a11y.test.jsx — narrow selector**: narrowed the decorative-icon selector from `container.querySelectorAll('svg')` (which caught the unhidden `HardDrive` disk-size glyph) to `div.rounded-lg svg` — the status icon wrapper specifically. The test's stated intent ("Loader2, AlertTriangle, Check, default Icon") now matches what it actually asserts.

### Backend — `routers/setup.py` exception-path sentinel
Wraps the `run_setup_diagnostics` generator body in try/except. **Re-raises `CancelledError` and `OSError`** (transport-level concerns — client disconnect, broken pipe — should propagate normally). Catches generic `Exception`, logs with `logger.exception()`, yields a `FAIL:1` sentinel so the frontend parser never gets stuck in partial-state UI.

The `except Exception` is tagged `# noqa: BLE001 — sentinel contract requires *some* terminal signal`. CG reviewed and accepted this as a CLAUDE.md §2 I/O-boundary exception — narrow, logged, and the wire contract with the React parser is explicit.

### Tests (3 new files, 18 assertions)
- **pytest — `test_setup_sentinel.py`** (5 cases): asserts PASS:0 on success, FAIL:`<rc>` on non-zero exit, fallback sentinel when script missing, machine-parseable regex format, auth required. 5/5 pass locally.
- **Vitest — `SetupWizard.test.jsx`** (6 cases): sentinel parser happy path (PASS and FAIL with rc), fallback-scan on `"All tests passed!"` without sentinel, default-failure when neither signal is present, AbortController unmount cleanup, mocked ReadableStream chunk split across sentinel boundary.
- **Vitest — `TemplatePicker.a11y.test.jsx`** (7 cases): `isApplied` render branch + `aria-hidden="true"` on decorative Lucide icons (scoped to status icon wrapper) + defensive undefined-template handling.

## Testing
- [x] pytest: **5/5 pass** (`python3 -m pytest tests/test_setup_sentinel.py -v`)
- [x] Vitest: **13/13 pass** on the rebased branch (`npx vitest run SetupWizard.test.jsx TemplatePicker.a11y.test.jsx`)
- [x] Pre-commit hooks clean
- [x] Sentinel format matches frontend parser regex at `SetupWizard.jsx:104`: `^__DREAM_RESULT__:(PASS|FAIL):(-?\d+)$`

## Platform Impact
- **All three**: dashboard runs in browser; dashboard-api in Docker. Same behavior across macOS / Linux / Windows (WSL2).

## Known Considerations (will file as fork follow-up issues if not addressed in review)
1. The exception-path yield is not directly exercised by the 5 pytest cases — forcing it requires monkeypatching `asyncio.create_subprocess_exec` to raise. Small follow-up test worth adding.
2. The `HardDrive` disk-size glyph in `TemplatePicker` does not currently carry `aria-hidden="true"`. Production a11y could be tightened by adding it; this PR's test scope is decorative status icons, so the test selector was narrowed instead. A separate small follow-up could add the marker.

### DIFF
--- a/dream-server/CONTRIBUTING.md
+++ b/dream-server/CONTRIBUTING.md
@@ -15,6 +15,26 @@ No CLA. No committee. No waiting for permission. If it makes Dream Server better
 If you're adding or extending services, read these first:
 - [docs/EXTENSIONS.md](docs/EXTENSIONS.md) — how to add a new service in 30 minutes
 - [docs/INSTALLER-ARCHITECTURE.md](docs/INSTALLER-ARCHITECTURE.md) — how the installer works under the hood
+- [docs/DASHBOARD-API-DEVELOPMENT.md](docs/DASHBOARD-API-DEVELOPMENT.md) — how to actually iterate on the dashboard-api FastAPI backend without your changes silently no-op'ing
+
+## Dashboard API development
+
+If you're touching `extensions/services/dashboard-api/`, read [docs/DASHBOARD-API-DEVELOPMENT.md](docs/DASHBOARD-API-DEVELOPMENT.md) **before** you start editing.
+
+The short version: the dashboard-api `Dockerfile` copies the Python source into `/app/` at image build, and uvicorn imports from there. The compose service mounts `./extensions:/dream-server/extensions:ro` — but that bind-mount is for manifest and config discovery, not Python imports. Editing files under `extensions/services/dashboard-api/` on the host does **not** reload the running container. Your changes silently do nothing until the image is rebuilt.
+
+The recommended workflow is to run uvicorn natively on the host with `--reload`:
+
+```bash
+cd dream-server/extensions/services/dashboard-api
+pip install -r requirements.txt
+DREAM_INSTALL_DIR=/path/to/dream-server \
+  uvicorn main:app --host 127.0.0.1 --port 3002 --reload
+```
+
+Other services already reach the host via `host.docker.internal:host-gateway`, which is wired into the dashboard-api compose service, so the rest of the stack keeps working while you iterate. Hot-reload works natively on macOS and Linux; on WSL2, keep the repo on the WSL2 filesystem (not `/mnt/c/...`) for the watcher to behave.
+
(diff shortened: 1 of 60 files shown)


Lightheartdevs:
Audit follow-up: needs test fixes before merge.

The sentinel contract work is directionally good, but the new SetupWizard tests mock the wrong fetch call: Step 2 consumes `/api/templates` and `/api/extensions/catalog` before the diagnostic POST, so one-shot stream mocks get used too early. The a11y assertion also currently catches an unhidden `HardDrive` icon in `TemplatePicker`. Please mock by URL or queue Step 2 responses, and mark/narrow the decorative icon test.

yasinBursali:
Rebased on current `main` (#1003 has merged) and addressed both audit asks in a follow-up commit:

1. **Mock by URL** — replaced `fetch.mockImplementationOnce` with a URL-dispatched fetch helper. Step 2's `/api/templates` + `/api/extensions/catalog` calls no longer consume the diagnostic-stream mock prematurely.
2. **Narrow a11y selector** — scoped the decorative-icon assertion from `container.querySelectorAll('svg')` to `div.rounded-lg svg` (the status icon wrapper). `HardDrive` disk-size glyph is no longer caught by this test.

Re-ran on the rebased branch: pytest 5/5 + Vitest 13/13 pass. Description updated; left a note about adding aria-hidden to HardDrive as a separate small follow-up.

PR #1030, opened 2026-04-24
fix(host-agent): install flow — built-in hooks, bind-mount anchor, post-up state verify

yasinBursali:
## What
Fix five clustered defects in the host agent's extension install flow that combined to produce silent install failures.

## Why
- Built-in extensions declaring a `post_install` hook had their hooks silently skipped — the ext_dir was hardcoded to the user-extensions directory so `_resolve_hook` never found the manifest.
- `_precreate_data_dirs` had the same hardcode, plus a prefix filter that only caught `./data/` bind-mount sources. Community extensions like label-studio (`./upload`, `./media`, `./www`), continue and sillytavern (`./config/...`) were skipped, Docker auto-created those dirs as `root:root`, and non-root containers broke on Linux.
- Relative paths in extension compose files were resolved against `INSTALL_DIR`, but Docker Compose resolves them against the compose file's directory — dirs were created in the wrong place.
- Post-`compose up -d` progress was written as "started" unconditionally on return-code zero, which Docker emits for Exited/Created/Restarting containers as well as Running. The dashboard showed green for crash-looping services.

## How
1. `_run_install` and `_precreate_data_dirs` now call `_find_ext_dir(service_id)`.
2. Prefix filter widened: any relative bind-mount source (no leading `/`, contains `/`) is now pre-created; named volumes are still skipped.
3. `dir_path` and the traversal safety check both anchor on `ext_dir`.
4. `_handle_install` polls `docker inspect --format '{{.State.Status}}|{{.State.Error}}'` for up to 15s (1s interval). Non-running -> `status=error` with `State.Error`; running -> `status=started` as before.

## Testing
- `python3 -m py_compile` clean.
- `ruff check` — all checks passed.
- `pytest tests/test_host_agent.py` — 44/44 pass (includes 6 new tests).
- `pytest tests/` — rest of suite green (two pre-existing Python-3.14 asyncio baseline failures in `test_main.py` and `test_gpu_detailed.py` are unrelated and reproduce on `main`).
- `make lint` — clean.
- `pre-commit` — gitleaks / private-key / large-file all pass.

## Review
Critique Guardian: APPROVED WITH WARNINGS. No required changes. CG flagged nice-to-have follow-ups tracked as separate fork issues.

## Known Considerations
- **15s post-up window** is a tuning choice. If a cold-boot image (e.g. qdrant / whisper on first disk-cache miss) takes longer to reach `running` state (not healthy — just running), operators can widen via follow-up.
- **Per-service lock held during poll** — previously fire-and-forget, now serializes concurrent enables for the same service for up to 15s. This is an improvement (no more races) but callers that polled rapidly will see the latency.

## Platform Impact
- **macOS:** defects #1 and #5 fully active; #2/#3/#4 partially masked by Docker Desktop's auto-creation of bind-mount dirs. Fix corrects behavior on all three regardless.
- **Linux:** all five defects fully active; rootless Docker / strict permissions make #2/#3 a hard fail. Fix required.
- **Windows (WSL2):** same as Linux (WSL2 Docker behaves as Linux).

### DIFF
--- a/dream-server/bin/dream-host-agent.py
+++ b/dream-server/bin/dream-host-agent.py
@@ -15,6 +15,7 @@ import signal
 import subprocess
 import sys
 import threading
+import time
 from datetime import datetime, timezone
 from http.server import HTTPServer, BaseHTTPRequestHandler
 from pathlib import Path
@@ -167,7 +168,9 @@ def resolve_compose_flags() -> list:
 
 def _precreate_data_dirs(service_id: str):
     """Pre-create data directories for an extension with correct ownership."""
-    ext_dir = USER_EXTENSIONS_DIR / service_id
+    ext_dir = _find_ext_dir(service_id)
+    if ext_dir is None:
+        return
     compose_path = ext_dir / "compose.yaml"
     if not compose_path.exists():
         return
@@ -200,10 +203,14 @@ def _precreate_data_dirs(service_id: str):
             continue
         for vol in volumes:
             vol_str = str(vol).split(":")[0]
-            if vol_str.startswith("./data/") or vol_str.startswith("data/"):
-                dir_path = (INSTALL_DIR / vol_str.lstrip("./")).resolve()
+            # Accept any relative bind-mount source (e.g. "./data/state",
+            # "./upload", "config/stuff"). Skip named volumes (no "/") and
+            # absolute paths ("/etc/..."). Compose resolves relative paths
+            # against the compose file's directory, so anchor on ext_dir.
+            if vol_str and not vol_str.startswith("/") and "/" in vol_str:
+                dir_path = (ext_dir / vol_str.lstrip("./")).resolve()
                 try:
-                    dir_path.relative_to(INSTALL_DIR.resolve())
+                    dir_path.relative_to(ext_dir.resolve())
                 except ValueError:
                     logger.warning("Skipping out-of-tree volume path in %s: %s", service_id, vol_str)
                     continue
@@ -1100,10 +1107,15 @@ class AgentHandler(BaseHTTPRequestHandler):
             try:
                 flags = resolve_compose_flags()
 
+                ext_dir = _find_ext_dir(service_id)
+                if ext_dir is None:
+                    _write_progress(service_id, "error", "Installation failed",
+                                    error=f"Extension directory not found for {service_id}")
+                    return
+
                 # Step 1: Setup hook (if requested)
                 if run_setup_hook:
                     _write_progress(service_id, "setup_hook", "Running setup...")
-                    ext_dir = USER_EXTENSIONS_DIR / service_id
                     hook_path = _resolve_hook(ext_dir, "post_install")
                     if hook_path:
                         # Minimal allowlist env — mirror _execute_hook (L856-866)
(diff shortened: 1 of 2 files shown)


yasinBursali:
**Follow-up commit on this branch — anchor correction.**

While preparing a downstream change that depends on `_precreate_data_dirs`, I verified its behavior against a live install and discovered the `ext_dir` anchor introduced in this PR doesn't match where Compose actually mounts. Docker Compose v2's project directory is the parent of the first `-f` file (always `docker-compose.base.yml` → `INSTALL_DIR` in our setup), not the individual fragment's directory.

Live `docker inspect` on a running install confirms every relative bind mount resolves under `INSTALL_DIR/...`:

```
dream-token-spy  /Volumes/X/dreamserver/data/token-spy         -> /app/data
dream-whisper    /Volumes/X/dreamserver/data/whisper           -> /home/ubuntu/.cache/...
dream-paperless  /Volumes/X/dreamserver/data/paperless/media   -> /usr/src/paperless/media
dream-gitea      /Volumes/X/dreamserver/data/gitea             -> /var/lib/gitea
dream-crewai     /Volumes/X/dreamserver/data/crewai            -> /app
```

With the previous `ext_dir` anchor, `_precreate_data_dirs` was writing empty dirs into the extension source tree while Compose silently auto-created the real mount paths as root under `INSTALL_DIR` — guaranteed to break uid-sensitive containers (postgres, clickhouse) on Linux and cluttering the repo with unused `./data/` dirs.

This commit reverts the anchor to `INSTALL_DIR`, keeps the other two improvements from this PR intact (`_find_ext_dir` for built-ins, broadened relative-path filter), and updates the tests to assert the correct mount location. 44/44 tests pass locally.

Lightheartdevs:
Audit follow-up: needs revision before merge.

The bind precreation/state-verification goals are useful, but the branch has two blockers: its own regression test asserts the literal string `did not reach running state within 15s` while the implementation uses an f-string with `{startup_timeout}`, and the running-state verifier can break intentional one-shot extensions. Please fix the test and clarify state semantics for one-shot/setup-hook extensions.

yasinBursali:
Pushed audit follow-up (`61744614` + `a2b2e859`).

Both blockers from the 2026-04-28 audit are addressed:

**1. Test/impl mismatch** — the regression test was asserting the literal `did not reach running state within 15s`, but the implementation builds the message via f-string with `{startup_timeout}` (introduced in commit `9e751fbb`). Empirically verified the failure: `pytest -k test_install_writes_error_when_state_not_running` was reporting AssertionError on this branch. The assertion is now split: `did not reach running state within` (the constant prefix) AND `{startup_timeout}s` (the f-string template marker). Sabotage-resistant against a regression to either a hardcoded duration or non-f-string concat.

**2. State semantics for one-shot / setup-hook extensions** — `_handle_install._run_install` now reads `service.startup_check` from the manifest (default `True` for back-compat). When set to `False`, the post-`up -d` `docker inspect` poll is skipped entirely and the install completes after `compose up -d` returns 0:

```python
startup_check = install_service_def.get("startup_check", True)
if startup_check:
    # ... existing state-poll loop ...
    if state != "running":
        msg = f"Container did not reach running state within {startup_timeout}s ..."
        ...
        return
# Step 4: Success
_write_progress(service_id, "started", "Service started")
```

The opt-out covers both audit cases:
- One-shot init containers that exit cleanly after compose up
- Extensions whose value is purely the `setup_hook` (the container itself doesn't need to remain running)

No existing extension declares `startup_check` — verified with `grep -rn 'startup_check' --include='*.yaml'` returns zero hits across 17 built-in extensions and 33 community library entries. So back-compat is unchanged for everything currently shipping.

New test `test_install_supports_startup_check_opt_out` asserts both the manifest field is read with the True default and that the state poll is gated on `if startup_check:`. 47/47 host-agent tests pass.

Local CG **APPROVED** (no blockers). Three non-blocking items folded in or deferred:
- Comment honesty — folded into `a2b2e859` (the comment now mentions the opt-out path).
- Python truthiness on a quoted-string `startup_check: "false"` would re-enable the verifier (consistent with how the rest of the codebase reads manifest booleans; deferred to the documentation/schema PR).
- Behavioural test exercising the False branch via mocked `_read_manifest` — deferred (current source-grep tests follow the established `TestInstallStartCommandNoDeps` convention).

This unblocks the chain: #1039 (host-agent-enable-retry, currently CONFLICTING) and #1040 (langfuse-setup-hook, BLOCKED) can now rebase onto this branch's foundation, and #1052 (langfuse manifest test, BLOCKED) can stack on #1040.

PR #1031, opened 2026-04-24
fix(dashboard-api): error-progress parity, restart guidance, unhealthy status

yasinBursali:
## What
Three defect fixes across the dashboard-api extension lifecycle and the Extensions.jsx UI.

## Why
1. `enable_extension` never wrote `_write_error_progress` on host-agent failure, so the UI card hung at "installing" for 2 minutes before silently reverting to "stopped". `install_extension` already handled this correctly — parity gap only.
2. `install_extension` did return actionable restart guidance in the HTTP response body (`data.message = "Run 'dream restart' to recover"`) but the frontend's `handleInstall` fired `fetchCatalog() + pollProgress(serviceId)` and ignored `data.message`. The poller surfaced only the shorter `_write_error_progress` text, so operators never saw the "run dream restart" hint.
3. `_compute_extension_status` conflated two very different failure modes into "stopped" — a container that was intentionally off and a container that was crash-looping or health-probe-failing both showed the same Start button. Clicking it on the crashed case just replayed the failure.

## How
1. `enable_extension`'s stopped-path now mirrors `install_extension`: writes `_write_error_progress(service_id, <msg>)` on agent failure.
2. Both failure paths now write a single richer error message — `"Host agent failed to start extension. Run 'dream restart' to recover."` — so the progress poller surfaces the guidance. No frontend change needed.
3. `_compute_extension_status` now returns a new `"unhealthy"` status when the service-health poller reports `status="unhealthy"` (HTTP 4xx/5xx). Other failure modes (`down`, `degraded`, `not_deployed`) continue to return `"stopped"` — `"unhealthy"` is reserved for the clearest crash signal. `Extensions.jsx` adds `"unhealthy"` to STATUS_STYLES / DESCRIPTIONS / FILTERS / LABELS (all amber, reusing the existing toggle-stopped theme token), keeps `isStopped` as `=== 'stopped'` so the Start button isn't shown for crashed containers, adds a "Check Logs" CTA wired to the existing `onConsole` handler, and includes `unhealthy` in the summary bucket.

## Testing
- `pytest tests/test_extensions.py` — 140/140 pass (3 new tests).
- `pytest tests/` — full suite green (one pre-existing asyncio/Py3.14 failure in `test_gpu_detailed.py`, confirmed unrelated).
- `make lint` + `ruff` + `py_compile` — all clean.
- `pre-commit` — gitleaks / private-key / large-file pass.
- JSX grep audit: every status-branching location in `Extensions.jsx` symmetrically handles `"unhealthy"` (badge, filters, labels, summary, toggle, Check Logs, logs icon color). Zero inconsistencies found.

## Review
Critique Guardian: APPROVED WITH WARNINGS. No required changes.

## Known Considerations
- **Services without a health probe never reach `"unhealthy"`** — they get auto-`healthy` in `extensions_catalog`. Detecting "container crashed but no health endpoint" would require the host agent to expose `docker ps`/`docker inspect` state, which is out of scope.
- **Stale-window quirk:** during the first 5 minutes after enable-stopped, a 4xx-throwing container still reads as "installing" (existing grace window). After 5 minutes it correctly flips to `"unhealthy"`.

## Platform Impact
- **macOS / Linux / Windows (WSL2):** identical. Dashboard-api runs in a container on all three; `Extensions.jsx` runs in the browser. Amber Tailwind tokens render identically across engines.

### DIFF
--- a/dream-server/extensions/services/dashboard-api/routers/extensions.py
+++ b/dream-server/extensions/services/dashboard-api/routers/extensions.py
@@ -198,6 +198,13 @@ def _compute_extension_status(ext: dict, services_by_id: dict) -> str:
             svc = services_by_id.get(ext_id)
             if svc and svc.status == "healthy":
                 return "enabled"
+            # HTTP 4xx/5xx from the health endpoint is the clearest "container
+            # is up but broken" signal — surface it as "unhealthy" so the UI
+            # can prompt a log check. Timeouts / connection refused / DNS
+            # failures stay "stopped" because they don't distinguish a crashed
+            # container from an intentionally-stopped one.
+            if svc and svc.status == "unhealthy":
+                return "unhealthy"
             return "stopped"
         if (user_dir / "compose.yaml.disabled").exists():
             return "disabled"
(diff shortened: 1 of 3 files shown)


yasinBursali:
Superseded by #1038, which is a strict superset of this PR.

#1038 contains every hunk from this PR byte-for-byte (`routers/extensions.py`, `tests/test_extensions.py`, `Extensions.jsx`) and adds additional handling for `pre_start` return values and `post_start` warning surfacing. Closing this in favor of reviewing #1038 directly to avoid duplicate review effort.

PR #1033, opened 2026-04-24
fix(librechat): align MONGO_URI guard with sibling secrets

yasinBursali:
## ✅ Rebased on current `main` 2026-04-28 — Jupyter half dropped

Per the audit ask: the Jupyter half is now superseded by #1049 (which converted Jupyter's `command:` to exec-form list). This PR is reduced to the still-needed single-line LibreChat fix. Title updated.

## What
One-line guard fix in `resources/dev/extensions-library/services/librechat/compose.yaml`.

## Why
The `librechat-mongodb` service uses `${LIBRECHAT_MONGO_PASSWORD:?}` at initdb, enforcing a real password. The `librechat` service's `MONGO_URI` interpolated the same variable without `:?` — so when unset, `librechat` connected with `librechat:@` (empty password) while the DB required a real one. Operators hit silent MongoDB auth errors with no clear diagnostic.

## How
`librechat/compose.yaml:11` — change the MONGO_URI password reference to `${LIBRECHAT_MONGO_PASSWORD:?LIBRECHAT_MONGO_PASSWORD must be set}`, matching the convention already used everywhere else in the same file (JWT_SECRET, CREDS_KEY, CREDS_IV, LIBRECHAT_MEILI_KEY, and librechat-mongodb's own init password).

## Testing
- `unset LIBRECHAT_MONGO_PASSWORD && docker compose -f librechat/compose.yaml config` → explicit `required variable LIBRECHAT_MONGO_PASSWORD is missing a value: LIBRECHAT_MONGO_PASSWORD must be set` error at parse time (was silent runtime failure).
- `LIBRECHAT_MONGO_PASSWORD=test docker compose ... config` → parse advances to the next `:?` guard (CREDS_KEY), confirming our guard accepts a present value.
- YAML parse, pre-commit — all clean.

## Notes
- `LIBRECHAT_MONGO_PASSWORD` is auto-generated by `librechat/setup.sh` via `openssl rand`, so fresh installs are unaffected. Visible behaviour change is only on broken/edge-case setups where the password was somehow lost — they now fail at parse time with an explicit message instead of at runtime with a cryptic auth error.

## Platform Impact
- **macOS / Linux / Windows (WSL2):** Docker Compose variable-substitution is engine-side; behavior identical across all three platforms.

### DIFF
--- a/resources/dev/extensions-library/services/jupyter/compose.yaml
+++ b/resources/dev/extensions-library/services/jupyter/compose.yaml
@@ -11,7 +11,7 @@ services:
       - LLM_API_URL=${LLM_API_URL:-http://localhost:8000}
       - JUPYTER_TOKEN=${JUPYTER_TOKEN:?JUPYTER_TOKEN must be set}
       - NB_PREFIX=/
-    command: start.sh jupyter lab --NotebookApp.token=${JUPYTER_TOKEN:?JUPYTER_TOKEN must be set} --NotebookApp.password='' --NotebookApp.port=8888 --NotebookApp.ip=0.0.0.0 --NotebookApp.open_browser=False
+    command: start.sh jupyter lab --NotebookApp.token=${JUPYTER_TOKEN} --NotebookApp.password='' --NotebookApp.port=8888 --NotebookApp.ip=0.0.0.0 --NotebookApp.open_browser=False
     restart: unless-stopped
     security_opt:
       - no-new-privileges:true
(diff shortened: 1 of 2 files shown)


Lightheartdevs:
Audit follow-up: needs rebase/split after #1049.

The LibreChat `MONGO_URI` guard is useful, but the Jupyter half is now superseded by #1049 and still leaves required interpolation in the environment block on this branch. Please rebase on current `main`, drop the Jupyter half, and keep/retest only the LibreChat fix if it is still needed.

yasinBursali:
Rebased on current `main` and dropped the Jupyter half per the audit ask — that's superseded by #1049 (exec-form list conversion). PR is now a single-line LibreChat MONGO_URI guard fix. Title and description updated.

Verified: with `LIBRECHAT_MONGO_PASSWORD` unset, `docker compose config` now fails at parse time with an explicit message; with it set, parse advances normally to the next `:?` guard (CREDS_KEY).

PR #1039, opened 2026-04-24
fix(host-agent): retry install failure + DRY post_install hook executor

yasinBursali:
## Summary
Clicking **Enable** on an extension whose install left `{"status":"error"}` in its progress file silently took a different path from the original install: `docker compose up` without re-running the `post_install` hook and without writing the progress file. UI stayed stuck on the old error; extensions that failed because a hook didn't write required env vars failed again on retry for the same root cause.

A second commit on this PR extracts the post_install hook executor into a shared `_run_post_install_hook` helper used by BOTH the install path (`_handle_install._run_install`) and the new enable-retry path (`_enable_retry_work`). This closes the "shared helper would unify env allowlist + timeout handling" known-gap that was originally filed as a follow-up.

### Commit 1 — retry path (`_enable_retry_work`)
When `_handle_extension("start")` sees an error-status progress file, it now delegates to a daemon-thread retry path that:

1. Responds **202** with `{"status": "retrying"}` immediately (matches `_handle_install`'s 202-then-poll shape).
2. Writes `{"status": "starting"}` transitionally.
3. Re-runs the `post_install` hook if declared, with the **same argv / cwd / env allowlist** as `_handle_install` (no secrets leak — PATH, HOME, SERVICE_ID, SERVICE_PORT, SERVICE_DATA_DIR, DREAM_VERSION, GPU_BACKEND, HOOK_NAME only).
4. On hook success → `docker compose up`; writes `{"status": "started"}` or `{"status": "error"}` with compose stderr.
5. On hook failure → writes `{"status": "error"}` with hook stderr; compose not called.

**Non-retry paths** (stop, start-without-error, start-with-no-progress-file) preserve the existing synchronous `docker_compose_action` behavior. The `_service_locks[service_id]` is acquired before the progress check, so double-click Enable is still 409-safe.

### Commit 2 — DRY post_install hook executor (`_run_post_install_hook`)
Both `_run_install` and `_enable_retry_work` had byte-similar inline blocks running the `post_install` hook with the same env allowlist, the same `subprocess.run` shape, and the same failure-handling. Extracted the shared logic into:

```python
def _run_post_install_hook(service_id: str, ext_dir: Path) -> tuple[bool, str]:
    ...
```

**Behavior changes (intentional)**:
- Helper writes `setup_hook` / "Running setup..." progress **only when the hook actually resolves**. Previously `_run_install` wrote that progress message even when no hook was declared, producing a misleading "Running setup..." UI state for extensions without a `post_install` hook. The helper's no-hook branch returns `(True, "")` with no progress write.
- Helper uses tail-500 stderr slice + None guard (`(result.stderr or "")[-500:]`) — the tail of the stderr buffer is the most actionable failure context, and the None guard is defensive. `_enable_retry_work` previously used head-500 with a None guard; `_run_install` previously used tail-500 without the guard. The helper unifies both call sites on tail-500 + guard.
- Helper writes a specific `"post_install hook timed out (Ns)"` error on `subprocess.TimeoutExpired`. Previously `_run_install` relied on its outer `subprocess.TimeoutExpired` handler with a generic "timed out (Ns)" message; the helper produces the more specific message inside its own try/except. `_run_install`'s outer handler still fires for non-hook timeouts (`docker pull` / `up` / `inspect`).

The 8-key env allowlist (PATH, HOME, SERVICE_ID, SERVICE_PORT, SERVICE_DATA_DIR, DREAM_VERSION, GPU_BACKEND, HOOK_NAME) and the `subprocess.run` args are preserved verbatim from both call sites.

The third hook path, `_execute_hook` (`Popen` + `os.setsid` + `os.killpg`, `HOOK_TIMEOUT=120s`), is intentionally untouched — different shape, different concern.

## Platform Impact
- **macOS** (launchd-managed host-agent): identical — pure Python stdlib (subprocess/threading/json). No platform branches.
- **Linux** (systemd host-agent): same.
- **Windows-WSL2** (systemd-in-WSL host-agent): same.

## Testing
- `pytest tests/test_host_agent.py` → all retry-path + env-allowlist tests pass against the new structure. 4 retargeted env-allowlist source-level tests now grep `_run_post_install_hook` as the canonical home of the env allowlist, with a cross-check that the call-site (`_handle_install`) stays secret-free.
- 12 unrelated pre-existing failures (asyncio/Py3.14 compat) unchanged — verified via stash-and-baseline.
- ruff + `make lint` clean.
- `TestEnableRetry` tests cover: retry + hook success, retry + hook failure (compose skipped), no progress file (sync path), progress status=started (sync path).

Manual reproducer:
1. Install an extension with a `post_install` hook that writes env vars. Delete one of those env vars from `.env` manually to corrupt the post-install state.
2. Dashboard shows the extension in `error` state (write the progress file directly if needed).
3. Click **Enable** → `POST /v1/extension/start` returns 202. Dashboard polls progress: `starting → setup_hook → starting → started`. Missing env vars are repopulated by the re-run hook; container starts clean.

## Known gaps (still filed as fork follow-ups)
- ~~Hook execution duplicated between install and retry paths (~25 lines). Shared helper would unify env allowlist + timeout handling.~~ **Addressed in commit 2 of this PR.**
- 4 additional edge-case tests (no hook declared, corrupted JSON, mid-install statuses, lock-already-held) worth pinning — these are covered by an in-flight test-coverage branch.


Lightheartdevs:
Audit follow-up: keep draft / blocked by #1030.

The retry/progress-state work depends on the install-state semantics in #1030, and #1030 still needs revision. Please keep this draft blocked until #1030 is fixed, then rebase and prove the retry flow with tests covering both normal services and one-shot/setup-hook services.

Lightheartdevs (changes requested):
Thanks for pushing this retry path forward. I can't merge it in the current state because the PR's own Dashboard `api` job is red on `tests/test_host_agent.py`.

The live run for this head has three Linux API failures:

- `TestPrecreateDataDirs::test_creates_dirs_for_non_data_prefix` expects `./upload:/upload` to materialize under `INSTALL_DIR/upload`, but it is not created.
- `TestInstallRunningStateVerification::test_install_writes_error_when_state_not_running` no longer finds the expected dynamic `startup_timeout` error path in `_handle_install`.
- `TestInstallRunningStateVerification::test_install_supports_startup_check_opt_out` no longer finds the manifest `startup_check` opt-out path in `_handle_install`.

I also reproduced `test_host_agent.py` locally. Some local Windows-only symlink/chmod failures are environment noise, but the branch also surfaces new enable-retry/progress-file issues under Windows because `_write_progress` uses `os.rename` against an existing progress file and raises `FileExistsError` instead of replacing it.

Please rebase/fix the host-agent tests against the current merged host-agent flow and make the Dashboard API check green. Once the branch proves its own retry contract again, it is worth another pass.

yasinBursali:
Pushed fixes for the three failing tests + the Windows `_write_progress` issue.

Branch rebased onto current `main` (`d4124d6a`); 5 commits ahead.

**Tests now passing** — verified locally with `pytest tests/test_host_agent.py` → **65/65 PASS**:

- `TestPrecreateDataDirs::test_creates_dirs_for_non_data_prefix` — fixed by broadening `_precreate_data_dirs` to accept any relative bind source. Previously it only created host dirs for `./data/...` and `data/...` paths; now `./upload:/upload` correctly materializes `INSTALL_DIR/upload`. Path-traversal guard preserved (still resolves both sides and runs `relative_to(INSTALL_DIR.resolve())`).
- `TestInstallRunningStateVerification::test_install_writes_error_when_state_not_running` — passes; the rebase restored the `startup_check` opt-out + `startup_timeout` block in `_handle_install` that the earlier branch had dropped.
- `TestInstallRunningStateVerification::test_install_supports_startup_check_opt_out` — passes for the same reason.

**Windows `_write_progress` `FileExistsError`** — replaced `os.rename` with `os.replace` at all atomic-rename sites in `dream-host-agent.py` (lines 444 + 1204). `os.replace` overwrites atomically on both POSIX and Windows; `os.rename` on Windows raises `FileExistsError` when the destination exists, which is exactly what bit retries on existing progress files. Verified zero `os.rename` calls remain in the file.

#1030's install-state semantics that this PR depends on are now in `main`, so the stack-blocker is resolved.

Ready for fresh review.

Lightheartdevs (approved):
I took this one over because Yasin is inactive, merged current main into the branch, and audited the retry flow end to end.

Found and fixed one remaining blocker before merge: the enable-retry worker reran setup and called compose, then marked progress as started without verifying the container actually reached `running`. That could leave the UI with a false success when `docker compose up -d` returns 0 for an exited/restarting container. The branch now mirrors the install path's manifest-driven startup check, including `startup_check: false` opt-out and `startup_timeout`.

Validation:
- `python -m pytest dream-server/extensions/services/dashboard-api/tests/test_host_agent.py -q -k "EnableRetry or InstallRunningStateVerification or InstallHookEnvAllowlist or PrecreateDataDirs"` -> 17 passed
- `git diff --check`
- GitHub Actions all green on `f604a8a`

The full local `test_host_agent.py` run still has existing Windows-harness limitations around symlink creation/executable-bit checks, but the targeted retry/host-agent coverage is clean and Linux CI is green. Looks ready now.

PR #1040, opened 2026-04-24
fix(langfuse): chown postgres/clickhouse data dirs to image uids on Linux

yasinBursali:
> **DRAFT: must merge AFTER #1030.** This branch is based on `fix/host-agent-install-flow` (#1030) and depends on `_find_ext_dir` for built-in hook discovery. Promote to ready-for-review after #1030 merges.

## Summary
langfuse's `postgres:17.9-alpine` (uid **70**) and `clickhouse/clickhouse-server:26.2.4.23` (uid **101**) containers crash on Linux native Docker when their bind-mounted data dirs are owned by the install user. Phase 06 pre-creates the dirs but its generic chown loop restores install-user ownership — not the per-image uids these containers expect. macOS Docker Desktop and Windows WSL2 Docker Desktop mask the issue (osxfs/virtiofs translates uids transparently); **Linux native Docker does not**.

## How
Adds `extensions/services/langfuse/hooks/post_install.sh`:
- **Darwin**: short-circuits with a log message, exits 0 (macOS Docker Desktop masks uid).
- **Linux / WSL2**: `chown -R 70:70` on the postgres dir and `101:101` on the clickhouse dir. Uses `sudo` when available; falls back to direct chown when root.
- **Fail-hard on any of 3 failure paths** (mkdir fails without sudo, sudo chown fails, plain chown fails without sudo): exits non-zero with an actionable ERROR message that includes the exact manual-recovery command. Host agent surfaces this to the dashboard via `progress.error`, so users see the error immediately instead of a silent install-success followed by a postgres crash-loop.

Also adds `tests/reproducers/langfuse-uid-check.sh` as a documented two-phase reproducer (install-user-owned → chowned-to-70) to demonstrate the bug and verify the fix.

## Platform Impact
| Platform | Behavior |
|---|---|
| macOS Docker Desktop | Hook exits 0 (Darwin short-circuit). osxfs masks uid. |
| Linux native Docker | Hook runs chown; fails loud with actionable message on failure. |
| Windows WSL2 Docker Desktop | Hook runs chown (virtiofs masks uid so it's harmless-but-unnecessary). |

## Testing
- `shellcheck hooks/post_install.sh` → 0 findings.
- `shellcheck tests/reproducers/langfuse-uid-check.sh` → 0 findings.
- `bash -n` on both → OK.
- Darwin live test (macOS dev box): `bash post_install.sh /tmp/fake-install cpu` → prints "macOS Docker Desktop handles uid translation; no chown needed" and exits 0.
- `pytest tests/test_host_agent.py` → 44/44 pass.
- `make lint` + `make test` clean.

Manual (Linux / WSL2):
1. Fresh `dream install`.
2. Enable langfuse from the dashboard → hook runs, postgres/clickhouse containers start cleanly.
3. Optional diagnostic: `bash tests/reproducers/langfuse-uid-check.sh` — two-phase reproducer, expect phase 1 fail + phase 2 pass on Linux.

## Known gaps (filed as fork follow-ups)
- No structural pytest asserting manifest `setup_hook` field and hook file coexist.
- Hardcoded uids (70/101) are tied to current pinned image tags; future image bumps should re-verify.


yasinBursali:
Pushed a follow-up commit (`45ca9983`) addressing a data-integrity concern flagged in review:

The original `chown -R` ran unconditionally, which is a low-probability but real corruption vector when re-invoked on a live install (the hook is independently re-callable via the host-agent's setup_hook endpoint). The follow-up adds an `is_container_running` guard that skips chown when the postgres/clickhouse container is already up — a running container proves ownership is already correct (the image refuses to start otherwise), so re-chowning is wasted I/O at best and races with WAL fsync at worst.

**One trade-off worth flagging:** the helper uses `docker ps ... 2>/dev/null` to suppress errors from a missing/stopped Docker daemon, falling through to the chown path. This technically conflicts with CLAUDE.md §4 ("never silence stderr"), but the alternative (letting docker errors crash the hook) would regress installer UX during the common first-install case where Docker may be transitioning. Internal critique-guardian review: APPROVED-WITH-WARNINGS, no follow-up required.

Lightheartdevs:
Audit follow-up: keep draft / stacked on #1030.

The Langfuse UID hook direction is good, but this branch is explicitly stacked on #1030, which still needs work. Please keep this draft until the #1030 install-state foundation is corrected, then rebase and rerun the Langfuse hook tests.

Lightheartdevs (changes requested):
﻿Thanks for getting the Langfuse UID hook stacked onto the now-merged #1030 path. I rechecked this branch before approving and found one blocker:

The hook only skips `Darwin`, so on native Windows/Git Bash it treats `MINGW64_NT-*` as a Linux-family host and runs the `sudo chown` path. I reproduced that directly on the Windows DreamServer checkout:

```text
$ uname -s
MINGW64_NT-10.0-26200
$ bash dream-server/extensions/services/langfuse/hooks/post_install.sh "$tmp" cpu
langfuse post_install: chown -R 70:70 .../data/langfuse/postgres
Sudo is disabled on this machine...
langfuse post_install: ERROR: 'sudo chown -R 70:70 ...' failed
rc=1
```

Because dashboard installs pass `run_setup_hook: true`, this would make Langfuse fail during setup on native Windows before Docker gets a chance to start it. The UID problem is Linux-native Docker specific; Windows Docker Desktop should skip the chown the same way macOS does. Please gate the chown path to real Linux/WSL only, or explicitly skip `MINGW*`, `MSYS*`, and `CYGWIN*` platforms with a clear log line.

What did pass locally: `bash -n` on the hook and reproducer, `python -m py_compile dream-server/bin/dream-host-agent.py`, `pytest tests/test_hooks.py -q` (11/11), and `git diff --check`. `test_host_agent.py` still has Windows-local symlink/chmod failures unrelated to this branch.

yasinBursali:
Pushed the platform-gate fix you flagged for native Windows / Git Bash. Commit `1b12d492` switches the hook from a Darwin-only blacklist to a Linux-only whitelist:

```diff
-if [[ "$PLATFORM" == "Darwin" ]]; then
-    log "macOS Docker Desktop handles uid translation; no chown needed"
+if [[ "$PLATFORM" != "Linux" ]]; then
+    log "Skipping chown on $PLATFORM (uid alignment only needed on native Linux Docker)"
     exit 0
 fi
```

WSL2 reports `uname -s` = `Linux` per Microsoft docs, so it stays covered. macOS Docker Desktop, Windows Git Bash (`MINGW64_NT-*`), MSYS2 (`MSYS_NT-*`), Cygwin (`CYGWIN_NT-*`), and the BSDs all now correctly skip with a clear log line — no more `Sudo is disabled on this machine` failure on dashboard-driven Windows installs.

Verified the platform-skip behavior against all six relevant `uname -s` strings; only `Linux` enters the chown path.

Branch also rebased onto current `main` (`d4124d6a`); 4 commits ahead.

Ready for fresh review.

Lightheartdevs (approved):
Approved after re-auditing on current main.

What I verified:
- merged current main into the branch with no PR-file conflicts
- previous Windows/Git-Bash blocker is fixed: native Windows now skips the Linux-only chown path
- hook contract matches host-agent setup_hook/post_install execution
- simulated Linux first-install path creates/chowns postgres and clickhouse dirs with 70:70 and 101:101
- simulated running-container path skips recursive chown to avoid live data races
- simulated docker-query failure fails closed before chowning data

Local checks:
- bash -n on hook and reproducer
- python -m pytest dream-server/extensions/services/dashboard-api/tests/test_hooks.py -q: 11 passed
- direct YAML parse confirms langfuse setup_hook path
- git diff --check

GitHub Actions are green on the updated branch.

PR #966, opened 2026-04-15
docs(platform): sync Windows and macOS support docs

boffin-dmytro:
## Summary

This cleans up the remaining Windows/macOS support doc contradictions in the shipped user-facing docs.

The main mismatch before this change was that the root README and support matrix described Windows as supported today, while the Windows quickstart still described a preflight-only future state.

## What changed

- rewrote `WINDOWS-QUICKSTART.md` to describe the current supported Windows path through Docker Desktop + WSL2
- removed stale "coming soon", "preflight only", and past-dated placeholder language
- updated `SUPPORT-MATRIX.md` so Windows/macOS support language is internally consistent
- refreshed the root `README.md` Windows note to match the current installer and CLI path
- updated `WINDOWS-INSTALL-WALKTHROUGH.md` to remove stale installer details such as the old download URL, old install path, and nonexistent flags

## Impact

- users now get one consistent support story across the main docs
- Windows support is described accurately without implying unsupported native-only behavior
- macOS remains documented as supported without overstating full Linux parity

## Validation

- `git diff --check`
- targeted stale-phrase search across the touched docs for old Windows support wording

PR #1042, opened 2026-04-25
feat(support): add redacted diagnostics bundle generator

boffin-dmytro:
## Summary

This adds a standalone support bundle generator for DreamServer.

The new `scripts/dream-support-bundle.sh` command creates a redacted `.tar.gz` bundle under `artifacts/support/` so users can share one consistent diagnostics package when installs, Docker, GPU detection, extension startup, or runtime health go wrong.

## What changed

- added `scripts/dream-support-bundle.sh`
- added `tests/test-support-bundle.sh`
- added `docs/SUPPORT-BUNDLE.md`

The bundle collects:
- Dream Doctor output when available
- extension audit JSON
- compose resolution and validation output
- Docker version/info/container summary when Docker is available
- short DreamServer container log tails
- platform, git, disk, memory, port, manifest, env schema, and redacted env details

It is best-effort by design:
- missing Docker does not fail the whole bundle
- failing diagnostic subcommands are recorded in the manifest instead of aborting the run
- raw `.env` is never included


Lightheartdevs:
Audit follow-up: needs a Git Bash path fix.

The support bundle feature is useful, but the audit found `--json` can emit Windows-style paths under Git Bash that the PR's own Bash test cannot feed back into `tar`. Please make the JSON path contract consumable by the shell test environment, or emit both native and POSIX-safe paths with tests proving extraction works.

boffin-dmytro:
## Update — Git Bash path fix pushed

Addressed the path portability issue raised in the audit follow-up.

**Root cause:** `write_summary_json()` was using Python's `str(Path(...))` which returns backslash paths on Windows/Git Bash. The Bash test then tried to pass those paths to `[[ -d ]]` and `tar`, which don't handle Windows-style paths in a Bash context.

**Fix (ca1ed1f):**
- Changed `str(bundle_dir)`, `str(archive_path)`, `str(manifest)` → `.as_posix()` in the `write_summary_json()` heredoc — always emits forward-slash paths regardless of platform
- Added a POSIX-path assertion to `test-support-bundle.sh` that fails if any backslash appears in the JSON path fields

All 15 tests pass locally.

---

**Re: `integration-smoke` CI failure** — the failing BATS test (`not ok 67: _docker_cmd_arr: returns sudo docker when DOCKER_CMD is sudo docker`, `docker-phase.bats:100`) is pre-existing and unrelated to this PR. The same test was failing on PR #966 (run 2026-04-18) before this branch existed. This PR touches no installer phases or docker-phase code.

PR #1043, opened 2026-04-25
fix(installer): custom menu's 'n' answers were not actually disabling services

y-coffee-dev:
Fixes two compounding bugs caused users in Custom mode to get almost every optional service installed regardless of their answers:

1. The prompt logic used '[[ $REPLY =~ ^[Nn]$ ]] || ENABLE_X=true', which only SET flags to true and never to false. Since install-core defaults were already true (VOICE/WORKFLOWS/RAG/OPENCLAW/COMFYUI/ DREAMFORGE), pressing 'n' was a no-op. Replaced with explicit if/else branches that set the flag from the user's answer.

2. Only ComfyUI, DreamForge, and Langfuse had the compose.yaml → .disabled rename logic. For whisper, tts, n8n, qdrant, openclaw the ENABLE_* flag only gated cosmetic things (image pre-pull, health checks, summary URLs) — the compose file was always picked up by resolve-compose-stack.sh, so the service started anyway. Refactored the three existing rename blocks into a reusable helper and extended coverage to voice/workflows/rag/openclaw.

Also added the missing CLI equivalents --no-voice, --no-workflows, --no-rag, --no-openclaw so non-interactive installs get the same opt-out coverage as the Custom menu.

### DIFF
--- a/dream-server/install-core.sh
+++ b/dream-server/install-core.sh
@@ -121,9 +121,13 @@ Options:
     --tier N          Force specific tier (1-4) instead of auto-detect
     --cloud           Cloud mode: skip GPU detection, use LiteLLM + cloud APIs
     --voice           Enable voice services (Whisper + Kokoro)
+    --no-voice        Disable voice services
     --workflows       Enable n8n workflow automation
+    --no-workflows    Disable n8n workflow automation
     --rag             Enable RAG with Qdrant vector database
+    --no-rag          Disable RAG / Qdrant
     --openclaw        Enable OpenClaw AI agent framework
+    --no-openclaw     Disable OpenClaw
     --comfyui         Enable ComfyUI image generation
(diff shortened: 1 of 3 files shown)


Lightheartdevs:
Audit follow-up: needs one more optional-service sync fix.

This correctly starts syncing optional compose files for the custom menu, but the RAG opt-out only disables `qdrant`. `embeddings` still has `compose.yaml`, so the resolver can still include/pull/start TEI after the user answers no to RAG. Please add `embeddings` to the `ENABLE_RAG` sync list and cover the opt-out with a regression check.

y-coffee-dev:
Good catch, I  fixed it by adding embeddings to the ENABLE_RAG sync block in 03-features.sh so opting out of RAG now disables both the qdrant compose file and the embeddings (TEI) compose file.

PR #1045, opened 2026-04-25
fix(dashboard-api,host-agent): route extension config sync through host agent

yasinBursali:
## Summary

`_sync_extension_config` in `routers/extensions.py` calls `shutil.copytree` to copy `<ext_dir>/config/*` into `INSTALL_DIR/config/`. The dashboard-api container has `/dream-server/config` bind-mounted **read-only** (`docker-compose.base.yml:176`, present since v2.0.0), so the copy fails:

```
OSError: [Errno 30] Read-only file system: '/dream-server/config/<svc>'
```

The helper was added later, in commit `4dbabc53` (\"fix(extensions): sync config/ subdir on install\"), without coordinating with the read-only mount. Any user-installed extension that ships a `config/<id>/` subdirectory has been silently broken on install since that change. In the current community library this affects `continue` and `sillytavern`. Built-ins are unaffected because their configs are pre-created by the installer at phase 06.

This fix follows the established pattern (see `_handle_extension_compose_toggle` / `_call_agent_compose_rename`): the dashboard-api delegates host-side mutations to the host agent, which runs on the writable host filesystem.

## What this changes

**Host agent** — new `POST /v1/extension/sync_config`:
- Bearer auth (`check_auth`), JSON body `{\"service_id\": \"...\"}`, validated against `SERVICE_ID_RE`
- Resolves under `USER_EXTENSIONS_DIR / sid` only — built-ins return a 200 no-op (their bootstrap is installer-managed; resyncing risks overwriting user-modified files)
- If `<ext>/config/` is absent, returns 200 no-op
- **Rejects all symlinks** anywhere in the config tree (top-level `config/` itself, symlinked dirs at any depth, symlinked files). `_copytree_safe` strips symlinks at install time, so legitimate user extensions never have any — a symlink at sync time implies tampering. Iterating `dirs + files` (not just `files`) is required because `os.walk(followlinks=False)` does not recurse into symlinked directories
- `copytree(symlinks=False)` into `INSTALL_DIR/config/<entry>` with a path-traversal guard
- chmods `*.sh` files in the synced service tree executable
- Per-service `threading.Lock` to prevent concurrent sync races

**Dashboard-api** — `_sync_extension_config` becomes a thin wrapper around new `_call_agent_sync_config(service_id) -> bool`. The function signature changes from `-> None` to `-> bool`; the only call site in `install_extension` ignores the return, preserving fire-and-forget semantics with a warning log on failure.

## Why reject all symlinks (not just escaping ones)

A first cut only rejected symlinks whose target escaped the extension directory. Empirically this missed **symlinked directories** entirely: `os.walk(followlinks=False)` does not recurse into them — they appear in the parent's `dirs` list and get skipped — and `child.is_dir()` later in the function follows the link, takes the directory branch, and `shutil.copytree(symlinks=False)` then dereferences the link target. Reproduced locally: `config/leak -> ~/.ssh` resulted in SSH keys being copied to `INSTALL_DIR/config/leak/`. The unconditional-reject policy closes the gap and is justified by `_copytree_safe`'s install-time stripping.

## Test plan

**Dashboard-api side** (3 unit tests in `test_extensions.py::TestSyncExtensionConfig`):
- [x] Delegates to `_call_agent_sync_config` with the service id
- [x] Returns False on agent failure
- [x] HTTP request shape: URL, method, bearer auth header, body content

**Host agent side** (8 wire tests in `test_host_agent.py::TestSyncExtensionConfigWire`, with a real `HTTPServer` + `AgentHandler`):
- [x] Happy path: config tree copied, .sh files become executable
- [x] No-op when extension has no `config/` subdir
- [x] Auth rejection (no copy attempted)
- [x] **Symlinked directory** in `config/<svc>/leak -> /tmp/secret/` rejected with 400; no files leak into `INSTALL_DIR/config/`
- [x] **Symlinked file** rejected (regression lock for the previously-covered case)
- [x] **Top-level `config/` itself a symlink** rejected (covers the upfront `is_symlink` guard)
- [x] Invalid `service_id` rejected (path traversal, uppercase, space, empty, `..`)
- [x] Built-in-only service returns 200 no-op (pins the \"do not overwrite installer-managed configs\" decision)

**End-to-end on macOS Apple Silicon:**
- [x] Clean install of `continue` extension via dashboard returns 200; container `dream-continue` Up (healthy) on `127.0.0.1:8890`
- [x] `INSTALL_DIR/config/continue/{entrypoint.sh,nginx.conf}` synced by host agent; entrypoint.sh has `-rwxr-xr-x`
- [x] All 8 wire tests pass against the deployed `bin/dream-host-agent.py`


Lightheartdevs:
Audit follow-up: boundary is right, copy contract needs tightening.

Moving extension config sync into the host agent makes sense, but copying every top-level child of `<extension>/config` into `INSTALL_DIR/config/<child.name>` lets an extension overwrite unrelated service config trees such as `open-webui` or `litellm`. Please restrict the default contract to `config/<service_id>` or add an explicit manifest allowlist for cross-service config writes.

yasinBursali:
Pushed follow-up commit `6fc24167` per the audit ask: default copy contract restricted to `<ext>/config/<service_id>/` only. Took the maintainer's first option (default-only restriction) — manifest-allowlist escape hatch deferred until a real cross-service use case surfaces.

Sibling entries (e.g. `<ext>/config/open-webui/`, top-level files) are inventoried and reported in the response's `skipped` array, but never written to INSTALL_DIR.

**Order matters and is verified:** the whole-tree symlink rejection runs BEFORE the contract-restriction inventory. Without that ordering, an `ext_config` that is itself a symlink would have its target's contents enumerated by `iterdir()` before the upfront `is_symlink()` guard fires. All 3 existing symlink tests still pass against the reordered code.

New regression test `test_default_contract_only_copies_own_service_subdir` exercises the audit's exact scenario: an extension with both its own legitimate `config/<sid>/` AND a malicious `config/open-webui/` payload. Asserts byte-for-byte preservation of the pre-existing `INSTALL_DIR/config/open-webui/config.json` after the call — the strongest possible security oracle.

`pytest tests/test_host_agent.py` → 47/47. `pytest tests/test_extensions.py` → 138/138. ruff clean.

PR #1051, opened 2026-04-26
fix(resolver): hoist yaml import, guard empty manifests, align user-ext loop

yasinBursali:
## ✅ Rebased on current `main` 2026-04-29 + audit feedback addressed

This PR went through the maintainer audit comparing it against sibling PR #1029. Maintainer's verdict on #1029: rejected — it silently drops legacy/custom user extensions that have `compose.yaml` without a manifest. Maintainer's verdict on this PR: "better than #1029, but still needs the GPU filter."

I rebased on current `main` (#1004's apple-deadlock guard now visible in the built-in loop) and added two follow-up commits to close the audit-flagged gaps.

## What

Three resolver-hygiene improvements in `scripts/resolve-compose-stack.sh`, plus the audit follow-ups:

1. Hoist the optional PyYAML `import yaml` above both extension loops.
2. Add an `isinstance(manifest, dict)` guard after every `yaml.safe_load` / `json.load` so empty manifests warn-and-skip instead of crashing the resolver with `AttributeError`.
3. Align the user-extension loop's structure with the built-in loop (compose_file from manifest, `.disabled` handling, `compose.local.yaml` + `compose.multigpu.yaml` overlays).
4. **(NEW commit `50ae35f0`)** Apply the `gpu_backends` filter to the user-extension path, gated on `isinstance(manifest, dict)` so the manifest-less compat carve-out is preserved.
5. **(NEW commit `3572c973`)** Apple-guard the user-ext `compose.local.yaml` overlay — same predicate as the built-in loop got from PR #1004 (`gpu_backend != "apple"`). Without this, a user extension shipping `compose.local.yaml` with `depends_on: llama-server: condition: service_healthy` would deadlock the stack on macOS (llama-server runs natively, replicas: 0, the healthcheck never satisfies).

## Why

**1–3:** see the original PR motivation — yaml hoist consolidates two duplicate try/except blocks, isinstance guard prevents `AttributeError` on empty/comment-only YAML, user-ext loop alignment honors `.disabled` + overlays the same way built-ins do.

**4. gpu_backends filter on user-ext path:** the audit exposed that a user extension marked AMD-only could still be appended to an NVIDIA / CPU / Apple stack because the user-ext loop skipped the filter. The new code mirrors the built-in loop's lines 177–180 verbatim. **Gated on `isinstance(manifest, dict)`** so the manifest-less compat carve-out (legacy user extensions that pre-date the manifest convention) keeps working — those have no `gpu_backends` to filter on and continue to be included regardless of host backend.

**5. Apple guard on user-ext compose.local.yaml:** the original branch was on a pre-#1004 base, so locally the user-ext loop's symmetry-by-accident with the built-in loop hid this gap. Post-rebase the built-in loop has the guard back; the user-ext loop has to mirror it explicitly. Without the guard, any user extension shipping a `compose.local.yaml` with `depends_on: llama-server: condition: service_healthy` (the canonical pattern — all 7 built-in `compose.local.yaml` fragments use it) would hang the stack on Apple Silicon.

## How

- `import yaml` block lifted to a single shared try/except above the `if ext_dir.exists():` guard.
- `if not isinstance(manifest, dict): print("WARNING: empty/non-dict manifest for ...", file=sys.stderr); continue` added in both loops.
- User-extension loop now mirrors the built-in loop:
  - Read `service.compose_file` (default `"compose.yaml"`)
  - Skip when `compose_rel.endswith(".disabled")` OR sibling `<base>.disabled` file exists
  - Apply `compose.local.yaml` overlay when `dream_mode in ("local", "hybrid", "lemonade")` **AND `gpu_backend != "apple"`**
  - Apply `compose.multigpu.yaml` overlay when `gpu_count > 1`
- New `gpu_backends` filter inserted after the manifest parse, before `compose_rel`:
  ```python
  if isinstance(manifest, dict):
      backends = service.get("gpu_backends", ["amd", "nvidia"])
      if gpu_backend not in backends and "all" not in backends and "none" not in backends:
          continue
  ```
- `manifest = None` initialized at the top of the per-extension try block so the new isinstance gate is safe whether `manifest_path` is None or the file parses to non-dict.

**Compat carve-out:** user extensions WITHOUT a manifest still work — the `manifest_path is None` branch falls through to `service = {}` and the default `compose.yaml`. PyYAML-unavailable + manifest_path-is-None both end up with `manifest=None`, both intentionally bypass the gpu_backends filter.

## Testing

- `bash -n` + `shellcheck` (zero new warnings) → PASS
- All resolver tests pass:
  - `tests/test-resolve-compose-resilient.sh` — 9/0
  - `tests/test-extension-integration.sh` — 32/0 (1 skipped)
  - `tests/test-hardware-compatibility.sh` — 32/0
  - `tests/test-cpu-only-path.sh` — 9/0

**Smoke matrix** (3 fixtures × 4 backends, run on the rebased branch):

| | amd | nvidia | apple | cpu |
|---|:---:|:---:|:---:|:---:|
| amd-only-ext (manifest, gpu_backends:[amd]) | IN | OUT | OUT | OUT |
| legacy (no manifest) | IN | IN | IN | IN |
| with-local/compose.yaml | IN | IN | IN | IN |
| with-local/compose.local.yaml | IN | IN | **OUT** | IN |

The apple cell for `with-local/compose.local.yaml` is the deadlock-prevention proof.

Manual (operator on macOS dev install):
1. `cd /Volumes/X/dreamserver && ./dream-cli list` — extension list outputs cleanly
2. Empty user-ext manifest: `mkdir -p /Volumes/X/dreamserver/data/user-extensions/empty-test && touch .../empty-test/manifest.yaml && echo "services: {}" > .../empty-test/compose.yaml && bash dream-server/scripts/resolve-compose-stack.sh` → prints `WARNING: empty/non-dict manifest ...` to stderr, NOT a stack trace
3. Cleanup: `rm -rf /Volumes/X/dreamserver/data/user-extensions/empty-test`

## Relationship with PR #1029

This PR and #1029 both target the user-ext gpu_backends filter, but take different approaches:

- **#1029** drops legacy compose-only (no manifest) user extensions silently. Maintainer rejected this approach.
- **This PR** preserves the manifest-less compat carve-out AND now applies the gpu_backends filter when a manifest IS present. Maintainer's audit on this PR explicitly approved the "better than #1029" framing.

Once this PR lands, **#1029 should be closed as superseded.**

## Known Considerations

**1. Intentional empty-manifest semantic asymmetry between loops:**
- Built-in loop: empty manifest → `continue` (skip).
- User-ext loop: only NON-None non-dict manifests trigger the warning; truly empty (manifest=None) falls through to the manifest-less compat default (`service = {}`).

This preserves backward compatibility for user extensions that historically shipped without any manifest at all. Built-in extensions ship with manifests by convention; user-installed extensions did not always.

**2. compose_file default differs:** built-in loop uses `service.get("compose_file", "")`; user-ext loop uses `service.get("compose_file", "compose.yaml")`. Required for the manifest-less compat carve-out above.

**3. Path-traversal hardening on `compose_file`** (`.resolve()` + explicit subpath check) is a pre-existing concern in both loops. Tracked as fork follow-up issue #507.

**4. Stacked test PR for the user-extension loop scenarios** is tracked as fork follow-up issue #508.

## Platform Impact

| Platform | Impact |
|---|---|
| macOS | **Active.** The apple guard on user-ext `compose.local.yaml` prevents the llama-server-deadlock from extending to user-installed extensions. |
| Linux (NVIDIA / AMD / CPU / Intel) | Active — gpu_backends filter prevents wrong-backend user extensions from joining the stack. |
| Windows (WSL2) | Same as Linux. |

### DIFF
--- a/dream-server/scripts/resolve-compose-stack.sh
+++ b/dream-server/scripts/resolve-compose-stack.sh
@@ -287,8 +287,14 @@ if user_ext_dir.exists():
                 if gpu_overlay.exists():
                     resolved.append(str(gpu_overlay.relative_to(script_dir)))
 
-                # Mode-specific overlay — depends_on for local/hybrid mode only
-                if dream_mode in ("local", "hybrid", "lemonade"):
+                # Mode-specific overlay — depends_on for local/hybrid mode only.
+                # Skip on Apple Silicon: macOS runs llama-server natively on the host
+                # (Docker service has replicas: 0), so `depends_on: llama-server:
+                # service_healthy` inside compose.local.yaml overlays can never be
+                # satisfied and deadlocks the stack. The real LLM-ready gate on macOS
+                # is the `llama-server-ready` sidecar defined in the macOS overlay.
+                # Mirrors the same guard in the built-in loop above (PR #1004).
+                if dream_mode in ("local", "hybrid", "lemonade") and gpu_backend != "apple":
                     local_mode_overlay = service_dir / "compose.local.yaml"
                     if local_mode_overlay.exists():
                         resolved.append(str(local_mode_overlay.relative_to(script_dir)))
--- a/dream-server/scripts/resolve-compose-stack.sh
+++ b/dream-server/scripts/resolve-compose-stack.sh
@@ -137,15 +137,16 @@ if not resolved:
 if gpu_count > 1 and (script_dir / "docker-compose.multigpu.yml").exists():
     resolved.append("docker-compose.multigpu.yml")
 
+# Optional PyYAML — shared by built-in and user-installed extension loops.
+try:
+    import yaml
+    yaml_available = True
+except ImportError:
+    yaml_available = False
+
 # Discover enabled extension compose fragments via manifests
 ext_dir = script_dir / "extensions" / "services"
 if ext_dir.exists():
-    try:
-        import yaml
-        yaml_available = True
-    except ImportError:
-        yaml_available = False
-
     for service_dir in sorted(ext_dir.iterdir()):
         if not service_dir.is_dir():
             continue
@@ -166,6 +167,9 @@ if ext_dir.exists():
                     manifest = yaml.safe_load(f)
                 else:
                     continue  # skip YAML manifests when PyYAML unavailable
+            if not isinstance(manifest, dict):
+                print(f"WARNING: empty/non-dict manifest for {service_dir.name} at {manifest_path}, skipping", file=sys.stderr)
+                continue
             if manifest.get("schema_version") != "dream.services.v1":
                 continue
             service = manifest.get("service", {})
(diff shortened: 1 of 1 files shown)


yasinBursali:
**MUST MERGE AFTER #1029.** This PR shares `scripts/resolve-compose-stack.sh` with #1029 (resolver gpu_backends sweep). DRAFT until #1029 merges; rebasing on top is straightforward — keep #1029's gpu_backends filter and drop the duplicate yaml import that #1029 added inside the user-ext loop (this PR hoists it above both loops).

Lightheartdevs:
Audit follow-up: better than #1029, but still needs the GPU filter.

This preserves the user-extension fallback better than #1029, but the user-extension path does not apply `service.gpu_backends` before appending `compose_rel`. A user extension marked AMD-only can still be included in an NVIDIA/CPU/Apple stack. Please apply the same GPU-backend filter used for built-ins before adding user-extension compose files.

yasinBursali:
Rebased on current `main` and pushed two follow-up commits:

1. **`50ae35f0`** — apply gpu_backends filter to user-extension path, gated on `isinstance(manifest, dict)` so the manifest-less compat carve-out is preserved (the discriminator vs #1029 — that PR drops legacy extensions silently; this one keeps them).
2. **`3572c973`** — CG re-review caught that the user-ext loop's new `compose.local.yaml` inclusion was missing the apple-deadlock guard from #1004. Added the same `gpu_backend != "apple"` predicate the built-in loop uses, with a comment cross-referencing #1004.

Smoke matrix (3 fixtures × 4 backends): `amd-only-ext` correctly excluded on nvidia/apple/cpu, `legacy` (manifest-less) included on all 4, `with-local/compose.local.yaml` correctly skipped on apple only. `bash -n` + `shellcheck` clean. Resolver test suite 82/0.

Once this lands, #1029 should be closed as superseded — the maintainer's audit on #1029 explicitly preferred this PR's approach. Description fully refreshed.

yasinBursali:
Promoting from draft. Two pieces of context:

**#1029 closed today.** It was the predecessor PR this branch superseded (per maintainer's own audit on #1029, which preferred this PR's approach). With #1029 out of the queue, the original "wait for #1029 disposition" condition is met.

**Test coverage situation (full transparency):** an independent audit pass flagged "no in-PR tests for a resolver change" as a quality concern. The PR's existing claims hold — the 82-case existing resolver test suite passes (`test-resolve-compose-resilient.sh` 9/0, `test-extension-integration.sh` 32/0+1skip, `test-hardware-compatibility.sh` 32/0, `test-cpu-only-path.sh` 9/0), plus the documented 3 fixtures × 4 backends smoke matrix manually verified — but a focused unit test for the new `gpu_backends` filter and `compose.local.yaml` apple-guard branches isn't in this PR. Fork issue #508 tracks that as a follow-up "stacked test PR for the user-extension loop scenarios."

If you'd prefer the test to land in this PR rather than as the follow-up, I can scaffold a focused 4-case regression (gpu_backends filter on user-ext, apple-guard on compose.local.yaml, empty-manifest skip, `gpu_backends: ["all"]` on every backend) and push as one more commit before merge. Otherwise the smoke matrix + #508 follow-up is the current contract.

Lightheartdevs (approved):
Re-audited against current main. The earlier user-extension GPU backend gap is fixed, manifestless user-extension fallback is preserved, and the Apple compose.local guard now mirrors the built-in extension path. Local checks: git diff --check clean, Git Bash bash -n clean, test-resolve-compose-resilient.sh passed 9/0, test-cpu-only-path.sh passed 7/0 with the existing dry-run skip, and a synthetic user-extension matrix confirmed AMD-only exclusion plus Apple local-overlay skip. The live integration-smoke failure is the known unrelated _docker_cmd_arr issue. Approving for admin squash merge.

PR #1035, opened 2026-04-24
fix(openclaw): trigger open-webui recreate on install; simplify volume layout

yasinBursali:
## Summary
- Host agent now recreates `open-webui` after a successful `openclaw` install so the overlay's `OPENAI_API_BASE_URLS` + `OPENAI_API_KEYS` env actually reaches the running container (previously silently broken until a manual `dream restart`).
- Drops the `openclaw-home` named Docker volume entirely. Verified against `ghcr.io/openclaw/openclaw:2026.3.8` — openclaw only writes two paths under `/home/node/.openclaw/`: `openclaw.json` (regenerated by inject-token.js every start) and `workspace/` (bind-mounted). The named volume held only regenerated content.
- Phase 06 directory-setup comment updated; stale `./data/openclaw/home` row removed from `openclaw/README.md`.
- Follow-up commit (`e9e992fa`) drops three stale `--no-deps` references from explainer comments — PR #1021 dropped that flag from the install codepath, so the comments now describe the present-day install command (`up -d <ext>` without `--force-recreate`) accurately.

## Platform Impact
- **macOS / Linux / Windows-WSL2**: identical. Python host-agent + Docker Compose mount semantics are platform-agnostic. No platform branches introduced.

## Testing
- `pytest tests/test_host_agent.py` → **45 passed** on the rebased branch (40 pre-existing including #1021's `TestInstallStartCommandNoDeps` that came in via rebase + 5 new covering the `_post_install_core_recreate` helper).
- `docker compose -f docker-compose.base.yml -f extensions/services/searxng/compose.yaml -f extensions/services/openclaw/compose.yaml config` merges cleanly; `grep -c openclaw-home` in merged output → 0.
- `make lint` clean.
- Manual (any platform): enable openclaw via dashboard → `docker exec dream-webui env | grep OPENAI_API_BASE_URLS` shows openclaw URL without a full stack restart; `docker inspect dream-openclaw` shows 3 bind mounts and no named volume.

## Existing-install migration
The old `openclaw-home` Docker volume becomes orphaned after upgrade. Contents are non-critical (regenerated `openclaw.json` + misc caches). Users can reclaim disk with `docker volume rm openclaw-home` post-upgrade. No data loss.

## Review notes
- `_post_install_core_recreate` is hardcoded to `openclaw` (helper exists next to `docker_compose_recreate`). Kept scope-tight rather than introducing a manifest-level `post_install_requires_recreate` field for a single use case.
[...]

PR #1053, opened 2026-04-26
ci(openclaw): filesystem-write gate to detect new openclaw write paths

yasinBursali:
> **DRAFT: must merge AFTER #1035.** The workflow's mechanism is independent of named-volume vs bind-mount layout (it overrides volume layout in `docker run`), so this PR can run before #1035 lands — but the gate's *purpose* (catching upstream additions to openclaw's persistent write paths) only makes semantic sense once #1035's simplified layout is the production baseline. Promote to ready-for-review after #1035 merges; this PR has no file overlap so the rebase is trivial.

## What
Add `.github/workflows/openclaw-image-diff.yml`. When a PR touches openclaw's compose.yaml or manifest.yaml, the workflow runs the openclaw image with a host bind-mount over `/home/node/.openclaw`, exercises the entrypoint briefly, then walks the bind-mount directory and fails if any file other than the expected `openclaw.json` shows up.

## Why
Without a runtime gate, a future openclaw image bump that introduces a NEW persistent state path (e.g. `/home/node/.openclaw/sessions/`, `/home/node/.openclaw/cache.db`) would silently land in the container's ephemeral overlay and be lost on every recreate, with no CI signal until users complain.

**Why not `docker diff` (as the original issue proposed):** `docker diff` reports changes to the container's union-FS writable layer only. Writes to mounted volumes — named OR bind — are invisible to it. The originally-proposed `docker diff openclaw-test | grep "/home/node/\.openclaw/"` would have always passed (zero matches), making it a false-green CI gate. Verified during review.

## How
Pivoted detection mechanism: the workflow creates its OWN bind-mount over `/home/node/.openclaw` inside the container (overriding whatever's in compose.yaml), runs the production entrypoint briefly, then walks the host directory.

```yaml
on:
  pull_request:
    paths:
      - 'dream-server/extensions/services/openclaw/compose.yaml'
      - 'dream-server/extensions/services/openclaw/manifest.yaml'
      - '.github/workflows/openclaw-image-diff.yml'

permissions:
  contents: read

jobs:
  filesystem-write-gate:
    runs-on: ubuntu-latest
    timeout-minutes: 10
    env:
      OPENCLAW_IMAGE: ghcr.io/openclaw/openclaw:2026.3.8
```

Steps:
1. **Pull image** — `docker pull "$OPENCLAW_IMAGE"`
2. **Prepare bind-mount layout** — host directories for home / workspace / config; stub `inject-token.js` (the real script only patches Control UI HTML — irrelevant to the filesystem audit)
3. **Start openclaw with bind-mounted home** — `docker run -d` mirroring the production entrypoint (`chown -R 1000:1000` + `setpriv` + `docker-entrypoint.sh node openclaw.mjs gateway --allow-unconfigured --bind lan`); workspace mounted separately so it doesn't pollute the audit
4. **Inspect bind-mount for unexpected writes** — `find ./openclaw-test-home -mindepth 1 -maxdepth 5 -path './openclaw-test-home/workspace' -prune -o -type f ! -name 'openclaw.json' -print`; non-empty result → `::error::` annotation + `exit 1`. Plus a positive `[ -f openclaw.json ]` check that emits `::warning::` if openclaw.json was never written (catches silent crash-before-write false-greens)

Image tag matches `extensions/services/openclaw/compose.yaml`. Dummy `OPENCLAW_TOKEN=ci-test-token` satisfies the `:?` env guard. `--rm` is deliberately omitted with explicit `docker rm -f` cleanup so a startup crash still leaves logs available in the workflow output.

## Testing
- YAML parse + workflow schema check (name / on / jobs / permissions / runs-on / timeout-minutes / paths-filter present): PASS
- `bash -n` on each `run: |` block (4/4): PASS
- Image tag matches compose.yaml: `ghcr.io/openclaw/openclaw:2026.3.8` ✓
- `actions/checkout@v6` confirmed real (v6.0.2 released 2026-01-09; project already uses both `@v6` and SHA-pinned forms)
- Trigger scope verified: only `pull_request` + paths-filter, no `push` / `schedule` / `workflow_dispatch`
- Secrets scan: only `ci-test-token` dummy literals
- `actionlint` not installed locally; GitHub will gate on PR
- Cannot smoke-test locally (would need `act` or push-to-fork); manual test plan below for real CI run

Manual (operator on real CI run, post-merge):
1. Push a comment-only change to `extensions/services/openclaw/compose.yaml` from a fork branch. Confirm the new workflow fires in the Actions tab.
2. Verify the green tick (no unexpected writes detected by the gate).
3. **Adversarial test:** in a throwaway branch, modify the workflow's `find` allowlist temporarily so it expects `openclaw.json` PLUS some fake file. Push and confirm the gate fails with the expected `::error::` annotation: `openclaw is writing to new paths under /home/node/.openclaw/. Update compose.yaml volume layout to include these paths or update this test's allowlist.`

## Review
Critique Guardian APPROVED (post-polish-fix re-review). Both polish recommendations from the first CG pass — `OPENCLAW_CONFIG_PATH` divergence comment and positive `openclaw.json` assertion — were applied before commit.

## Known Considerations
- The workflow deliberately omits `OPENCLAW_CONFIG_PATH=/tmp/openclaw-config.json` (which production compose.yaml exports) so the audit observes any default in-tree config writes openclaw might make. A 4-line `# NOTE:` in the workflow source flags this for future maintainers.
- Gate is semantically meaningful only after #1035 sets the simplified production volume layout; pre-#1035 false greens are theoretically possible (the named volume swallows the writes the gate wants to audit). DRAFT label + must-merge-after #1035 is the safeguard.

## Platform Impact
| Platform | Impact |
|---|---|
| macOS / Linux / Windows (WSL2) | The workflow runs on `ubuntu-latest` only. CI gate; doesn't affect any production runtime path. The bug it would catch (openclaw writing to ephemeral overlay) affects all 3 platforms equally — one Linux gate is sufficient. |


yasinBursali:
**MUST MERGE AFTER #1035.** This workflow's gate is semantically meaningful only after #1035's simplified openclaw volume layout is the production baseline. Mechanically the workflow can run before #1035 (it overrides volume layout in `docker run`), but pre-#1035 false-greens are theoretically possible because the named volume swallows the writes the gate wants to audit. No file overlap — rebase is trivial.

Lightheartdevs:
Audit follow-up: CI gate needs to fail on missing expected write.

This fixes the original `docker diff` trap, but the positive assertion only warns when `openclaw.json` is missing. For a write-observation CI gate, a container that exits before doing meaningful work should fail the job, not pass with an empty bind mount. Please make the missing expected write a hard failure.

yasinBursali:
Pushed follow-up commit `edc5cf0c` per the audit ask: positive `openclaw.json` assertion converted from warning-only to hard failure (`::error::` + `exit 1`). The unexpected-files arm (line 117) was already a hard fail; the missing-baseline-write arm is now symmetric.

Comment block expanded to explain the write-observation invariant, so a future maintainer doesn't soften this back to a warning without understanding why.

Two non-blocking observations worth flagging:

1. **30s startup is now a stricter contract.** With warning-only, an image bump that increased startup time would silently false-green; with hard-fail, it'll fail visibly. This is the correct trade-off (false-fails > false-greens for a write-observation gate), but if a future openclaw release adds startup work and pushes past 30s, this contract may need to be widened (or converted to a poll loop).
2. **No `actionlint` locally.** Local validation covered YAML parse + `bash -n` on all 4 run blocks. GHA-specific syntax (action versions, expression syntax) is gated by CI on push.

YAML parses cleanly, bash -n clean across all 4 run blocks. Still DRAFT pending #1035.

Lightheartdevs (approved):
Approval contingent on Actions actually running the gate green.

Code review: new `.github/workflows/openclaw-image-diff.yml` workflow. Mechanism is sound — comment correctly notes that `docker diff` cannot see writes into volumes (which is what original issue #498 missed), so this approach instead bind-mounts host directory over `/home/node/.openclaw/` and walks it after running. Asserts only `openclaw.json` shows up. Triggers on changes to openclaw compose/manifest or this workflow itself. Static review confirms the previous false-green gap is fixed: missing `openclaw.json` now `::error::` exits 1, and unexpected files exit 1.

**Companion to #1035** (the openclaw home-dir-removed volume layout). The gate is only fully meaningful after #1035 (or its successor) sets the production layout, but the workflow can run before then because it overrides the volume layout in `docker run`.

**Required before merge:** GitHub-hosted runner must be able to pull/start the openclaw test image. Let Actions prove the workflow.

Lightheartdevs (changes requested):
The false-green issue is fixed now, but the workflow still is not mergeable because its own OpenClaw write-gate job fails for the workflow harness, not for a meaningful OpenClaw write-path finding.

The failing job logs show the container never gets to the baseline write:

```text
chown: changing ownership of '/home/node/.openclaw/workspace': Operation not permitted
chown: changing ownership of '/home/node/.openclaw': Operation not permitted
setpriv: setgroups failed: Operation not permitted
::error::openclaw.json was never written
```

That means this would add a permanently red required/visible workflow for any PR touching OpenClaw paths. Please make the harness runnable on GitHub-hosted Docker first: avoid chowning the nested workspace mount if the runner/userns forbids it, or run the container with a user/group strategy that does not need `setpriv --clear-groups` on the mounted host dirs. Once the workflow can actually observe `openclaw.json` on the current image, this gate becomes valuable.

Lightheartdevs (approved):
Approved after repairing the Yasin branch and re-auditing the result.

What changed during the repair:
- the write-gate now starts OpenClaw through the production config/inject-token path instead of a stubbed config
- the gate fails on crashed/incomplete startup and normalizes bind-mount ownership before inspection
- the gate uncovered real current OpenClaw writes under agents/, canvas/, and cron/, so the branch now persists those state paths explicitly in compose/installer/docs instead of pretending openclaw.json is the only write
- session-cleanup now defaults to the host path that Token Spy/systemd already expect

Verification:
- local workflow YAML parse + bash syntax for all workflow run blocks
- bash -n for changed shell scripts
- docker compose config --quiet for base + OpenClaw overlay
- dream-server/tests/test-session-cleanup.sh: 11 passed
- GitHub Actions: full PR suite green, including Detect openclaw writes outside known paths

PR #1054, opened 2026-04-26
fix(dashboard-api): require deployable compose.yaml to mark extension installable

yasinBursali:
## What
Tighten `_is_installable` in `routers/extensions.py` so an extension is only advertised as installable when it actually has a deployable `compose.yaml` on disk.

## Why
The library currently ships three entries that have only `compose.yaml.disabled` or `compose.yaml.reference` files:

- `dify`, `jan` — only `compose.yaml.disabled`
- `fooocus` — only `.reference` files

The predicate did `(library_dir / ext_id).is_dir()`, so the dashboard rendered Install buttons for them. Clicking Install copied the directory into `user-extensions/`, but the host agent could not start a service because no compose was deployable. Users saw cryptic errors with no clear cause.

`aider` keeps a real `compose.yaml` with `entrypoint: ["echo"]` and `restart: "no"` — that's intentional CLI-tool surface and remains installable, matching existing UX.

## How
Single-function change. New body checks both directory presence AND a deployable `compose.yaml`:

```python
def _is_installable(ext_id: str) -> bool:
    ext_dir = EXTENSIONS_LIBRARY_DIR / ext_id
    return ext_dir.is_dir() and (ext_dir / "compose.yaml").exists()
```

No schema change, no manifest field added, no frontend change. Catalog response shape is unchanged — only the `installable` boolean flips for the three affected entries. Both call sites consume the bool directly; no signature change.

## Testing
- `python3 -m py_compile` PASS.
- `ruff check` PASS.
- `pytest tests/` (dashboard-api) — 137/137 PASS, no new failures.
- Sanity-checked all 33 catalog entries: 3 correctly hidden (dify, jan, fooocus); 30 still installable, including aider.
- Live runtime verification on macOS Apple Silicon after rebuild: `/api/extensions/catalog` returns `installable: false` for dify/jan/fooocus and `installable: true` for aider/milvus/anythingllm. Extensions page UI shows no Install button for the three blocked entries.

## Review
None required from external reviewers — change is mechanical and observable. Local CG (Critique Guardian) review approved cleanly: five-pillar pass, no warnings.

## Known considerations
The predicate hardcodes the filename `compose.yaml`. Today every catalog entry's `compose_file` field is either `"compose.yaml"` or `""`, so this matches reality. A future extension that ships a different filename would need its own treatment — worth keeping in mind, not load-bearing here.

## Platform impact
- **macOS**: Active. UI no longer offers Install buttons for the three broken extensions on Apple Silicon (where `gpu_backends` already filtered some of them; this is a second layer of defense). Verified live.
- **Linux**: Same code path. dify/jan are GPU-compatible on AMD/NVIDIA, so this filter is the primary mechanism that hides their Install buttons there.
- **Windows/WSL2**: Same. Pure Python `pathlib`; no platform-specific behavior.


Lightheartdevs:
Audit follow-up: catalog/UI fix is incomplete without the server-side install gate.

This correctly marks non-deployable library entries as not installable in the catalog/UI, but a direct `POST /api/extensions/{id}/install` still calls `_install_from_library(service_id)` without requiring a deployable `compose.yaml`. Please reject missing `compose.yaml` in `_install_from_library` or before calling it, and add an API regression for `dify`/`jan`/`fooocus`-style entries.

yasinBursali:
Pushed follow-up commit `408c2e56` per the audit ask: server-side install gate in `_install_from_library`. The catalog/UI predicate (`_is_installable`) and the install endpoint now use the same truth function — direct `POST /api/extensions/{id}/install` for non-deployable entries (.disabled-only or .reference-only) returns 400 instead of copytree-without-effect.

Two new HTTP-level regression tests cover the dify/jan shape (`.disabled` only) and the fooocus shape (`.reference` only). Each asserts status 400 + "compose.yaml" in detail + nothing copied to user-extensions/.

`pytest tests/test_extensions.py` → 139/139 (was 137 + 2). `test_templates.py` → 19/19 (sibling caller already gated by `_is_installable`). ruff clean.

PR #1055, opened 2026-04-26
docs(dashboard-api): add development workflow guide for the bake-vs-bind-mount trap

yasinBursali:
## What
Document how to iterate on the dashboard-api FastAPI backend without your changes silently no-op'ing — and explain why the obvious bind-mount and \`--reload\` overlay alternatives weren't shipped.

Two files:
- New: \`dream-server/docs/DASHBOARD-API-DEVELOPMENT.md\` — full guide (~140 lines).
- Updated: \`dream-server/CONTRIBUTING.md\` — short pointer to the new doc, slipped into the existing Getting Started link list and a brief inline summary.

## Why
The dashboard-api Dockerfile bakes the Python source into \`/app/\` at image build, and uvicorn imports from there. The compose service mounts \`./extensions:/dream-server/extensions:ro\` — but that mount exists for manifest and config discovery, not for Python imports. Editing files under \`extensions/services/dashboard-api/\` on the host therefore does **not** reload the running container; changes silently no-op until the image is rebuilt. Multiple contributors have hit this trap (and lost time tracking down why their change "did nothing"), so the gap is worth a focused doc rather than another retread.

## How
Pure documentation. No code, Dockerfile, compose, or installer change.

The guide:
1. States the trap up front (TL;DR + a "The Trap" section).
2. Recommends running uvicorn natively on the host with \`--reload\`, leveraging the existing \`host.docker.internal:host-gateway\` wiring on the dashboard-api compose service so the rest of the stack continues to talk to the host-side process.
3. Documents \`docker cp\` as a stop-gap until the next image rebuild.
4. Explains why **Option B** (bind-mount the source as the working directory) was not shipped — \`__pycache__\` writes, standalone-image breakage, dual prod/dev image divergence.
5. Explains why **Option C** (uvicorn \`--reload\` via a dev compose overlay) was not shipped — uvicorn launched from \`WORKDIR /app\` re-imports baked code regardless of \`--reload-dir\`; making it actually work would also require overriding \`WORKDIR\` (or wrapping the command), at which point you're effectively re-implementing Option B with extra steps.
6. Cross-platform notes — macOS / Linux / WSL2 (with the \`/mnt/c/...\` inotify caveat).

The doc references structural patterns (\`WORKDIR /app\`, \`COPY ... ./\`, \`CMD uvicorn ...\`) rather than specific line numbers, so it doesn't rot when the Dockerfile shifts.

## Testing
- markdownlint not installed locally, skipped.
- Internal links resolve (verified \`docs/EXTENSIONS.md\`, \`docs/HOST-AGENT-API.md\` exist).
- Technical accuracy cross-checked against the actual \`Dockerfile\` and \`docker-compose.base.yml\` — \`WORKDIR\`, COPY commands, CMD shape, \`extra_hosts\`, volumes block all match.
- No emojis, all code blocks language-tagged, single H1, no TODO/FIXME.
- \`cut -d= -f2-\` in the env-extraction example uses the range form (not \`-f2\`) so values containing \`=\` aren't truncated.
- Both \`DASHBOARD_API_KEY\` and \`DREAM_AGENT_KEY\` shown in the example so a copy-paster running native uvicorn doesn't immediately hit 401 from the host agent (since #979, the two keys are independent).

## Review
None required from external reviewers — observable on read. Local CG approved with the warnings above already addressed before submission.

## Known considerations
\`CONTRIBUTING.md\` is currently rebased frequently (open PR #973 may overwrite the link list and prose). If that lands first, this PR will need a small rebase touching only the link-list section and the new "Dashboard API development" subsection — no semantic conflict.

## Platform impact
- **macOS**: Recommended workflow (native uvicorn) works without caveats. Documentation is platform-aware.
- **Linux**: Same. inotify is reliable on the WSL2 filesystem.
- **Windows/WSL2**: Documentation includes the standard caveat to keep the repo on the WSL2 filesystem (\`~/...\`) rather than \`/mnt/c/...\` for uvicorn's watchfiles to behave.

### DIFF
--- a/dream-server/CONTRIBUTING.md
+++ b/dream-server/CONTRIBUTING.md
@@ -15,6 +15,26 @@ No CLA. No committee. No waiting for permission. If it makes Dream Server better
 If you're adding or extending services, read these first:
 - [docs/EXTENSIONS.md](docs/EXTENSIONS.md) — how to add a new service in 30 minutes
 - [docs/INSTALLER-ARCHITECTURE.md](docs/INSTALLER-ARCHITECTURE.md) — how the installer works under the hood
+- [docs/DASHBOARD-API-DEVELOPMENT.md](docs/DASHBOARD-API-DEVELOPMENT.md) — how to actually iterate on the dashboard-api FastAPI backend without your changes silently no-op'ing
+
+## Dashboard API development
+
+If you're touching `extensions/services/dashboard-api/`, read [docs/DASHBOARD-API-DEVELOPMENT.md](docs/DASHBOARD-API-DEVELOPMENT.md) **before** you start editing.
+
+The short version: the dashboard-api `Dockerfile` copies the Python source into `/app/` at image build, and uvicorn imports from there. The compose service mounts `./extensions:/dream-server/extensions:ro` — but that bind-mount is for manifest and config discovery, not Python imports. Editing files under `extensions/services/dashboard-api/` on the host does **not** reload the running container. Your changes silently do nothing until the image is rebuilt.
+
+The recommended workflow is to run uvicorn natively on the host with `--reload`:
+
+```bash
+cd dream-server/extensions/services/dashboard-api
+pip install -r requirements.txt
+DREAM_INSTALL_DIR=/path/to/dream-server \
+  uvicorn main:app --host 127.0.0.1 --port 3002 --reload
+```
+
+Other services already reach the host via `host.docker.internal:host-gateway`, which is wired into the dashboard-api compose service, so the rest of the stack keeps working while you iterate. Hot-reload works natively on macOS and Linux; on WSL2, keep the repo on the WSL2 filesystem (not `/mnt/c/...`) for the watcher to behave.
+
+If you can't run native uvicorn for some reason, `docker cp <file> dream-dashboard-api:/app/<file>` followed by `docker compose restart dashboard-api` is a survivable stop-gap until the next image rebuild. The full guide covers why we did **not** ship a bind-mount overlay or a `uvicorn --reload` compose mode as the default.
 
 ## What We Care About Right Now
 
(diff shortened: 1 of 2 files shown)


Lightheartdevs:
Audit follow-up: docs workflow currently breaks the normal dashboard proxy.

The guide correctly identifies the baked `/app` trap, but it tells contributors to stop the `dashboard-api` container and then says the dashboard container will reach host uvicorn via `host.docker.internal`. In current compose, dashboard nginx still proxies `/api` to `http://dashboard-api:3002`; stopping that container breaks the normal UI unless the contributor also runs Vite or changes the proxy. Please update the workflow to match the actual proxy path.

yasinBursali:
Pushed audit follow-up (`3c906fbe` + `0ce7f49f`).

The audit was right: the doc told contributors to stop only `dashboard-api` and assume the Docker dashboard nginx would reach a host uvicorn via `host.docker.internal`. That path is structurally impossible — `dashboard/nginx.conf:21` hardcodes `proxy_pass http://dashboard-api:3002`, and the dashboard nginx container doesn't have `extra_hosts: host.docker.internal:host-gateway` (only the dashboard-api container does, in `docker-compose.base.yml:151-152`). Stopping dashboard-api with the dashboard nginx still in front means every `/api/*` request 502s.

**Corrected workflow:** stop **both** the `dashboard` and `dashboard-api` containers (frees host ports 3001 and 3002), then run Vite dev server on 3001 + native uvicorn on 3002. Vite's existing proxy at `vite.config.js:9-13` routes `/api` to `localhost:3002`, completing the chain:

```
browser → http://localhost:3001 (Vite dev) → /api/* → http://localhost:3002 (host uvicorn)
```

Other compose services (llama-server, host-agent, etc.) keep running. After dev session, `docker compose start dashboard dashboard-api` restores production state.

CG returned APPROVED WITH WARNINGS. One substantive item folded in (`0ce7f49f`):
- `DREAM_AGENT_HOST=127.0.0.1` added to the Terminal 2 env block. Without it, `config.py:301` defaults `AGENT_HOST=host.docker.internal` which doesn't resolve on the host; any host-agent-backed dashboard-api route would silently 502.

The doc now also explicitly calls out the prior advice as wrong (`"Stopping only dashboard-api (an earlier draft of this guide's recommendation) is wrong"`) so the audit is reflected in the fix rather than papered over.

CG verified all line/file citations in the doc against the actual code: `nginx.conf:21`, `docker-compose.base.yml:154/219`, `vite.config.js:6-15`, `Dockerfile:9/21/40`. All match.

PR #1056, opened 2026-04-26
fix(dashboard-api): catalog timeout, orphaned whitelist, GPU passthrough scan, health_port

yasinBursali:
## What
Four scoped fixes to the dashboard-api extension paths, plus three new unit tests covering the new compose-scan rule.

- **#A — Catalog/detail health-fan-out timeout:** add a separate short timeout (5 s) for the catalog and \`extension_detail\` user-extension health probes, distinct from the existing 30 s \`_HEALTH_TIMEOUT\` used elsewhere.
- **#B — Orphaned-storage whitelist:** add \`extension-progress\` and \`config-backups\` to the system-dirs set used by \`/api/storage/orphaned\`.
- **#C — GPU passthrough scan:** reject \`deploy.resources.reservations.devices\` (Compose v2 GPU syntax) in the user-extension compose security scanner, gated so built-ins can still legitimately request it.
- **#D — User-extension \`health_port\`:** read the manifest \`health_port\` field in \`scan_user_extension_services\`, mirroring the built-in \`config.py\` pattern.

## Why
**A.** \`_HEALTH_TIMEOUT\` is a session-level 30 s default. The catalog endpoint fan-outs every user-extension health probe with that timeout. The frontend \`Extensions.jsx\` aborts \`/api/extensions/catalog\` after 8 s. A single slow user-extension health endpoint pushes the whole catalog response past the frontend abort, the page renders "Failed to load extensions catalog — Retry", every extension card vanishes, and ongoing polling drops out. \`extension_detail\` has the same fan-out shape and the same DoS surface.

**B.** \`/api/storage/orphaned\` whitelists \`{models, config, user-extensions, extensions-library}\`. \`extension-progress\` is created at runtime by \`_write_initial_progress\` / \`_write_progress\`; \`config-backups\` is created by the host agent's \`.env\` backup writer. Any system with install history shows them as "orphaned data that can be removed". An operator following the suggestion deletes in-flight progress state.

**C.** \`_scan_compose_content\` blocks the short-form \`devices:\` directive but does not walk \`deploy.resources.reservations.devices\` — the Compose v2 standard for GPU passthrough. A user extension can request NVIDIA GPU access without explicit operator grant. Built-ins (e.g. \`docker-compose.nvidia.yml\`) legitimately use this construct, so the rejection must be gated by the existing \`is_builtin\` check rather than blanket.

**D.** \`config.py:136\` reads \`health_port\` for built-in extensions but \`user_extensions.py:scan_user_extension_services\` does not — it builds the probe URL from the primary \`port\` only. Extensions like \`milvus\` declare \`port: 19530\` (gRPC primary) and \`health_port: 9091\` (HTTP \`/healthz\`); the dashboard-api currently probes \`http://milvus:19530/healthz\` (HTTP request against a gRPC port) and reports them perpetually unhealthy.

### A — short catalog/detail timeout
- Add \`_CATALOG_HEALTH_TIMEOUT = aiohttp.ClientTimeout(total=5)\` in \`helpers.py\`.
- Add an optional \`timeout=\` keyword to \`check_service_health\`. Forward it to \`session.get(url, ...)\` only when set, so existing callers keep the 30 s session default.
- The \`extensions_catalog\` and \`extension_detail\` user-extension fan-outs now pass \`_CATALOG_HEALTH_TIMEOUT\`.

### B — orphaned whitelist
- Add \`extension-progress\` and \`config-backups\` to the \`system_dirs\` set in \`orphaned_storage\`. (Both directories audited as the only runtime-created directories under \`DATA_DIR\` — other entries are JSON files or lock files.)

### C — \`deploy.resources.reservations.devices\` rejection
- Add a \`skip_gpu_passthrough_check: bool = False\` parameter to \`_scan_compose_content\`, mirroring the existing \`skip_name_collision\`.
- New gate inside the per-service loop walks \`svc_def.get("deploy") -> resources -> reservations -> devices\` and raises HTTPException(400) when present.
- All three existing call sites of \`_scan_compose_content\` audited and updated:
  1. \`_install_from_library\` (line ~947, library install) — \`trusted=True\` only; \`skip_gpu_passthrough_check\` STAYS DEFAULT FALSE because the staged source is a community/user extension, not a built-in. Library extensions must NOT request GPU passthrough.
  2. \`_activate_service\` (line ~1112) — passes \`skip_name_collision=is_builtin, skip_gpu_passthrough_check=is_builtin\`.
  3. \`enable_extension\` re-scan (line ~1161) — same pair.

### D — \`health_port\` mirror
- \`user_extensions.py:scan_user_extension_services\` now propagates the manifest \`health_port\` into the service config when present, exactly mirroring the \`config.py:136\` pattern (\`**({"health_port": int(svc["health_port"])} if "health_port" in svc else {})\`). \`helpers.py:check_service_health\` already falls back to \`port\` when \`health_port\` is absent, so this is a forward-compatible add.

## Testing
- \`make lint\` PASS (shell + Python compile).
- \`make test\` PASS.
- \`ruff check\` PASS.
- \`pytest tests/test_extensions.py\` — 140/140 PASS (was 137; added 3 new tests):
  - \`TestComposeScanEdgeCases.test_scan_rejects_deploy_resources_devices\` — install-path negative.
  - \`TestScanComposeSkipGpuPassthroughCheck.test_rejects_deploy_devices_by_default\` — direct-unit negative.
  - \`TestScanComposeSkipGpuPassthroughCheck.test_allows_deploy_devices_when_skipped\` — direct-unit positive (locks in built-in legitimacy).
- Live runtime verification on macOS Apple Silicon after image rebuild + container recreate:
  - \`/api/storage/orphaned\` returns only legitimate orphans; \`extension-progress\` and \`config-backups\` confirmed absent.
  - \`_CATALOG_HEALTH_TIMEOUT\` and \`skip_gpu_passthrough_check\` symbols verified live in the running container.
  - Extensions page renders 70 cards, 0 error banners, 0 console errors.

## Review
Local CG approved with two warnings, both addressed before submission (test for #C added; \`extension_detail\` timeout added to match \`extensions_catalog\` for symmetry).

## Known considerations
- \`extension_detail\` previously deferred timeout was absorbed into this PR per CG feedback — same DoS shape, one-line change, makes the per-detail-view path safe too.
- \`deploy: {resources: null}\` (where \`reservations\` would be \`None\`) would AttributeError on the new \`.get(...)\` chain. Crash-not-bypass; Docker Compose itself rejects that YAML shape upstream of us. Defended by the existing \`isinstance(deploy, dict)\` guard before the chain. Left a single-isinstance hardening for a future PR if anyone files a real reproduction.

## Platform impact
- **macOS**: All changes active. Server-side Python in a Docker container — platform-neutral. Live-verified.
- **Linux**: Same. Built-in NVIDIA GPU passthrough still works correctly through the gated path (most exposed platform for #C).
- **Windows/WSL2**: Same. No POSIX-only assumptions in the changed paths.


Lightheartdevs:
Audit follow-up: scanner hardening needs one more malformed-shape guard.

The GPU scanner improvement is directionally right, but it still assumes `deploy.resources` is a dict before calling `.get('reservations')`. A malformed user extension with `deploy: { resources: null }` can 500 instead of returning a clean scanner error. Please type-check `resources` before dereferencing and add a malformed-compose regression.

yasinBursali:
Pushed two follow-up commits per the audit + a CG re-review of my work:

- `9e274066` — isinstance-guards each level of `deploy → resources → reservations` so a malformed `deploy: { resources: null }` returns a clean scanner pass-through instead of 500. The chained `.get(key, {})` was vulnerable because `dict.get(key, default)` returns `None` when the key is present-with-null, breaking the next `.get()` call.
- `7e90bae4` — clarifies the regression-vs-defense-in-depth framing in the new tests' docstrings (per CG note: only `test_handles_null_resources_without_500` is a strict regression for the maintainer-flagged shape; the other two are defensive lock-ins).

`pytest tests/test_extensions.py` → 143/143 pass. ruff clean.

PR #1057, opened 2026-04-26
fix(host-agent): runtime hygiene — narrow pull, surface failures, normalize bind volumes

yasinBursali:
## What
Seven small surgical edits in `dream-server/bin/dream-host-agent.py` that harden the host agent's install and model-validation paths against silent failures, misleading error surfacing, and a couple of long-form Compose volume cases the pre-create logic was missing. Plus an audit follow-up making the install pull narrowing dependency-aware.

## Why
The host agent is the dispatcher between the dashboard-api and Docker. Most of its current failure paths either truncate diagnostics (head-truncation of stderr drops the actual error in favour of the layer-pull preamble), swallow OSError silently (frozen progress bars, no journal record), or conflate distinct error classes (HTTP 403 vs 500). When something goes wrong, operators see a misleading symptom and have to read the journal — often to find that nothing was logged because the failure was swallowed.

Concretely:

1. **Narrow `docker compose pull`.** Currently runs with the FULL merged compose stack, so a single `${VAR:?}` guard in any unrelated extension's compose aborts the pull for the target extension with an error message naming the wrong extension.
2. **`stderr[:N]` head-truncation at three sites in `_handle_install`.** Docker-compose puts the real error at the END of stderr — head-truncation drops exactly the diagnostic bytes the operator needs.
3. **`_write_model_status` silent OSError swallow.** Disk-full or permission errors freeze the dashboard progress bar with no journal trace.
4. **`_recreate_llama_server` silent docker-run failure.** Logs but does not raise; caller waits the full 5-minute health-check timeout before rolling back. Also reachable on Linux native via the `_compose_restart_llama_server` no-`.compose-flags` fallback (not just Windows/in-container).
5. **HTTP 403/500 conflation in model-library validation.** A corrupt or missing `config/model-library.json` returns the same 403 as a genuinely-denied model. The same silent-except exists in the parallel `_handle_model_list` handler.
6. **`_precreate_data_dirs` misses long-form Compose volumes.** `str(vol).split(":")[0]` works for the string short-form `./data/foo:/target` but stringifies the dict long-form `{type: bind, source: ./foo, target: /bar}` into a Python repr and misses the bind source. Extensions using long-form volumes hit the "Docker auto-creates as root:root" failure mode.
7. **`_precreate_data_dirs` doesn't skip non-pre-expanded sources.** `~`, `$VAR`, backtick, backslash get fed to the relative-bind logic and produce literal garbage directories on disk.

## How
Single file: `dream-server/bin/dream-host-agent.py`.

1. `_handle_install._run_install` derives a narrow `pull_flags` list by filtering out `-f` entries pointing at OTHER extensions' compose fragments under `EXTENSIONS_DIR` / `USER_EXTENSIONS_DIR`. The pull step uses `pull_flags`; the `up` step keeps full `flags` so cross-service `depends_on` still resolves. Conservative: when a path can't be resolved it stays in the list (better to over-pull than to silently drop a needed file).
2. Three `stderr[:N]` truncations flipped to `[-N:]` in `_handle_install`.
3. `_write_model_status`: `except OSError: pass` -> `except OSError as e: logger.warning(...)`. Behavior preserved (does not raise) so the activate flow continues, but the failure now lands in the journal.
4. `_recreate_llama_server`: `raise RuntimeError(f"docker run failed: {result.stderr[-500:]}")` after the existing `logger.error`. The outer try at the activate-flow level catches and surfaces a real error within seconds instead of the 5-minute hang. This single change covers both call sites — the in-container path (Windows/WSL2) and the no-`.compose-flags` fallback path (Linux native).
5. `_handle_model_download` adds a `catalog_ok` sentinel and returns HTTP 500 (`"Model catalog unavailable"`) when the library file is missing or unreadable. HTTP 403 (`"Model not in library catalog"`) is now reserved strictly for the genuine policy-denial path. The same `logger.exception` + 500 pattern is also applied to `_handle_model_list`'s parallel silent except.
6 + 7. `_precreate_data_dirs` handles dict-form Compose volumes (`type: bind` only; extracts `source`) and skips sources starting with `~`, `$`, backtick, or backslash before the existing relative-bind filter.

## Audit follow-up (commits 488f5b48 + ff20aadf)

Maintainer review on the original commit flagged that the narrowed pull set drops dependency services some extensions reference across extension boundaries. Empirically reproduced: `extensions/services/perplexica/compose.yaml` and `extensions/services/openclaw/compose.yaml` both declare `depends_on: searxng`. Filtering out `extensions/services/searxng/compose.yaml` makes `docker compose pull perplexica` fail at config-parse time with `service "perplexica" depends on undefined service "searxng": invalid compose project`.

Fix:

- Promote `_is_other_ext_compose` from inline closure to module-level helper (signature: `fpath, service_id, ext_roots`).
- Factor the filter loop into `_narrow_install_pull_flags(flags, service_id) -> list`.
- Add `_narrowed_compose_set_resolves(narrowed_flags, service_id, cwd, timeout) -> bool` which runs `docker compose [narrowed] config --services` and returns True only if exit 0 AND the target service is present in stdout.
- `_handle_install._run_install` attempts narrowing first; if `narrowed != flags` and the validator passes, use narrowed flags. Otherwise fall back to the full flag set and log at INFO. Validator timeout is 30s (matches `resolve_compose_flags`'s subprocess), avoids waiting 600s on a hung Docker daemon.

This implements the second of the two options the maintainer offered (validate via `config --services`) rather than building a dependency-aware filter.

## Testing
- `make lint` PASS (shell syntax + Python compile).
- `make test` PASS (9/9).
- `python3 -m py_compile` + AST parse PASS.
- `pytest tests/test_host_agent.py` PASS (49/49 — including 8 new tests across `TestNarrowInstallPullFlags`, `TestNarrowedComposeSetResolves`, `TestInstallPullFallsBackOnUnresolvedNarrow`).
- Full dashboard-api `pytest tests/`: 663/675 PASS; the 12 pre-existing failures in `test_workflows.py`, `test_gpu_detailed.py`, and `test_main.py::test_gpu_cached_falsy` are Python 3.14 asyncio shim issues unrelated to this change.
- Empirical bug repro: `docker compose -f perplexica.yaml config --services` exits 1 with `depends on undefined service "searxng"`; full set with searxng included exits 0.
- Live end-to-end on macOS Apple Silicon (original commit): `dream stop` then `install.sh --non-interactive` against an existing install. Exit 0, all 6 phases complete, all 11 core services come up healthy.

## Review
Local CG **APPROVED** the audit follow-up:

> The fix correctly addresses the maintainer audit. The validate-and-fallback approach matches one of the two options the maintainer explicitly offered, the empirical reproduction is solid, the seven original surgical edits are intact, and all 49 tests pass.

CG flagged one latent edge case (non-blocking, matches upstream/main behaviour): when a `${VAR:?}`-broken unrelated extension AND a cross-ext `depends_on` co-exist, the fall-back hits the broken `:?` and pull still fails. Filed as a follow-up consideration; not addressed in this PR since the maintainer explicitly chose validate-and-fallback over dependency-aware filtering.

## Known considerations
**Behavior change.** `_handle_model_download` previously returned HTTP 403 "Model not in library catalog" for both genuine policy denials AND for missing/unreadable catalog files. Clients that relied on 403 to detect "no catalog yet" must update — they will now see HTTP 500 with `"Model catalog unavailable"` in that case. The 403 remains for the legitimate "model exists in catalog but not allowed" path.

**Asymmetry between `_handle_model_list` and `_handle_model_download` on missing catalog.** `_handle_model_list` returns HTTP 200 + empty library (browse semantics — a fresh install legitimately has no catalog yet, and the dashboard splash should still render). `_handle_model_download` returns HTTP 500 (validate-this-specific-model semantics — catalog absence breaks the security check; lying with 403 would conflate a broken install with a policy denial). Inline comment in the diff explains the choice.

## Platform impact
- **macOS** (native llama-server, Docker for the rest): All edits active except `_recreate_llama_server` itself (Apple Silicon uses `_launch_native_llama_server`); the raise still applies if the in-container path is ever triggered. Manual reinstall verified.
- **Linux Docker**: All edits active. `_recreate_llama_server` raise covers the previously-undocumented Linux native fallback path (when `.compose-flags` is missing) — operator sees a real error within seconds instead of the 5-minute hang.
- **Windows/WSL2**: All edits active. `_recreate_llama_server` is the primary activate path here and was the most user-visible offender for the silent-fail bug.


Lightheartdevs:
Audit follow-up: pull narrowing needs dependency awareness.

Narrowing `docker compose pull` to avoid unrelated `${VAR:?}` failures is the right instinct, but removing compose files for other extensions can also remove required dependency services that `up` still expects, such as a web app depending on `searxng`. Please either make the filter dependency-aware or validate the selected service set with `docker compose config --services` before pull.

yasinBursali:
Pushed audit follow-up (`488f5b48` + `ff20aadf`).

Empirically reproduced the dependency-loss case: `extensions/services/perplexica/compose.yaml` and `extensions/services/openclaw/compose.yaml` both declare `depends_on: searxng`. Narrowing without validation made `docker compose pull perplexica` fail at config-parse time with `depends on undefined service "searxng"`.

Implemented option 2 from the audit (validate via `config --services` rather than dependency-aware filter):

- `_is_other_ext_compose` promoted from inline closure to module-level helper.
- New `_narrow_install_pull_flags(flags, service_id) -> list` for the filter.
- New `_narrowed_compose_set_resolves(narrowed, service_id, cwd, timeout) -> bool` runs `docker compose [narrowed] config --services` and returns True only if exit 0 AND target service is in stdout.
- `_handle_install._run_install` falls back to the full flag set whenever the narrowed view fails to resolve; logs at INFO. Validator timeout is 30s (matches `resolve_compose_flags`).

8 new tests across `TestNarrowInstallPullFlags`, `TestNarrowedComposeSetResolves`, `TestInstallPullFallsBackOnUnresolvedNarrow`. 49/49 host-agent tests pass; full dashboard-api suite green except the 12 pre-existing Python 3.14 asyncio shim failures unrelated to this change.

Local CG approved with one non-blocking observation: when a `${VAR:?}`-broken unrelated extension AND a cross-ext `depends_on` co-exist, the fallback hits the broken `:?` and pull still fails. This matches upstream/main behaviour today — not a regression introduced by this PR. A truly dependency-aware filter would handle it; the chosen `config --services` validation cannot.

Lightheartdevs:
Audit pass (2026-05-02): direction is right (narrow pull to enabled service set, surface failed starts, normalize bind volumes). Codex's static review confirms the narrowed pull validates with `docker compose ... config --services` and falls back to full flag set when dependencies are missing. Cannot review further while CONFLICTING — please rebase. Will re-review after.

Lightheartdevs (dismissed):
Re-audited after the follow-up commits. The original dependency-filter concern is addressed: the narrowed pull set is validated with docker compose config --services and falls back to full flags when a cross-extension depends_on would be dropped. Focused host-agent tests pass locally (49/49), py_compile passes, Ruff passes for the touched Python files, and the live integration-smoke failure is the known unrelated _docker_cmd_arr expectation mismatch. Approving for admin squash merge.

Lightheartdevs:
Substantive re-audit passed and I approved this, but GitHub cannot merge it into current `main` now. A local merge simulation reports conflicts in:

```text
dream-server/bin/dream-host-agent.py
dream-server/extensions/services/dashboard-api/tests/test_host_agent.py
```

Please rebase/merge current `main` and preserve the narrowed-pull validator + fallback behavior. Once the conflicts are resolved, this should be close to merge-ready; the focused local checks I ran before the conflict check were `test_host_agent.py` 49/49, `py_compile` for `dream-host-agent.py`, and Ruff on the touched Python files.

Lightheartdevs (approved):
Re-audited after merging current `main` into Yasin's branch. The host-agent runtime-hygiene code from this PR is already present on current `main`; the rescued branch now contributes the missing narrowed-pull regression tests only.

Local verification:
- `python -m pytest extensions/services/dashboard-api/tests/test_host_agent.py -q -k "NarrowInstallPullFlags or NarrowedComposeSetResolves or InstallPullFallsBackOnUnresolvedNarrow"` -> 9 passed
- `python -m py_compile bin/dream-host-agent.py`
- `python -m ruff check bin/dream-host-agent.py extensions/services/dashboard-api/tests/test_host_agent.py`
- `git diff --check`

Full local `test_host_agent.py` remains 72 passed / 4 failed on this Windows workstation due the known local executable-bit and symlink privilege harness issues in sync_config tests; CI is green.

PR #1070, opened 2026-05-01
fix(windows-amd): add readiness sidecar gating open-webui on native inference (Lemonade or llama-server.exe)

yasinBursali:
## What
Add a busybox readiness sidecar (`llama-server-ready`) for the Windows AMD install path that polls the native inference server's health endpoint and gates `open-webui` on it becoming healthy. Delivered as a NEW overlay file `installers/windows/docker-compose.windows-amd.local.yml` loaded conditionally by `install-windows.ps1` ONLY in the AMD non-cloud branch.

## Why
Windows AMD installs zero out the `llama-server` container (Lemonade runs natively on the Windows host) but had no readiness gate. On cold start, `open-webui` came up immediately while Lemonade was still compiling its first-boot binary (~3-5 min) — users saw "connection refused" / blank model dropdown for the entire first-compile window.

The macOS overlay solves the equivalent problem for native llama-server with a `llama-server-ready` busybox sidecar that polls `host.docker.internal:${OLLAMA_PORT}/health` and gates `open-webui.depends_on`. This PR mirrors that pattern for Windows AMD with three Windows-specific adaptations:

1. **Dual-backend probe**: Lemonade exposes `/api/v1/health` (per `installers/windows/lib/constants.ps1:36`); the llama-server.exe fallback exposes `/health` (per `installers/windows/lib/llm-endpoint.ps1:124`). The SAME `docker-compose.windows-amd.yml` overlay is used for both backends. So the sidecar uses an OR-probe (`wget /api/v1/health || wget /health`), avoiding any need for env-var-driven endpoint configuration or duplicating the centralized endpoint-resolution logic from #987.

2. **Generous timing**: `retries: 90 × interval: 10s = 15min` healthy-window headroom for Lemonade's first-boot compile, consistent with `docker-compose.amd.yml`'s `start_period: 300s` (#878 design).

3. **Separate overlay file, conditional inclusion**: cloud mode reuses the SAME `docker-compose.windows-amd.yml` overlay (Lemonade isn't running, but the windows-amd.yml file's `replicas: 0` for llama-server is still applicable). The sidecar must NOT load in cloud mode (would block open-webui forever waiting on a non-existent native server). Solution: a separate `docker-compose.windows-amd.local.yml` file, loaded via an additional `-f` flag in `install-windows.ps1`'s `elseif ($gpuInfo.Backend -eq "amd")` branch only. Since cloud mode is the FIRST `if ($cloudMode)` branch in the if/elseif chain, AMD non-cloud is reached only when cloud is false — implicit conditional, no new env var.

This naming/loading pattern mirrors the maintainer's PR #913 (`compose.amd.yaml` overlay for AMD-specific OpenClaw routing) — separate file, conditional via the resolver/installer, not a runtime gate.

### New overlay file: `dream-server/installers/windows/docker-compose.windows-amd.local.yml` (47 lines)
- `llama-server-ready`: busybox:1.36.1 (pinned tag), `host.docker.internal:host-gateway`, OR-probe entrypoint, healthcheck same OR-probe, `retries: 90`, minimal `cpus: 0.1` / `memory: 32M`, `security_opt: no-new-privileges:true`.
- `open-webui.depends_on.llama-server-ready.condition: service_healthy` — additive merge into base's open-webui.

### `dream-server/installers/windows/install-windows.ps1` (+6 lines)
Inside the `elseif ($gpuInfo.Backend -eq "amd")` branch (line ~494), after the existing `windows-amd.yml -f` flag, append the new `windows-amd.local.yml -f` flag with a 4-line comment explaining cloud-mode exclusion.

## Backward compatibility
- Existing Windows AMD installs that re-run the installer will get the new overlay automatically (installer regenerates compose flags on every run).
- `OLLAMA_PORT` defaults to `8080` (matching `installers/windows/lib/constants.ps1` and the existing `windows-amd.yml`'s hardcoded `host.docker.internal:8080`).
- Cloud mode and other backends (NVIDIA, CPU fallback) are unaffected — their if/elseif branches don't add the new overlay.

### Mechanical
- YAML parse: PASS
- `make lint`: PASS (shell + Python compile)
- `docker compose config --quiet` with full Windows AMD non-cloud stack (base + windows-amd.yml + windows-amd.local.yml): exit 0; merged graph confirms `llama-server-ready` defined and `open-webui.depends_on.llama-server-ready.condition: service_healthy` resolves cleanly.
- `docker compose config --quiet` with cloud-mode stack (base + windows-amd.yml only): exit 0; `llama-server-ready` correctly ABSENT.
- pre-commit (gitleaks, private-key, large-file): PASS
- ASCII grep on .ps1 added lines: clean (no em-dash, no smart quotes per maintainer commit `e4d0ca44`).
- ASCII grep on new YAML: clean.

### CI gating
- `lint-powershell.yml` will lint the .ps1 change (no path filter; runs on every PR).
- `validate-compose.yml` does NOT cover `installers/**/*compose*.yml` paths — this is a pre-existing CI gap affecting both the existing `windows-amd.yml` and this new file. Manual `docker compose config` validation closes the gap for this PR; a follow-up issue will track extending the CI path filter.

### Manual (operator on a Windows AMD host)
1. **Lemonade primary**: fresh install → accept Lemonade → `dream-llama-ready` becomes healthy after 3-5 min compile → open-webui starts cleanly without "connection refused" period.
2. **llama-server.exe fallback**: same install but decline Lemonade → sidecar OR-probe hits `/health` → open-webui still gated correctly.
3. **Cloud mode**: install with cloud → `dream-llama-ready` does NOT appear in `docker compose ps`; open-webui starts independently.
4. **Other platforms** (macOS/Linux/Windows-NVIDIA): unaffected — different overlay paths.

## Review
Independent verification confirmed the bug and surfaced the dual-backend / cloud-mode complications. Implementation follows the maintainer's separate-overlay-file pattern from #913, generous-start_period stance from #878, and avoids duplicating the centralized endpoint resolution from #987. Holistic review approved on first pass.

## Known Considerations / Optional follow-ups (not in this PR)
- **Latent OLLAMA_PORT mismatch (cross-platform issue)**: if a user customizes `OLLAMA_PORT` in `.env`, the sidecar honors it via `${OLLAMA_PORT:-8080}` interpolation, BUT the native servers (Lemonade hardcoded 8080, macOS native llama-server) may not. The macOS sidecar shares this latent gap. Will be tracked as a separate fork issue.
- **CI path filter gap on `installers/**/*compose*.yml`**: pre-existing — the existing `installers/windows/docker-compose.windows-amd.yml` was already not validated by CI. Will be tracked as a separate fork issue.

## Platform Impact
- **Windows AMD non-cloud (Lemonade or llama-server.exe)**: bug fixed.
- **Windows AMD cloud mode**: no impact (sidecar correctly excluded).
- **Windows NVIDIA / Windows CPU fallback**: no impact (different overlays).
- **macOS / Linux NVIDIA / Linux AMD / Linux CPU**: no impact (don't load `docker-compose.windows-amd.yml`).


Lightheartdevs (approved):
Conditional approval — needs Windows AMD smoke before merge.

Code review: new `docker-compose.windows-amd.local.yml` overlay adds a busybox readiness sidecar (`dream-llama-ready`) that probes both `/api/v1/health` (Lemonade) and `/health` (llama-server.exe) so the same overlay works for both backends without env-var configuration. `extra_hosts: host-gateway` is correct for container-to-host. Codex confirmed compose validates with placeholder env, and the installer only adds the local overlay for non-cloud AMD (avoiding obvious cloud-mode deadlock).

**Required before merge:** one Windows AMD install smoke with Lemonade or native `llama-server.exe` to verify the readiness gating actually fires before `open-webui` tries to connect.

Lightheartdevs (changes requested):
This is a good AMD UX target, but the current overlay probes the wrong port for the Windows AMD local-inference contract.

`docker-compose.windows-amd.local.yml` uses `${OLLAMA_PORT:-8080}` in the readiness sidecar. The Windows env generator still writes `OLLAMA_PORT=11434`, while the existing Windows AMD overlay points dashboard-api and Open WebUI at `host.docker.internal:8080` for native Lemonade / llama-server.exe. A direct compose render with `.env.example` shows the mismatch:

```text
llama-server-ready -> http://host.docker.internal:11434/api/v1/health
open-webui        -> http://host.docker.internal:8080/v1
```

That can make `open-webui` wait forever on `llama-server-ready` even though the native backend is healthy on 8080. Please key the sidecar off the same native endpoint contract as the Windows AMD overlay, or introduce a dedicated native-inference port variable and update the overlay/env generator consistently. This one is AMD-facing, so I�d like a compose-render regression check before merge.

Lightheartdevs (approved):
Re-audited after carrying Yasin's branch onto current `main` and fixing the requested-change item.

The previous blocker was real: the readiness sidecar inherited `OLLAMA_PORT=11434`, while the Windows AMD native-inference contract uses `host.docker.internal:8080`. I changed the sidecar entrypoint and healthcheck to probe port 8080 directly, matching `docker-compose.windows-amd.yml`, and added a compose-render contract that sets `OLLAMA_PORT=11434` and fails if that port leaks into the rendered Windows AMD local overlay.

Local checks run:
- `bash -n dream-server/tests/contracts/test-windows-amd-local-compose.sh dream-server/tests/contracts/test-installer-contracts.sh`
- `bash dream-server/tests/contracts/test-windows-amd-local-compose.sh`
- `docker compose --env-file dream-server/.env.example -f dream-server/docker-compose.base.yml -f dream-server/installers/windows/docker-compose.windows-amd.yml -f dream-server/installers/windows/docker-compose.windows-amd.local.yml config --quiet`
- PowerShell parser on `dream-server/installers/windows/install-windows.ps1`

GitHub checks are green on the repaired head, including Docker Compose validation.

PR #1071, opened 2026-05-01
fix(installer): NVIDIA Container Toolkit keyring fail-loud + apt-only

yasinBursali:
## What
Move the NVIDIA Container Toolkit GPG-key download into the apt branch and replace the silent `|| true` / `2>/dev/null` pair with a fail-loud empty-file guard.

## Why
`installers/phases/05-docker.sh` ran `curl … | gpg --dearmor … 2>/dev/null || true` *before* the `case "$PKG_MANAGER"` block. The unconditional silencers swallowed pipe failures and ran a pointless curl on pacman/zypper/dnf. On apt, a silent failure left the keyring missing, and the next `apt-get update` died with a confusing `NO_PUBKEY` error nobody could trace back. Project rule (CLAUDE.md) explicitly forbids `|| true` / `2>/dev/null`.

## How
- The `curl | gpg` pipe now lives inside the `apt)` branch (the only consumer of `signed-by=…/nvidia-container-toolkit-keyring.gpg`).
- Both silencers removed; pipefail (inherited from `install-core.sh`) catches the pipe-side failure.
- Defensive `[[ ! -s … ]]` guard calls the existing `error()` helper to abort with a specific message if the file lands empty.

## Testing
**Automated**: \`bash -n\` PASS · \`make lint\` PASS · \`make simulate\` PASS · secrets-scan clean. shellcheck not installed locally — CI \`lint-shell.yml\` is the authoritative gate.

**Manual**: apt happy-path → keyring present + signed-by reference intact; apt failure-path (block \`nvidia.github.io\`) → install aborts with the new error message; dnf/pacman/zypper → no spurious \`curl: Failed writing body\` log line and install proceeds normally.

## Review
Critique Guardian: APPROVED (5-pillar pass). One latent edge case noted but pre-existing: \`gpg --dearmor -o <existing-file>\` refuses to overwrite on recent gpg versions, so a re-install after a previous broken run may need manual cleanup of the empty file. Not introduced by this fix.

## Known Considerations
None.

## Platform Impact
- **macOS Apple Silicon**: not affected — `installers/phases/05-docker.sh` is sourced only by the Linux installer; macOS uses `install-macos.sh` and has no NVIDIA Container Toolkit path.
- **Linux**: apt + WSL2-Ubuntu paths now fail loudly with actionable error; dnf / pacman / zypper paths now skip the unnecessary curl entirely.
- **Windows/WSL2**: same as Linux apt when WSL2 distro is Ubuntu/Debian.


Lightheartdevs (approved):
Conditional approval — needs Linux NVIDIA install smoke before merge.

Code review: removes the silent `|| true` mask on the NVIDIA keyring download, then explicitly checks the keyring file is non-empty before proceeding to add the apt repo. Errors out if the download was empty. Direct CLAUDE.md alignment — no more "silently broken apt repo gives mysterious install failures." The fix is correctly scoped to the apt branch only (dnf/pacman branches unchanged).

**Required before merge:** install on one apt-based NVIDIA host. The PR title says "keyring fail-loud" — please don't let a future title revision imply this removes dnf/pacman support, since those branches still exist.

Lightheartdevs (changes requested):
Thanks for tightening this path. I found one reinstall correctness issue that should be fixed before merge.

The new apt branch still runs:

```bash
sudo gpg --dearmor -o /usr/share/keyrings/nvidia-container-toolkit-keyring.gpg
```

without overwrite handling. On a rerun where the keyring already exists, `gpg --dearmor -o existing-file` exits non-zero in non-interactive mode before the new empty-file guard can run. I reproduced this locally with the NVIDIA key: the first dearmor succeeds, the second dearmor to the same output path exits `2` with `gpg: cannot open 'no tty'`, and the pipe then fails. In the installer, that means an ordinary rerun/reinstall can fail even when the existing keyring is healthy.

Please make the keyring write rerun-safe, e.g. `gpg --batch --yes --dearmor -o ...`, or remove/replace the target atomically before writing, and keep the empty-file guard after the write. #1119 seems to be pointed at this same edge, so this may just need that fix folded in or stacked cleanly.

yasinBursali:
Pushed the rerun-safety fix. Commit `c979f9ff` adds `--batch --yes` to the `gpg --dearmor` line:

```diff
 curl -fsSL --max-time 60 https://nvidia.github.io/libnvidia-container/gpgkey | \
-    sudo gpg --dearmor -o /usr/share/keyrings/nvidia-container-toolkit-keyring.gpg
+    sudo gpg --batch --yes --dearmor -o /usr/share/keyrings/nvidia-container-toolkit-keyring.gpg
```

Reproduced the failure locally with the actual NVIDIA key:
- bare `--dearmor`: first invocation exit 0, second invocation exit 2 with `gpg: cannot open '/dev/tty': Device not configured` (matches your repro)
- with `--batch --yes`: both invocations exit 0, idempotent overwrite — empty-file guard from this PR still meaningful

Aware of #1119 — same edge, narrower fix (just `--yes`). `--batch --yes` is a strict superset, so this PR is self-sufficient and idempotent with #1119 if either order lands. Happy to drop one as a duplicate post-merge — your call which.

Fingerprint pin from the earlier commit (`d3e8431e`) is unaffected by the change; trust anchor remains the published NVIDIA fingerprint, not the HTTPS chain alone.

Branch rebased onto current `main` (`d4124d6a`); 3 commits ahead.

Ready for fresh review.

Lightheartdevs (approved):
Re-audited after carrying Yasin's branch onto current `main`.

The requested-change item is fixed: the apt keyring write is now rerun-safe with `gpg --batch --yes --dearmor`, and the empty-file guard remains after the write. I also verified the current NVIDIA Container Toolkit gpgkey still extracts to the pinned fingerprint `C95B321B61E88C1809C4F759DDCAE044F796ECB0`, and repeated dearmor into the same output path succeeds with the new flags.

Local checks run:
- `bash -n dream-server/installers/phases/05-docker.sh dream-server/tests/test-network-timeouts.sh`
- live NVIDIA gpgkey fingerprint extraction and repeated `gpg --batch --yes --dearmor` overwrite smoke

I also ran `bash dream-server/tests/test-network-timeouts.sh`; the NVIDIA key/repo timeout checks passed, but the suite currently fails on `main` for an unrelated `installers/phases/11-services.sh` GGUF retry-message pattern, so I did not treat that as a #1071 blocker. GitHub checks are green on the repaired head.

PR #1072, opened 2026-05-01
fix(installer): pre-create token-spy data dir + macOS extension parity

yasinBursali:
## What
Pre-create `data/token-spy` on Linux (with chown) and `data/{dreamforge,ape,token-spy}` on macOS so Docker doesn't lazy-create those dirs as root and crash uid-1000 containers.

## Why
Live-confirmed failure on a fresh Linux install: `dream-token-spy` crash-looped with `PermissionError: [Errno 13] Permission denied: '/app/data/token-spy-api-key.txt'` because Phase 06 didn't pre-create the host bind-mount and Docker created it as uid 0. `token-spy/Dockerfile` bakes uid 1000. macOS Docker-Desktop virtiofs hides the issue but Podman / OrbStack-without-virtiofs share the bug, and the macOS list was missing three extension dirs that exist in the Linux list.

## How
- `installers/phases/06-directories.sh:42` — add `token-spy` to the data-dirs brace expansion.
- `installers/phases/06-directories.sh:175` — `chown -R 1000:1000 data/token-spy || warn "…"` (sibling to the openclaw chown). The `|| warn` follows the documented project rule for tolerated failures so a non-uid-1000 installer still completes; the warn message names the consequence.
- `installers/macos/install-macos.sh:430-432` — add `mkdir -p` for `dreamforge`, `ape`, `token-spy`. No chown (virtiofs).

## Testing
**Automated**: bash -n PASS · make lint PASS · tests/contracts/test-installer-contracts.sh PASS (incl. token-spy contract) · secrets-scan clean.

**Manual**: Linux fresh install → `stat $INSTALL_DIR/data/token-spy` shows uid 1000 before compose-up; container starts without PermissionError. Re-install → idempotent. macOS fresh install → `ls -la data/` shows all three new dirs.

## Review
Critique Guardian: APPROVED. One coherence note: the legacy openclaw chown one block earlier still uses `2>/dev/null || true` (the banned anti-pattern); we apply the new `|| warn` rule to new code only. Worth a follow-up sweep PR; not a blocker for this fix.

## Known Considerations
- Chown line is unconditional; if installer runs as a user whose uid ≠ 1000 (rare; documented install path is `sudo bash install.sh`), chown fails and the warn fires, install completes, container may crash at runtime. Documented in the warn message.
- Embeddings / ape-uid concerns surfaced during verification (different uid scheme) are out of scope and tracked separately.

## Platform Impact
- **Linux**: bug fixed (live-verified failure mode).
- **macOS Apple Silicon**: parity restored; virtiofs continues to hide the bug on Docker Desktop but pre-creation matters for Podman / OrbStack-without-virtiofs.
- **Windows/WSL2**: install path uses Linux scripts inside WSL2 — Linux fix applies.


Lightheartdevs (approved):
Sorry — missed this one in my first Phase B sweep.

Two changes: (1) Linux installer's directory-create line adds `token-spy` to the brace expansion (matches existing dreamforge/ape pattern); (2) macOS installer adds explicit `mkdir -p` for `token-spy`, `dreamforge`, `ape` — bringing macOS to parity with the Linux directory set. Plus a `chown -R 1000:1000 data/token-spy` since the container UID is baked into the Dockerfile. Non-fatal warn pattern correctly used (matches openclaw chown above it).

Pre-creating with the right UID avoids first-start crashes from container-vs-host UID mismatch. `bash -n` clean per Codex. Ship after rebase + Linux/macOS installer smoke.

Lightheartdevs (approved):
Re-audited for merge. The change only pre-creates token-spy data on Linux and brings macOS directory creation into parity for token-spy/dreamforge/ape. Local bash -n passed for both touched installer scripts and the expected directory/chown lines are present. Approving for squash merge.

PR #1067, opened 2026-04-30
fix(compose): drop CPU overlay AUDIO_STT_MODEL literal; make AMD memory limit env-driven

yasinBursali:
## What
Two related compose-overlay drift fixes:
- `dream-server/docker-compose.cpu.yml`: drop the now-empty `open-webui:` block (was overriding `AUDIO_STT_MODEL` to turbo)
- `dream-server/docker-compose.amd.yml`: replace literal `memory: 110G` with `${LLAMA_SERVER_MEMORY_LIMIT:-110G}`

## Why
**CPU overlay**: PR #985 made `AUDIO_STT_MODEL` env-driven (`${AUDIO_STT_MODEL:-Systran/faster-whisper-base}` in `docker-compose.base.yml`). The cpu overlay's residual literal `AUDIO_STT_MODEL: "deepdml/faster-whisper-large-v3-turbo-ct2"` overrode that interpolation, so on CPU-only Linux installs:
1. Phase 06 of the installer correctly writes `AUDIO_STT_MODEL=Systran/faster-whisper-base` to `.env`
2. Phase 12 pre-downloads only the base model
3. Open WebUI's container env is forced to turbo by the overlay literal
4. Whisper 404s on every transcription

The NVIDIA overlay's identical literal is **intentional** — for NVIDIA, Phase 06 also picks turbo, so the literal is a redundant safety net (per #985's design). Only the cpu overlay drifted.

**AMD overlay**: 5 of 6 GPU overlays use `${LLAMA_SERVER_MEMORY_LIMIT:-N}` (NVIDIA 64G, Apple 32G, Intel/Arc 24G, CPU 6G). `.env.schema.json` registers the variable as a tunable string. The amd overlay's literal `memory: 110G` made AMD the only platform where users couldn't tune llama-server's memory cap.

`tier0.yml`'s literal `memory: 4G` is a deliberate hard cap for <8 GB RAM hosts and was left alone.

## How
- cpu.yml: removed the entire `open-webui:` block (3 + leading-blank lines). Container inherits the base's env-var interpolation.
- amd.yml: 1-line substitution. The 110G default is preserved as the AMD-specific fallback so existing AMD installs without the env var keep current byte-for-byte behavior.

## Testing
- YAML parse, `make lint`, `docker compose config --quiet`, pre-commit: all PASS
- Functional substitution checks:
  - cpu env unset → resolves `Systran/faster-whisper-base`
  - cpu env override → propagates correctly
[...]

PR #1073, opened 2026-05-01
fix(nvidia-overlay): allow .env override of AUDIO_STT_MODEL default

yasinBursali:
## What
One-line change to `docker-compose.nvidia.yml`: replace the hardcoded `AUDIO_STT_MODEL` literal with `\${AUDIO_STT_MODEL:-deepdml/faster-whisper-large-v3-turbo-ct2}` so users can override via `.env`.

## Why
Compose merging replaces overlay env keys at the key level. The overlay's bare literal silently overrode both `docker-compose.base.yml`'s `\${AUDIO_STT_MODEL:-Systran/faster-whisper-base}` and any user `.env` value. NVIDIA users couldn't switch Whisper models without editing the overlay file directly. AMD overlay correctly omits the key already.

## How
\`\`\`diff
-      AUDIO_STT_MODEL: "deepdml/faster-whisper-large-v3-turbo-ct2"
+      AUDIO_STT_MODEL: "\${AUDIO_STT_MODEL:-deepdml/faster-whisper-large-v3-turbo-ct2}"
\`\`\`
Default unchanged when `.env` is unset. `.env.schema.json` already registers `AUDIO_STT_MODEL`, so no schema update needed. This completes the env-var introduction work that originally added `AUDIO_STT_MODEL` to base.yml + env-generators but missed the NVIDIA overlay.

## Testing
**Automated**: `docker compose -f base -f nvidia config` validated for both unset (default `deepdml/...`) and override (`AUDIO_STT_MODEL=Systran/faster-whisper-base` resolves correctly). AMD overlay regression-guarded (still no `AUDIO_STT_MODEL` reference). make lint PASS · contract tests PASS.

**Manual**: `docker exec dream-webui env | grep AUDIO_STT_MODEL` should reflect `.env` value.

## Review
Critique Guardian: APPROVED. Same `\${VAR:-default}` shape as base.yml. Round-trip preservation through `.env` write path intact.

## Known Considerations
The CPU overlay (`docker-compose.cpu.yml`) has the identical bug and is tracked as a separate issue — intentionally not bundled here to keep scope tight.

## Platform Impact
- **macOS Apple Silicon**: not affected — uses native llama-server, no compose overlay.
- **Linux NVIDIA / Windows-WSL2-NVIDIA**: bug fixed.
- **AMD / CPU**: not affected (different overlays; CPU has its own tracker).

### DIFF
--- a/dream-server/docker-compose.nvidia.yml
+++ b/dream-server/docker-compose.nvidia.yml
@@ -28,5 +28,5 @@ services:
 
   open-webui:
     environment:
-      AUDIO_STT_MODEL: "deepdml/faster-whisper-large-v3-turbo-ct2"
+      AUDIO_STT_MODEL: "${AUDIO_STT_MODEL:-deepdml/faster-whisper-large-v3-turbo-ct2}"
       ENABLE_IMAGE_GENERATION: "${ENABLE_IMAGE_GENERATION:-true}"


Lightheartdevs (approved):
Single-line: NVIDIA `AUDIO_STT_MODEL` becomes `${AUDIO_STT_MODEL:-deepdml/...}` so the strong default is preserved while `.env` overrides work. Symmetric with PR #1067's CPU overlay cleanup. Ship after rebase.

Lightheartdevs (approved):
Re-audited for merge. The NVIDIA overlay keeps the turbo default but now respects AUDIO_STT_MODEL overrides. Local NVIDIA compose config passed for both default and AUDIO_STT_MODEL=custom/model cases. Approving for squash merge.

PR #1078, opened 2026-05-01
fix(scripts): BSD-compat for memory-shepherd, llm-cold-storage, migrate-config

yasinBursali:
## What
Three macOS shell-portability bugs in standalone diagnostic/maintenance scripts (`memory-shepherd.sh`, `llm-cold-storage.sh`, `migrate-config.sh`), plus a dormant bug where every config migration was silently skipped on all platforms.

## Why
DreamServer ships these scripts to all three platforms (macOS BSD, Linux GNU, Windows-WSL2). With `set -uo pipefail` (no `-e`) the GNU-only flag failures substitute empty strings into arithmetic comparisons and size guards — silent misbehavior on macOS. The migrate-config root cause was unrelated: `MIGRATIONS_DIR` was pointed at the wrong directory, so `ls migrate-v*.sh` always exit-failed and `sort -V` was never even reached.

## How
- `memory-shepherd.sh`: add `_stat_mtime`/`_stat_size` helpers (Darwin `stat -f %m|%z` vs GNU `stat -c %Y|%s`); replace 6 call sites
- `llm-cold-storage.sh:66`: `find -printf '%A@\n' | sort -rn | head -1` → `find -type f -exec stat ({+}) | sort -rn | sed -n '1p'` (SIGPIPE-safe)
- `llm-cold-storage.sh:56`: `pgrep -af` → `pgrep -f` (`-a` only adds discarded output formatting)
- `migrate-config.sh:22`: `MIGRATIONS_DIR="${SCRIPT_DIR}"` → `${SCRIPT_DIR}/../migrations`
- `migrate-config.sh:227`: `sort -V` → `sort`

## Testing
- New `tests/bats-tests/macos-bsd-compat.bats` (5/5 pass on macOS Apple Silicon) — PATH-stubbed `uname`+`stat` verify per-platform helper dispatch; static grep guards against re-introducing GNU-only constructs
- `make lint` clean; `make test` clean (115 tests, 0 failures)
- Negative test: reverting `sort -V` removal causes BATS test 5 to fail as expected
- Live macOS sanity: `_stat_mtime /etc/hosts` returns `1763819368`, `_stat_size /etc/hosts` returns `213`

## Known Considerations
Future migrations beyond `migrate-v0.9.x` would lex-sort incorrectly (`v0.10.0` before `v0.2.0`). Plain `sort` is correct for the current single migration; if multi-digit components arrive, switch to portable `sort -t. -k1,1n -k2,2n -k3,3n` or zero-pad components.

## Platform Impact
- **macOS** — primary target; three latent bugs become functional
- **Linux** — no behavior change (GNU branch was already correct)
- **Windows/WSL2** — no behavior change (uses Linux/GNU branch inside WSL2)

### DIFF
--- a/dream-server/memory-shepherd/memory-shepherd.sh
+++ b/dream-server/memory-shepherd/memory-shepherd.sh
@@ -3,6 +3,23 @@
 # Usage: memory-shepherd.sh [agent-name|all]
 set -uo pipefail
 
+# Cross-platform stat helpers — BSD (Darwin) / GNU diverge on -c vs -f
+_stat_mtime() {
+    if [[ "$(uname -s)" == "Darwin" ]]; then
+        stat -f %m "$1"
+    else
+        stat -c %Y "$1"
+    fi
+}
+
+_stat_size() {
+    if [[ "$(uname -s)" == "Darwin" ]]; then
+        stat -f %z "$1"
+    else
+        stat -c %s "$1"
+    fi
+}
+
 TIMESTAMP=$(date '+%Y-%m-%d_%H%M')
 LOCKFILE=/tmp/memory-shepherd.lock
 SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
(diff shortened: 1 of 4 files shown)


Lightheartdevs (approved):
Conditional approval — needs macOS BATS run before merge.

Code review: adds `_stat_mtime` and `_stat_size` helpers that branch on `uname -s` for BSD vs GNU `stat` flag differences (`-f` vs `-c`). Standard macOS-portability pattern. Also touches `llm-cold-storage` and `migrate-config` per Codex's note. Static review and `bash -n` clean per Codex.

**Required before merge:** run the new BATS file in CI or via `tests/run-bats.sh` on macOS — the local environment here is Git Bash on Windows, which doesn't exercise the Darwin branch.

Lightheartdevs (approved):
Re-audited against current main. The BSD/GNU portability fixes are narrowly scoped to the three affected scripts, syntax checks pass under Git Bash for the touched shell files, and the existing migrate-config test failure reproduces on baseline too. The live integration-smoke failure is the known unrelated _docker_cmd_arr expectation mismatch, while the rest of the PR checks are green. Approving for admin squash merge.

PR #1079, opened 2026-05-01
fix(installer): clean compose orphans during install

yasinBursali:
## What
Two macOS-installer UX bugs:
1. Re-install leaves orphan containers from previously-enabled extensions running. `docker compose up` lacks `--remove-orphans` in the macOS installer (`install-macos.sh:871`) and Linux phase 11 (`phases/11-services.sh:345/364`).
2. The macOS opencode LaunchAgent plist runs `opencode web` which auto-opens a browser to `localhost:3003`, hijacking the post-install UX before the dashboard URL is even shown. `opencode serve` is the headless variant.

## Why
- `--remove-orphans` is already used on `down` calls (`dream-update.sh`, `dream-cli`, `dream-uninstall.sh`); the gap is on `up`. Symmetric fix.
- `opencode web` per opencode docs starts the server **and opens a browser**. There's no `--no-browser` flag. `opencode serve` skips the auto-open.

## How
- `install-macos.sh:871`: `up -d` → `up -d --remove-orphans`
- `phases/11-services.sh:345/364`: same flag on both Linux up calls
- `install-macos.sh:971`: plist `<string>web</string>` → `<string>serve</string>`
- `install-macos.sh:1010`: manual-start hint updated
- `install-macos.sh:994/996`: log filename `opencode-web.log` → `opencode-serve.log` (no external readers — confirmed via repo-wide grep)

## Testing
- `bash -n` + `shellcheck -x` clean on both files
- `plutil -lint` confirms plist remains valid; ProgramArguments contains `serve`, log paths use new filename
- `tests/contracts/test-installer-contracts.sh` 10/10 pass; `tests/contracts/test-plist-log-paths.sh` 9/9 pass (filename-agnostic)
- `make lint` + `make test` green

**Manual smoke** (recommended verification):
- macOS: re-install with prior `dream-aider` container present → assert it's gone post-install
- macOS: fresh install → assert no browser tab opens to `localhost:3003` during phase 5
- Linux: re-run `phases/11-services.sh` against an install with stale containers → assert removed

## Known Considerations
- Linux's `opencode/opencode-web.service:9` uses the same `opencode web` subcommand. Tracked as a follow-up — kept out of this PR for minimal/focused scope.
- Windows installer (`install-windows.ps1:610`) has the same orphan gap. Follow-up.
- 8 `dream-cli` `up -d` invocations also lack `--remove-orphans`. Different scope, separate PR.

## Platform Impact
- **macOS**: both fixes apply
- **Linux**: orphan fix in phase 11
- **Windows/WSL2**: not in this PR (deferred follow-up)


Lightheartdevs (approved):
Conditional approval — needs install/update smoke before merge.

Code review: two changes. (1) Adds `--remove-orphans` to `docker compose up -d` on macOS install — correctly addresses stale containers from previous extension sets. (2) Changes the launchd plist's opencode invocation from `web` to `serve` — same fix as #1097 for systemd; `web` auto-opens a browser which is wrong on a headless install. Both are right.

**Note on `--remove-orphans`:** moderate-risk because it affects compose project state across runs. If the user has manually-orchestrated containers with the same project name, those would be removed. Acceptable risk for the macOS installer's controlled scope.

**Required before merge:** one install + one update smoke confirming the expected extension containers are still present after the orphan cleanup.

Lightheartdevs (changes requested):
﻿The orphan cleanup half looks useful, but I do not think the OpenCode LaunchAgent change is merge-safe as written.

DreamServer advertises this service as the OpenCode web IDE on port 3003 (`docs/MACOS-QUICKSTART.md`, `extensions/services/opencode/README.md`, phase 12 health, phase 13 summary). The current OpenCode docs distinguish the commands this way:

- `opencode web` starts the web interface and opens a browser.
- `opencode serve` starts a headless HTTP/API server.

So changing the macOS LaunchAgent from `opencode web --port 3003 --hostname 127.0.0.1` to `opencode serve --port 3003 --hostname 127.0.0.1` likely fixes the unwanted browser auto-open by removing the web UI itself. That makes the advertised `http://localhost:3003` OpenCode IDE link questionable/broken rather than merely headless.

Please split this or revise the OpenCode half: keep the `--remove-orphans` installer cleanup if desired, but solve the macOS auto-open problem without changing the service from web UI to API-only, or update the product/docs/health checks if the intended product is now headless API instead of the browser IDE.

Local checks: `bash -n` passed for `installers/macos/install-macos.sh` and `installers/phases/11-services.sh`. The live `integration-smoke` red is the shared `_docker_cmd_arr` expectation mismatch and appears unrelated to this PR.

yasinBursali:
Pushed two changes addressing the review:

1. **Windows `--remove-orphans` parity** — added inline at `install-windows.ps1:610` so native Windows installs match the macOS/Linux behavior this PR introduced. The `$composeFlags` array is deliberately untouched (it's persisted to `.compose-flags` and consumed by `dream.ps1` for non-install operations; appending there would change semantics for `dream start`/`restart`/`stop`).

2. **`opencode web → serve` rationale** — re #1097 which already merged the identical change for the systemd unit: per that PR's body, `opencode serve` is the canonical headless variant that calls `Server.listen()` and never `open()`. The web UI is **still served** on `--hostname 127.0.0.1 --port 3003`; the only behavioral delta is no host-browser auto-launch. The user reaches the IDE manually at `http://localhost:3003`, which is what the existing `MACOS-QUICKSTART.md` already says ("OpenCode (IDE): http://localhost:3003"). Hence no docs change is required — the docs remain factually correct after the change. The LaunchAgent label `com.dreamserver.opencode-web` is also preserved deliberately to keep the upgrade path's `launchctl bootout` working on existing installs.

Branch is now rebased onto `upstream/main` (`d4124d6a`).

Lightheartdevs (approved):
Re-audited after carrying Yasin's branch forward onto current `main`.

The original requested-changes concern was valid: current OpenCode docs still distinguish `opencode web` as the browser UI from `opencode serve` as the headless HTTP/OpenAPI server, so I removed the macOS LaunchAgent `serve` change from this PR rather than changing DreamServer's advertised OpenCode IDE behavior. The repaired diff now keeps the worthwhile installer behavior only: install-time `docker compose up` on Linux, macOS, and Windows runs with `--remove-orphans` while preserving the newer `--no-build` flow from `main`.

Local checks run:
- `bash -n dream-server/installers/macos/install-macos.sh dream-server/installers/phases/11-services.sh dream-server/tests/contracts/test-plist-log-paths.sh`
- `bash dream-server/tests/contracts/test-plist-log-paths.sh`
- PowerShell parser on `dream-server/installers/windows/install-windows.ps1`
- `git diff --check --cached`

PR #1080, opened 2026-05-01
fix(openclaw): drop dangerouslyAllowHostHeaderOriginFallback flag

yasinBursali:
## What
Removes the `dangerouslyAllowHostHeaderOriginFallback` controlUi flag from openclaw — the only one of three shipped dangerous flags that's safe to remove without an auth-flow redesign.

## Why
The flag weakens DNS-rebinding / CSRF protection by accepting the Host header as the origin when no Origin header is sent. There's no functional need for it: `inject-token.js` already populates `allowedOrigins` with the correct localhost / 127.0.0.1 entries, so legitimate browsers can authenticate normally without the fallback.

The other two flags (`allowInsecureAuth`, `dangerouslyDisableDeviceAuth`) were verified as required: TLS termination isn't available in this stack (so HTTP-only auth must be permitted), and the device-pairing flow is incompatible with the Docker-internal auto-token injection (so device-auth must be skipped). Removing those would require a separate auth-flow redesign.

The compose entrypoint's `--bind lan` is **intentionally preserved**: it controls the container's *internal* listen address. Docker's port-NAT requires the container service to listen on `0.0.0.0`, which `--bind lan` provides. Host-side LAN exposure is gated separately by the `${BIND_ADDRESS:-127.0.0.1}` ports pattern (PR #964).

## How
Removes `dangerouslyAllowHostHeaderOriginFallback` from all 5 sites:
- `config/openclaw/openclaw.json` (default)
- `config/openclaw/openclaw-strix-halo.json` (AMD Strix Halo APU)
- `config/openclaw/pro.json` (pro tier)
- `config/openclaw/inject-token.js` Part 1 (writes `~/.openclaw/openclaw.json`)
- `config/openclaw/inject-token.js` Part 3 (writes `/tmp/openclaw-config.json` — the runtime config the gateway loads via `OPENCLAW_CONFIG_PATH`)

The two `inject-token.js` sites use `delete` (not line-removal) so the upgrade path actively defangs any pre-existing flag that's persisted in `~/.openclaw/openclaw.json` via the `openclaw-home` named Docker volume.

## Testing
New test `tests/test-openclaw-inject-token.sh` runs `inject-token.js` with mock env in a temp HOME and asserts:
1. Part 3 merged config (`/tmp/openclaw-config.json`) lacks the flag
2. Part 1 home config (`~/.openclaw/openclaw.json`) lacks the flag
3. **Upgrade scenario**: pre-seeds the home config with `dangerouslyAllowHostHeaderOriginFallback: true`, re-runs `inject-token.js`, confirms the flag is defanged
4. Regression guards: `dangerouslyDisableDeviceAuth: true`, `allowInsecureAuth: true`, `allowedOrigins` contains `http://localhost:7860` + `http://127.0.0.1:7860`, `gateway.auth.mode === 'token'`, `gateway.auth.token` populated
5. Built-in negative test: synthetic regressed copy that re-introduces the flag fails the suite

15/15 assertions pass. Cross-platform-safe (`sed -i` BSD vs GNU branch, `mktemp` portable form, jq `// empty` for absence checks). Pre-commit hooks (gitleaks/private-key/large-files) all green.

**Manual** (recommended):
- Linux/macOS: `docker compose restart openclaw`; `docker exec openclaw cat /tmp/openclaw-config.json | jq '.gateway.controlUi'` — assert `dangerouslyAllowHostHeaderOriginFallback` is absent; `dangerouslyDisableDeviceAuth: true` and `allowInsecureAuth: true` are preserved
- DNS-rebinding test: `curl -H "Host: evil.example.com" http://127.0.0.1:7860/` — assert non-2xx (no Host-header fallback)

## Known Considerations
- **LAN-mode access from peer hosts**: with the flag removed, peer LAN hosts accessing `http://192.168.x.y:7860` are no longer accepted via the Host-header fallback. `inject-token.js` populates `allowedOrigins` with `localhost`, `127.0.0.1`, and the container's `os.hostname()` (which inside Docker is the container ID hex, NOT the host's LAN IP). Users who relied on peer-LAN access via openclaw will need a follow-up that adds the host LAN IP to `allowedOrigins` when `BIND_ADDRESS=0.0.0.0`. Defense-in-depth wins this trade-off.
- **Runtime DNS-rebinding test**: not added here (would require full container start in CI). Tracked as a separate test PR.

## Platform Impact
- **macOS**: same openclaw image, identical fix
- **Linux NVIDIA / CPU**: same
- **Linux AMD / Strix Halo**: same (separate JSON config patched symmetrically)
- **Windows/WSL2**: same — openclaw runs identically inside WSL2's Docker

### DIFF
--- a/dream-server/config/openclaw/inject-token.js
+++ b/dream-server/config/openclaw/inject-token.js
@@ -61,7 +61,7 @@ try {
   // Ensure controlUi flags are set for local use
   config.gateway.controlUi.allowInsecureAuth = true;
   config.gateway.controlUi.dangerouslyDisableDeviceAuth = true;
-  config.gateway.controlUi.dangerouslyAllowHostHeaderOriginFallback = true;
+  delete config.gateway.controlUi.dangerouslyAllowHostHeaderOriginFallback;  // defang upgrades from pre-PR installs that wrote this flag to the persistent volume
 
   // Keep token auth (required for LAN bind) with token from env
   if (token) {
@@ -217,7 +217,7 @@ try {
     if (!primary.gateway.controlUi) primary.gateway.controlUi = {};
     primary.gateway.controlUi.allowInsecureAuth = true;
     primary.gateway.controlUi.dangerouslyDisableDeviceAuth = true;
-    primary.gateway.controlUi.dangerouslyAllowHostHeaderOriginFallback = true;
+    delete primary.gateway.controlUi.dangerouslyAllowHostHeaderOriginFallback;  // defang carry-over (mirrors Part 1)
     const extPort = process.env.OPENCLAW_EXTERNAL_PORT || '7860';
     const origins = primary.gateway.controlUi.allowedOrigins || [];
     for (const o of [`http://localhost:${extPort}`, `http://127.0.0.1:${extPort}`]) {
--- a/dream-server/config/openclaw/openclaw-strix-halo.json
+++ b/dream-server/config/openclaw/openclaw-strix-halo.json
@@ -32,8 +32,7 @@
     "mode": "local",
     "controlUi": {
       "allowInsecureAuth": true,
-      "dangerouslyDisableDeviceAuth": true,
-      "dangerouslyAllowHostHeaderOriginFallback": true
+      "dangerouslyDisableDeviceAuth": true
     }
   }
 }
--- a/dream-server/config/openclaw/openclaw.json
+++ b/dream-server/config/openclaw/openclaw.json
@@ -32,8 +32,7 @@
     "mode": "local",
     "controlUi": {
       "allowInsecureAuth": true,
-      "dangerouslyDisableDeviceAuth": true,
-      "dangerouslyAllowHostHeaderOriginFallback": true
+      "dangerouslyDisableDeviceAuth": true
     }
   }
 }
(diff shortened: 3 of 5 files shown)


Lightheartdevs (approved):
Conditional approval — needs jq-backed CI test green before merge.

Code review: removes `dangerouslyAllowHostHeaderOriginFallback: true` from both OpenClaw config-injection paths (Part 1 and Part 2 in inject-token.js). The `delete config.gateway.controlUi.dangerouslyAllowHostHeaderOriginFallback` form correctly defangs the flag on existing installs that already wrote it to the persistent volume — important for upgrade safety. Same removal in the static `openclaw.json` and `openclaw-strix-halo.json` configs. This was an outstanding finding from the original SECURITY_AUDIT.md (H3).

**Required before merge:** `tests/test-openclaw-inject-token.sh` in CI with `jq` available (the local Windows env doesn't have it; full assertion path skipped).

Lightheartdevs (approved):
Approved after re-audit against current main. I simulated the merge result to make sure the newer #1123 no-token-output/no-token-static-asset behavior is preserved, then ran a focused Node smoke: inject-token.js parses, stdout does not include OPENCLAW_GATEWAY_TOKEN, the merged config removes dangerouslyAllowHostHeaderOriginFallback, and allowedOrigins keeps localhost/127.0.0.1. The live integration-smoke red is the shared _docker_cmd_arr expectation mismatch and is unrelated to this OpenClaw change.

PR #1059, opened 2026-04-28
refactor(dashboard-api): extract pure settings helpers into settings.py

boffin-dmytro:
## Summary

- `main.py` was ~1478 lines (~3× the 500-line guideline). This PR extracts **9 env-related constants** and **14 pure leaf functions** into a new `settings.py` module, reducing `main.py` to ~1170 lines (~21% reduction).
- Functions that call `_resolve_install_root`, `_resolve_runtime_env_path`, or `_resolve_template_path` remain in `main.py` so that monkeypatches in `test_settings_env.py` continue to intercept them correctly (Python resolves names at the module where they are defined, not the caller's module).
- Zero behavioural change — only the file boundary moves.

## What moved to `settings.py`

**9 constants:** `_ENV_ASSIGNMENT_RE`, `_ENV_COMMENTED_ASSIGNMENT_RE`, `_SENSITIVE_ENV_KEY_RE`, `_SETTINGS_APPLY_ALLOWED_SERVICES`, `_LLAMA_APPLY_KEYS`, `_OPEN_WEBUI_APPLY_KEYS`, `_TOKEN_SPY_APPLY_KEYS`, `_PRIVACY_SHIELD_APPLY_KEYS`, `_MANUAL_RESTART_KEYS`

**14 functions:** `_strip_env_quotes`, `_parse_env_text`, `_read_env_map_from_path`, `_normalize_bool`, `_humanize_env_key`, `_is_secret_field`, `_slugify`, `_build_env_fields`, `_validate_env_values`, `_serialize_form_values`, `_match_apply_service`, `_build_apply_summary`, `_compute_env_apply_plan`, `_check_host_agent_available`

## Test plan

- [x] `tests/test_settings_env.py` — all 16 tests pass unchanged
- [x] `python3 -c "import settings"` — imports cleanly
- [ ] No new tests added (pure refactor — no logic changed)

PR #1088, opened 2026-05-01
fix(dashboard-api): preserve explicit empty values for commented-default env keys

yasinBursali:
## What
`_render_env_from_values` now writes `KEY=` unconditionally for commented-assignment lines, matching the assignment branch above it. The previous `if value != \"\":` guard silently dropped explicit empty values on round-trip, reverting the rendered `.env` to the comment form.

## Why
A user who set `LLAMA_ARG_TENSOR_SPLIT=\"3,1\"` via the dashboard, then cleared it (set to `\"\"`), would see the `.env` end up with `# LLAMA_ARG_TENSOR_SPLIT=` instead of `LLAMA_ARG_TENSOR_SPLIT=`. The key was also added to `seen` so it didn't appear in the extras section either — the explicit clear vanished. Downstream consumers relying on `${VAR+set}` distinction would see the user's intent reverted.

## How
4-line fix in `main.py:704` — drop the `if value != \"\":` guard, mirror the assignment branch's unconditional `f\"{key}={values.get(key, '')}\"` write.

## Testing
- 19/19 `test_settings_env.py` pass after fix
- 3 new regression tests cover the actual `commented_assignment` branch via `LLAMA_ARG_TENSOR_SPLIT` (commented at `.env.example:184`).
- New `commented_example_template` fixture monkeypatches `_resolve_template_path` so tests actually traverse the patched branch — without it, the existing `test_render_env_preserves_extras_with_empty_values` silently fell through to the extras branch and never exercised the regression target.
- Negative-test verified: reverting the production fix causes 2 of 3 new tests to fail (the empty-value and absent-from-values cases — exactly the regression scope).

## Manual test
1. Pick a commented-default key in `.env.example` (e.g., `# LLAMA_ARG_TENSOR_SPLIT=`).
2. PUT `/api/settings/env` with that key set to `\"\"`.
3. Read `.env` — expect `LLAMA_ARG_TENSOR_SPLIT=` (uncommented, empty).
4. GET `/api/settings/env` — expect round-trip preserves `\"\"`.

## Known Considerations
This is a small intentional behavior change for keys that appear as `# KEY=` in `.env.example` and are absent from `values`: the rendered output is now `KEY=` (uncommented empty) instead of `# KEY=`. Verified no consumer in the tree distinguishes set-but-empty from unset on these keys (all `\${...:-}` colon-form, which treats both identically; the two `\${...+set}` callers I checked apply to keys not in `.env.example`).

File overlap with #1059 (settings-module refactor): both touch `main.py`. If #1059 moves `_render_env_from_values` into `settings.py`, whichever lands second will need a small rebase.

## Platform Impact
Pure Python — identical on macOS, Linux, Windows-WSL2.

### DIFF
--- a/dream-server/extensions/services/dashboard-api/main.py
+++ b/dream-server/extensions/services/dashboard-api/main.py
@@ -706,11 +706,7 @@ def _render_env_from_values(values: dict[str, str]) -> str:
         if commented_assignment:
             key = commented_assignment.group(1)
             seen.add(key)
-            value = values.get(key, "")
-            if value != "":
-                output_lines.append(f"{key}={value}")
-            else:
-                output_lines.append(line)
+            output_lines.append(f"{key}={values.get(key, '')}")
             continue
 
         output_lines.append(line)
--- a/dream-server/extensions/services/dashboard-api/main.py
+++ b/dream-server/extensions/services/dashboard-api/main.py
@@ -706,7 +706,10 @@ def _render_env_from_values(values: dict[str, str]) -> str:
         if commented_assignment:
             key = commented_assignment.group(1)
             seen.add(key)
-            output_lines.append(f"{key}={values.get(key, '')}")
+            if key in values:
+                output_lines.append(f"{key}={values[key]}")
+            else:
+                output_lines.append(line)
             continue
 
         output_lines.append(line)
--- a/dream-server/extensions/services/dashboard-api/tests/test_settings_env.py
+++ b/dream-server/extensions/services/dashboard-api/tests/test_settings_env.py
@@ -311,6 +311,83 @@ def test_render_env_preserves_extras_with_empty_values():
     assert "GPU_UUID=GPU-abc123" in rendered
 
 
+@pytest.fixture()
+def commented_example_template(tmp_path, monkeypatch):
+    """Patch ``_resolve_template_path`` so .env.example resolution returns a
+    controlled file containing a commented-assignment line.
+
+    Required because the default test environment cannot resolve the real
+    ``.env.example`` (INSTALL_DIR is /tmp/dream-test-install which does not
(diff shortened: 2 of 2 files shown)


Lightheartdevs (changes requested):
﻿This fixes the explicit-empty case, but it broadens the behavior too much.

In the commented-assignment branch, the new code always writes `KEY=` even when `key` is absent from `values`:

```py
output_lines.append(f"{key}={values.get(key, '')}")
```

That means any dashboard save can turn optional commented defaults from `.env.example` into active empty assignments, even if the user never had or submitted that setting. The new test `test_render_env_uncomments_commented_key_absent_from_values` locks that broader behavior in, but I think that is the wrong contract: absent should preserve the commented template line; explicit empty should render as `KEY=`.

A narrower fix would be:

```py
if key in values:
    output_lines.append(f"{key}={values[key]}")
else:
    output_lines.append(line)
```

That still preserves a user's explicit clear because `_prepare_env_save()` merges submitted values into `merged_values`, but it avoids changing unrelated optional commented defaults from unset/commented to set-empty. It also keeps future `${VAR+set}` consumers from being surprised by dashboard saves that made an absent key present.

Lightheartdevs (approved):
Re-audited and pushed the narrower fix requested in the prior review: commented .env.example defaults are only uncommented when the submitted/merged values explicitly contain the key. Explicit empty values still round-trip as KEY=, while absent optional defaults remain commented.\n\nLocal checks: python -m pytest dream-server/extensions/services/dashboard-api/tests/test_settings_env.py -q; python -m py_compile for main.py and test_settings_env.py; git diff --check. Full dashboard API pytest timed out locally after 3 minutes in this workspace, but CI's API job is green, along with mypy and the rest of the checks.

PR #1089, opened 2026-05-01
fix(installer): rebuild locally-built images on macOS, Windows, and via dream-cli

yasinBursali:
## What
Brings macOS and Windows installers + `dream-cli` to feature parity with Linux phase-11's image-rebuild step:
- `installers/macos/install-macos.sh` — parallel build loop with PID-array per-build error checking via `ai_warn`; switches to `up -d --no-build`.
- `installers/windows/install-windows.ps1` — sequential build loop with `\$LASTEXITCODE` per-service `Write-AIWarn`; switches to `up -d --no-build`.
- `dream-cli` — opt-in `--rebuild-images` flag for `restart`, `start`, and `update`, backed by `_dream_cli_rebuild_images` helper.

## Why
On Linux, `installers/phases/11-services.sh:311–340` rebuilds the 5 locally-built services (dashboard, dashboard-api, ape, token-spy, privacy-shield, plus conditional comfyui/llama-server-amd/dreamforge) before `up -d --no-build`. macOS and Windows installers were running `up -d` with no preceding build — contributor edits to bake-in services silently no-op'd on reinstall. Same trap PR #1055's dev-workflow doc warns about, addressed from the installer side.

## How
- **macOS**: parallel `&` with explicit PID-array tracking + per-PID `wait \$pid || ai_warn \"Build failed for \$svc\"`. Bare `wait` returns 0 even on child failure under `set -e`, so per-PID checks are required to surface failures by service name.
- **Windows**: sequential `docker compose ... build --no-cache <svc>`, each followed by `\$LASTEXITCODE` check + non-fatal `Write-AIWarn`. CRLF preserved.
- **dream-cli**: helper appends conditional services based on `.env` state (mirroring phase-11 logic). Documented in `cmd_help`. Default behavior unchanged — `--rebuild-images` is opt-in. Argparse handles flag-first / flag-last / `--` / unknown-flag cases without breaking existing positional usage.

## Testing
- `bash -n` clean on `dream-cli` and `install-macos.sh`
- `shellcheck` no new warnings on edited regions
- PowerShell visual review (no pwsh on dev host); CRLF preserved

## Manual test plan (per platform)
- **Linux** (regression-only — phase-11 unchanged): full `install.sh` should still rebuild as before.
- **macOS**: edit `routers/extensions.py`, run `install-macos.sh`, confirm \"Rebuilding local-built images\" log + `docker exec dream-dashboard-api` reflects the change.
- **macOS failure path**: break a Dockerfile, run installer, confirm `[!!] Build failed for <name>` warning fires + `up -d --no-build` produces a clean missing-image error.
- **Windows**: run `install-windows.ps1` in a Windows VM, confirm sequential build output to `compose-build.log`.
- **dream-cli flag** (any platform): `dream restart --rebuild-images`, `dream start <svc> --rebuild-images`, `dream update --rebuild-images`, `dream restart` (default — no rebuild).

## Known Considerations
- The `up -d --no-build` switch on macOS and Windows means a failed build now surfaces as a missing-image error at compose-up time rather than silently rebuilding from cached layers — matches Linux's deliberate semantics.
- File overlap with #1079 (macos installer orphan + opencode serve): both touch `install-macos.sh` at different sections. Whichever lands second will rebase cleanly.

## Platform Impact
- **Linux**: NOT affected (regression-only).
- **macOS**: build loop newly added before `up -d --no-build`.
- **Windows**: build loop newly added before `up -d --no-build`.
- **dream-cli `--rebuild-images`**: opt-in across all platforms; ComfyUI/llama-server/dreamforge conditionally included based on `.env` state.

### DIFF
--- a/dream-server/dream-cli
+++ b/dream-server/dream-cli
@@ -607,6 +607,25 @@ _compose_run_with_summary() {
 }
 
 #=============================================================================
+# Warn (don't exit) when the current user can't reach the Docker daemon on
+# Linux. Phase 05 of the installer sets DOCKER_CMD="sudo docker" for users
+# not in the docker group, but dream-cli calls bare `docker compose` and
+# can't see that. Without this check, --rebuild-images fails with an opaque
+# "permission denied while trying to connect to the Docker daemon socket"
+# and users have no idea what to do next. macOS Docker Desktop grants
+# socket access to the logged-in user, so the check is Linux-only.
+#=============================================================================
+_check_docker_access() {
+    [[ "$(uname -s)" == "Darwin" ]] && return 0
+    if id -nG 2>/dev/null | grep -qw docker && docker ps >/dev/null 2>&1; then
+        return 0
+    fi
+    warn "Current user cannot reach the Docker daemon (not in 'docker' group, or daemon unreachable)."
+    warn "If the next step fails with 'permission denied', run:  newgrp docker  (current shell)"
+    warn "or log out and back in (all shells), then re-run your dream command."
+}
+
+#=============================================================================
 # Rebuild local-built images (opt-in via --rebuild-images on restart/start/update).
 # Always rebuilds: dashboard, dashboard-api, ape, token-spy, privacy-shield.
 # Conditionally rebuilds based on .env state:
@@ -627,6 +646,7 @@ _dream_cli_rebuild_images() {
     [[ "${ENABLE_COMFYUI:-}" == "true" ]] && svcs+=(comfyui)
     [[ "${GPU_BACKEND:-}" == "amd" ]] && svcs+=(llama-server)
     [[ "${ENABLE_DREAMFORGE:-}" == "true" ]] && svcs+=(dreamforge)
+    _check_docker_access
     log "Rebuilding local-built images (--rebuild-images)..."
     if ! docker compose "${flags[@]}" build --no-cache "${svcs[@]}"; then
         warn "One or more images failed to rebuild (continuing — see compose output above)"
--- a/dream-server/dream-cli
+++ b/dream-server/dream-cli
@@ -607,6 +607,33 @@ _compose_run_with_summary() {
 }
 
 #=============================================================================
+# Rebuild local-built images (opt-in via --rebuild-images on restart/start/update).
+# Always rebuilds: dashboard, dashboard-api, ape, token-spy, privacy-shield.
+# Conditionally rebuilds based on .env state:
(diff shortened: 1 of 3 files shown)


Lightheartdevs (changes requested):
﻿Blocking this until the Windows installer parses cleanly.

The new warning line in `installers/windows/install-windows.ps1` contains mojibake for an em dash (`â€”`) inside a double-quoted string:

```powershell
Write-AIWarn "$_svc build failed (non-fatal â€” see $_buildLog)"
```

PowerShell treats the embedded smart-quote character in that sequence as a string delimiter, so parsing the file fails before the installer can run. Reproduction:

```powershell
$src = Get-Content -Raw dream-server/installers/windows/install-windows.ps1
[scriptblock]::Create($src)
```

That reports missing closing braces / missing try-catch starting at the new build loop because the string is prematurely closed. Please replace that dash with plain ASCII (`-`) or otherwise make the file parse-safe. Local Bash syntax for `dream-cli` and `install-macos.sh` passed; the blocker is Windows parseability.

yasinBursali:
Pushed the parseability fix you flagged. Commit `fd35c986` replaces the UTF-8 em-dash on `installers/windows/install-windows.ps1:625` with an ASCII hyphen:

```diff
- Write-AIWarn "$_svc build failed (non-fatal — see $_buildLog)"
+ Write-AIWarn "$_svc build failed (non-fatal - see $_buildLog)"
```

Verified line 625 contains zero non-ASCII bytes after the fix. The other em-dashes in the file are inside `#` comments (banner decorators, lines 59/64/75/97/202/220/263/277/326/338/383/473/589 — pre-existing, not parsed as strings).

Branch also rebased onto current `main` (`d4124d6a`); 3 commits ahead.

Ready for fresh review on the install-time rebuild loop.

Lightheartdevs (approved):
Re-audited after Yasin's parse fix and pushed one follow-up correction: macOS now prebuilds DreamForge when the generated .env sets DREAMFORGE_PULL_POLICY=build, so Apple Silicon installs do not hit up --no-build without the required native image. Also narrowed the dream-cli Docker group warning to Linux only and fixed the Windows log text to match --no-build.\n\nLocal checks: bash -n dream-server/dream-cli dream-server/installers/macos/install-macos.sh; PowerShell scriptblock parse for install-windows.ps1; git diff --check; targeted dream-cli fake-docker smoke for --rebuild-images service selection. CI is green.

PR #1090, opened 2026-05-01
fix(host-agent,dashboard): surface one-shot CLI extensions as installed

yasinBursali:
## What
Aider extension installs cleanly via the dashboard for the first time. Adds first-class dashboard surfacing for one-shot CLI extensions whose containers exit 0 by design.

- `resources/dev/extensions-library/services/aider/manifest.yaml` — sets `service.startup_check: false` so the host agent's running-state poll exits early on clean container exit.
- `extensions/schema/service-manifest.v1.json` + `README.md` — documents the existing `startup_check` (boolean) and `startup_timeout` (integer 1–600s) fields the host agent reads.
- `routers/extensions.py` — adds `_is_one_shot_extension` helper + `cli_installed` status branches in `_compute_extension_status` (in-flight install with recent 'started' progress, and post-install steady state). Catalog summary's `installed` count now includes `cli_installed`, plus a top-level `cli_installed` counter.
- `dashboard/src/pages/Extensions.jsx` — `cli_installed` extended through STATUS_STYLES, STATUS_DESCRIPTIONS, pollProgress success branch + differentiated success toast, isToggleable, and ExtensionCard color classes.
- `tests/test_extensions.py` — 2 regression tests with negative-test verification.

## Why
Aider's compose is intentionally one-shot: `entrypoint: [\"echo\"]`, `restart: \"no\"`, `port: 0`, no `health_endpoint`. The container exits 0 cleanly within seconds, but the host agent's `_handle_install` was waiting for `State.Status == \"running\"` for 15 seconds before failing with \"Container did not reach running state within 15s (state=exited)\". The host-agent already supports this case via `service.startup_check: false` (`bin/dream-host-agent.py:1454-1462`) — the aider manifest just never opted in. Even after opting in, the dashboard would never resolve the install to a success state because `_compute_extension_status` had no notion of one-shot CLI tools — the toast never fired, the badge cycled \"installing\" → \"stopped\".

## How
- `_is_one_shot_extension` uses catalog `port == 0` as the proxy (no manifest read on hot path; aider is the only `port: 0` extension in the catalog today).
- `_compute_extension_status` returns `cli_installed` in two places: in-flight install (replacing the \"installing\" → \"stopped\" cycle) and post-install steady state when `compose.yaml` exists in `USER_EXTENSIONS_DIR`. `cli_installed` is added BEFORE the healthy/unhealthy checks so the conditional ladder remains coherent.
- Frontend extends every status-handling surface symmetrically with the existing `enabled` precedent.
- Catalog summary integrates with the recently-merged `unhealthy` counter.

## Testing
- `python3 -m py_compile` clean
- `ruff check` clean
- 185/185 `test_extensions.py` pass (including 2 new `cli_installed` tests)
- `npm run lint` 0 errors / 5 pre-existing warnings (none in `Extensions.jsx`)
- `npm run build` clean
- `bash scripts/validate-manifest-schema.sh` — aider manifest validates 0/0
- Negative-test verified: reverting either of the two `_compute_extension_status` `cli_installed` branches causes the corresponding new test to fail.

## Manual test
1. Install Aider via the dashboard — expect \"CLI INSTALLED\" badge + success toast \`Aider installed — run via 'docker compose run --rm aider'.\` (no error, no 15s timeout).
2. Regression: install n8n → expect normal \"installing\" → \"enabled\" flow.
3. Toggle/disable aider after install → expect normal disable behavior.

## Known Considerations / Follow-ups (out of scope for this PR)
- The toast/STATUS_DESCRIPTIONS text references \`docker compose run --rm aider\` — since aider's compose has \`entrypoint: [\"echo\"]\`, that exact command actually hits the echo stub. Invoking aider needs \`--entrypoint aider\`. Pre-existing aider compose design; can be addressed separately.
- `_is_one_shot_extension` uses `port == 0` as a proxy because the extensions catalog doesn't carry `startup_check`. A future enhancement could extend the catalog generator to expose the flag — accepted as a limitation today since aider is the only one-shot extension.
- `scripts/validate-manifest-schema.sh` has a pre-existing `set -euo pipefail` + `((TOTAL++))` bug that aborts the script on the first manifest. Pre-existing, unrelated to this PR.

## Platform Impact
**macOS / Linux / Windows-WSL2**: identical behavior. Host-agent is host-side Python; dashboard-api is Python; frontend is JSX. Aider container behavior is Docker-only (no native path).


Lightheartdevs (dismissed):
Conditional approval — needs frontend + API CI green before merge.

Code review: adds two new manifest schema fields (`startup_check`, `startup_timeout`) for one-shot CLI extensions whose containers intentionally exit after initialization (e.g. aider). Currently those show as failed because the host agent's running-state poll times out. New `cli_installed` status threads through schema → API → frontend. Schema is backwards-compatible (both fields default to current behavior).

**Required before merge:** dashboard-api tests + frontend Vitest both green. The status-state addition is user-visible across catalog count, badges, and toggle behavior — full surface needs CI confirmation. Couldn't run frontend tests locally (Vitest deps not installed in this audit env).

Lightheartdevs:
﻿Merge-pass recheck: this is not mergeable right now because GitHub reports conflicts and the live `Check catalog is up-to-date` job is red.

The one-shot CLI-extension state is a real UX gap, but this needs a rebase onto current `main`, conflict resolution, and catalog regeneration/cleanup before review can finish. Once the branch is clean and checks are green, the main review question will be whether `port == 0` is the right long-term proxy for one-shot CLI tools or whether the catalog should expose `startup_check` directly.

Lightheartdevs (approved):
Re-audited after repairing Yasin's branch onto current `main` and resolving the dashboard conflict against the newer recovery-tracker flow.

I made two corrections before merge:
- exposed `service.startup_check` in the generated extensions catalog and made the dashboard prefer that explicit signal, keeping `port: 0` only as a fallback for older catalogs
- fixed the `cli_installed` card toggle so it renders as an active/green installed state and added a frontend regression test for that surface

Local checks run:
- `python resources/dev/scripts/generate-extensions-catalog.py`
- `python -m json.tool` for the generated catalog and manifest schema
- `python -m pytest dream-server/extensions/services/dashboard-api/tests/test_extensions.py -k "cli_installed or catalog_includes_user_extension_health"`
- `npm ci`
- `npm test -- --run src/pages/Extensions.test.jsx`
- `npm run lint -- src/pages/Extensions.test.jsx` (existing warnings only in unrelated hooks)
- `git diff --check`

PR #1092, opened 2026-05-02
fix(installer): add INTERACTIVE/sudo-n guard to Phase 01 jq install

yasinBursali:
## What
Add an INTERACTIVE / `sudo -n` guard before the bare-sudo `jq` install in `installers/phases/01-preflight.sh:49`.

## Why
In `--non-interactive` mode without NOPASSWD sudo, the existing bare-sudo block hangs on a TTY password prompt that never arrives. Phase 05 already has the canonical guard; Phase 01 lacked it.

## How
Insert the same `[[ "${INTERACTIVE:-true}" != "true" ]] && ! sudo -n true 2>/dev/null` guard (verbatim from `05-docker.sh:60-68`) before the case block. On guard hit, call `error` for fail-fast.

## Testing
- `bash -n` clean
- `shellcheck` delta-clean (no new warnings)
- `make lint` clean
- Manual Linux: `INTERACTIVE=false` on a sudoless container hard-errors with the new message instead of hanging. With NOPASSWD: jq installs normally.

## Review
Critique Guardian: APPROVED. Verbatim copy of the established Phase 05 pattern; no security/quality regression.

## Platform Impact
- **Linux**: fixed. Bug only triggers in `--non-interactive` without NOPASSWD; pure UX/reliability improvement.
- **macOS**: not applicable. macOS uses `installers/macos/install-macos.sh` (separate preflight, no jq sudo path).
- **Windows / WSL2**: not applicable. Windows installer (`installers/windows/phases/01-preflight.ps1`) has no jq install path.

### DIFF
--- a/dream-server/installers/phases/01-preflight.sh
+++ b/dream-server/installers/phases/01-preflight.sh
@@ -46,6 +46,11 @@ log "curl: $(curl --version 2>/dev/null | sed -n '1p')"
 
 if ! command -v jq &> /dev/null; then
     log "jq not found - attempting auto-install..."
+    # In non-interactive mode, fail fast if sudo requires a password
+    # (otherwise the bare `sudo <pkgmgr>` below hangs on a TTY prompt that never comes).
+    if [[ "${INTERACTIVE:-true}" != "true" ]] && ! sudo -n true 2>/dev/null; then
+        error "Cannot install jq: sudo password required but running in --non-interactive mode. Re-run interactively or configure NOPASSWD sudo for the package manager."
+    fi
     case "$PKG_MANAGER" in
         dnf)    sudo dnf install -y jq ;;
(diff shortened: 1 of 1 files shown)


Lightheartdevs (approved):
Adds `INTERACTIVE`/`sudo -n` guard before bare `sudo <pkgmgr>` install of jq in Phase 01. Mirrors the same pattern in 05-docker.sh. Fails fast in non-interactive mode instead of hanging on a TTY prompt. Ship after rebase.

Lightheartdevs (approved):
Re-audited for merge. The Phase 01 jq install now has the non-interactive sudo -n guard, matching the existing installer pattern and avoiding password-prompt hangs. Local bash -n passed and the guard is present. Approving for squash merge.

PR #1094, opened 2026-05-02
fix(installer): subshell-scope .env umask, mkdir+warn openclaw chown, repoint missing tier configs

yasinBursali:
## What
Three independent installer fixes bundled into one PR (all touch Phase 06 + Phase 03):

1. **`.env` umask race** — wrap the heredoc in `( umask 077; cat > .env << EOF ... EOF )` so the file is created at mode 600 directly, closing the create-then-chmod race for other UIDs.
2. **OpenClaw chown silencer** — pre-create `data/openclaw` then convert `chown ... 2>/dev/null || true` to `chown ... || warn`.
3. **Missing OpenClaw tier configs** — Tier 1/2/3/wildcard mappings referenced `minimal.json`/`entry.json`/`prosumer.json` which were deleted in v2.0.0. Repoint to existing `openclaw.json` (Bash + PowerShell parity). Drop the also-missing `openclaw.json.example` fallback; replace with a hard `error` so packaging corruption surfaces immediately.

Adds `dream-server/tests/bats-tests/phase-06-perms.bats` with two regression tests.

## Why
1. Without `umask 077`, the heredoc creates `.env` at 0644 (Ubuntu/Debian default) for ~141 lines until `chmod 600` runs. Race for other UIDs to read all secrets.
2. CLAUDE.md bans `2>/dev/null || true`. The silencer was load-bearing because `data/openclaw` wasn't pre-created — chown crashed on fresh installs without it.
3. Tier 1/2/3 covers the majority of consumer GPUs. Currently produces a silent broken OpenClaw with no config and a green healthcheck.

## How
1. Subshell-scope the `umask 077` so it doesn't leak to subsequent `mkdir -p config/litellm`, `mkdir -p config/searxng`, or to phases 09/11/13 bind-mount dirs (which would otherwise become 700 and block container UIDs like SearXNG 977 / OpenClaw 1000). Comment block explains the constraint.
2. `mkdir -p data/openclaw` before the chown; mirror the established `mkdir + warn` pattern from token-spy.
3. Bash case + PowerShell switch repointed identically. Hard `error` / `Write-AIError + exit 1` on missing config.

## Testing
- `bash -n` / `shellcheck` / `make lint` clean (PowerShell visual review against existing PS idiom).
- New BATS tests pass locally (1..2; ok 1; ok 2).
- Manual Linux Tier 1: `.env` is 0600, `data/openclaw` exists with uid 1000:1000, `config/openclaw/openclaw.json` is the generic config. Renaming `openclaw.json` triggers the new hard error.
- Windows manual test: deferred to release smoke.

## Review
Critique Guardian: APPROVED (round 2 after subshell-wrap fix + 2 BATS tests added per round-1 required-before-merge findings).

## Known Considerations
The `.env` umask race is platform-contingent: triggered when `\$HOME` is mode 0755 (Ubuntu/Debian default) or `INSTALL_DIR` is outside `\$HOME`. Arch/CachyOS/Fedora's 0700 `\$HOME` makes the race unreachable, but the fix is universal/harmless. A follow-up issue is recommended for adding a hard-error-path BATS test on missing OpenClaw config (advisory, not blocking).

## Platform Impact
- **Linux**: fully fixed.
- **macOS**: not affected. `installers/macos/lib/env-generator.sh:348` heredoc-generates `openclaw.json` inline and is not touched.
- **Windows**: tier-config repoint applied to PowerShell (`installers/windows/phases/03-features.ps1`); hard-error fallback applied to `installers/windows/phases/06-directories.ps1`. The `.env` umask + chown sub-fixes are Linux-only (Windows installer creates `.env` differently).
- **WSL2**: same as Linux.


Lightheartdevs (approved):
Three coupled changes: (1) subshell-scope umask 077 ensures generated `.env` is created 0600 from inception (not chmod-after); (2) mkdir+`ai_warn` for openclaw chown failures (was silent); (3) repoint missing tier configs (`minimal.json`/`entry.json`/`prosumer.json`) to canonical `openclaw.json`. Worth a Linux + Windows installer smoke with OpenClaw enabled before merge. Ship after that.

Lightheartdevs (approved):
Re-audited for merge. The OpenClaw tier mappings now point only at profiles that exist in config/openclaw, Linux phase 03/06 syntax checks pass, Windows phase 03/06 PowerShell parsing passes, and the PR's Linux BATS coverage is included in the green Test Linux / integration-smoke job. I also verified all mapped OpenClaw profiles resolve to existing files. Approving for squash merge.

PR #1096, opened 2026-05-02
fix(memory-shepherd): add -e to set flags so failed cp/mv/scp aborts visibly

yasinBursali:
## What
One-character fix in `dream-server/memory-shepherd/memory-shepherd.sh:4`: `set -uo pipefail` → `set -euo pipefail`.

## Why
Without `-e`, a failed `cp`/`mv`/`scp` mid-reset doesn't abort the script. The most critical scenario: line 161's `cp \$baseline \$tmpfile` could silently fail (disk full, permissions), producing an empty `\$tmpfile`; line 162's atomic `mv -f \$tmpfile \$memory_file` would then overwrite live agent memory with empty content — silent data loss.

CLAUDE.md mandates `set -euo pipefail` everywhere; this script was the residual non-compliant case (the `dream-cli` portion of the original concern is covered by separate upstream changes).

## How
Single-character flag flip. Pre-flight audit confirmed: 2 `grep` sites (L135, L202) and 1 `scp` site (L188) that legitimately exit non-zero are already structurally safe under `-e` (inside command-substitution short-circuits or `if !` conditions). No `|| true` or `|| warn` masks added per CLAUDE.md "Let It Crash > KISS" — failed I/O should crash.

## Testing
- `bash -n` / `shellcheck` / `make lint` clean.
- Manual Linux smoke: script runs with `--help` / unknown-agent invocation cleanly.
- Manual sabotage test: `chmod 000` a memory file path → script crashes visibly with permission-denied (Let-It-Crash confirmed).

## Review
Critique Guardian: APPROVED WITH WARNINGS (advisory).

## Known Considerations
After this fix, two bare `scp` statements (L190 and L228 — baseline-push to remote sync target) will abort the function on push failure. This is the intended Let-It-Crash semantic per CLAUDE.md, NOT a regression. The systemd timer (`Type=oneshot`, no `Restart=`, fires every 3h with `Persistent=true`) records the failure cleanly and retries on the next cycle — no "until reset" lockout. Similarly, the `for agent in "\${AGENTS[@]}"` loop in `main()` will halt on the first failing agent rather than continuing past data-integrity issues; this is also intended.

## Platform Impact
- **Linux**: fixed. Script is deployed via systemd timers installed by `installers/phases/10-amd-tuning.sh:74` (gated `GPU_BACKEND==amd`), so production exposure is AMD-only Linux installs. Script body is portable (no AMD-specific calls).
- **macOS**: not applicable. Script not deployed on macOS (uses BSD `stat` incompatible with the script's GNU `stat -c` calls anyway).
- **Windows**: not applicable.

### DIFF
--- a/dream-server/memory-shepherd/memory-shepherd.sh
+++ b/dream-server/memory-shepherd/memory-shepherd.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 # memory-shepherd.sh — Periodic memory baseline reset for LLM agents
 # Usage: memory-shepherd.sh [agent-name|all]
-set -uo pipefail
+set -euo pipefail
 
 TIMESTAMP=$(date '+%Y-%m-%d_%H%M')
 LOCKFILE=/tmp/memory-shepherd.lock
--- a/dream-server/memory-shepherd/memory-shepherd.sh
+++ b/dream-server/memory-shepherd/memory-shepherd.sh
@@ -108,15 +108,15 @@ reset_agent() {
     local archive_dir="$4"
 
     if [ ! -f "$baseline" ]; then
-        log "CRITICAL: Baseline missing for $agent at $baseline — aborting"
-        return 1
+        log "ERROR: Baseline missing for $agent at $baseline — skipping"
+        return 0
     fi
 
     local baseline_size
     baseline_size=$(stat -c %s "$baseline")
     if [ "$baseline_size" -lt "$MIN_BASELINE_SIZE" ]; then
-        log "CRITICAL: Baseline for $agent is suspiciously small (${baseline_size} bytes, min: ${MIN_BASELINE_SIZE}) — aborting"
-        return 1
+        log "ERROR: Baseline for $agent is suspiciously small (${baseline_size} bytes, min: ${MIN_BASELINE_SIZE}) — skipping"
+        return 0
     fi
 
     if [ ! -f "$memory_file" ]; then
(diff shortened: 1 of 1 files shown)


Lightheartdevs (approved):
Single-character fix: `set -uo pipefail` → `set -euo pipefail` adds the missing `-e` so failed cp/mv/scp aborts visibly. Aligns memory-shepherd with CLAUDE.md "`set -euo pipefail` everywhere" rule. Existing `|| true` guards on non-critical find cleanup are preserved per Codex's audit. Ship after rebase + reset/archive smoke.

Lightheartdevs (changes requested):
Thanks for tightening the failure behavior here. I found one regression that should be fixed before merge: adding global set -e makes process_agent return paths abort the entire all-agent run, even for cases the script logs as 'skipping'. Repro: create a config with [bad] missing baseline and [good] valid baseline/memory, then run MEMORY_SHEPHERD_CONF=... memory-shepherd.sh all. On main, bad is logged/skipped and good is reset; on this branch the script exits after bad with STATUS=1 and good is never processed. The fix should make true transfer/write failures fail loud without changing intended per-agent skip semantics, e.g. handle process_agent failures in the all loop explicitly while preserving nonzero summary status if desired.

Lightheartdevs (approved):

Audit notes:
- The `set -euo pipefail` change is worthwhile: real failed writes/transfers now stop visibly instead of risking silent partial reset behavior.
- Fixed the requested regression: validation cases that the script logs as skipped now return success, so `memory-shepherd.sh all` continues to later agents when one agent has a missing/invalid baseline or incomplete config.
- Preserved fail-loud behavior for actual I/O: a baseline copy into a missing memory-file directory now exits nonzero under `set -e`.
- Local checks: `bash -n` passed; all-agent skip smoke passed; transfer/write failure smoke passed. Local ShellCheck is unavailable in this Windows environment, but refreshed GitHub Actions ShellCheck is green.

PR #1103, opened 2026-05-02
refactor: extract _run_post_install_hook + createRecoveryTracker

yasinBursali:
## What
Two refactors that DRY up the post-install hook executor in the host-agent and the failure-counter / poll-recovery pattern in the dashboard:

- Extracts a shared `_run_post_install_hook(service_id, ext_dir) -> tuple[bool, str]` helper consolidating the hook-execution logic that was duplicated between `_run_install` and `_enable_retry_work`.
- Introduces a `createRecoveryTracker` factory utility consolidating the failure-counter pattern shared between `Extensions.jsx`'s progress poller and `ConsoleModal`'s log poller.

## Why
- `_run_install` and `_enable_retry_work` had three behavioral divergences in their hook-execution code: progress write timing (one wrote "Running setup..." even when no hook existed — a misleading-message bug), inner timeout handling (one had a specific message, the other relied on a generic outer handler), and stderr slice direction (head vs tail). Three diverging code paths with subtly different UX = easy to forget to update one when fixing the other.
- The dashboard's `pollProgress` (per-service interval-based) and `ConsoleModal.poll` (singleton recursive-with-backoff) shared a "count failures, fire callback at threshold, reset on success" primitive but each implemented it inline with slightly different counter semantics. A change to the threshold (3) or the recovery semantics required updating both sites.

## How
**`bin/dream-host-agent.py`** — extracted helper at line 450 (placed between `_resolve_hook` and `_enable_retry_work`):

```python
def _run_post_install_hook(service_id: str, ext_dir: Path) -> tuple[bool, str]:
    """8-key env allowlist mirrors _execute_hook (L1488-1498) ..."""
    hook_path = _resolve_hook(ext_dir, "post_install")
    if not hook_path:
        return (True, "")  # no hook declared — caller proceeds without setup progress
    _write_progress(service_id, "setup_hook", "Running setup...")
    # ... 8-key env allowlist (PATH, HOME, SERVICE_ID, SERVICE_PORT, SERVICE_DATA_DIR,
    #     DREAM_VERSION, GPU_BACKEND, HOOK_NAME) preserved verbatim from both call sites ...
    try:
        result = subprocess.run([...], timeout=SUBPROCESS_TIMEOUT_START)
    except subprocess.TimeoutExpired:
        msg = f"post_install hook timed out ({SUBPROCESS_TIMEOUT_START}s)"
        _write_progress(service_id, "error", "Setup failed", error=msg)
        return (False, msg)
    if result.returncode != 0:
        msg = (result.stderr or "")[-500:]   # tail-500 + None guard
        _write_progress(service_id, "error", "Setup failed", error=msg)
        return (False, msg)
    return (True, "")
```

Helper writes progress (not caller-writes) so the callers stay compact and the no-hook bug fix is centralized.

`_run_install` and `_enable_retry_work` now both invoke `ok, _ = _run_post_install_hook(service_id, ext_dir); if not ok: return`. The pre-existing `if run_setup_hook:` guard at the `_run_install` call site is preserved. The pre-existing outer `subprocess.TimeoutExpired` handler in `_run_install` still fires for non-hook timeouts (`docker pull`, `docker compose up`, docker inspect) — the helper's inner try/except absorbs only the hook-subprocess timeout.

**`extensions/services/dashboard/src/utils/recoveryTracker.js`** (new file):

```javascript
export function createRecoveryTracker({ threshold = 3, onThresholdReached, onRecovered } = {}) {
  let counter = 0;
  let pastThreshold = false;
  return {
    recordFailure() {
      counter += 1;
      if (counter >= threshold && !pastThreshold) {
        pastThreshold = true;
        onThresholdReached?.();
      }
      return counter;
    },
    recordSuccess() {
      const wasPastThreshold = pastThreshold;
      counter = 0;
      pastThreshold = false;
      if (wasPastThreshold) onRecovered?.();
    },
  };
}
```

Closure-based, not a React hook — keeps `pollProgress`'s per-serviceId Map shape and `ConsoleModal`'s singleton shape both clean.

`Extensions.jsx` `pollProgress` now stores trackers in `recoveryTrackers.current` (Map keyed by serviceId, lazy-created) and calls `recordFailure()` / `recordSuccess()` from the existing fetch handlers. Cleanup mirrors `activePollers.current` cleanup at every clear-poll site (error path, success path, unmount). `ConsoleModal.poll` uses a singleton tracker created when the effect runs; the existing `setTimeout`-with-backoff mechanism keeps a local `failCount = tracker.recordFailure()` for the backoff math.

**Intentional behavior changes** (each operator-approved):
1. `_run_install` no longer writes "Running setup..." progress when no hook is declared — bug fix.
2. `_run_install`'s hook-timeout message changes from generic "timed out (600s)" to specific "post_install hook timed out (600s)" — UX improvement.
3. `_enable_retry_work`'s stderr slice changes from head-500 to tail-500 — unifies with `_run_install` post the in-flight runtime-hygiene branch's similar change; tail is more useful (errors typically at the end of stderr).

**Out of scope** (intentionally untouched):
- `_execute_hook` (third hook path with `Popen + os.setsid + os.killpg`, 120s timeout) — different shape, unrelated concern.
- POSIX-only `setsid`/`killpg` promotion in the new helper — kept as a separate follow-up.
- `setInterval` (pollProgress) vs recursive `setTimeout+backoff` (ConsoleModal.poll) unification — the scheduling difference is intentional and stays.
- `AbortSignal.timeout(8000)` parity — only ConsoleModal.poll has it; pollProgress's progress endpoint is fast enough that adding it would be over-engineering.

## Testing
- `make lint` — pass.
- `pytest extensions/services/dashboard-api/tests/` — 698 passed, 13 failed (all 13 pre-existing baseline failures verified by stash-pop on the worktree; not introduced by this PR).
- `pytest tests/test_host_agent.py` — 63/63 pass, including 4 retargeted env-allowlist source-level tests (now grep `_run_post_install_hook` instead of `_handle_install` since the env construction moved into the helper — same regression invariant, more truthful target).
- `npx vitest run` — 54/54 pass across 11 files, including 6 new `recoveryTracker.test.js` cases (threshold-fires-once, recordSuccess-resets, onRecovered-only-when-was-past-threshold, default-threshold-3, etc.).
- `npm run lint` — clean.
- Live host-agent smoke skipped — would require `launchctl kickstart -k` while operator is not present, not safe under dangerously-skip mode. Source-level + unit tests cover the regression surface; runtime smoke can be operator-driven post-merge.

## Review
Critique Guardian: **APPROVED** (re-verified after a 1-line docstring count fix from "7-key" → "8-key"). Two non-blocking observations from CG:
1. Docstring count fixed.
2. New `src/utils/` directory was created; existing `src/lib/` could have hosted it. Defensible as-is, not worth re-spinning.

## Known Considerations
- This PR is **DRAFT** and must merge after the in-flight host-agent + dashboard one-shot CLI extension branch (which touches `pollProgress` and is the highest-risk overlap with the `recoveryTracker` refactor). Mechanical safeguard.
- Branch base is the local staging branch with the full upstream-PR queue merged; until the prerequisite branches land, the PR diff includes those changes too. After they merge, a rebase onto `upstream/main` will reduce the diff to just this PR's scope.
- No POSIX-only / Windows-only platform code was introduced. The host-agent already runs on macOS launchd, Linux systemd, and Windows-WSL2 systemd; the helper changes are platform-agnostic.

## Platform Impact
- **macOS** (launchd-managed host-agent + Apple Silicon dashboard): no impact — `subprocess.run` semantics identical; React tree unchanged shape; tracker is pure JS.
- **Linux** (systemd host-agent + Docker dashboard): no impact — same.
- **Windows-WSL2** (systemd-in-WSL host-agent + Docker Desktop dashboard): no impact — same.


yasinBursali:
Closing in favor of two separately-shippable PRs that split this change cleanly:

- The **host-agent half** (extracting `_run_post_install_hook` and refactoring both `_run_install` and `_enable_retry_work` to use it) was consolidated into PR Light-Heart-Labs/DreamServer#1039 as a second commit, since it depends on `_enable_retry_work` which #1039 introduces. The known-gap noted in #1039's original description ("shared helper would unify env allowlist + timeout handling") is now addressed inside the same PR.

- The **dashboard half** (extracting `createRecoveryTracker` and refactoring `pollProgress` + `ConsoleModal.poll`) ships independently as PR Light-Heart-Labs/DreamServer#1105 — it has no dependency on the host-agent work and can merge against current `upstream/main` today.

The original PR was filed off the local staging branch which produced an unreviewable 91-commit diff. Closing instead of force-pushing because the work is now better organized as two coherent PRs.

PR #1105, opened 2026-05-02
refactor(dashboard): consolidate failure-counter into createRecoveryTracker

yasinBursali:
## What
Consolidates the "count consecutive failures, fire callback at threshold, reset on success" primitive that was duplicated between two pollers in `Extensions.jsx`:
- `pollProgress` (per-service interval-based progress poller)
- `ConsoleModal.poll` (singleton recursive-with-backoff log poller)

Both inline implementations used threshold=3, reset-on-success, with slightly different recovery callbacks. Extracted into a small factory utility so future pollers can re-use the same primitive without copy-paste.

## Why
Threshold + counter logic that's right twice and wrong twice if anyone mis-edits one site. The primitive is small but the duplication makes it footgun-shaped — change the threshold (3) on one side and the other silently drifts.

## How
New file `dream-server/extensions/services/dashboard/src/utils/recoveryTracker.js`:

```javascript
export function createRecoveryTracker({ threshold = 3, onThresholdReached, onRecovered } = {}) {
  let counter = 0;
  let pastThreshold = false;
  return {
    recordFailure() {
      counter += 1;
      if (counter >= threshold && !pastThreshold) {
        pastThreshold = true;
        onThresholdReached?.();
      }
      return counter;
    },
    recordSuccess() {
      const wasPastThreshold = pastThreshold;
      counter = 0;
      pastThreshold = false;
      if (wasPastThreshold) onRecovered?.();
    },
  };
}
```

**Closure-based on purpose, NOT a React hook.** The two callers want fundamentally different storage shapes:
- `pollProgress` wants per-serviceId trackers in a `useRef` Map (lazy-created on first poll per service).
- `ConsoleModal.poll` wants a singleton tracker per modal instance.

A single hook signature can't cover both shapes cleanly; a factory utility does. Each caller wraps the factory in their own state-management mechanism (Map ref vs local closure).

**`pollProgress` refactor** (Extensions.jsx L107-160): replaces `consecutiveFailuresRef` with `recoveryTrackers.current` (Map keyed by serviceId). Lazy-created on first poll per service. Cleanup at every clear-poll site (error path, success path, unmount) deletes the tracker entry alongside the existing `setInterval` cleanup. Threshold callback retains the original `setPollingLost(true) + fetchCatalog()`. `onRecovered` uses functional setter `setPollingLost(prev => prev ? false : prev)` to avoid extra renders.

**`ConsoleModal.poll` refactor** (Extensions.jsx L1047-1081): singleton tracker created when the effect runs. `recordFailure()` returns the new count which feeds the existing exponential-backoff math `Math.min(2000 * Math.pow(2, failCount-1), 30000)`. Threshold callback sets `disconnected`. Success path retains `setDisconnected(false)` + error clear directly (no `onRecovered` callback needed).

## Intentionally NOT changed
- `setInterval(3000)` cadence in `pollProgress`.
- Recursive `setTimeout` + exponential backoff in `ConsoleModal.poll`.
- `AbortSignal.timeout(8000)` only on `ConsoleModal.poll` (the log endpoint is slow; the progress endpoint is fast — adding it to `pollProgress` would be over-engineering).
- Threshold value (3) on either side.
- `pollingLost` banner UI.

## Testing
- New file `src/utils/__tests__/recoveryTracker.test.js` — 6 cases covering:
  - Counter increments on failure.
  - `onThresholdReached` fires once at threshold (not multiple times on subsequent failures).
  - `recordSuccess` resets counter.
  - `onRecovered` fires when crossing back from past-threshold.
  - `onRecovered` does NOT fire if never past threshold.
  - Default threshold = 3.
- `npx vitest run` previously verified at 54/54 pass with these exact files (was 48/48 before this PR's tests + the in-flight Extensions UI test PR).
- `npm run lint` clean.

## Review
The backend half of this refactor (extraction of the host-agent's `_run_post_install_hook` helper that DRYs the install + enable-retry hook paths) was consolidated into PR Light-Heart-Labs/DreamServer#1039 since it depends on `_enable_retry_work` which that PR introduces. This frontend half is independent — works against current upstream/main.

## Platform Impact
- **macOS**: no impact — pure JS / closure factory + React component refactor; Vitest runs identically.
- **Linux**: no impact — same.
- **Windows-WSL2**: no impact — same.

### DIFF
--- a/dream-server/extensions/services/dashboard/src/pages/Extensions.jsx
+++ b/dream-server/extensions/services/dashboard/src/pages/Extensions.jsx
@@ -7,6 +7,7 @@ import { useState, useEffect, useRef } from 'react'
 import { DependencyBadges, DependencyConfirmDialog, DisableDependentWarning } from '../components/DependencyBadges'
 import { TemplatePicker } from '../components/TemplatePicker'
 import { getTemplateStatus } from '../lib/templates'
+import { createRecoveryTracker } from '../utils/recoveryTracker'
 
 // Re-export so existing importers of getTemplateStatus from this module keep working.
 export { getTemplateStatus }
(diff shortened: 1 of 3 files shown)


Lightheartdevs (approved):
Extracts the per-service consecutive-failure counter into a reusable `createRecoveryTracker` utility with its own unit tests. Pulls duplicated polling-failure logic out of Extensions.jsx and makes the recovery banner behavior testable. Ship after rebase + frontend CI.

Lightheartdevs (approved):
Re-audited for merge. The refactor pulls the duplicated failure-counter/polling recovery behavior into a small closure helper and keeps the backoff count behavior intact for the console poller. Local targeted Vitest passed on the branch, then I temporarily merged current main to include the newly merged Extensions tests and reran 
pm.cmd test -- src/utils/__tests__/recoveryTracker.test.js src/pages/Extensions.test.jsx: 2 files, 12 tests passed. Current CI is green.

PR #1110, opened 2026-05-02
fix(host-agent): rewrite build.context to absolute path on staging

yasinBursali:
## What
Fix the install path for library extensions that ship `build:` directives. Three previously-broken extensions now install correctly: bark, audiocraft, open-interpreter.

- `dream-server/extensions/services/dashboard-api/routers/extensions.py` — new `_rewrite_build_context()` helper invoked inside `_install_from_library` between `_scan_compose_content` and the atomic `os.rename`. It rewrites relative `build.context` (`.`, `./foo`, short-form `build: <str>`) to the absolute path of the post-rename extension directory. Idempotent on absolute paths.
- `resources/dev/extensions-library/services/audiocraft/compose.yaml`, `…/open-interpreter/compose.yaml` — add `platform: linux/amd64` so Apple Silicon Docker Desktop pulls amd64 layers via QEMU emulation. Bark untouched (multi-arch already).
- `dream-server/extensions/services/dashboard-api/tests/test_install_from_library_build_context.py` — 10-case unit-test module exercising the rewrite logic (relative dot, relative non-dot, short-form, missing context, idempotent absolute, no-build, multi-service, invalid YAML root, malformed YAML).

## Why
Docker Compose v2 resolves `build.context: .` against the project directory — the directory of the FIRST `-f` file. The host agent invokes `docker compose -f <base.yml> -f <gpu.yml> -f data/user-extensions/<svc>/compose.yaml up -d <svc>` from `cwd=INSTALL_DIR`, so the project directory becomes `INSTALL_DIR`. A relative `context: .` then resolves to `INSTALL_DIR/.` instead of the extension's staged directory at `INSTALL_DIR/data/user-extensions/<svc>/`. Result: `unable to evaluate symlinks in Dockerfile path: lstat INSTALL_DIR/Dockerfile: no such file or directory`. All three affected extensions failed to install via the dashboard.

The fix lives at the staging layer (post-`_scan_compose_content`, pre-`os.rename`) where the final extension dir's absolute path is known. Failure at this layer cleans up via the surrounding `tempfile.TemporaryDirectory` — partial state cannot leak into `USER_EXTENSIONS_DIR`. Closes the issue for all current and future library extensions with `build:` directives, not just the three named ones.

## How
1. Read the staged compose.yaml with `yaml.safe_load(encoding="utf-8")`.
2. For each service with a `build` field, normalize short-form (`build: <str>`) and dict-form to dict-form, then rewrite `context` to `str((USER_EXTENSIONS_DIR / service_id).resolve())` if the current value is relative. Absolute paths are left alone (idempotency).
3. Write back via `yaml.safe_dump(data, sort_keys=False)`. `logger.info` per service rewritten so operators can audit.
4. Failure modes: `yaml.YAMLError` propagates (per CLAUDE.md error-handling — no swallowing). Atomic rename hasn't happened yet; broken staging copy is in tmpdir cleaned by `shutil.rmtree(tmpdir, ignore_errors=True)`.

`_scan_compose_content(staged_compose, trusted=True)` is unchanged — `build:` capability for library extensions remains intact.

## Testing
- 10 new pytest cases — all pass. Coverage: relative dot context; relative non-dot; short-form; missing context; idempotent absolute (dict and short forms); no-build no-op; multi-service mixed; invalid YAML root; malformed YAML propagation.
- Existing dashboard-api test suite: 70 pre-existing failures (asyncio deps, unrelated), 659 passes — zero regression.
- `python3 -m py_compile` on the changed file.
- Both modified compose files parse via `docker compose config`.

## Review
Critique Guardian: APPROVED across two rounds. Round 1 adjudicated the "absolute-path security question" (preserved trust assumption, not widened). Round 2 re-validated the polish: encoding kwargs, README QEMU notes (factual + calibrated wording).

## Platform Impact
- **Linux amd64 / Linux WSL2 amd64**: `platform: linux/amd64` is a no-op; native execution. Build-context rewrite resolves to a normal Linux absolute path. Both audiocraft and open-interpreter install end-to-end.
- **macOS Apple Silicon**: `platform: linux/amd64` triggers Docker Desktop's QEMU x86_64 emulation. Build is functional but slower (5–10x build time, 2–5x runtime CPU). Each extension's README documents this so users know what to expect. Bark untouched (multi-arch).
- **macOS Intel**: native amd64; no emulation; functional.
- **Windows native**: not a target for the dashboard install flow.

## Known Considerations
- The `build:` trust path remains gated by `_scan_compose_content(trusted=True)` for library extensions. This PR neither widens nor narrows that trust assumption. Defense-in-depth (rejecting absolute paths and `..` traversal in `build.context` even for `trusted=True`) is a future hardening opportunity, out of scope here.
- No existing pytest harness exercised `_install_from_library`'s body; the new test module covers `_rewrite_build_context` in isolation. An end-to-end integration harness for `_install_from_library` is a future improvement.

### DIFF
--- a/dream-server/extensions/services/dashboard-api/routers/extensions.py
+++ b/dream-server/extensions/services/dashboard-api/routers/extensions.py
@@ -1120,10 +1120,83 @@ def _install_from_library(service_id: str) -> None:
         staged_compose = staged / "compose.yaml"
         if staged_compose.exists():
             _scan_compose_content(staged_compose, trusted=True)
+            # Rewrite build.context to the absolute final extension dir.
+            # Compose resolves relative contexts against the project dir
+            # (INSTALL_DIR), not the extension dir, so "context: ." would
+            # look for the Dockerfile in INSTALL_DIR/Dockerfile and fail.
+            _rewrite_build_context(staged_compose, dest.resolve())
         os.rename(str(staged), str(dest))
     finally:
         if Path(tmpdir).exists():
             shutil.rmtree(tmpdir, ignore_errors=True)
+
+
+def _rewrite_build_context(compose_path: Path, final_dir: Path) -> None:
+    """Rewrite build.context in a staged compose.yaml to an absolute path.
+
+    Library extensions ship `build: { context: ., ... }` so they're portable
+    inside the library, but `docker compose` resolves relative contexts
+    against the compose project directory (INSTALL_DIR), not the extension's
+    own directory. After staging, rewrite each service's build.context to
+    the absolute path of the final post-rename extension directory.
+
+    Idempotent: absolute paths are left alone (in case the compose was
+    already rewritten on a previous attempt).
+    """
+    with open(compose_path, encoding="utf-8") as f:
+        data = yaml.safe_load(f)
+
+    if not isinstance(data, dict):
+        return
+
+    services = data.get("services")
+    if not isinstance(services, dict):
+        return
+
+    final_dir_str = str(final_dir)
+    changed = False
+
+    for service_name, service in services.items():
+        if not isinstance(service, dict):
+            continue
+        build = service.get("build")
+        if build is None:
+            continue
+
+        if isinstance(build, str):
+            # Short-form `build: <path>` — normalize to dict form
+            if os.path.isabs(build):
+                continue
+            service["build"] = {"context": final_dir_str}
+            logger.info(
+                "Rewrote build context for service '%s' from '%s' to '%s'",
+                service_name, build, final_dir_str,
+            )
+            changed = True
+            continue
+
+        if isinstance(build, dict):
+            context = build.get("context")
+            if context is None:
+                # Compose default is `.`; make it explicit and absolute
+                build["context"] = final_dir_str
+                logger.info(
+                    "Set build context for service '%s' to '%s' (was implicit)",
+                    service_name, final_dir_str,
+                )
+                changed = True
+                continue
+            if isinstance(context, str) and not os.path.isabs(context):
+                build["context"] = final_dir_str
+                logger.info(
+                    "Rewrote build context for service '%s' from '%s' to '%s'",
+                    service_name, context, final_dir_str,
+                )
+                changed = True
+
+    if changed:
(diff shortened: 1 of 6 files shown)


Lightheartdevs (approved):
Adds `_rewrite_build_context` helper that converts `build.context: .` (relative) to the absolute final extension dir on staging. Required for build-backed library extensions like Audiocraft and Open Interpreter — `docker compose` resolves relative contexts against INSTALL_DIR, not the extension dir, so "context: ." looks for the Dockerfile in the wrong place. Idempotent (absolute paths left alone). Ship after rebase + API tests.

Lightheartdevs (changes requested):
This fixes the current context: . library entries, but the generalized helper is wrong for non-dot relative contexts. _rewrite_build_context() rewrites uild.context: ./subdir to <final_dir>; Docker Compose semantics would make that <final_dir>/subdir once the extension is installed. The new test 	est_rewrites_other_relative_context currently asserts the incorrect behavior. Please preserve the relative suffix, e.g. resolve (final_dir / context) for relative contexts, and reject/avoid paths that escape the extension dir if you want this to stay safe. Current library entries only use ., but the helper/tests should not encode the wrong behavior before we merge it.

Lightheartdevs (approved):

Audit notes:
- Confirmed the original fix is useful: library extensions with `build.context: .` need staging-time absolute contexts because Compose resolves relative build paths against the project directory, not the extension directory.
- Addressed the requested review change: relative build contexts now preserve their suffix, so `./subdir` becomes `<final extension dir>/subdir` instead of the extension root.
- Added a safety guard so relative contexts that resolve outside the final extension directory, such as `../bar`, are rejected with a 400 before install rename.
- Added test coverage for relative suffix preservation, short-form build paths, and escape rejection; the full build-context test module now has 12 passing cases.
- Local checks: `py_compile`, `ruff`, targeted pytest, Audiocraft compose config, and Open Interpreter compose config with a dummy required API key all pass. Refreshed GitHub Actions are green across API, Ruff, mypy, shellcheck, matrix smoke, and integration smoke.

PR #1112, opened 2026-05-02
fix(host-agent): rollback all three config backups on model-activate failure

yasinBursali:
## What
`_do_model_activate` in the host agent now restores all three config backups (`.env`, `models.ini`, and `lemonade.yaml`) before returning a 500 when an unexpected exception fires during model activation.

## Why
The outer `except Exception` handler at the bottom of `_do_model_activate` previously returned a 500 response without restoring any of the three files it had already mutated. On AMD-Linux and WSL2-AMD (where all three files are written) this left `.env`, `models.ini`, and `lemonade.yaml` in a partially-modified state, with litellm routing pointing at a model that may or may not have loaded. On macOS and NVIDIA-Linux (no lemonade.yaml) two files were left dirty. A subsequent activation attempt would then start from an inconsistent baseline.

## How
Three changes to `dream-server/bin/dream-host-agent.py`:

1. **Hoist backup variables before the outer `try`** — `env_backup`, `ini_backup`, `lemonade_backup`, and `committed` are now initialised to `""`, `""`, `None`, and `False` respectively above the try block. This prevents the rollback handler from hitting a `NameError` when an exception fires before the in-try capture assignments complete.

2. **Set `committed = True` on the success path** — placed immediately after the `json_response(self, 200, …)` call. The outer except checks `if not committed:` before restoring, so a post-commit failure (e.g. a secondary write in the `dream-mode=lemonade` branch) cannot undo a healthy activation.

3. **Rollback with nested `try/except OSError`** — the three `write_text` calls in the outer except are wrapped in a narrow `OSError` catch. If a rollback write itself fails the original 500 error message is preserved in the response and the rollback failure is logged via `logger.exception`.

The macOS pre-launch guard (L1991-2004) and the existing `healthy=False` rollback (L2119-2122) are deliberately untouched; each has its own documented restore pattern.

## Testing
- Automated: `py_compile` clean on `dream-server/bin/dream-host-agent.py`; no automated unit test (host-agent module is not importable by the dashboard-api pytest scaffolding — see Known Considerations)
- Manual: On AMD-Linux, induce `_recreate_llama_server` failure by setting an invalid `LLAMA_SERVER_IMAGE` value; confirm all three files are restored to their pre-activate checksums and the API returns a 500 with the original error message

## Review
Critique Guardian: APPROVED. No required-before-merge items. Informational: the empty-string default for `env_backup`/`ini_backup` is a near-zero-probability TOCTOU corner — a read failure between file-open and assignment could cause the rollback to write `""` over an existing file. A `None`-sentinel would be cleaner but applying it here without also hardening the existing `healthy=False` rollback at L2119 would be inconsistent; accepted as-is and tracked as a future holistic cleanup.

## Known Considerations
Two follow-up enhancements should be filed post-merge:

1. **Testable helper extraction** — extract the rollback logic from `_do_model_activate` into a standalone helper so it can be covered by a unit test without requiring the full host-agent HTTP server.
2. **macOS guard consolidation** — the pre-launch guard at L1991-2004 currently returns early without calling the outer-except rollback. A future change should convert that `return` to `raise` so all failure paths flow through the same restore logic.

These are enhancements, not deficiencies in this fix.

## Platform Impact
- macOS: affected — exercises 2-file rollback (`.env` + `models.ini`; `lemonade_backup` is `None` and skipped)
- Linux (NVIDIA): affected — same 2-file rollback as macOS
- Linux (AMD) + WSL2-AMD: affected — exercises full 3-file rollback including `lemonade.yaml`
- Windows (WSL2-NVIDIA): affected — same 2-file rollback as NVIDIA-Linux

### DIFF
--- a/dream-server/bin/dream-host-agent.py
+++ b/dream-server/bin/dream-host-agent.py
@@ -1900,6 +1900,13 @@ class AgentHandler(BaseHTTPRequestHandler):
         models_ini = INSTALL_DIR / "config" / "llama-server" / "models.ini"
         lemonade_yaml = INSTALL_DIR / "config" / "litellm" / "lemonade.yaml"
 
+        # Hoisted so the outer except's rollback can reference them even if
+        # an exception fires before the in-try captures complete.
+        env_backup = ""
+        ini_backup = ""
+        lemonade_backup = None
+        committed = False
+
         try:
             # Read current env BEFORE modification — needed for gpu_backend guard
             env_pre = load_env(env_path)
@@ -2105,6 +2112,7 @@ class AgentHandler(BaseHTTPRequestHandler):
                     subprocess.run(["docker", "restart", svc],
                                    capture_output=True, timeout=60)
                 json_response(self, 200, {"status": "activated", "model_id": model_id})
+                committed = True  # success-path complete; outer except must not rollback
             else:
                 # Rollback
                 logger.warning("Model activation failed — rolling back")
@@ -2148,6 +2156,14 @@ class AgentHandler(BaseHTTPRequestHandler):
                 json_response(self, 500, {"error": "Health check failed — rolled back to previous model", "rolled_back": True})
 
         except Exception as exc:
+            if not committed:
+                try:
+                    env_path.write_text(env_backup, encoding="utf-8")
+                    models_ini.write_text(ini_backup, encoding="utf-8")
+                    if lemonade_backup is not None:
+                        lemonade_yaml.write_text(lemonade_backup, encoding="utf-8")
+                except OSError:
+                    logger.exception("Rollback write failed during model-activate failure handling")
             json_response(self, 500, {"error": f"Model activation failed: {exc}"})
 
     def _handle_model_delete(self):
--- a/dream-server/bin/dream-host-agent.py
+++ b/dream-server/bin/dream-host-agent.py
@@ -1900,12 +1900,20 @@ class AgentHandler(BaseHTTPRequestHandler):
         models_ini = INSTALL_DIR / "config" / "llama-server" / "models.ini"
         lemonade_yaml = INSTALL_DIR / "config" / "litellm" / "lemonade.yaml"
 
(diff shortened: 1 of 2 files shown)


Lightheartdevs (changes requested):
This is very close, but I think it needs one small ordering fix before merge.

On the healthy path the code currently does:

```py
json_response(self, 200, {status: activated, model_id: model_id})
committed = True
```

If activation and dependent-service restarts succeed but the client disconnects while the 200 response is being written, `json_response()` can raise `BrokenPipeError`/`ConnectionResetError`. Because `committed` is still `False`, the new outer `except` will restore `.env`, `models.ini`, and `lemonade.yaml`, rolling back an activation that actually completed successfully. That is especially plausible here because model activation can take minutes and dashboard clients may navigate away or timeout.

Please set `committed = True` after the last real mutation/restart succeeds but before writing the response, e.g. immediately before the `json_response(self, 200, ...)` call. The rollback guard should represent whether the system state is committed, not whether the HTTP response was delivered.

Local verification so far: `py_compile` passes; the issue is behavioral and not covered by current CI.

Lightheartdevs (approved):

Audit notes:
- Confirmed the original rollback fix is valuable: unexpected activation failures now restore `.env`, `models.ini`, and `lemonade.yaml` instead of leaving a half-activated model config behind.
- Addressed the requested review change: `committed` is now set after the last state mutation/restart succeeds and before the HTTP 200 response write, so a client disconnect cannot roll back a successful activation.
- Hardened the backup handling with `None` sentinels so an exception before snapshots are captured cannot overwrite existing config files with empty defaults.
- Added focused regression coverage for client disconnect after success, unexpected restart failure rollback across all three config files, and pre-snapshot failure safety.
- Local checks: `py_compile`, `ruff`, `test_model_activate.py` all pass. Running `test_host_agent.py + test_model_activate.py` on local Windows has unrelated symlink/chmod failures in existing host-agent tests; refreshed GitHub Actions are green, including API, Ruff, mypy, shellcheck, and integration smoke.

PR #1113, opened 2026-05-02
fix(dashboard-api): narrow broad except handlers in main.py per CLAUDE.md rule 1

yasinBursali:
## What
Narrow four `except Exception:` handlers in `dashboard-api/main.py` to specific exception tuples, and update the corresponding test to reflect the new contract.

## Why
`except Exception:` at I/O probe sites masks programming errors (AttributeError, KeyError, TypeError, RuntimeError) alongside the transient I/O failures those handlers actually intended to tolerate. A `KeyError` inside `_build_api_status` would silently return a `{"services": [], "gpu": null, …}` dict of zeros to the dashboard instead of surfacing as a 500 where it can be caught by tests and monitoring. This directly violates CLAUDE.md rule 1 ("No broad or silent catches") and the project's "Let It Crash" priority.

## How
`main.py` — four targeted one-line narrowings:

- `_check_host_agent_available` (L774): `except Exception` → `except (urllib.error.URLError, urllib.error.HTTPError, OSError)`. The `False` return path is unchanged; only the guard is tightened.
- `/api/status` top-level handler (L1097): `except Exception` → `except (asyncio.TimeoutError, OSError)`. Docstring updated to document the new "programming errors propagate" contract.
- `api_settings_env_save` inner json.loads (L1351): `except Exception` → `except (json.JSONDecodeError, UnicodeDecodeError, AttributeError)`. The outer `raise HTTPException(503)` is unconditional — this was never a silent swallow, just an over-broad guard on a malformed host-agent error body.
- `api_settings_env_apply` inner json.loads (L1411): same as above.

`_poll_service_health` background loop (L1449) is intentionally left untouched. Loop death from an unhandled exception is worse than an over-broad catch, and `logger.exception` already captures the full stack trace. CLAUDE.md rule 2 covers this case explicitly.

`tests/test_main.py`:

- `test_fallback_on_exception` renamed `test_fallback_on_oserror`; injection changed from `RuntimeError` to `OSError`; still asserts 200 + safe-fallback dict — the intended I/O-error path still works.
- New `test_runtime_error_propagates_as_500` documents the new contract: a `RuntimeError` inside `_build_api_status` now reaches the client as 500.

## Testing
- Automated: `py_compile` clean on both changed files. Test contracts verified by inspection: `OSError` is in the caught tuple → 200 fallback path intact; `RuntimeError` is not → propagates as 500. Full pytest suite covers remaining `TestApiStatusFallback` class.
- Manual:
  1. Start dashboard-api with host agent down → `GET /api/status` → expect 200 with `{"services": [], "gpu": null, "model": null, "tier": "Unknown"}`.
  2. Corrupt `data/bootstrap-status.json` (e.g. write `{invalid`) → `GET /api/status` → expect 500 (was 200-with-zeros before this fix). This is the intended "Let It Crash" outcome.
  3. Restart host agent → `GET /api/status` → expect full payload restored.

## Review
Critique Guardian: APPROVED. No required-before-merge items. Informational: `get_bootstrap_status` with a corrupt status file now surfaces 500 instead of 200-with-zeros — this is the intended "Let It Crash" outcome, worth flagging so reviewers understand the observable behavior change. The `AttributeError` catch at L1351/L1411 is consistent with the file's narrow-at-I/O-boundary idiom; an alternative `isinstance(err_payload, dict)` guard would be slightly clearer but is a style question only.

## Known Considerations
- `get_bootstrap_status` corrupt-file 500 is a behavior change (see Testing step 2 above). It is intentional and correct per the project's error-handling rules, but a stale `data/bootstrap-status.json` will become a visible error rather than a silent zero in the dashboard.
- Follow-up opportunities (not blocking): add dedicated tests for the L774 host-agent-unreachable path and the L1351/L1411 JSON-decode fallthrough; add a one-line comment at L1449 documenting why `_poll_service_health`'s broad except stays.

## Platform Impact
- macOS: not affected — `dashboard-api` runs in Docker on macOS; exception narrowing is platform-agnostic Python.
- Linux: not affected — same reasoning.
- Windows (WSL2): not affected — same reasoning.

### DIFF
--- a/dream-server/extensions/services/dashboard-api/main.py
+++ b/dream-server/extensions/services/dashboard-api/main.py
@@ -771,7 +771,7 @@ def _check_host_agent_available() -> bool:
     try:
         with urllib.request.urlopen(f"{AGENT_URL}/health", timeout=3) as response:
             return response.status == 200
-    except Exception:
+    except (urllib.error.URLError, urllib.error.HTTPError, OSError):
         return False
 
 
@@ -1093,13 +1093,14 @@ async def status(api_key: str = Depends(verify_api_key)):
 async def api_status(api_key: str = Depends(verify_api_key)):
     """Dashboard-compatible status endpoint.
 
-    Wrapped in a top-level try/except so that a transient failure in any
-    sub-call (GPU, health checks, llama metrics …) never returns a raw 500
-    to the dashboard — the frontend would flash "0/17" otherwise.
+    Catches transient I/O failures from sub-calls (GPU, health checks,
+    llama metrics …) and returns a safe fallback. Programming errors
+    (AttributeError, KeyError, TypeError) propagate so they surface in
+    tests instead of being masked.
     """
     try:
         return await _build_api_status()
-    except Exception:
+    except (asyncio.TimeoutError, OSError):
         logger.exception("/api/status handler failed — returning safe fallback")
         return {
             "gpu": None, "services": [], "model": None,
@@ -1348,7 +1349,7 @@ async def api_settings_env_save(
         try:
             err_payload = json.loads(exc.read().decode("utf-8"))
             detail = err_payload.get("error", detail)
-        except Exception:
+        except (json.JSONDecodeError, UnicodeDecodeError, AttributeError):
             pass
         raise HTTPException(status_code=503, detail={"message": detail}) from exc
     except urllib.error.URLError as exc:
@@ -1408,7 +1409,7 @@ async def api_settings_env_apply(
         try:
             payload = json.loads(exc.read().decode("utf-8"))
             detail = payload.get("error", detail)
-        except Exception:
+        except (json.JSONDecodeError, UnicodeDecodeError, AttributeError):
             pass
         raise HTTPException(status_code=503, detail={"message": detail}) from exc
     except urllib.error.URLError as exc:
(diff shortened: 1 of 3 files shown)


Lightheartdevs (dismissed):
Narrows two broad `except Exception` handlers in main.py: `_check_host_agent_available` → `(URLError, HTTPError, OSError)`; `api_status` → `(asyncio.TimeoutError, OSError)`. Programming errors (AttributeError, KeyError, TypeError) now propagate to surface in tests instead of being masked. Direct application of CLAUDE.md rule 1. Ship after rebase + API tests.

Lightheartdevs:
Holding this during the merge pass because the live pi check is failing. I am not merging PRs with a current red API check; please fix/rebase and rerun CI, then we can re-review.

Lightheartdevs (approved):

Audit notes:
- Reviewed the narrowed exception boundaries in `main.py`; the production behavior matches the intended contract: transient I/O errors still fall back, programming/configuration errors now surface instead of becoming 200-with-zeros.
- Fixed the stale failing API tests: `test_runtime_error_propagates_as_500` now uses a no-raise TestClient, and the router smoke test mocks `_build_api_status` so it does not depend on an intentionally incomplete test registry.
- Ran `python -m py_compile main.py tests/test_main.py tests/test_routers.py`.
- Ran targeted pytest for `TestApiStatusFallback` and `test_api_status_authenticated`: 3 passed locally.
- Full local pytest timed out / was polluted by this Windows Python 3.13 environment missing async pytest support; refreshed GitHub Actions are green, including the API job and mypy.

PR #1114, opened 2026-05-02
fix(dashboard-api): make /api/update/dry-run non-blocking via httpx

yasinBursali:
## What
Replace synchronous `urllib.request.urlopen` in `get_update_dry_run` with
an async `httpx.AsyncClient` call, eliminating a blocking I/O hazard in a
FastAPI async handler.

## Why
`get_update_dry_run` is declared `async def` but contained a synchronous
`urllib.request.urlopen` call with an 8-second timeout. Any GitHub
rate-limit or network failure blocked the FastAPI event loop for the full
timeout duration, stalling all concurrent async requests — GPU stats
polling, service health checks, and log streaming — until the DNS/TCP
timeout elapsed.

## How
- `routers/updates.py`: replaced the `urllib.request.Request` +
  `urlopen` block with `async with httpx.AsyncClient(timeout=8.0)` +
  `await client.get(...)`. The `resp.json()` call is placed outside the
  `async with` block, consistent with the `_refresh_release_cache` call
  site in the same file (httpx buffers the body before context exit).
  Reuses the existing module-level `_GITHUB_HEADERS` constant instead of
  an inline dict. The two inline `import urllib.*` statements at the top
  of the function body are removed; they had no other consumers.
  Exception tuple updated from `urllib.error.*` types to `httpx.HTTPError,
  httpx.TimeoutException` — same realistic failure surface.
- `tests/test_updates.py`: 5 dry-run tests updated from
  `patch("urllib.request.urlopen", side_effect=OSError(...))` (now a dead
  stub — urllib is no longer called) to
  `patch("routers.updates.httpx.AsyncClient.get", side_effect=httpx.ConnectError(...))`.
  `httpx.ConnectError` is a subclass of `httpx.HTTPError`, so it is caught
  by the new exception tuple.

## Testing
- Automated: 19/19 tests in `test_updates.py` pass after migration.
- Manual: hit `/api/update/dry-run` while concurrent dashboard requests are
  in flight (GPU stats, service health). With an `iptables` block on
  outbound 443, concurrent requests should complete in <500ms; the dry-run
  itself returns after the httpx connect timeout (~2s default) rather than
  blocking the loop.

## Review
Critique Guardian: APPROVED WITH WARNINGS (informational only — CG confirmed proceed).
- Optional polish: adding `resp.raise_for_status()` before `resp.json()` would restore
  strict 4xx/5xx parity with urllib's old behavior (urllib raised `HTTPError` on non-2xx;
  httpx by default does not). Not done — `httpx.HTTPStatusError` is a subclass of
  `httpx.HTTPError` and is already caught by the exception tuple; practical effect on
  the response payload is unchanged (version fields default, `version_check_error` populates).
- Pre-existing test gap: no positive-path test asserting `latest_version` populates
  correctly when GitHub returns 200 with valid JSON. This is a pre-existing gap on
  upstream/main, not introduced here.

## Known Considerations
- The `resp.raise_for_status()` omission is intentional for now: the existing logic
  already degrades gracefully on any exception (fills `version_check_error`, returns
  200 with default version fields). A follow-up can add strict status checking if desired.
- No positive-path test coverage for the 200 success branch — pre-existing gap, noted
  for future enhancement.

## Platform Impact
- macOS: affected (dashboard-api runs in Docker on all platforms; this handler fires on
  the `/api/update/dry-run` endpoint regardless of GPU backend)
- Linux: affected — primary fix target
- Windows (WSL2): affected — same dashboard-api container path

### DIFF
--- a/dream-server/extensions/services/dashboard-api/routers/updates.py
+++ b/dream-server/extensions/services/dashboard-api/routers/updates.py
@@ -200,9 +200,6 @@ async def get_update_dry_run():
     that the update process reads or writes.  No containers are started,
     stopped, or re-created.
     """
-    import urllib.request
-    import urllib.error
-
     install_path = Path(INSTALL_DIR)
 
     # ── current version ───────────────────────────────────────────────────────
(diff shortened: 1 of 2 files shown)


Lightheartdevs (approved):
Re-audited. The change removes synchronous urllib I/O from an async FastAPI handler and replaces it with httpx.AsyncClient, which is the right fix for event-loop blocking in /api/update/dry-run. Local verification: py_compile passed and the changed dry-run tests passed locally (6 passed, 13 deselected). The full file has Windows-local failures around cp1252 source reading and executing .sh files as native Windows processes, but the live Linux/API CI is green.

PR #1115, opened 2026-05-02
fix(dashboard-api): allow built-in extensions to declare init-time root user

yasinBursali:
## What
Adds `skip_root_user_check: bool = False` to `_scan_compose_content` so
built-in extensions can declare `user: "0:0"` for a legitimate init-time
chown + privilege-drop pattern without being rejected by the security scanner.

## Why
`_scan_compose_content` rejects any compose service that declares
`user: "0:0"` or `user: "root"`. The openclaw built-in uses this pattern
legitimately: its entrypoint runs `chown -R 1000:1000 /home/node/.openclaw`
to fix bind-mount ownership, then immediately drops to uid/gid 1000 via
`setpriv --reuid=1000 --regid=1000 --clear-groups`. The container also
declares `security_opt: no-new-privileges: true`, so the chown step cannot
escalate. The existing scanner had no mechanism to permit this pattern for
repo-shipped built-ins while continuing to reject it for user and library
extensions.

## How
Mirrors the existing `skip_name_collision` and `skip_gpu_passthrough_check`
parameter pattern already present on `_scan_compose_content`. The new
`skip_root_user_check: bool = False` parameter is passed as `is_builtin` at
both internal call sites (`_activate_service` and `enable_extension`), where
`is_builtin` is the pre-existing
`ext_dir.is_relative_to(EXTENSIONS_DIR.resolve())` gate. User-installed and
library extensions default to `skip_root_user_check=False` — the root-user
check is unchanged for them.

Files changed:
- `routers/extensions.py` — new parameter + guard + updated docblock on both
  call sites (+16/-9)
- `tests/test_extensions.py` — new `TestScanComposeSkipRootUserCheck` class
  (+53)

## Testing
- Automated: py_compile clean on both files; new test class
  `TestScanComposeSkipRootUserCheck` follows the established skip-class
  pattern (positive: skip=True accepts root user; negative: default rejects;
  regression-guard: skip=True does NOT bypass the `privileged: true` check);
  `test_install_allows_library_build_context` still passes (install path is
  unchanged — `trusted=True` retained at the library install call site).
- Manual: Enable openclaw via the dashboard Extensions page on a clean install;
  confirm the extension activates without a 400 "runs as root" error. On Linux
  (NVIDIA or CPU tier), verify `docker inspect openclaw` shows the container
  running as uid 1000 after startup.

## Review
Critique Guardian: APPROVED. No required-before-merge items. Informational
only: an optional ergonomics follow-up — a one-line comment in
`extensions/services/openclaw/compose.yaml` near `user: "0:0"` documenting
the init-time chown + setpriv handoff and the `skip_root_user_check` trust
boundary would help future readers. Pure documentation; not blocking.

## Known Considerations
This PR is intentionally scoped narrowly to the built-in root-user-check
exemption. A related broader question about tightening the `trusted` parameter
at the library install path was split out and deferred for separate review.
That work is not part of this PR and is not a prerequisite for merge.

`is_builtin` derives from `EXTENSIONS_DIR.resolve()`, which canonicalises
symlinks before comparison. A symlink-into-builtins path substitution would
require pre-existing write access to the install tree — outside the relevant
threat model.

## Platform Impact
- macOS: affected — extension scanning runs on all platforms; the exemption
  applies equally to Apple Silicon installs.
- Linux: affected — primary target platform; openclaw is Linux-deployable.
- Windows (WSL2): affected — same Python code path runs under WSL2.

### DIFF
--- a/dream-server/extensions/services/dashboard-api/routers/extensions.py
+++ b/dream-server/extensions/services/dashboard-api/routers/extensions.py
@@ -303,6 +303,7 @@ def _scan_compose_content(
     trusted: bool = False,
     skip_name_collision: bool = False,
     skip_gpu_passthrough_check: bool = False,
+    skip_root_user_check: bool = False,
 ) -> None:
     """Reject compose files containing dangerous directives."""
     try:
@@ -395,12 +396,13 @@ def _scan_compose_content(
                 status_code=400,
                 detail=f"Service '{svc_name}' uses host user namespace",
             )
-        user = svc_def.get("user")
-        if user is not None and str(user).split(":")[0] in ("root", "0"):
-            raise HTTPException(
-                status_code=400,
-                detail=f"Service '{svc_name}' runs as root",
-            )
+        if not skip_root_user_check:
+            user = svc_def.get("user")
+            if user is not None and str(user).split(":")[0] in ("root", "0"):
+                raise HTTPException(
+                    status_code=400,
+                    detail=f"Service '{svc_name}' runs as root",
+                )
         if not trusted and "build" in svc_def:
             raise HTTPException(
                 status_code=400,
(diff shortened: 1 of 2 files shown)


Lightheartdevs (approved):
Adds `skip_root_user_check` parameter to `_scan_compose_content`, defaulting to `False` so user/library extensions still reject root users. Built-in extensions (resolved under EXTENSIONS_DIR) can opt in via `skip_root_user_check=True` — required for OpenClaw-style init-time `user: "0:0"` followed by privilege drop. Codex's pytest run shows 3/3 passing on `TestScanComposeSkipRootUserCheck`. Ship after rebase + API tests.

Lightheartdevs (approved):
Re-audited for merge. The new skip_root_user_check path is only applied for built-in extensions during activation/enable, so user/library extensions still reject user: 0:0. The added tests cover built-in root allowance, user extension rejection, and privileged still being blocked under the exemption. Local py_compile passed; broad test_extensions run had 183 passes and only the known Windows symlink-privilege failures, while current CI is green.

PR #1116, opened 2026-05-02
refactor(host-agent): migrate systemd unit from user-mode to system-mode

yasinBursali:
## What
Migrate the dream-host-agent systemd unit from user-mode (`systemctl --user`) to system-mode (`/etc/systemd/system/`) with `User=<installer-user>` and `SupplementaryGroups=docker`.

## Why
User-mode systemd (PID 1 per-user manager) spawns at login with a frozen supplementary group set. When Phase 05 runs `usermod -aG docker $USER`, the change takes effect for future login sessions but the user manager is never re-exec'd — so the host-agent process inherits a stale group set and lacks docker socket access at runtime. `loginctl enable-linger` keeps the unit alive past logout but does not solve the group-inheritance race; the group set is still frozen at the moment the user manager first started.

A system-mode unit with `User=<installer-user>` and `SupplementaryGroups=docker` makes docker group membership explicit at unit-load time, completely bypassing the group-inheritance race.

## How
**`dream-server/scripts/systemd/dream-host-agent.service`** — added `User=__INSTALL_USER__` and `SupplementaryGroups=docker` to `[Service]`; flipped `WantedBy` from `default.target` to `multi-user.target` (system-mode target).

**`dream-server/installers/phases/07-devtools.sh`** — replaced the `--user`-mode install block with a system-mode block. Key decisions:
- Detection guard: `systemctl status >/dev/null 2>&1 || [[ -d /run/systemd/system ]]` (covers WSL2-with-systemd and standard Linux).
- Idempotent migration: stops, disables, and removes any pre-existing `~/.config/systemd/user/dream-host-agent.service`, emitting `✓ Migrated host agent from --user mode to system mode`.
- Hard-fail sudo guard in non-interactive mode (mirrors Phase 05 pattern).
- User resolution: `${INSTALL_USER:-${SUDO_USER:-$(whoami)}}` — INSTALL_USER wins (test harnesses), then SUDO_USER (the documented `sudo bash install.sh` path), then `whoami` as final fallback.
- Root-user guard: emits `ai_warn` (not `error`) so appliance installs are flagged but not blocked.
- `__INSTALL_USER__` placeholder substituted alongside existing placeholders in the same sed loop.
- `sudo install -m 644` to `/etc/systemd/system/`.
- All `systemctl` invocations use the `if sudo X 2>&1 | tee -a "$LOG_FILE" >/dev/null; then ... else ... fi` form (pipefail-aware; shellcheck SC2024-clean — avoids sudo redirection antipattern).
- `loginctl enable-linger` removed for the host-agent block; preserved for opencode-web and memory-shepherd which remain user-mode.

**`dream-server/dream-cli`** — `cmd_agent` start/stop/status-detection: `systemctl --user X` → `sudo systemctl X` (3 lines). Detection condition updated to match the installer guard (`systemctl status || [[ -d /run/systemd/system ]]`).

**`dream-server/dream-uninstall.sh`** — existing user-mode cleanup block preserved (handles migration-leftover case). Appended idempotent system-unit cleanup: `is-enabled`-gated `disable --now`, `rm -f /etc/systemd/system/dream-host-agent.service`, `daemon-reload`.

## Testing
- Automated: shellcheck clean (warning-parity with baseline); `make lint` passes; no SC2024 warnings introduced.
- Manual (live install on CachyOS sandbox, `--all --non-interactive` via `sudo bash install.sh`):
  - System-mode unit installed at `/etc/systemd/system/dream-host-agent.service`
  - Unit active and running (`Active: active (running)`)
  - `User=yasin` confirmed in unit status (resolved via SUDO_USER — correct for the `sudo bash install.sh` path)
  - `SupplementaryGroups=docker` present in unit file
  - `WantedBy=multi-user.target` confirmed; cgroup under `/system.slice/dream-host-agent.service`
  - Migration block ran cleanly with `✓ Migrated host agent from --user mode to system mode` log line (pre-existing user-mode unit was removed)
  - Port 7710 listening on 172.17.0.1 (docker bridge)
  - `/health` returns `{"status":"ok","version":"1.0.0"}`
  - `dream agent stop` and `dream agent start` work via cached sudo
  - Full stack came up (19+ containers healthy)
  - Dashboard HTTP 200 on :3001

## Review
Critique Guardian: APPROVED WITH WARNINGS (informational only). No required-before-merge items.

## Known Considerations

**Behavior change:** Non-interactive Linux installs without passwordless sudo now hard-fail at Phase 07. This mirrors the existing Phase 05 pattern — any CI flow that already passes Phase 05 will pass Phase 07.

**Pre-existing SIGTERM-handler issue surfaced (not introduced by this PR):** `dream agent stop` may end in `Active: failed (Result: timeout)` instead of a clean `inactive` state. The host-agent SIGTERM handler at `bin/dream-host-agent.py` blocks on `server.shutdown()` (a `ThreadingHTTPServer` that waits for in-flight requests); CachyOS systemd `DefaultTimeoutStopUSec=10s` escalates to SIGKILL after 10s. The identical handler exists on `upstream/main` — this is a pre-existing issue, not introduced by this change. The service still stops; just not cleanly. Filed as a separate follow-up for tracking.

**Follow-up issues to file post-merge:**
- Pre-start port-7710 collision check (prevent silent bind failure on reinstall).
- `dream-cli` non-tty sudo UX improvement: `sudo -n` probe + clear error message when cached credentials have expired.
- BATS coverage for the migration path (coordinate with PR #1102's test framework).
- Optional `unset _agent_user` at end of the install block for shell hygiene.

## Platform Impact
- **macOS Apple Silicon:** NOT AFFECTED — `installers/macos/install-macos.sh` is untouched; macOS uses launchd for the host agent separately.
- **Linux:** PRIMARY TARGET. All Linux installs migrate to the system-mode unit on next install. Any pre-existing `--user` unit is cleaned by the migration block.
- **Windows (WSL2):** AFFECTED (compatible). The `systemctl status || [[ -d /run/systemd/system ]]` guard covers WSL2-with-systemd. WSL2 without systemd falls through to `ai_warn "No systemd detected"` — same graceful behavior as before this change.

### DIFF
--- a/dream-server/dream-cli
+++ b/dream-server/dream-cli
@@ -2521,7 +2521,7 @@ cmd_agent() {
     case "$(uname -s)" in
         Darwin) daemon_type="launchd" ;;
         Linux)
-            if systemctl --user status >/dev/null 2>&1; then
+            if systemctl status >/dev/null 2>&1 || [[ -d /run/systemd/system ]]; then
                 daemon_type="systemd"
             fi
             ;;
@@ -2546,7 +2546,7 @@ cmd_agent() {
                         success "Agent started (LaunchAgent)" || warn "Agent start failed"
                     ;;
                 systemd)
-                    systemctl --user start dream-host-agent.service 2>/dev/null && \
+                    sudo systemctl start dream-host-agent.service 2>/dev/null && \
                         success "Agent started (systemd)" || warn "Agent start failed"
                     ;;
                 *)
@@ -2572,7 +2572,7 @@ cmd_agent() {
                         success "Agent stopped" || warn "Agent not running"
                     ;;
                 systemd)
-                    systemctl --user stop dream-host-agent.service 2>/dev/null && \
+                    sudo systemctl stop dream-host-agent.service 2>/dev/null && \
                         success "Agent stopped" || warn "Agent not running"
                     ;;
                 *)
--- a/dream-server/dream-uninstall.sh
+++ b/dream-server/dream-uninstall.sh
@@ -126,6 +126,14 @@ for unit in opencode-web.service openclaw-session-cleanup.timer \
     fi
 done
 systemctl --user daemon-reload 2>/dev/null || true
+
+# Remove system-mode dream-host-agent unit (migrated from --user mode).
+# Idempotent — no-op if the unit was never installed (e.g. older user-mode installs).
+if systemctl is-enabled dream-host-agent.service >/dev/null 2>&1; then
+    sudo systemctl disable --now dream-host-agent.service 2>/dev/null || true
+fi
+sudo rm -f /etc/systemd/system/dream-host-agent.service 2>/dev/null || true
+sudo systemctl daemon-reload 2>/dev/null || true
 log_ok "Systemd services removed"
 
 # 3. Remove CLI symlink
--- a/dream-server/installers/phases/07-devtools.sh
+++ b/dream-server/installers/phases/07-devtools.sh
@@ -220,49 +220,99 @@ OPENCODE_EOF
 fi
 
 # ── Dream Host Agent (extension lifecycle management) ──
+# System-mode systemd unit (was --user mode pre-#573). Installs to
+# /etc/systemd/system, runs as the installing user with SupplementaryGroups=docker
+# so the agent can manage Docker without socket-mounting into a container.
 if [[ -f "$INSTALL_DIR/bin/dream-host-agent.py" ]]; then
     AGENT_PYTHON="$(command -v python3)"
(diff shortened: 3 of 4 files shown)


Lightheartdevs (approved):
Re-audited the system-mode migration. This is a real fix for the user systemd stale supplementary-group race: the unit now runs as the installer user with SupplementaryGroups=docker, the installer removes the old user-mode unit idempotently, and uninstall cleans up both paths. Local verification: bash syntax passed for dream-cli, dream-uninstall.sh, and Phase 07; the live CI matrix is green. The sudo UX/follow-up items are real but documented and not merge blockers.

PR #1082, opened 2026-05-01
harden(resolver): boundary-check compose_file against extension dir

yasinBursali:
## What
Hardens the inline-Python extension loop in `scripts/resolve-compose-stack.sh` against two related vulnerabilities in the manifest's `compose_file` field:
1. `..`-traversal: a malicious user-installed manifest can point `compose_file` at a host file outside its own extension directory (`/etc/passwd`, etc.) and the resolver will include it in the `-f` flags passed to `docker compose`.
2. Absolute path: `compose_file: /etc/shadow` triggers an unhandled `ValueError` in `Path.relative_to()` that propagates to the existing `else: raise` branch, crashing the resolver (DoS).

## Why
`Path.relative_to()` uses **lexical part-prefix matching**, NOT path resolution. A path like `service_dir/../../../../../../etc/passwd` retains its `..` segments unresolved; `relative_to(script_dir)` still finds `script_dir`'s components at the front and returns successfully. The traversal segments persist, `compose_path.exists()` follows them, and the resolved file is appended to `COMPOSE_FLAGS`. Trust boundary: built-in extensions ship from this repo (low risk), but post-PR #909/#926/#907 user-installed extensions widen the surface considerably.

## How
Adds a `try: compose_path.resolve().relative_to(service_dir.resolve())` boundary check between the join and the existence check:
```python
compose_path = service_dir / compose_rel
try:
    compose_path.resolve().relative_to(service_dir.resolve())
except ValueError:
    print(f"WARNING: {service_dir.name}: compose_file '{compose_rel}' "
          f"escapes the extension directory; skipping", file=sys.stderr)
    continue
if compose_path.exists():
    resolved.append(str(compose_path.relative_to(script_dir)))
```
- **Resolves both sides** to handle macOS `/var → /private/var` symlink dance (an unresolved `script_dir` would false-positive-reject legitimate `/var/folders/...` paths)
- **Bound to `service_dir`** (stricter than `script_dir`) — each extension's compose must live in its own directory
- **Narrow `try/except`** — catches only `ValueError`, no Exception swallowing
- **Emit unchanged** — `relative_to(script_dir)` still produces the relative path docker-compose expects

## Testing
- `tests/test-resolve-compose-resilient.sh`: 13/13 assertions pass (9 prior + 4 new — traversal rejection + WARNING + absolute-path no-crash + WARNING)
- Negative test: removing the boundary check causes 2 WARNING-assertion tests to fail
[...]

PR #1118, opened 2026-05-02
harden(resolver): scan user-extension and override compose content + boundary-check compose_file

yasinBursali:
## What
Three layered hardenings of `dream-server/scripts/resolve-compose-stack.sh`:

1. **Boundary check on `compose_file`** in the user-extension loop — a malicious manifest with `compose_file: ../../../etc/passwd` can no longer make the resolver include host files in the stack.
2. **Content scan for user-extension composes** — mirrors the dashboard-api `_scan_compose_content` scan (privileged, host network/PID/IPC, docker.sock mount, root user, dangerous capabilities, GPU passthrough, non-loopback port bindings, named-volume bind-mount backdoors, build: directive). Closes the bypass where a compose dropped under `data/user-extensions/` outside the install API was never scanned.
3. **Same scan applied to `docker-compose.override.yml`** — operator-placed but may carry hostile content from cloned repo or restored backup.

Plus three small user-ext-loop alignments needed by the above:
- Module-scope yaml import (shared with built-in loop).
- `gpu_backends` manifest filter on user extensions (mirrors built-in predicate).
- Apple Silicon guard on the user-ext `compose.local.yaml` overlay.

## Why
- **Boundary check**: defense against path-traversal in a (currently hypothetical) malicious user extension's manifest. Cheap; matches the boundary-check pattern Light-Heart-Labs/DreamServer#1082 introduces for the built-in loop.
- **Content scan at resolver layer**: dashboard-api's `_scan_compose_content` runs at install time only. The resolver runs on every `dream` invocation. A compose dropped into `data/user-extensions/` outside the install API (manual placement, restored backup, cloned-repo deployment) would otherwise bypass content scanning entirely. Closing this gap moves the scan to a TOCTOU-safe location.
- **Override.yml**: same threat surface as user extensions; auto-included by the resolver if present; same scan applies for the same reasons.

## How
- New `_scan_user_compose_content` helper, structurally identical to dashboard-api's scanner (without the FastAPI HTTPException dependency). Returns `(ok, warnings)`.
- User-ext loop: resolves manifest's `compose_file` through `path.resolve().relative_to(service_dir.resolve())` and rejects on `ValueError`. Then scans content; rejects on any dangerous directive.
- GPU overlay (`compose.<gpu>.yaml`) gets the same scan since its filename is fixed but its content isn't.
- Override.yml gets a final scan pass before being added to the resolved stack.
- Tests: `tests/test-resolve-compose-resilient.sh` extended with new cases — traversal rejection, content rejection per dangerous directive, override.yml positive and negative cases.

## Testing
- `bash dream-server/scripts/resolve-compose-stack.sh --script-dir <fixture> --tier 1 --gpu-backend nvidia --skip-broken` against the new traversal/content fixtures (covered by the test-suite extensions).
- `bash dream-server/tests/test-resolve-compose-resilient.sh` — new test cases.
- Full `make test` parity expected once dependency PR lands.

## Review
Pre-submission verifier ✅ confirmed (orchestrator's deep-evaluation pass on 2026-05-02).

This PR has not yet gone through the full Critique Guardian cycle for the consolidated diff. The `_scan_user_compose_content` content-scan helper is the largest surface and warrants careful review of its check-list against the dashboard-api equivalent to ensure parity.

## Known Considerations

**⚠ This PR must merge AFTER `Light-Heart-Labs/DreamServer#1082`** (`harden(resolver): boundary-check compose_file against extension dir`). Reasons:

1. PR #1082 introduces the same `compose_file` boundary-check pattern but for the **built-in extension loop**. This PR applies the same pattern to the **user-extension loop**.
2. Both PRs touch the same two files (`resolve-compose-stack.sh` and `tests/test-resolve-compose-resilient.sh`). The diffs overlap in two places:
   - `resolve-compose-stack.sh` near the top: this PR's squashed commit chain includes a duplicate of #1082's built-in-loop boundary-check (the layered-fix history was developed on top of #1082).
   - `tests/test-resolve-compose-resilient.sh`: both PRs add new test cases at the same line range.

**Resolution path:**
- Merge #1082 first (it's the smaller, narrower fix).
- Maintainer or PR author rebases this branch on the post-#1082 main; the rebase will collapse the duplicate boundary-check content cleanly.
- The user-extension hardening (the unique L11a contribution) lands on top.

**This PR is opened as DRAFT to make this dependency mechanically explicit** — GitHub will not allow merging a draft, so the merge order is enforced by the draft state.

## Platform Impact
- **macOS Apple Silicon**: affected — Apple Silicon guard on user-ext `compose.local.yaml` is a new branch in the user-ext path; the existing macOS install path (which uses native llama-server, not Docker) gets correct skip semantics.
- **Linux**: primary target — all three hardenings apply.
- **Windows (WSL2)**: same as Linux — Python-based resolver, no platform branch.

### DIFF
--- a/dream-server/installers/phases/03-features.sh
+++ b/dream-server/installers/phases/03-features.sh
@@ -134,7 +134,7 @@ fi
 # which were just renamed to .disabled.
 if [[ -x "$SCRIPT_DIR/scripts/resolve-compose-stack.sh" ]]; then
     _refreshed_flags=$("$SCRIPT_DIR/scripts/resolve-compose-stack.sh" \
-        --script-dir "$SCRIPT_DIR" --tier "${TIER:-1}" --gpu-backend "${GPU_BACKEND:-nvidia}" 2>/dev/null) || true
+        --script-dir "$SCRIPT_DIR" --tier "${TIER:-1}" --gpu-backend "${GPU_BACKEND:-nvidia}") || true
     if [[ -n "$_refreshed_flags" ]]; then
         COMPOSE_FLAGS="$_refreshed_flags"
         log "Compose flags refreshed after feature selection"
--- a/dream-server/scripts/resolve-compose-stack.sh
+++ b/dream-server/scripts/resolve-compose-stack.sh
@@ -144,6 +144,172 @@ try:
 except ImportError:
     yaml_available = False
 
+import re
+
+_LOOPBACK_VAR_DEFAULT_RE = re.compile(
+    r"^\$\{[A-Za-z_][A-Za-z0-9_]*:-127\.0\.0\.1\}$",
+)
+
+# Capabilities and security_opt strings that grant container escape primitives.
+_DANGEROUS_CAPS = {
+    "SYS_ADMIN", "NET_ADMIN", "SYS_PTRACE", "NET_RAW",
+    "DAC_OVERRIDE", "SETUID", "SETGID", "SYS_MODULE",
+    "SYS_RAWIO", "ALL",
+}
+_DANGEROUS_SECURITY_OPTS = {
+    "seccomp:unconfined", "apparmor:unconfined", "label:disable",
+}
+
+
+def _host_part_is_loopback(host):
+    if host == "127.0.0.1":
+        return True
+    return bool(_LOOPBACK_VAR_DEFAULT_RE.fullmatch(host))
+
+
+def _split_port_host(port_str):
+    """Mirror dashboard-api/_split_port_host: handle ${VAR:-127.0.0.1}: prefix."""
+    if port_str.startswith("${"):
+        end = port_str.find("}")
+        if end == -1 or end + 1 >= len(port_str) or port_str[end + 1] != ":":
+            return port_str, ""
+        return port_str[: end + 1], port_str[end + 2:]
+    if ":" not in port_str:
+        return None, port_str
+    host, _, rest = port_str.partition(":")
+    if host.isdigit():
+        return None, port_str
+    return host, rest
+
+
+def _scan_user_compose_content(compose_path):
+    """Reject compose fragments containing dangerous directives.
+
+    Mirrors dashboard-api/routers/extensions.py:_scan_compose_content (without
+    the FastAPI HTTPException dependency). Returns ``(ok, warnings)``: ``ok``
+    is False on any rejection, ``warnings`` is a list of human-readable
+    messages. User-extension contexts are always untrusted at the resolver
+    layer — no ``trusted=True`` exemption.
+    """
+    if not yaml_available:
+        return (False, [f"cannot scan {compose_path} — PyYAML unavailable; user-extension compose skipped for safety"])
+    try:
+        data = yaml.safe_load(compose_path.read_text(encoding="utf-8"))
+    except (yaml.YAMLError, OSError) as e:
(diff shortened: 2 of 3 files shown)


Lightheartdevs:
﻿Merge-pass recheck: this still cannot be merged because it is draft and currently conflicts with `main`.

The direction is valuable now that #1082 has landed, but the branch needs the dependency cleanup described in the PR body: rebase onto current `main`, drop any duplicated #1082 content, resolve the `resolve-compose-stack.sh` / `test-resolve-compose-resilient.sh` conflicts, then undraft for a fresh review. The current CI is green, but the conflict/draft state is the blocker.

yasinBursali:
Pushed the merge-pass cleanup you asked for:

**Rebased onto current `main` (`d4124d6a`)**, dropped #1082-duplicated logic, resolved the `resolve-compose-stack.sh` / `test-resolve-compose-resilient.sh` conflicts. Branch is now 2 commits ahead of `upstream/main` with no stranger commits.

**Three corrections fold into the new fixup commit (`0f2d78ef`):**

1. Extended `_scan_user_compose_content` to `compose.local.yaml` and `compose.multigpu.yaml` under user extensions — PR's original scan reached `compose.yaml` and `compose.<gpu>.yaml` but missed those two overlay paths, so a malicious user-ext could bypass via the default-mode (`local`) overlay.
2. Reverted the silent PyYAML `try/except` to a hard `import yaml`. On systems without PyYAML the scan would silently return False and user extensions would disappear from the stack with no UI feedback. Loud `ImportError` is the right failure mode.
3. **Added the `CORE_SERVICE_IDS` name-collision check** the docstring already claimed parity for. Loads `config/core-service-ids.json` at module init with a hardcoded fallback matching `helpers.CORE_SERVICE_IDS_FALLBACK`. Without this a hand-dropped or backup-restored extension declaring services named e.g. `dashboard-api` could shadow core services in the merge — caught and verified in tests.

`tests/test-resolve-compose-resilient.sh` now passes 29/29 (was 27 — added 2 assertions for the core-service-collision path against a synthetic `shadow-core` user extension).

Behavioral notes for review:
- `docker-compose.override.yml` is now content-scanned. Operators with absolute-host-path bind-mounts in their override.yml will see them rejected; matches the user-ext contract. No paired test for that specific rejection — happy to add if you'd prefer it explicit.
- Hard-fail on missing PyYAML: Phase 03's `|| true` keeps the install path unaffected (falls back to phase-02 flags), but stderr now surfaces the `ImportError`.

Ready for fresh review.

Lightheartdevs (approved):

Audit notes:
- Reviewed resolver-layer scanning against the dashboard install-time scanner and checked the user-extension, overlay, override.yml, compose_file boundary, GPU backend, and core service collision paths.
- Found and fixed two issues before merge:
  - Test assertions were path-separator-sensitive and could miss Windows-style resolver output; added `contains_path` normalization.
  - Long-form Compose bind mounts (`type: bind` / `source: /...`) bypassed the new resolver scan; added rejection plus a regression fixture.
- Ran `git diff --check`.
- Ran `bash -n` for `resolve-compose-stack.sh` and `test-resolve-compose-resilient.sh`.
- Ran `bash dream-server/tests/test-resolve-compose-resilient.sh`: 31/31 passed locally.
- Refreshed GitHub Actions are green, including ShellCheck and integration-smoke.

PR #1120, opened 2026-05-02
fix(forge): defense-in-depth — disable bundled cloud-tunnel processes

yasinBursali:
## What
Extend `SUPERVISOR_NO_AUTOSTART` in forge `compose.yaml` from `syncthing` to `syncthing,quicktunnel,serviceportal,sshd`. Add a Privacy & Defense-in-depth section to forge's README.

## Why
The upstream `ai-dock` Forge image bundles four processes that autostart by default and would otherwise expose:
- `syncthing` — public-relay outbound + inbound listener (already disabled by Light-Heart-Labs/DreamServer#1077)
- `quicktunnel` — outbound cloudflare tunnel
- `serviceportal` — inbound HTTP UI listing services (port 1111/11111)
- `sshd` — inbound SSH on `${SSH_PORT}`

Light-Heart-Labs/DreamServer#1077 disables only `syncthing`. This PR closes the wider attack surface for default installs.

## How
Single env-var value extension in `compose.yaml`. README gains a 6-line Privacy section that documents the bundled processes, the override path, and the `cloudflared` carve-out.

`cloudflared` is excluded from the disable list — it is gated on `CF_TUNNEL_TOKEN` upstream and is a no-op without the token.

## Testing
- yamllint clean
- `docker compose -f resources/dev/extensions-library/services/forge/compose.yaml config --quiet` exit 0
- Manual: `docker exec dream-forge supervisorctl status` should show `sshd`, `cf_quicktunnel`, `serviceportal`, `syncthing` as STOPPED  
  (Note the supervisorctl output displays `quicktunnel` as `cf_quicktunnel` — that's the program name in `quicktunnel.conf`.)

## Review
Critique Guardian: APPROVED. Mechanism verified against ai-dock `init.sh` (env-var keys off conf filenames; the disabled list is exactly the inbound-listening / cloud-relay subset of ai-dock's own `SERVERLESS=true` defaults).

## Known Considerations
- `caddy` (the in-container reverse proxy serving forge itself) is intentionally NOT disabled.

## Platform Impact
- macOS: Docker-only image; identical behavior on Docker Desktop.
- Linux: GPU-accelerated extension, primary target. NVIDIA only.
- Windows: WSL2 Docker Desktop, identical.

Must merge after Light-Heart-Labs/DreamServer#1077.

### DIFF
--- a/resources/dev/extensions-library/services/forge/README.md
+++ b/resources/dev/extensions-library/services/forge/README.md
@@ -2,6 +2,14 @@
 
 The original Stable Diffusion web UI (based on Automatic1111). Features extensive model support, extensions, inpainting, outpainting, and professional-grade image generation capabilities.
 
+## Privacy & Defense-in-depth
+
+The upstream `ai-dock` Forge image bundles `syncthing`, `quicktunnel`, `serviceportal`, and `sshd` for cloud workflows. These are autostarted by the image's bundled supervisord and would otherwise expose the host to remote relays, tunnel endpoints, and inbound SSH on default install.
+
+DreamServer disables them via `SUPERVISOR_NO_AUTOSTART=syncthing,quicktunnel,serviceportal,sshd` in `compose.yaml`. To re-enable any of them, override that env var in your `.env` or a compose override.
+
+`cloudflared` is also bundled but only autostarts when `CF_TUNNEL_TOKEN` is set, so it stays a no-op by default.
+
 ## Requirements
 
 - **GPU:** NVIDIA (min 8 GB VRAM)
--- a/resources/dev/extensions-library/services/forge/README.md
+++ b/resources/dev/extensions-library/services/forge/README.md
@@ -4,9 +4,9 @@ The original Stable Diffusion web UI (based on Automatic1111). Features extensiv
 
 ## Privacy & Defense-in-depth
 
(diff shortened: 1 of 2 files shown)


Lightheartdevs:
Holding this during the merge pass because GitHub reports the branch as conflicting with current main. Please rebase/resolve conflicts, then we can re-review.

yasinBursali:
Rebased onto current `main` (`d4124d6a`); merge conflicts resolved. Branch is now 2 commits ahead of `upstream/main`.

The follow-up commit (`f24bf689`) adds `jupyter` to the `SUPERVISOR_NO_AUTOSTART` list — defense-in-depth: ai-dock Forge image autostarts a Jupyter notebook inside the container, and while there's no published port mapping for it (forge compose only exposes 7861), running unnecessary services is worth disabling. README documents all five suppressed services + why each one. Final value: `syncthing,quicktunnel,serviceportal,sshd,jupyter`.

Ready for fresh review.

Lightheartdevs (approved):

Audit notes:
- Verified the ai-dock base-image init logic uses `SUPERVISOR_NO_AUTOSTART` to disable matching supervisor conf filenames, and its serverless default covers the same service family.
- Found and fixed one mismatch before merge: README said `.env` can override the disable list, but the compose entry was hardcoded. I pushed `fix(forge): allow supervisor disable-list override` so `SUPERVISOR_NO_AUTOSTART=${SUPERVISOR_NO_AUTOSTART:-...}` keeps the secure default while allowing an intentional override.
- Ran `git diff --check`.
- Ran `docker compose -f resources/dev/extensions-library/services/forge/compose.yaml config --quiet`.
- Verified `SUPERVISOR_NO_AUTOSTART=sshd docker compose ... config` renders the override as expected.
- Refreshed GitHub Actions are green.

PR #1124, opened 2026-05-02
test(preflight): BATS fixtures for UFW/firewalld active-state warning

yasinBursali:
## What
Add `dream-server/tests/bats-tests/preflight-firewall.bats` covering the `FIREWALL_CHECK` introduced by Light-Heart-Labs/DreamServer#1076. Five test cases:

1. neither installed → pass
2. ufw installed but inactive → pass
3. ufw active → warn (mentions UFW)
4. firewalld active → warn (mentions firewalld)
5. both active → warn (UFW first-detected wins)

## Why
#1076 adds firewall-state preflight detection but ships without test coverage. Without a regression guard, a future change to the dual-path `if/elif` over `command -v ufw && systemctl is-active --quiet` could silently break detection.

## How
Black-box drives `scripts/linux-install-preflight.sh --json`; parses `FIREWALL_CHECK` from `.checks[]` via `jq`. PATH-prepended stub bin holds:
- a parameterised `systemctl` stub (reads `$SYSTEMCTL_ACTIVE_UNITS`)
- per-test `ufw` / `firewall-cmd` no-op stubs (presence-only) installed via a `make_stub` helper

Real tools (`python3`, `jq`, `df`, `awk`, `mktemp`) remain reachable because `$STUB_BIN` is PREpended to `$PATH`, not replacing it.

## Testing
- `bash dream-server/tests/run-bats.sh dream-server/tests/bats-tests/preflight-firewall.bats`: 5/5 PASS

## Review
Critique Guardian: APPROVED. Production message strings (`"No restrictive host firewall detected"`, `"UFW"`, `"firewalld"`) verified against #1076's `linux-install-preflight.sh`.

## Known Considerations
- Test 5 ("both active — UFW first-detected wins") locks in the production code's `if/elif` ordering.
- A 6th case for `firewall-cmd` installed-but-inactive would round out symmetry but is duplicative of the ufw-inactive case.

## Platform Impact
- macOS: BATS runs on macOS shell; tests target Linux preflight only (UFW/firewalld are Linux). No-op on macOS preflight.
- Linux: primary target.
- Windows: N/A.

Must merge after Light-Heart-Labs/DreamServer#1076. Until #1076 lands, the test will not find FIREWALL_CHECK in `--json` output and will fail.

### DIFF
--- a/dream-server/installers/phases/01-preflight.sh
+++ b/dream-server/installers/phases/01-preflight.sh
@@ -73,6 +73,19 @@ if [[ -n "$OPTIONAL_TOOLS_MISSING" ]]; then
     esac
 fi
 
+# Warn about host firewalls that commonly block LAN access to the dashboard or
+# WebUI. This is informational: loopback-only installs are fine, and users may
+# intentionally keep a firewall active.
+if command -v ufw >/dev/null 2>&1 && command -v systemctl >/dev/null 2>&1 && systemctl is-active --quiet ufw; then
+    warn "UFW is active and may block Dream Server dashboard/WebUI ports."
+    echo "  Allow the configured Dream Server ports in UFW or keep BIND_ADDRESS on loopback."
+elif command -v firewall-cmd >/dev/null 2>&1 && command -v systemctl >/dev/null 2>&1 && systemctl is-active --quiet firewalld; then
+    warn "firewalld is active and may block Dream Server dashboard/WebUI ports."
+    echo "  Allow the configured Dream Server ports in firewalld or keep BIND_ADDRESS on loopback."
+else
+    log "Host firewall: no active UFW/firewalld detected"
+fi
+
 # Check source files exist
 if [[ ! -f "$SCRIPT_DIR/docker-compose.yml" ]] && [[ ! -f "$SCRIPT_DIR/docker-compose.base.yml" ]]; then
     error "No compose files found in $SCRIPT_DIR. Please run from the dream-server directory."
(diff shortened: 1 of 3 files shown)


Lightheartdevs:
I re-checked this after #1076 landed. The added fixture is directionally good and only targets the new `FIREWALL_CHECK`, but the current CI failure is real for the commit GitHub tested: the branch/check ran without the #1076 implementation, so the BATS selector finds no `FIREWALL_CHECK` entry and all five status assertions return empty.

Please refresh/rebase this draft onto current `main` (or rerun after GitHub rebuilds the merge ref with #1076 included). I also tried the targeted BATS file locally after merging current `main`; the only local blocker was my Git Bash environment lacking `jq`, while CI has `jq`, so I don't see a test-logic issue beyond the branch needing to be current. Not mergeable while draft/red, but likely approveable once refreshed and green.

Lightheartdevs (approved):
Took this across the finish line since Yasin is no longer active.

The original test fixture was useful, but the branch was stale relative to the `FIREWALL_CHECK` implementation now on `main`. I pushed follow-up commits to the PR branch to keep the BATS fixture and add a matching installer-phase warning without duplicating the canonical structured `linux-install-preflight.sh` check from `main`.

Validation:
- `git diff --check`
- Bash syntax check for the touched shell scripts
- direct `linux-install-preflight.sh --json` firewall cases with stubbed `ufw`/`firewall-cmd`/`systemctl`
- targeted BATS fixture locally with a temporary `jq` shim: 5/5 passed
- GitHub Actions after the branch fix: all checks passing, including `integration-smoke`

Looks good to merge.

PR #1125, opened 2026-05-02
fix(openclaw): include host LAN IP in allowedOrigins when BIND_ADDRESS=0.0.0.0

yasinBursali:
## What
After Light-Heart-Labs/DreamServer#1080 drops `dangerouslyAllowHostHeaderOriginFallback`, LAN-mode installs (`BIND_ADDRESS=0.0.0.0`) lose Control UI access because `allowedOrigins` lists only `localhost` / `127.0.0.1`. `os.hostname()` inside the openclaw container returns the container ID, not the host LAN IP.

This PR makes the host LAN IP available to the container as `HOST_LAN_IP` env var, registers it in the env schema, detects it in the installer (Linux + macOS), and threads it through `inject-token.js` so allowedOrigins includes the LAN address when binding to `0.0.0.0`.

## Why
With #1080 alone, LAN access to the openclaw Control UI breaks for any operator who picked LAN mode at install time. This PR completes the LAN-mode path.

## How
**6 changes**:
- `inject-token.js` Part 1 + Part 3: append `http(s)://${HOST_LAN_IP}:${EXTERNAL_PORT}` to allowedOrigins when `HOST_LAN_IP` is non-empty; empty-string guard preserves localhost-only behavior.
- `compose.yaml`: pass `HOST_LAN_IP=${HOST_LAN_IP:-}` from host to container.
- `.env.schema.json`: register `HOST_LAN_IP` (string, non-secret).
- `installers/phases/06-directories.sh` (Linux): detect host LAN IP via `hostname -I` (probe-gated) → `ip route get 1.1.1.1` → `ifconfig` fallback chain; only writes when `BIND_ADDRESS=0.0.0.0`; uses `_env_get` so operator overrides in `.env` survive re-installs.
- `installers/macos/lib/env-generator.sh`: ifconfig-first BSD-correct detection with the same fallback chain; only when `BIND_ADDRESS=0.0.0.0`.
- `tests/test-openclaw-inject-token.sh`: 8 new assertions (positive `HOST_LAN_IP` → http+https origins on both merged and home configs; negative empty/unset → no malformed `http:///` entries).

## Testing
- `node --check inject-token.js` clean
- `python3 -c json.load .env.schema.json` clean
- `bash -n` clean on both shell files
- `bash tests/test-openclaw-inject-token.sh`: 23/23 pass (16 pre-existing + 7 new HOST_LAN_IP cases)
- Live cross-platform IP detection on a Linux/CachyOS host: returns correct LAN IP via `ip route` fallback (host's `hostname -I` exits 64; the probe correctly skips it). The macOS detection function (ifconfig-first), invoked on Linux, returns the same IP.

## Review
Critique Guardian: APPROVED (round 2). Round 1 flagged three required-before-merge items; round 2 addressed all three:
1. `_env_get HOST_LAN_IP "$HOST_LAN_IP"` added in 06-directories.sh — operator overrides survive re-installs
2. 8 new test assertions in test-openclaw-inject-token.sh
3. Schema entry style: `secret: false` removed (convention is omit field for non-secrets)

## Known Considerations
- **Windows native installer is unchanged** — `HOST_LAN_IP` applies to WSL2 via the Linux phase 06 path.
- **WSL2 caveat**: `ip route get 1.1.1.1` returns the WSL2 vNIC IP, which may differ from the Windows host's LAN IP visible to non-WSL2 LAN clients. Operators may need to override `HOST_LAN_IP` in `.env` to the actual Windows-side IP.
- **Multi-homed Linux host**: detection picks the first interface; operator override in `.env` is preserved by `_env_get` across re-installs.
- **macOS BIND_ADDRESS post-install**: pre-existing limitation (not introduced here) — operators who change `BIND_ADDRESS` post-install must re-run env-generator to populate `HOST_LAN_IP`.

## Platform Impact
- macOS: ifconfig-first detection (BSD-correct).
- Linux: 3-fallback chain validated live.
- Windows: WSL2 covered via Linux phase 06; Windows native PowerShell installer untouched.

This branch contains two commits: the first is a rebased copy of #1080's commit on current upstream/main (so this PR's diff is self-contained and reviewable); the second is the actual fix. **Must merge after Light-Heart-Labs/DreamServer#1080** — when #1080 lands, rebase this branch onto upstream/main and the #1080 commit drops out cleanly.


Lightheartdevs (changes requested):
Blocking this for now because it is stacked on an older OpenClaw change set and would regress the token-disclosure fix already merged in #1123.

On this branch `config/openclaw/inject-token.js` still writes the raw gateway token into logs and `/app/dist/control-ui/auto-token.js` (`http://localhost:${EXTERNAL_PORT}/#token=${token}` and `s.token = ...`). On current `main`, #1123 replaced that with a placeholder `auto-token.js` specifically to avoid exposing the gateway token as an unauthenticated static asset.

The HOST_LAN_IP allowedOrigins idea is valuable, but this branch needs to be rebased onto current `main` and reworked so it preserves #1123's no-token-output/no-token-static-asset behavior. After that, the useful part to retest is: BIND_ADDRESS=0.0.0.0 + HOST_LAN_IP set adds the LAN http/https origins without recreating auto-login token injection.

yasinBursali:
Rebased onto current `main` (`d4124d6a`) — #1123 token-disclosure protections fully preserved. Branch is now 2 commits ahead of `upstream/main`.

Verified the diff vs `main` line-by-line on `config/openclaw/inject-token.js`:

- ✅ `delete config.gateway.controlUi.dangerouslyAllowHostHeaderOriginFallback` defang preserved (Part 1 + Part 3)
- ✅ `auto-token.js` placeholder content unchanged (no real token written to the unauthenticated static asset)
- ✅ Token-redaction log line preserved: `Token redacted; paste from .env to sign in.`
- ✅ No new `s.token = …` localStorage assignments
- ✅ No new `http://localhost:.../#token=…` log lines
- ✅ `if (hostLanIp)` truthy-guard ensures empty-string + unset both produce no `http:///` malformed origins

The HOST_LAN_IP additions are surgical: two symmetric blocks in inject-token.js Part 1 (home config) and Part 3 (merged config), gated on `process.env.HOST_LAN_IP` truthy. `tests/test-openclaw-inject-token.sh` → 23/23 pass including 5 new assertions covering positive (http+https variants land on both home and merged configs), empty-string, and unset cases.

The follow-up commit (`e45ca7ab`) closes a macOS upgrade-path gap I noticed during the rebase: existing installs with `BIND_ADDRESS=0.0.0.0` weren't getting `HOST_LAN_IP` backfilled because the macOS env-generator's upgrade branch returned early before the detection block. Now upserts when `BIND_ADDRESS=0.0.0.0` AND `HOST_LAN_IP` is missing, mirroring the existing `DREAM_AGENT_KEY` and `DREAMFORGE_PULL_POLICY` upsert pattern.

Ready for fresh review.

Lightheartdevs (approved):
Re-audited after bringing the branch up to current `main` and resolving the OpenClaw compose conflict by keeping both `HOST_LAN_IP` and `BIND_ADDRESS` env passthroughs.

The #1123 token-disclosure protections remain intact: LAN-bound `auto-token.js` stays token-free, while `HOST_LAN_IP` adds the expected `http`/`https` LAN origins to both home and merged OpenClaw configs.

Local checks run:
- `node --check dream-server/config/openclaw/inject-token.js`
- `bash -n` on touched shell/test scripts
- `tests/test-openclaw-inject-token.sh` skipped locally because `jq` is unavailable
- custom Node smoke verified LAN origin insertion, empty/unset `HOST_LAN_IP` guard, and no LAN token leak

PR #1083, opened 2026-05-01
feat(preflight): warn when INSTALL_DIR is on networked filesystem

yasinBursali:
## What
Adds warn-only detection for networked filesystems on **all three platforms**:
- **macOS** (`stat -f %T`): `nfs`, `smbfs`, `afpfs`, `webdav`
- **Linux** (`stat -fc %T`): `nfs`, `nfs4`, `cifs`, `fuse.smbnetfs`, `fuse.glusterfs`, `ocfs2`
- **Windows** (PowerShell DriveInfo + UNC regex): mapped network drives + `\\server\share\...` UNC paths

Surfaces the gotcha that POSIX/NTFS permissions are advisory on networked FS — `chmod 600 .env` (or `icacls` equivalent) is enforced locally but governed by the share's ACL across the network.

## Why
On NFS/SMB/CIFS shares, the local kernel accepts `chmod 600` and reports back the new mode, but the server's ACL system (NFSv4 ACLs, SMB ACLs, NTFS ACLs) is the actual access control boundary. A user installing on a NAS share will set `chmod 600 .env` and believe their secrets are locked — but another client of the same share may read the file regardless of local permissions bits. This is well-known to sysadmins; less so to non-technical users (DreamServer's target audience).

## How
- `installers/macos/lib/preflight-fs.sh`: add `INSTALL_FS_NETWORKED` boolean alongside existing `INSTALL_FS_FATAL`; case-match on `nfs|smbfs|afpfs|webdav`
- `installers/macos/install-macos.sh`: 3× `ai_warn` block after the existing fatal consumer (line ~209) — non-fatal continuation
- `installers/phases/01-preflight.sh`: add second case in `check_install_dir_filesystem()` matching Linux personality strings; uses `warn` (not `error`)
- `installers/windows/phases/01-preflight.ps1`: 31-line block after existing FS-FATAL detection; checks UNC regex (`^\\\\`) for UNC paths, then DriveInfo `DriveType -eq 'Network'` for mapped SMB drives. Reuses existing `$_di` short-circuit-guarded for null safety. 3× `Write-AIWarn` lines mirror macOS/Linux wording.

Existing FATAL behavior for non-POSIX FS (exfat, msdos, fat, vfat, ntfs, ntfs-3g, 9p, drvfs) is preserved unchanged.

## Testing
- New `tests/bats-tests/preflight-fs-networked.bats` — 7 cases (Linux nfs/cifs/ext4/exfat, macOS nfs/smbfs/apfs); all pass
[...]

PR #1126, opened 2026-05-02
test(preflight): cover nfs4 mount type in preflight-fs-networked BATS

yasinBursali:
## What
Add one new BATS test case to `dream-server/tests/bats-tests/preflight-fs-networked.bats` (the file added by Light-Heart-Labs/DreamServer#1083) covering the `nfs4` filesystem type.

## Why
#1083's Linux production case statement at `01-preflight.sh:121` matches `nfs|nfs4|cifs|fuse.smbnetfs|fuse.glusterfs|ocfs2`, but the BATS fixture only exercises `nfs` / `cifs` / `ext` / `exfat` — leaving `nfs4` untested. A future change that broke the `nfs4` arm of the regex-OR would slip past CI.

## How
21-line test case mirroring the existing `linux preflight: nfs warns and does not exit fatally` test, with the stat stub returning `nfs4` instead of `nfs`. Asserts:
- the warning fires (`networked filesystem`)
- the type appears in the warning message (`nfs4`)
- the script does not exit fatally (`EXIT_OK`)

## Testing
- `bash dream-server/tests/run-bats.sh dream-server/tests/bats-tests/preflight-fs-networked.bats`: 8/8 PASS

## Review
Critique Guardian: APPROVED. Production warn line at `01-preflight.sh:122` includes `($fs_type)`, so the `nfs4` partial-match is reliable. Test scope honors the plan (`nfs4` only); `fuse.glusterfs` and `ocfs2` are also untested but out of scope here.

## Known Considerations
- macOS `preflight-fs.sh` does not include `nfs4` in its case statement — that's a separate production gap in #1083 itself, not a test-only gap, and is out of scope for this PR.
- `fuse.smbnetfs`, `fuse.glusterfs`, `ocfs2` Linux types are similarly untested by the fixture but match the same case statement; follow-up worth filing.

## Platform Impact
- macOS: BATS runs on macOS shell; this test targets Linux preflight only. No-op on macOS.
- Linux: primary target.
- Windows: N/A.

This branch contains two commits: the first is a rebased copy of #1083's commit on current upstream/main (so this PR's diff is self-contained); the second is the new BATS case. **Must merge after Light-Heart-Labs/DreamServer#1083** — when #1083 lands, rebase this branch onto upstream/main and the #1083 commit drops out cleanly.

### DIFF
--- a/dream-server/installers/macos/install-macos.sh
+++ b/dream-server/installers/macos/install-macos.sh
@@ -208,6 +208,16 @@ if $INSTALL_FS_FATAL; then
 fi
 ai_ok "Filesystem supports POSIX permissions"
 
+# Networked-filesystem advisory (warn-only).
+# chmod 600 still applies on NFS/SMB/AFP, but the actual access control is
+# enforced server-side by the share's ACL — other clients with access to the
+# share may read .env regardless of local permissions.
+if [[ "${INSTALL_FS_NETWORKED:-false}" == "true" ]]; then
+    ai_warn "INSTALL_DIR ($INSTALL_DIR) is on a networked filesystem ($INSTALL_FS_TYPE)."
+    ai_warn ".env permissions (chmod 600) are advisory — actual access control is governed by the share's ACL on the server."
+    ai_warn "If this share is exposed to other clients, sensitive credentials may be readable from those hosts."
+fi
+
 # Docker Desktop file-sharing allowlist check
 # Bind-mounts of paths outside the allowlist fail with cryptic OCI errors at
 # `docker compose up`. Probe with a throwaway container so we surface a clear
--- a/dream-server/installers/macos/lib/preflight-fs.sh
+++ b/dream-server/installers/macos/lib/preflight-fs.sh
@@ -8,7 +8,8 @@
 #          sharing allowlist gaps (which surface as cryptic OCI mount errors).
 #
 # Provides:
-#   test_install_dir_filesystem() -- sets INSTALL_FS_TYPE, INSTALL_FS_FATAL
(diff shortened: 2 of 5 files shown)


Lightheartdevs (changes requested):
Blocking this because the branch makes the Windows preflight script unparsable.

The new Windows advisory string in `installers/windows/phases/01-preflight.ps1` contains mojibake for an em dash (`â€”`). PowerShell treats the embedded smart-quote character in that sequence as a string delimiter, so parsing the file fails before the installer can run. Reproduction:

```powershell
$src = Get-Content -Raw dream-server/installers/windows/phases/01-preflight.ps1
[scriptblock]::Create($src)
```

That reports `The string is missing the terminator: '` and points at the later `causing 'model not found'` line because the earlier advisory string was prematurely closed. Please replace the advisory dash with plain ASCII (`-`) or otherwise ensure the file parses cleanly, and update the PR title/body because this is not test-only: it changes Linux, macOS, and Windows preflight behavior in addition to adding BATS coverage.

The Linux/macOS BATS target itself passed locally (`8/8`) and the shell syntax checks passed; the blocker is the Windows script parse failure.

yasinBursali:
Closing — verified the PR's intent is already in upstream/main @ `0c6d5f05`. After rebase, 0 net diff remains across both commits:

- `e45b0cfe feat(preflight): warn when INSTALL_DIR is on networked filesystem` — landed upstream as #1083
- `abcdd121 test(preflight): cover nfs4 mount type in preflight-fs-networked BATS` — coverage already incorporated alongside #1083

(The mojibake em-dash bug flagged in the May-02 review is therefore also moot — no PS1 file needs updating.)

Issue #1129, opened 2026-05-03
Feat: Add GB10 arm64 compatibility


Nordln:
### Description

Would be nice for Dreamserver to work on GB10 arm64 DGX Spark devices without having to fiddle  a lot.

More details at the bottom of my closed issue: https://github.com/Light-Heart-Labs/DreamServer/issues/871

### Use Case

The NVIDIA GB10 Grace CPU is an ARM64 (aarch64) processor, but these Docker images were built for x86_64). There is a CPU architecture mismatch between the Docker image and my host arm64 GB10 machine. 

```
~/dream-server$ docker logs dream-embeddings 
exec /usr/local/bin/text-embeddings-router: exec format error

~/dream-server$ docker logs dream-llama-server 
exec /app/llama-server: exec format error

~/dream-server$ docker logs dream-dreamforge 
exec /usr/local/bin/dreamforge-server: exec format error
```

This kills core functionality.

### Proposed Solution

Had a go at adding GB10 arm64 compatibility to the project myself. My build is now up and running apart from dream-embeddings. To get it working, I applied the following:

dream-llama-server: I incorporated https://github.com/cslev/llamacpp-cuda-arm64-docker as a local build solution. Note: Update the contained llama.cpp repo to head to gain gwen3.5 functionality. Also, an additional cURL install placement is needed on the final image build apt string of the build file, otherwise the health-check fails and dependant containers do not start.

dream-dreamforge: I just rebuilt the image manually and nudged it not to default to downloading.

@Team: Kudos on your design pattern for the containers. It made modding very easy

Lightheartdevs:
Hey my friend!  Great feedback and THANK YOU!  I picked up a GB10 to finally get this supported.  Lots of demand for it, but also one of those machines that's hard to build around unless you have your hands on it first for testing.  Just shipped PR# 1130 for this which should hopefully get us WAY closer for the next user attempting to setup Dream Server on a DGX Spark etc.  I'll run additional installer tests tomorrow to be sure we really locked it up.  Thank you again! 

P.S.  You are welcome on our dev Discord if you ever want to hang out: https://discord.gg/q7nFqhGa

PR #1130, opened 2026-05-05
arm64 / NVIDIA Grace Blackwell support — partial #1129

SSignall:
## Summary

Partially addresses #1129 — adds NVIDIA Grace Blackwell (GB10/GB200) support to the installer and fixes the arm64 install gap for `llama-server`, `whisper`, `dreamforge`, and (Apple Silicon) macOS. Validated end-to-end on a **DGX Spark** (GB10, aarch64, 128 GB unified, Ubuntu 24.04 / DGX OS, Docker 29 + Compose v5).

## Per-commit walkthrough

| Commit | Purpose |
|---|---|
| `78306c4` feat(detect): GB10/GB200 unified-memory tiering | Adds GB10/GB200 to `gpu-database.json` + 5 `nvidia_unified_*` rules with NV_ULTRA/T4/T3/T2/T1 tiering by RAM. `scripts/detect-hardware.sh` falls back to `/proc/meminfo` when `nvidia-smi` reports `[N/A]` for unified memory (mirrors `installers/lib/detection.sh:178-194`). |
| `e0de92f` fix(dreamforge): build locally on non-amd64 hosts | Removes the `platform: linux/amd64` pin so non-amd64 hosts don't fall into qemu emulation. (Refined later by `76f0301` to preserve the amd64 prebuilt-image fast-path.) |
| `ef95fa7` fix(installer): guard openclaw config self-copy when src=dst | Phase 06 was issuing `cp src=dst` when `OPENCLAW_CONFIG=openclaw.json`. Adds an explicit guard. |
| `3c7d0b5` chore(llama): bump llama.cpp server image to b9014 | `ghcr.io/ggml-org/llama.cpp:server-cuda-b9014` is multi-arch (verified via `docker manifest inspect`); the previous pin `b8248` was amd64-only. **This commit alone closes the arm64 gap for `llama-server`.** |
| `e92e591` feat(detect): add `detect_host_arch()` helper | New `installers/lib/host-arch.sh` — pure function mapping `uname -m` → `amd64\|arm64\|unknown`. Sourced from `install-core.sh`. |
| `8e9dc97` fix(whisper): support multi-arch speaches images on arm64 hosts | Adds `WHISPER_IMAGE` env-var indirection in compose as a forward-compat hatch + fixes a pre-existing VAD-patcher bug in `extensions/services/whisper/docker-entrypoint.sh` (the per-call check only saw the first line of multi-line `transcribe()` calls and produced "SyntaxError: keyword argument repeated: vad_filter" against arm64 speaches code that ships with built-in `vad_filter=`). The patcher fix is what actually unblocks arm64 — see also `e6d6813` below. |
| `e6d6813` fix(whisper): drop unneeded arm64 image override | Empirical re-test confirmed `0.9.0-rc.3-cuda` is now multi-arch (upstream republished the tag); with the patcher fix from `8e9dc97` the default works on arm64 directly. Removes the auto-set of `WHISPER_IMAGE` in phase 02; env-var hatch + schema entry preserved. Reported by @Lightheartdevs in review. |
| `e5d9324` fix(macos): migrate existing Apple Silicon installs to DREAMFORGE_PULL_POLICY=build | The macOS env-generator returns early when `.env` already exists, so `345bd2b`'s fresh-install heredoc never fired on upgrades. Adds an upsert in the existing-env branch mirroring the `DREAM_AGENT_KEY` migration just above. Reported by @Lightheartdevs in review. |
| `b823e0b` docs(env): bump LLAMA_SERVER_IMAGE example to b9014 | `.env.example` example pin was stale. |
| `76f0301` fix(dreamforge): preserve amd64 prebuilt-image fast-path | Compose v2.20+ interprets `pull_policy: build` as "always build, never pull" — so e0de92f inadvertently made amd64 users do a multi-minute Cargo build on first install. Switch to env-var indirection (`DREAMFORGE_PULL_POLICY`) mirroring the WHISPER_IMAGE pattern: amd64 default is `missing` (pull-if-not-local); installer sets `build` on arm64. |
| `345bd2b` fix(macos): set DREAMFORGE_PULL_POLICY=build on Apple Silicon | The macOS env-generator doesn't share `installers/phases/02-detection.sh`, so on Apple Silicon DREAMFORGE_PULL_POLICY would default to `missing` and Docker Desktop's amd64-only-image behaviour is version-dependent. Mirror the Linux phase-02 logic: when `uname -m == arm64` on macOS, write `DREAMFORGE_PULL_POLICY=build` into the generated .env so dreamforge builds a native linux/arm64 image. Intel Macs unchanged. |

## Scope: what's fixed vs what's not

| Service | Pre-PR on arm64 | Post-PR | Mechanism |
|---|---|---|---|
| `llama-server` | Exit 128 (amd64-only `b8248`) | ✅ Works | b9014 image is multi-arch |
| `whisper` | Crashed on arm64 (built-in `vad_filter=` collided with the existing entrypoint patcher) | ✅ Works | VAD-patcher early bail-out — default `0.9.0-rc.3-cuda` tag is multi-arch and pulls native arm64 |
| `dreamforge` | qemu emulation | ✅ Builds locally | env-var override sets `pull_policy: build` |
| `comfyui` | (Unrelated host issue, see below) | No change | Already arm64-compatible — `nvidia/cuda:12.8.0-runtime-ubuntu22.04` is multi-arch and torch cu128 stable supports Blackwell sm_120/sm_121 |
| `dashboard-api` | (Unrelated host issue, see below) | No change | Already arm64-compatible — `python:3.11-slim` + arm64 wheels for all deps |
| `embeddings` (TEI) | Restart 255 | ❌ **Still broken** | text-embeddings-inference publishes no arm64 image. Out of scope for this PR; tracked as a follow-up. |

## Test matrix (live install, 2026-05-05)

- `whisper` healthy on `linux/arm64` with `sha-c78b77d-cuda`, `/health 200`.
- `llama-server` b9014 arm64 binary loaded the model and served chat completions: `Device 0: NVIDIA GB10, compute capability 12.1, VRAM 124615 MiB`.
- `dashboard-api` served `/health 200 OK` on arm64 (verified pre-libEGL issue).
- `comfyui` server printed `To see the GUI go to: http://0.0.0.0:8188` on arm64.
- `docker compose config --quiet` against `.env.example` and against the merged stack — clean.
- Repo CI run locally:
  - `make lint` — exit 0
  - `make test` — exit 0
  - `tests/test-tier-map.sh` — 82/82
  - `tests/test-validate-env.sh` — 9/9
  - `tests/test-validate-manifests.sh` — 7/7
  - `tests/test-resolve-compose-resilient.sh` — 13/13
  - `tests/contracts/test-installer-contracts.sh` — pass
  - `tests/test-extension-integration.sh` — 32/32
  - `tests/integration-test.sh` — 67/70 (3 expected skips)
  - `tests/test-cpu-only-path.sh` — 10/10
  - `tests/test-hardware-compatibility.sh` — 32/32
  - `python3 scripts/audit-extensions.py` — 0 errors, 0 warnings
- Shellcheck (CI severity = error) on full repo — clean.

## What's not in this PR (with rationale)

- **Custom `llama-server/Dockerfile.cuda-arm64`** — not needed. Upstream b9014 publishes both amd64 and arm64 manifests on GHCR. The Nordln workaround referenced in #871 is no longer required.
- **`dashboard-api/Dockerfile` TARGETARCH plumbing** — not needed. `python:3.11-slim` is multi-arch and every dep in `requirements.txt` has arm64 wheels on PyPI. Verified the image builds and serves traffic on arm64.
- **`comfyui/Dockerfile` torch-nightly pin for sm_121** — not needed. `nvidia/cuda:12.8.0-runtime-ubuntu22.04` is multi-arch and torch `cu128` stable already supports Blackwell sm_120/sm_121. Pinning nightly would be a regression risk for amd64.
- **STT-engine swap on arm64** — not needed. Sticking with speaches preserves the OpenAI-compatible `/v1/audio/transcriptions` API surface that Open WebUI's `AUDIO_STT_OPENAI_API_BASE_URL` integration depends on.
- **Embeddings (TEI) arm64 build** — TEI publishes no arm64 image. A local-build path is technically feasible but needs deeper Dockerfile work (proper entrypoint, runtime deps, CUDA static-linking on aarch64). Tracked as a separate follow-up.

## Known unrelated host issue on the test box

After installing NVIDIA driver 580.142, the NVIDIA Container Toolkit cached the old library path and produced:

```
failed to fulfil mount request: open /usr/lib/aarch64-linux-gnu/libEGL_nvidia.so.580.82.09: no such file or directory
```

This kills any container with `--gpus all` (`llama-server`, `dashboard-api`, `comfyui`). Fix is host-side, not PR-scope:

```
sudo nvidia-ctk runtime configure --runtime=docker
sudo systemctl restart docker
```

`whisper` is unaffected because its compose entry uses `deploy.resources.reservations.devices` (not `--gpus all`).

## Things to verify on platforms I couldn't test

I built and tested this on a DGX Spark (Linux arm64 + NVIDIA). The arm64+nvidia path is exercised end-to-end. The following platforms would benefit from a maintainer/reviewer smoke:

1. **amd64 + Linux (NVIDIA)** — `dreamforge` with the new `${DREAMFORGE_PULL_POLICY:-missing}` default should pull `ghcr.io/light-heart-labs/dreamforge:v0.1.0` from GHCR (no local Cargo build) on first install with Compose v2.20+ / v5. `whisper` with the new `${WHISPER_IMAGE:-...0.9.0-rc.3-cpu/-cuda}` defaults should pull the existing pinned images and start healthy. The VAD patcher's new early bail-out is verified harmless on amd64 — `0.9.0-rc.3-cuda` ships an `stt.py` containing zero `vad_filter=` strings, so the bail-out can't trigger and the existing patch path runs unchanged.
2. **macOS Apple Silicon** — confirm the new `345bd2b` change writes `DREAMFORGE_PULL_POLICY=build` into the generated `.env` and that compose builds Dockerfile.rust to a native arm64 image (instead of pulling amd64 + Rosetta). Intel Macs should be unchanged.
3. **Low-end arm64+cpu hosts (e.g. Raspberry Pi)** — `ENABLE_DREAMFORGE=true` is the install default; on a Pi this means a multi-minute Cargo build on first install. Behaviour is correct (the upstream image is amd64-only so build is the only option) but Pi users may want to opt out via `--no-dreamforge`. Worth a docs note in a follow-up.

## Known coverage gap (follow-up issue)

There's currently no CI test that exercises the dreamforge compose path — `grep -rln dreamforge tests/` is empty. That's why the original `pull_policy: build` regression in `e0de92f` slipped past `make test`. Filing a separate follow-up to add a `tests/contracts/test-dreamforge-compose.sh` that asserts `pull_policy: missing` for the default case and `build` for the arm64 override. The same pattern would be valuable for `whisper`.

Refs #1129 (partial). Embeddings (TEI) remains amd64-only and is intentionally out of scope; please leave the upstream issue open after merge — it should track the embeddings follow-up until that lands separately.


Lightheartdevs (changes requested):
Thanks for the focused GB10 work. I audited this as the implementation PR for issue #1129. The shape is valuable and most local validation passed: JSON/schema syntax, bash syntax, extension audit, compose config with default and arm64 override values, tier map, hardware compatibility, CPU path, resolver resilience, and overlay-map coherence. I also verified live image manifests: `llama.cpp:server-cuda-b9014` is multi-arch while the old CUDA b8248 tag is amd64-only, and DreamForge's published image is amd64-only.

I think this is close, but I am requesting changes for two merge blockers and one documentation correction:

1. The PR body currently says `Closes #1129`, but the PR explicitly leaves embeddings/TEI broken and issue #1129 includes `dream-embeddings` exec-format failure. Please change this to `Refs #1129`, `Partially addresses #1129`, or split the embeddings follow-up into a separate issue before using a closing keyword.

2. Existing Apple Silicon installs do not get the new `DREAMFORGE_PULL_POLICY=build` migration because the macOS env generator returns early when `.env` already exists. In `dream-server/installers/macos/lib/env-generator.sh`, the existing-env branch returns at line 124 before the fresh-env `DREAMFORGE_PULL_POLICY=build` stanza can run. Please upsert `DREAMFORGE_PULL_POLICY=build` in that existing-env branch when `uname -m` is `arm64`, similar to the CPU limit and `DREAM_AGENT_KEY` migrations above.

3. The Whisper comment/justification should be updated. Live manifest inspection shows `ghcr.io/speaches-ai/speaches:0.9.0-rc.3-cuda` currently advertises both `amd64/linux` and `arm64/linux`, so the override may still be right for runtime behavior on GB10, but it is no longer accurate to describe the default tag as amd64-only without qualification. The comment in `dream-server/extensions/services/whisper/compose.nvidia.yaml` should explain the actual reason for preferring `sha-c78b77d-cuda` on arm64+nvidia.

Once those are fixed, I expect this to be mergeable as a partial GB10 support PR, with embeddings tracked separately.

SSignall:
Thanks @Lightheartdevs — really appreciate the thorough review. Pushed three follow-up commits addressing all three items:

1. **`Closes` → `Refs`**: PR body updated; #1129 will not auto-close on merge.
2. **macOS existing-install migration** (`e5d9324`): the existing-env branch in `installers/macos/lib/env-generator.sh` now upserts `DREAMFORGE_PULL_POLICY=build` on Apple Silicon when the key isn't already set, mirroring the `DREAM_AGENT_KEY` migration above it. Intel Macs and existing user overrides are untouched.
3. **Whisper override + comment** (`e6d6813`): you were right, my "amd64-only" claim was wrong — re-tested empirically on the GB10, `0.9.0-rc.3-cuda` pulls a native arm64 manifest and (with the patcher fix from `8e9dc97`) runs healthy on arm64 directly. Dropped the auto-set of `WHISPER_IMAGE` in phase 02 + the corresponding `.env` render in phase 06; the compose env-var indirection (`${WHISPER_IMAGE:-...}`) and schema entry stay as a forward-compat hatch but the installer no longer overrides the default. The actual value-add of the whisper commit is the docker-entrypoint patcher fix — the arm64 manifest of `0.9.0-rc.3-cuda` ships speaches code with built-in `vad_filter=` and the original patcher duplicated the kwarg, crashing the container on startup.

Live verification on the DGX Spark with the override removed: `dream-whisper Up X (healthy)` on `ghcr.io/speaches-ai/speaches:0.9.0-rc.3-cuda`, `/health 200`, `Application startup complete` in logs.

Lightheartdevs (changes requested):
Re-reviewed current head `e5d9324bce8c91143cb950d987ae8312b1b66210`.

The code changes address the substantive review items:

- Existing Apple Silicon installs now upsert `DREAMFORGE_PULL_POLICY=build` before the existing `.env` early return.
- The Whisper installer override was removed, and the compose/comment now correctly treats `WHISPER_IMAGE` as a forward-compat escape hatch rather than claiming the default image is amd64-only.
- Local focused validation passed: JSON/schema parsing, bash syntax, extension audit, `git diff --check`, default compose config, arm64 DreamForge policy compose config, tier-map tests, and overlay-map coherence.

One remaining blocker before approval: the PR body still starts with `Partially closes #1129`. That still contains the `closes #1129` closing keyword pattern, so it can still auto-close the issue even though embeddings/TEI remains unresolved. Please change that line to `Partially addresses #1129`, `Refs #1129`, or similar non-closing wording. The final footer already uses the right idea; the summary line needs the same treatment.

Once that wording is fixed, I am comfortable approving this as a partial GB10 support PR.

SSignall:
Good catch — fixed. PR body summary line now reads `Partially addresses #1129` (no closing keyword). Both the first line and the footer should be safe from auto-close now. No code changes; pushed only as a body edit via the REST API.

SSignall:
Belt-and-suspenders: also reworded the footer's "please don't auto-close #1129" line to "please leave the upstream issue open after merge" so the literal `close #1129` adjacency is gone too. Verified the rendered body has zero `(close|fix|resolve)s? #1129` matches end-to-end.

Lightheartdevs (approved):
Approved at `e5d9324bce8c91143cb950d987ae8312b1b66210`.

The requested fixes are now addressed:

- PR body no longer uses a closing keyword for #1129, so the embeddings follow-up issue can remain open.
- Existing Apple Silicon installs migrate `DREAMFORGE_PULL_POLICY=build` before the existing `.env` early return.
- Whisper no longer gets an unnecessary arm64 image override, and the comments now describe `WHISPER_IMAGE` as a forward-compat escape hatch.

Focused validation from the re-review passed: JSON/schema parsing, bash syntax, extension audit, `git diff --check`, default compose config, arm64 DreamForge policy compose config, tier-map tests, and overlay-map coherence.

This is good to merge as partial GB10/Grace Blackwell support, with embeddings/TEI arm64 support tracked separately.

PR #1133, opened 2026-05-07
fix(dashboard-api): unified-memory fallback for GB10/GB200

SSignall:
## Summary

On NVIDIA Grace Blackwell hardware (GB10, GB200), `nvidia-smi` reports `[N/A]` for `memory.total` / `memory.used` because the GPU shares system RAM. The dashboard-api was skipping any row with `[N/A]` values, so `get_gpu_info_nvidia()` and `get_gpu_info_nvidia_detailed()` returned `None` and the dashboard rendered **"no GPU detected"** on a working GB10.

This mirrors the fallback already in `installers/lib/detection.sh` (added in #875): when memory fields are `[N/A]`, read `MemTotal` / `MemAvailable` from `/proc/meminfo` and tag the GPU as `memory_type="unified"`. Utilization, temperature, and power keep coming from `nvidia-smi` — those fields work on GB10.

After this change on a DGX Spark:

```
$ curl -sH "Authorization: Bearer …" http://127.0.0.1:3001/api/status | jq .gpu
{
  "name": "NVIDIA GB10",
  "vramUsed": 12.3,
  "vramTotal": 121.7,
  "utilization": 8,
  "temperature": 45,
  "memoryType": "unified",
  "backend": "nvidia",
  "gpu_count": 1,
  "powerDraw": 9.3,
  "memoryLabel": "VRAM Partition"
}
```

Follow-on to #1130 (now merged); the runtime metrics path needed the same treatment as the installer detection path.

## Changes

- `extensions/services/dashboard-api/gpu.py`
  - New `_read_meminfo_mb()` helper (reads `MemTotal` / `MemAvailable`).
  - `get_gpu_info_nvidia()` falls back to `_read_meminfo_mb()` when memory is `[N/A]`; sets `memory_type="unified"` on the resulting `GPUInfo`.
  - `get_gpu_info_nvidia_detailed()` does the same per-GPU.
- `tests/test_gpu.py`, `tests/test_gpu_detailed.py`
  - Three new tests covering the fallback path and the case where `/proc/meminfo` is unavailable.

## Test plan

- [x] `make lint` passes
- [x] `make test` passes
- [x] `pytest tests/test_gpu.py tests/test_gpu_detailed.py` — 65 passed (was 62)
- [x] On DGX Spark (GB10): rebuilt `dashboard-api` image, `localhost:3001` shows the GB10 with correct util / temp / VRAM, `memoryType` reads `unified`.
- [ ] Regression check on a discrete-VRAM NVIDIA host (need a reviewer with an RTX-class card).

## Notes

- No model change; `IndividualGPU` does not gain `memory_type`. The aggregate built by `routers/gpu.py::_build_aggregate` still defaults to `"discrete"`. Propagating `memory_type` into per-GPU records and through the aggregate would be a follow-up — left out here to keep the diff focused on the "no GPU detected" symptom.
- `_read_meminfo_mb` returns `None` on `OSError` (missing/unreadable `/proc/meminfo`); callers fall through to skipping the row, matching the existing behavior on hosts with no meminfo.

### DIFF
--- a/dream-server/extensions/services/dashboard-api/gpu.py
+++ b/dream-server/extensions/services/dashboard-api/gpu.py
@@ -34,6 +34,31 @@ def _read_sysfs(path: str) -> Optional[str]:
         return None
 
 
+def _read_meminfo_mb() -> Optional[tuple[int, int]]:
+    """Return (used_mb, total_mb) from /proc/meminfo, or None on failure.
+
+    Used as a fallback for unified-memory NVIDIA GPUs (GB10, GB200) where
+    nvidia-smi reports [N/A] for memory.total/memory.used. Mirrors the
+    installer's detection.sh fallback.
+    """
+    try:
+        with open("/proc/meminfo") as f:
+            info = {}
+            for line in f:
+                parts = line.split()
+                if len(parts) >= 2:
+                    info[parts[0].rstrip(":")] = int(parts[1])
+    except OSError:
+        return None
+    total_kb = info.get("MemTotal", 0)
+    avail_kb = info.get("MemAvailable", 0)
+    if total_kb <= 0:
+        return None
+    used_mb = max(0, (total_kb - avail_kb)) // 1024
+    total_mb = total_kb // 1024
+    return used_mb, total_mb
+
+
 def _find_amd_gpu_sysfs() -> Optional[str]:
     """Find the sysfs base path for an AMD GPU device."""
     import glob
@@ -143,6 +168,7 @@ def get_gpu_info_nvidia() -> Optional[GPUInfo]:
 
     try:
         gpus = []
+        any_unified = False
         for line in lines:
             parts = [p.strip() for p in line.split(",")]
             if len(parts) < 5:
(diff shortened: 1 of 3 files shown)


Lightheartdevs (approved):
Audited the NVIDIA unified-memory fallback in both aggregate and detailed GPU paths. The change correctly stops dropping GB10/GB200 rows when `nvidia-smi` reports `[N/A]` for memory, falls back to `/proc/meminfo`, marks the aggregate as unified memory, and preserves the existing discrete NVIDIA parsing path.

The added tests cover the key cases: aggregate GB10 fallback, no-meminfo fallback failure, and detailed per-GPU fallback. The implementation also keeps `[N/A]` power handling intact.

Local checks:
- `python -m py_compile gpu.py tests/test_gpu.py tests/test_gpu_detailed.py`
- `python -m pytest tests/test_gpu.py tests/test_gpu_detailed.py`: 66 passed, 1 existing deprecation warning

I also attempted the full dashboard-api pytest suite, but it exceeded a 3-minute local timeout in this Windows environment. Remote CI is green, including API and type-check jobs, and the PR is mergeable. Looks good to merge.

PR #1135, opened 2026-05-07
fix(dashboard): allow setup wizard to be completed when diagnostics fail

SSignall:
## Summary

The setup wizard's Step 6 "Complete Setup" button was gated on `config.tested`, set true only when *every* functional test PASSed. A single genuinely-failing-but-optional service (embeddings on a host with no arm64 TEI image, whisper missing its model, a transient TTS hiccup, etc.) was enough to permanently lock a user out of the wizard. The only failure UI was a red "✗ Some tests failed — check logs" banner and a disabled green button — no skip path, no re-run affordance, no way to acknowledge the failure and proceed.

This was severe enough that on a fresh GB10 install the only way to escape the wizard was to open DevTools and `localStorage.setItem('dream-dashboard-visited','true')`.

## Fix

Loosen the gate: the bottom button is now disabled only while diagnostics haven't completed (`!testStatus.done`). Once the run finishes — pass or fail — the button is always clickable. Its label and color flip based on outcome:

| Outcome | Color | Label             |
| ------- | ----- | ----------------- |
| PASS    | green | Complete Setup    |
| FAIL    | amber | Continue Anyway   |

The failure banner copy is updated accordingly:

> ⚠ Some tests failed — review the log above. You can re-run, or continue anyway and revisit from the Diagnostics tab later.

A "Re-run Diagnostics" button is added next to that banner so the user can retry without leaving the step. The dead `config.tested` field is removed (it had no other readers).

## What this isn't

This isn't lying to the user — failures are still surfaced clearly with amber styling and explanatory copy. It isn't auto-skipping either: the user has to read the "Continue Anyway" label and click it deliberately. It just stops *blocking* progress on something that's almost always going to be a partial-fail in practice.

## Test plan

- [x] Unit tests updated. The previous `expect(...Complete Setup...).toBeDisabled()` assertion is replaced with three: button label flips to "Continue Anyway", that button is enabled, "Re-run Diagnostics" appears.
- [x] New test: clicking "Continue Anyway" after a FAIL sentinel calls `onComplete` once.
- [x] Full dashboard test suite — `vitest run` — 12 files, 61 tests, all green (was 59).
- [x] Rebuilt `dream-server-dashboard` image on a live install (DGX Spark / GB10) — `Continue Anyway`, `Re-run Diagnostics`, and `Complete Setup` strings all present in the deployed JS bundle.
- [ ] Manual click-through in the browser by a reviewer to confirm UX feel.

## Diff stat

```
SetupWizard.jsx           | 30 +++++++++++++++-------
SetupWizard.test.jsx      | 26 +++++++++++++++++--
2 files changed, 45 insertions(+), 11 deletions(-)
```

### DIFF
--- a/dream-server/extensions/services/dashboard/src/components/SetupWizard.jsx
+++ b/dream-server/extensions/services/dashboard/src/components/SetupWizard.jsx
@@ -9,7 +9,6 @@ export default function SetupWizard({ onComplete }) {
   const [config, setConfig] = useState({
     userName: '',
     voice: 'af_heart',
-    tested: false,
     preflightPassed: false
   })
   const [testStatus, setTestStatus] = useState({ running: false, output: [], done: false, success: false })
@@ -133,9 +132,6 @@ export default function SetupWizard({ onComplete }) {
         : collected.some(l => l.includes('All tests passed!'))
 
       setTestStatus(prev => ({ ...prev, running: false, done: true, success }))
-      if (success) {
-        setConfig(c => ({ ...c, tested: true }))
-      }
     } catch (err) {
       // Aborted fetches throw AbortError. That's the user cancelling or the
       // component unmounting — don't surface it as a user-visible error.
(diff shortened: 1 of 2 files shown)


Lightheartdevs (approved):
Audited the setup wizard behavior and tests. The change cleanly removes the old `config.tested` gate and replaces it with a simpler rule: diagnostics must finish, but failures can be acknowledged via an explicit `Continue Anyway` path. That matches the goal of avoiding first-run lockout when diagnostics fail for environmental reasons.

I checked that `dream-config`/`dream-dashboard-visited` are the only persisted setup values used by the dashboard, so removing `tested` does not break any downstream reader. The failure state now remains visible, offers a re-run button, and the completion path is covered by a focused test.

Local checks:
- `npm ci` (existing audit output: 6 vulnerabilities)
- `npm test`: 12 files, 61 tests passed
- `npm run build`: passed

Remote CI is green and the PR is mergeable. Looks good to merge.

PR #1136, opened 2026-05-07
fix: detect DGX Spark llama.cpp arch mismatch

SSignall:
## Summary

- add a `dream doctor` runtime diagnostic for DGX Spark / GB10 systems that compares detected compute capability 12.1 against llama-server's reported CUDA arch list
- surface a remediation hint when GB10 is served by a llama.cpp CUDA binary without `sm_121` / `121a` support
- stop the OpenCode user service from importing Dream Server's full `.env`, which unintentionally enabled Basic Auth through `OPENCODE_SERVER_PASSWORD`
- restore OpenClaw Control UI auto-connect for localhost-only installs while keeping the token-free placeholder behavior for LAN-bound installs
- pass `BIND_ADDRESS` into OpenClaw so the token bootstrap can distinguish localhost from LAN exposure
- add a DGX Spark field note documenting the Qwen3-Coder-Next question-mark output, CPU/GPU isolation, and the working `CMAKE_CUDA_ARCHITECTURES=121` build path

## Validation

- `bash -n dream-server/scripts/dream-doctor.sh`
- `bash -n dream-server/dream-cli`
- `node --check dream-server/config/openclaw/inject-token.js`
- `git diff --check`
- `./tests/test-openclaw-inject-token.sh` (18 passed)
- `./tests/test-bind-address-sweep.sh`
- ran source-tree `dream-server/scripts/dream-doctor.sh /tmp/dream-doctor-pr-check.json` on DGX Spark; report shows `dgx_spark_gpu: true`, `llama_cuda_archs: "1210"`, `dgx_spark_cuda_arch_check.status: "pass"`
- verified local OpenCode service on `http://127.0.0.1:3003/` returns `200 text/html` without Basic Auth
- recreated live `dream-openclaw`; container is healthy on `127.0.0.1:7860`, serves localhost-only `auto-token.js`, and logs `agent model: local-llama/qwen3-coder-next`

### DIFF
--- a/dream-server/config/openclaw/inject-token.js
+++ b/dream-server/config/openclaw/inject-token.js
@@ -18,6 +18,8 @@ const EXTERNAL_PORT = process.env.OPENCLAW_EXTERNAL_PORT || '7860';
 const LLM_MODEL = process.env.LLM_MODEL || '';
 const GGUF_FILE = process.env.GGUF_FILE || '';
 const OPENCLAW_LLM_URL = process.env.OPENCLAW_LLM_URL || '';
+const BIND_ADDRESS = (process.env.BIND_ADDRESS || '127.0.0.1').trim();
+const AUTO_TOKEN_ALLOWED = ['127.0.0.1', 'localhost', '::1'].includes(BIND_ADDRESS);
 
 // On AMD/Lemonade, compose.amd.yaml sets OLLAMA_URL to
 // "http://llama-server:8080/api" (Lemonade's Ollama-compat endpoint).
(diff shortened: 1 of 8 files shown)


Lightheartdevs (approved):
Audited this end to end. The security-sensitive OpenClaw change behaves as intended: with `BIND_ADDRESS=127.0.0.1`, `inject-token.js` writes a localhost-only `auto-token.js` that stores the token in the Control UI settings; with `BIND_ADDRESS=0.0.0.0`, it writes a token-free placeholder while still injecting the external script tag. I verified this with direct `node` execution using temporary Control UI files, and confirmed the home config still gets token auth while the dangerous Host-header fallback remains absent.

I also checked the compose output: OpenClaw now receives `BIND_ADDRESS=127.0.0.1` by default and publishes the gateway on host loopback by default, matching the auto-token safety boundary.

For the DGX Spark path, I ran `dream-doctor.sh` locally and also simulated a GB10 host with fake `nvidia-smi`/`docker` output. A bad llama.cpp arch list (`500,610,700,750,800,860,890,1200`) produces `runtime_warnings=1` and the expected remediation hint; a good `1210` arch list produces `runtime_warnings=0`. The `dream-cli doctor` renderer also surfaces the runtime warning when `DREAM_HOME` points at this checkout. The documented remediation matches NVIDIA's DGX Spark llama.cpp guidance for `-DCMAKE_CUDA_ARCHITECTURES=121`.

Local checks run:
- Bash syntax check for `dream-cli`, `scripts/dream-doctor.sh`, and `tests/test-openclaw-inject-token.sh`
- `tests/test-openclaw-inject-token.sh` (skips locally because `jq` is not installed here)
- manual Node assertions for localhost vs LAN OpenClaw token asset behavior
- Docker Compose config render for OpenClaw
- simulated DGX Spark doctor bad/good arch cases

Remote CI is green and the PR is mergeable. Looks good to merge.

PR #1137, opened 2026-05-07
fix(scripts): guard ((counter++)) sites + curl assignment against set -e abort

yasinBursali:
## What

Five operational scripts under `dream-server/scripts/` use `set -euo pipefail` together with the `((var++))` post-increment idiom to count PASS/FAIL/SKIP/ERROR/etc. Bash's `((expr))` arithmetic command exits 1 when the expression evaluates to 0 — and post-increment evaluates to the **pre**-increment value — so the **first** time any of these counters fires while still at 0, the whole script aborts under `set -e`.

This PR applies the established `((var++)) || true` guard already used in `dream-cli` (8 sites) to all affected scripts, plus closes one latent `set -e` abort that the counter fix exposed in `validate.sh`.

## Why

`validate.sh` is the canonical "is my install healthy" command. Pre-fix, it would print one PASS/FAIL line and silently exit 1, leaving operators thinking the very first check killed it. The other four scripts have similar single-line-then-exit failure modes that hide downstream check results.

Verified the bug semantics with a minimal reproducer:
```bash
$ bash -c 'set -euo pipefail; FAILED=0; ((FAILED++)); echo "never printed"'
# exits 1, "never printed" never appears
```

## How

**Counter guards** — append `|| true` to all `((var++))` sites in scripts that use `set -e`:

| File | Sites |
|---|---|
| `dream-server/scripts/validate.sh` | 4 (lines 55, 58, 88, 91 — both PASSED **and** FAILED) |
| `dream-server/scripts/validate-manifest-schema.sh` | 6 |
| `dream-server/scripts/first-boot-demo.sh` | 1 (line 132 — fired before any check ran) |
| `dream-server/scripts/migrate-config.sh` | 1 (line 252) |
| `dream-server/scripts/pre-download.sh` | 4 (lines 231, 234, 241, 247) |

`dream-server/scripts/llm-cold-storage.sh` uses `set -uo pipefail` (no `-e`) so its `((var++))` sites are unaffected — left alone.

**Latent abort fix in `validate.sh:78-84`** — the inference test does `RESPONSE=$(curl -sf ... )`. Under bash 4.4+, command-substitution exit codes propagate through simple-assignment under `set -e`, so when `curl` returns 7 (no llama-server running) the script aborted before reaching the optional-services loop and the summary block. Append `|| RESPONSE=""` so the assignment never aborts and the subsequent `grep -q content` correctly falls through to the FAIL branch.

## Testing

- Mechanical: `bash -n` + `shellcheck` on all 5 modified scripts — clean (no new warnings).
- `make lint` PASS, `make test` PASS (82 + 9 tests, no regressions).
- Functional reproducer for the counter bug: `bash -c 'set -euo pipefail; FAILED=0; ((FAILED++)) || true; echo "PASS FAILED=$FAILED"'` → `PASS FAILED=1` (was: silent exit 1).
- Live run on a partially-disabled DreamServer install: `validate.sh` now emits **all** PASS/FAIL/SKIP lines AND reaches the summary block. Pre-fix it died after one line; post-fix it correctly reports `⚠️ N test(s) failed, M passed`.
- Comprehensive grep confirms no remaining unguarded `((var++))` in scripts that use `set -e`:
  ```bash
  grep -rn '((.*++))' dream-server/scripts/ | grep -v '|| true' | grep -v 'for ((' | grep -v 'llm-cold-storage'
  # empty
  ```

## Review

Internal critique-guardian: APPROVED. Followed the project's existing `((var++)) || true` pattern from `dream-cli` (8 sites) for full consistency.

## Platform Impact

- **macOS**: bash 3.2+ semantics identical; project already uses Homebrew bash 4+ for `dream-cli`. No behavioral change beyond the bug fix.
- **Linux**: bash 4.x/5.x semantics identical. Affected scripts run during Phase 12 / install validation / model pre-download / config migration — all now reach their summary blocks under `set -e`.
- **Windows / WSL2**: WSL2 runs the same Linux scripts. No platform-specific code paths.

### DIFF
--- a/dream-server/scripts/first-boot-demo.sh
+++ b/dream-server/scripts/first-boot-demo.sh
@@ -129,25 +129,25 @@ SERVICES_OK=0
 SERVICES_TOTAL=0
 
 # Core services
-((SERVICES_TOTAL++))
+((SERVICES_TOTAL++)) || true
 if check_service "LLM (llama-server)" "$LLM_URL" "/health"; then
-    ((SERVICES_OK++))
+    ((SERVICES_OK++)) || true
     LLM_AVAILABLE=true
 else
     LLM_AVAILABLE=false
 fi
 
-((SERVICES_TOTAL++))
+((SERVICES_TOTAL++)) || true
 if check_service "Open WebUI" "$WEBUI_URL" "/"; then
-    ((SERVICES_OK++))
+    ((SERVICES_OK++)) || true
 fi
 
 # Optional services
 if curl -sf "${WHISPER_URL}/health" > /dev/null 2>&1; then
     success "Whisper STT is running (voice input enabled)"
     WHISPER_AVAILABLE=true
-    ((SERVICES_OK++))
-    ((SERVICES_TOTAL++))
+    ((SERVICES_OK++)) || true
+    ((SERVICES_TOTAL++)) || true
 else
     info "Whisper STT not running (voice input disabled)"
     WHISPER_AVAILABLE=false
@@ -156,8 +156,8 @@ fi
 if curl -sf "${PIPER_URL}" > /dev/null 2>&1; then
     success "OpenTTS TTS is running (voice output enabled)"
     PIPER_AVAILABLE=true
-    ((SERVICES_OK++))
-    ((SERVICES_TOTAL++))
+    ((SERVICES_OK++)) || true
+    ((SERVICES_TOTAL++)) || true
 else
     info "OpenTTS TTS not running (voice output disabled)"
     PIPER_AVAILABLE=false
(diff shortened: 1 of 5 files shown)


Lightheartdevs (approved):
Audited the Bash `errexit` guards in depth. The changed arithmetic increments are now safe under `set -euo pipefail`, the failed `curl` assignment path in `validate.sh` falls through to the normal failure accounting, and the touched scripts pass Bash syntax checks and whitespace checks locally.

I also exercised `scripts/validate.sh` with services down: it now reaches the final summary (`6 failed, 0 passed`) instead of aborting on the first zero-to-one counter increment or failed chat completion request. The manifest validator likewise reaches its summary; my local Git Bash environment lacks PyYAML, so it reports schema failures there, but that is an environment dependency rather than a regression from this patch. CI is green. Looks good to merge.

PR #1139, opened 2026-05-08
feat(Add dashboard integrations topology view): + Bug fixes and implementations with the current DreamServer structure.

gabsprogrammer:
image

### DIFF
--- a/dream-server/extensions/services/dashboard-api/main.py
+++ b/dream-server/extensions/services/dashboard-api/main.py
@@ -278,6 +278,7 @@ def _serialize_model(model_info) -> Optional[dict]:
 def _serialize_services(service_statuses: list[ServiceStatus], uptime: int) -> list[dict]:
     return [
         {
+            "id": service.id,
             "name": service.name,
             "status": service.status,
             "port": service.external_port,
@@ -294,6 +295,7 @@ def _fallback_services() -> list[dict]:
         if not external_port:
             continue
         links.append({
+            "id": service_id,
             "name": config.get("name", service_id),
             "status": "unknown",
             "port": external_port,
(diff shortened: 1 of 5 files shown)


Lightheartdevs (changes requested):
Thanks for the topology view. The build and existing dashboard tests pass locally, but I found one blocking integration issue.

`ServiceMap` builds every node from `service.id` and the known edges/categories are keyed by service id (`llama-server`, `open-webui`, `dashboard-api`, etc.). The page fetches `/api/status`, but that endpoint currently serializes services as `{ name, status, port, uptime }` and does not include `id`. In the actual dashboard payload, all of these new nodes therefore get `id === undefined`: classification falls through to `other`, React keys collide, `nodeById` only has the undefined key, and the known edges never render.

Fix options:
- add `id` to the `/api/status` service payloads, including fallback serialization, and keep this page keyed by ids; or
- have the page fetch/join against an endpoint that already exposes service ids.

Please also add a focused frontend test for `buildTopology` or the page using a representative `/api/status` payload, so this does not regress silently. I verified `npm test` and `npm run build` pass, so this is a runtime contract issue rather than a compile issue.

gabsprogrammer:
> Thanks for the topology view. The build and existing dashboard tests pass locally, but I found one blocking integration issue.
> 
> `ServiceMap` builds every node from `service.id` and the known edges/categories are keyed by service id (`llama-server`, `open-webui`, `dashboard-api`, etc.). The page fetches `/api/status`, but that endpoint currently serializes services as `{ name, status, port, uptime }` and does not include `id`. In the actual dashboard payload, all of these new nodes therefore get `id === undefined`: classification falls through to `other`, React keys collide, `nodeById` only has the undefined key, and the known edges never render.
> 
> Fix options:
> 
> * add `id` to the `/api/status` service payloads, including fallback serialization, and keep this page keyed by ids; or
> * have the page fetch/join against an endpoint that already exposes service ids.
> 
> Please also add a focused frontend test for `buildTopology` or the page using a representative `/api/status` payload, so this does not regress silently. I verified `npm test` and `npm run build` pass, so this is a runtime contract issue rather than a compile issue.

Done.

Lightheartdevs (approved):
Deep audit pass. The prior blocker is fixed: /api/status now serializes service id in both normal and fallback service payloads, and ServiceMap has explicit coverage for id-based and legacy name-only payloads. Local verification: tests/test_main.py::TestApiStatusServiceSerialization passed; ServiceMap Vitest passed; dashboard production build passed. Full test_main.py on this Windows host is still blocked by local pytest plugin/env issues unrelated to this PR, but the touched serialization contract is covered. No remaining blocker found.

PR #1140, opened 2026-05-08
Expose llama.cpp long-context and MoE tuning knobs

gabsprogrammer:
## Summary
- Exposes llama.cpp long-context tuning knobs through DreamServer environment configuration.
- Adds opt-in Flash Attention, KV cache type, and MoE expert CPU offload support.
- Wires the same knobs through Docker, Linux env generation, macOS native launchers, Windows native launchers, bootstrap upgrade, and the host agent.
- Documents safe long-context and MoE profiles for users who want to tune beyond the default model profile.

## Compatibility
- Defaults preserve existing behavior: Flash Attention remains `auto`, KV cache remains `f16`, and MoE CPU offload is unset.
- `LLAMA_ARG_N_CPU_MOE` is intentionally optional because llama.cpp exits if it receives a blank integer value.
- Dense models can use the long-context/Flash/KV knobs without setting the MoE-only option.

## Validation
- `python3 -m json.tool dream-server/.env.schema.json`
- `python3 -m py_compile dream-server/bin/dream-host-agent.py`
- `bash -n` on touched shell scripts
- PowerShell parser validation on touched `.ps1` files
- `git diff --check`
- `docker compose config` with long-context KV settings and with `LLAMA_ARG_N_CPU_MOE=25`


Lightheartdevs (approved):
Audited the llama.cpp flag plumbing across Docker, native macOS/Windows launchers, env generation, schema, and host-agent path. Verified against upstream llama.cpp b9014 docs that the exposed flags/env vars are supported, and local validation passed: diff check, JSON parse, Python compile, shell syntax via Git Bash, and docker compose config with explicit tuning env values. Looks good to merge.

PR #1147, opened 2026-05-11
docs: refresh main README accuracy

gabsprogrammer:
## Summary

This is a focused documentation accuracy pass for the root README, the nested `dream-server/README.md`, and the bootstrap installer comment.

It updates the docs to match the current project shape without removing the existing sections or contributor credits:

- replaces stale/broken one-line installer references with the current raw path under `main/dream-server/get-dream-server.sh`
- documents the host/container llama-server port split (`localhost:11434` on the host, `8080` inside Docker)
- updates Windows install guidance to avoid recommending Administrator by default, matching the installer warning
- expands the service list to include currently documented/manifests-backed services such as Dashboard API, Token Spy, APE, OpenCode, DreamForge, Brave Search, TEI embeddings, Langfuse, and Memory Shepherd
- updates hardware/model documentation for Qwen, Gemma 4, Apple Silicon, Intel Arc, AMD Strix Halo, CPU fallback, and cloud fallback
- fixes the nested README Wall of Heroes link target

## Why

The main page currently mixes older release docs with newer installer/runtime behavior. In particular:

- the root README still points at `v2.3.2`
- the nested README points at `v2.4.0/get-dream-server.sh`, but that raw URL is currently broken because the path is missing `dream-server/` and the tag is not present in the visible tag list
- service counts and hardware support no longer match the current manifests/tier maps
- Windows docs recommend Administrator, while the installer warns against it

I used the `main` raw installer path here because it is the only currently valid path for the current main-branch docs. If the preferred release policy is to pin README install commands to tags only, this can be switched once a matching tag exists.

## Validation

- `git diff --check`
- local README relative-link check for `README.md` and `dream-server/README.md`
- local Markdown table pipe-count check for both READMEs
- verified the new raw installer URL returns HTTP 200
- grep check confirmed the stale references were removed from the touched docs/scripts:
  - `v2.3.2`
  - `v2.4.0/get-dream-server`
  - `13 services`
  - `PowerShell as Administrator`
  - `localhost:8080/health`
  - `localhost:8080/v1`

## Maintainer note

This PR intentionally does not remove the existing Wall of Heroes or other narrative sections. It only updates claims, commands, support tables, and links that had drifted from the current repository behavior.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ A handful of companies control the vast majority of global AI traffic — and wi
 
 If AI is becoming critical infrastructure, it shouldn’t be rented. Self-hosting local AI should be a sovereign human right, not a career choice.
 
-**Dream Server is the exit.** A fully local AI stack — LLM inference, chat, voice, agents, workflows, RAG, image generation, and privacy tools — deployed on your hardware with a single command. No cloud. No subscriptions. No one watching.
+**Dream Server is the exit.** A local-first AI stack — LLM inference, chat, voice, agents, workflows, RAG, image generation, and privacy tools — deployed on your hardware with a single command. No cloud required. No subscriptions required. No one watching. Cloud and hybrid API modes are optional when you choose them.
 
 [![License: Apache 2.0](https://img.shields.io/badge/License-Apache%202.0-blue.svg)](LICENSE)
 [![GitHub Stars](https://img.shields.io/github/stars/Light-Heart-Labs/DreamServer)](https://github.com/Light-Heart-Labs/DreamServer/stargazers)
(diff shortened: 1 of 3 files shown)


Lightheartdevs (changes requested):
Deep audit found a docs accuracy blocker. The updated docs state the llama-server host API defaults to http://localhost:11434 as a general rule (README.md:62 and dream-server/README.md:57/98/376), but that is only true for the Linux Docker llama-server path generated by installers/phases/06-directories.sh (OLLAMA_PORT=11434). The macOS installer still writes OLLAMA_PORT=8080 in installers/macos/lib/env-generator.sh, and the Windows native/Lemonade path probes localhost:8080 in installers/windows/dream.ps1. So the new README text will send macOS/Windows-native users to the wrong health/API endpoint. Please qualify this as Linux/Docker-default behavior and keep the macOS/Windows native endpoint documented as 8080, especially in the troubleshooting health-check line.

Lightheartdevs:
Audit pass after follow-up fixes. The previous docs accuracy blocker is resolved:

- README endpoint guidance now distinguishes Linux Docker host `localhost:11434`, container `llama-server:8080`, and native macOS/Windows `localhost:8080` instead of implying `11434` universally.
- The nested `dream-server/README.md` component table, architecture notes, and troubleshooting curl examples were brought into the same contract.
- The current Hermes direction is reflected without erasing OpenClaw as an optional/legacy extension path.

Verification I ran locally:

- README table pipe-count validation
- install URL HEAD check for `dream-server/get-dream-server.sh`
- stale-string sweep for the old version/install/admin wording

GitHub checks are green on `91dfe6fb12500007ef52c2e6be9360f7f389b8da`. Merging this now.

PR #1148, opened 2026-05-11
fix(installer): validate llama-server image tags before startup

gabsprogrammer:
## Summary
- validate the configured llama-server Docker image before image pull / compose startup
- fail early with a clear message when the tag is missing instead of surfacing a late `docker compose up` manifest error
- support an explicit `LLAMA_SERVER_IMAGE_FALLBACK` override without silently substituting runtime images
- replace the missing Gemma 4 llama.cpp CUDA pin (`server-cuda-b8648`) with the available `server-cuda-b9014` pin and align the Linux pre-pull default with compose
- add focused BATS coverage for image validation and update tier-map expectations

## Why
Installs can currently fail late with an opaque error when a pinned llama.cpp image tag disappears from GHCR, for example:

```text
ghcr.io/ggml-org/llama.cpp:server-cuda-b8648: not found
```

This PR catches that before startup. It checks local images first, so preloaded/offline images still work, then verifies the remote manifest. If the image is unavailable, the installer stops with actionable guidance unless the user explicitly configured `LLAMA_SERVER_IMAGE_FALLBACK`.

## Maintainer notes
- No silent fallback is performed; the fallback is opt-in because runtime/model compatibility matters.
- The preflight is scoped to the configurable llama-server runtime image rather than every service image, keeping the change focused on the failure mode that blocks local inference startup.
- Windows preserves and reads `LLAMA_SERVER_IMAGE_FALLBACK` from `.env` as well as the process environment; Linux/WSL writes the variable when provided and reads it back before pull validation.

## Validation
- `bash -n install-core.sh installers/lib/docker-images.sh installers/lib/tier-map.sh installers/phases/06-directories.sh installers/phases/08-images.sh installers/macos/lib/tier-map.sh`
- `tests/bats/bats-core/bin/bats tests/bats-tests/docker-images.bats tests/bats-tests/tier-map.bats`
- `bash tests/test-tier-map.sh`
- PowerShell parser check for `installers/windows/install-windows.ps1`, `installers/windows/lib/env-generator.ps1`, and `installers/windows/lib/tier-map.ps1`
- `git diff --check`
- `docker manifest inspect ghcr.io/ggml-org/llama.cpp:server-cuda-b9014`
- `docker manifest inspect ghcr.io/ggml-org/llama.cpp:server-b8248`


Lightheartdevs (approved):
Deep audit pass. I reviewed the Linux/macOS image validation helper, phase 08 wiring, tier-map image/tag updates, Windows pre-start validation, and the new test coverage. Local verification: Git Bash syntax check passed for touched shell scripts; tests/test-tier-map.sh passed 83/83; PowerShell touched scripts parse; docker manifest inspect succeeds for ghcr.io/ggml-org/llama.cpp:server-cuda-b9014 and server-b8248. BATS is not installed on this Windows host, but CI is green. No blocker found.

PR #1152, opened 2026-05-12
feat(mdns): publish dream.local on the local network

Lightheartdevs:
# PR-1: mDNS — publish `dream.local` on the local network

**Branch:** `feat/mdns-announcer` (local, not pushed)
**Effort:** Medium — 4 files, 382 insertions
**Depends on:** PR-2 (`feat/webui-pwa-branding`) for the `DREAM_DEVICE_NAME` env-schema entry

## Summary

Adds a systemd-managed mDNS announcer that publishes the device on the LAN as `<DREAM_DEVICE_NAME>.local` (default `dream.local`). Once running, any phone/laptop on the same WiFi can browse to the device without knowing the IP.

Same UX as Sonos / Apple TV / any consumer device on a home network. **The foundation for the phone-first onboarding flow.**

## What's published

| mDNS name | Points to | What it serves |
|---|---|---|
| `<device>.local` | LAN IP | base hostname — bare URL in any browser |
| `<device>-chat._http._tcp.local` | port 3000 | Open WebUI chat |
| `<device>-dashboard._http._tcp.local` | port 3001 | Dashboard |
| `<device>-dashboard-api._http._tcp.local` | port 3002 | Dashboard API |

## Files

| File | Lines | What |
|---|---|---|
| `bin/dream-mdns.py` (new) | ~200 | Standalone Python service. Reads `.env` every 30s, calls `zeroconf.Zeroconf.register_service()` for each configured service. Best-effort IP detection via the UDP-to-public-IP trick (kernel chooses the right interface). Graceful SIGTERM shutdown. |
| `scripts/systemd/dream-mdns.service` (new) | 18 | System-mode unit matching `dream-host-agent.service`'s pattern (placeholders for `__INSTALL_DIR__` / `__HOME__` / `__PYTHON3__` / `__INSTALL_USER__`). |
| `docs/MDNS.md` (new) | ~80 | Operator docs: what gets announced, platform support, troubleshooting, how to disable. |
| `installers/phases/07-devtools.sh` (modified) | +60 | Install block after the host-agent install. Installs `python3-zeroconf` via apt/dnf/pacman/zypper, templates the systemd unit, enables --now. Non-fatal failure mode. |

## Platform support

- **Linux** — primary target. Uses `python3-zeroconf` against the system's `avahi-daemon` (already installed on virtually all desktop distros).
- **macOS** — script detects `platform.system() == "Darwin"` and exits cleanly. macOS announces hostname.local natively via mDNSResponder.
- **Windows** — script logs "not yet supported" and exits 0. Deferred to follow-up.

## Non-fatal failure mode

If `python3-zeroconf` install fails or the service fails to start, the install logs a `warn` but doesn't block. The device is still reachable by IP — mDNS is a quality-of-life feature, not a critical path. The dashboard always displays the IP prominently.

## Test plan

I cannot test from this Windows machine — these need a real Linux box on a real LAN:

- [ ] `systemctl status dream-mdns` shows active running after install
- [ ] On a phone on the same WiFi: `http://dream.local` resolves and shows the chat UI
- [ ] On a phone: dns lookup `dig @224.0.0.251 -p 5353 dream.local` returns the device IP
- [ ] On a phone: `avahi-browse -art` (or equivalent) lists the dream services
- [ ] Change `DREAM_DEVICE_NAME` in `.env`, wait 30s — `http://newname.local` resolves
- [ ] Two devices both running Dream — verify they don't collide if named differently

## Known limitations called out in docs

- Some routers / corporate WiFi block mDNS multicast (especially "guest" networks). Documented in `MDNS.md` troubleshooting section.
- Windows mDNS support is fragmented; deferred.
- Two devices both named `dream.local` is undefined behavior — documented as "give each one a unique name."

### DIFF
--- a/dream-server/bin/dream-mdns.py
+++ b/dream-server/bin/dream-mdns.py
@@ -5,8 +5,9 @@ Publishes the device on the local network as `<DREAM_DEVICE_NAME>.local`
 (default `dream.local`) plus per-service `_http._tcp` records so any device
 on the same LAN can find the dashboard and chat UI without knowing the IP.
 
-This makes the "open `dream.local` from any phone" UX work out of the box —
-the device boots, joins WiFi, and starts announcing itself within seconds.
+When paired with dream-proxy on port 80, this makes the "open `dream.local`
+from any phone" UX work: the device boots, joins WiFi, and starts announcing
+itself within seconds.
 
 Reads `DREAM_DEVICE_NAME` and the service ports from `.env`. Re-publishes
 when the file changes (poll-based, 30s cadence) so renaming the device or
@@ -72,7 +73,24 @@ logger = logging.getLogger(__name__)
 
 INSTALL_DIR = Path(os.environ.get("DREAM_INSTALL_DIR", "/opt/dream-server"))
 ENV_FILE = INSTALL_DIR / ".env"
-POLL_INTERVAL = int(os.environ.get("DREAM_MDNS_POLL_INTERVAL", "30"))
+
+
+def _safe_positive_int_env(key: str, default: int) -> int:
+    raw = os.environ.get(key, "")
+    if raw == "":
+        return default
+    try:
+        value = int(raw)
+    except (TypeError, ValueError):
+        logger.warning("Env %s=%r is not a valid integer; using default %d", key, raw, default)
+        return default
+    if value <= 0:
+        logger.warning("Env %s=%r is non-positive; using default %d", key, raw, default)
+        return default
+    return value
+
+
+POLL_INTERVAL = _safe_positive_int_env("DREAM_MDNS_POLL_INTERVAL", 30)
 
 # Hostname-safe pattern matches DREAM_DEVICE_NAME schema in .env.schema.json.
 _HOSTNAME_RE = re.compile(r"^[a-zA-Z0-9]([a-zA-Z0-9-]{0,30}[a-zA-Z0-9])?$")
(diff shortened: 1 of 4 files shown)


Lightheartdevs:
Deep audit notes: I can't mark this one merge-ready yet.

1. The user-facing promise is `http://dream.local`, but a default Dream install does not expose anything on LAN port 80. The compose defaults bind Open WebUI and the dashboard to `${BIND_ADDRESS:-127.0.0.1}` on ports 3000/3001, and existing docs explicitly say a fresh install is loopback-only with no reverse proxy/TLS bundled. Publishing an mDNS name for the host LAN IP does not make `http://dream.local` load chat from a phone: the browser will try `<LAN IP>:80`, while Dream is bound to `127.0.0.1:3000/3001` unless the user opts into LAN binding. This needs either a real port-80/reverse-proxy entrypoint, or the docs/installer should be narrowed to `http://<device>.local:<port>` and only report LAN URLs when `BIND_ADDRESS=0.0.0.0` is actually configured.

2. `dream-mdns.py` says macOS/Windows are clean no-ops, but `zeroconf` is imported before the platform check. On Windows without the package installed, `python bin/dream-mdns.py` exits 1 with the missing-package error instead of returning 0 via the Windows no-op branch. The platform gate should happen before requiring `zeroconf`, or the import should be delayed until the Linux path.

3. Port parsing can kill the daemon. `_build_services()` and `_config_signature()` call `int(env.get(...))`; if any of `WEBUI_PORT`, `DASHBOARD_PORT`, or `DASHBOARD_API_PORT` is blank/non-numeric, `ValueError` escapes the refresh loop because the loop only catches `OSError`/`RuntimeError`. Since `.env` is intentionally user-editable for ports/device name, this should log a warning and skip/fall back instead of crashing/restarting the systemd unit.

Checks I ran locally: `python -m py_compile bin/dream-mdns.py`, `bash -n installers/phases/07-devtools.sh`, direct Windows execution of `bin/dream-mdns.py`, and a mocked `zeroconf` import test for invalid port values.

Lightheartdevs:
Re-audit after the latest author update: better, but still not merge-ready.

Fixed from my previous pass:
- macOS/Windows now exit cleanly before requiring `zeroconf`; I verified `python bin/dream-mdns.py` exits 0 on Windows.
- User-editable service ports now go through `_safe_port()` instead of raw `int(...)`, so bad `WEBUI_PORT` / dashboard port values fall back instead of killing the daemon.
- Local syntax/lint checks pass: `python -m py_compile bin/dream-mdns.py`, `python -m ruff check bin/dream-mdns.py`, and `git diff --check origin/main...HEAD`.

Remaining blockers:

1. CI is still red. The docs link checker fails on `docs/MDNS.md:12 -> DREAM-PROXY.md` because this PR links to `docs/DREAM-PROXY.md`, but that file is not present on this branch/base. That doc appears to arrive in the new dream-proxy PR (#1172), so either #1152 needs to be stacked/retargeted after #1172, or the link needs to be removed until the proxy docs are in main.

2. The installer success message still overpromises `http://<device>.local`. `installers/phases/07-devtools.sh` prints “browse to http://${_device_name}.local from any device on your network” as soon as the mDNS service starts. But the updated docs correctly say that URL only works if `BIND_ADDRESS=0.0.0.0` and `dream-proxy` is enabled. With the default loopback bindings, or without the port-80 proxy, mDNS resolution succeeds but the browser still hits an empty/refused LAN port. The installer should say “mDNS name published” and only claim the bare URL works when those prerequisites are actually true.

3. Minor inherited mismatch: `docs/MDNS.md` now advertises a Hermes SRV record on port 9119, but in the newer Hermes proxy stack direct Hermes is intentionally internal-only and the user-facing port is 9120. If this lands before the Hermes proxy chain, it is just future-looking; if it lands after, it is stale. The mDNS record/docs should align with whichever Hermes exposure model wins.

So the platform/import and bad-port problems are fixed, but this branch is still red and still tied to dream-proxy assumptions that are not present in the branch.

Lightheartdevs:
Deep audit pass on the rebased/fixed head `0a0515bf27adafd013561e389df369c9fc78a612`.

I fixed the remaining integration issues before merge:

- Rebased onto current `main` after #1172 and #1155, so the mDNS docs now reference the landed dream-proxy behavior instead of a future PR.
- Removed the stale/global `BIND_ADDRESS=0.0.0.0` guidance. The LAN-facing requirement is now correctly `dream-proxy` with `DREAM_PROXY_BIND=0.0.0.0`; other services can stay loopback-bound.
- Made direct-port SRV records conditional on an explicitly LAN-facing `BIND_ADDRESS`, so the daemon does not advertise `LAN_IP:3000/3001/3002` while the backends are intentionally bound to `127.0.0.1`.
- Added an explicit proxy-root mDNS publication so the bare `dream.local` name still resolves even when direct-port records are skipped.
- Hardened port/env parsing: invalid `DREAM_MDNS_POLL_INTERVAL` no longer crashes at import, ports above 65535 fall back, and `HERMES_PORT` participates in the refresh signature.

Verification:

- `python -m ruff check dream-server/bin/dream-mdns.py` -> pass
- `python -m py_compile dream-server/bin/dream-mdns.py` -> pass
- Git Bash `bash -n dream-server/installers/phases/07-devtools.sh` -> pass
- `python dream-server/bin/dream-mdns.py` on Windows -> exits 0 no-op as intended
- Mocked ServiceInfo behavior check: default publishes only `dream.local` + proxy subdomains; direct SRV records appear only when `BIND_ADDRESS=0.0.0.0`
- `dream-server/tests/test-doc-links.sh` -> pass
- GitHub required checks on the pushed head are green.

This is now merge-ready. It gives the headless flow a real LAN name while keeping the safer loopback-by-default service posture intact.

PR #1155, opened 2026-05-12
feat(dashboard-api): add magic-link auth router

Lightheartdevs:
# PR-4: Magic-link auth (generate + redeem)

**Branch:** `feat/magic-link-auth` (local, not pushed)
**Effort:** Medium — 4 files (3 new, 1 modified), ~530 insertions / 2 deletions
**Depends on:** none (independent — the dashboard UI in PR-5 consumes this)

## Summary

Adds a single-use, time-boxed magic-link generator to the dashboard-api. An admin presses a button in the dashboard (PR-5), gets a URL and a QR code. The recipient scans / taps the link; the server redeems the token, sets an audited session cookie, and 302s them to the chat surface.

This is the **invite primitive** for the phone-first onboarding flow. Today an operator brings a phone home and either types in a 32-char URL or scans a QR — no account flow, no password setup, no app store.

## What's in scope for this PR

| Capability | Endpoint | Auth |
|---|---|---|
| Generate a token | `POST /api/auth/magic-link/generate` | admin (Bearer) |
| Get a QR for a URL | `GET  /api/auth/magic-link/qr` | admin (Bearer) |
| Redeem (sets cookie, 302) | `GET  /auth/magic-link/{token}` | **public** |
| List active + redeemed | `GET  /api/auth/magic-link/list` | admin (Bearer) |
| Revoke by hash prefix | `DELETE /api/auth/magic-link/{prefix}` | admin (Bearer) |

## Files

| File | Lines | What |
|---|---|---|
| `routers/magic_link.py` (new) | ~500 | The router. Pydantic schemas, JSON-file storage, rate limiting, QR generation, audit trail. |
| `tests/test_magic_link.py` (new) | ~315 | 25 tests: auth, generate validation, redeem happy path, redeem failure modes, reusable semantics, rate-limit, list, revoke, QR. |
| `main.py` (modified) | +2 | Add `magic_link` to the router imports + registration block. |
| `requirements.txt` (modified) | +2 | Add `qrcode[pil]` for QR generation (optional in code, soft-fail if absent). |

## Security posture

The token is the credential — every decision in the design is about not leaking it:

- **Tokens hash before persist.** `secrets.token_urlsafe(32)` is returned to the admin once; only the SHA-256 hash is written to disk. A read of `magic-links.json` reveals zero usable tokens.
- **Constant-shape failure response.** Invalid token, expired token, revoked token, already-redeemed token — all return the same `404 "Invalid or expired magic link"`. Holder cannot fingerprint which state caused the failure.
- **Rate-limited.** 5 failed redemptions per IP per 60s → 429. In-memory bucket (single-process dashboard-api makes this fine).
- **Single-use by default.** `reusable=True` is opt-in (family invites, share posters). Every redemption — including reusable ones — appends to the audit trail.
- **Audited.** Each redemption records timestamp, IP, and User-Agent (truncated to 200 chars).
- **Cookies are correctly hardened.** `HttpOnly` + `SameSite=Lax` always; `Secure` when the request scheme is `https`. Host-only (no `Domain=`).
- **Store file is 0600.** Best-effort `chmod` on write (some Docker volumes ignore it — documented behavior).
- **Atomic writes.** Write-tmp + rename, so a crash mid-write never corrupts the store.
- **Pruning.** Expired-but-never-redeemed tokens are dropped on the next operation. Audit records retained, capped at 200 to keep the file small.

## What's NOT in this PR (deliberately)

- **Bridging the redeemed session into Open WebUI's own user system.** Today the redirect lands on the chat surface with a `dream-target-user` hint cookie; Open WebUI's session is unaffected. The user still sees Open WebUI's signup/signin screen on first hit unless WEBUI_AUTH is already off. Bridging happens in a follow-up PR (likely via a small `auth/sso` shim in the chat container that reads our session cookie and provisions/signs-in the named user via Open WebUI's admin API).
- **Dashboard UI to invoke any of this.** That's PR-5 (`feat/invite-flow`). Today the endpoints are exercised via curl or the tests.
- **Multi-process / multi-worker rate-limit.** The dashboard-api is single-process; if that ever changes, the rate-limit table needs Redis or similar.
- **iOS Smart App Banner / Add-to-Home-Screen prompting after redemption.** Out of scope here — PWA install is PR-2 (chat) and PR-3 (dashboard).

## Test plan

```
pytest tests/test_magic_link.py -v
```

25 tests, all passing locally. Covers:

- [x] Every admin endpoint rejects unauthenticated requests with 401
- [x] Redemption endpoint is reachable without auth (the whole point)
- [x] Happy-path generate returns plaintext token + URL + expiry
- [x] Validation: rejects empty / non-matching-pattern username, bad scope, out-of-range expiry
- [x] Redemption sets the session cookie with the right attributes (HttpOnly, SameSite=Lax, dream-target-user)
- [x] Single-use tokens 404 on second redemption
- [x] Bogus / expired / revoked tokens all return the same 404 (no state oracle)
- [x] Reusable tokens redeem repeatedly; every redemption shows in the audit trail
- [x] Rate-limit kicks in after 5 failures from the same IP
- [x] List shows redemption count, last_redeemed_at, revoke status, note
- [x] Revoke rejects short prefixes (<4 chars) with 400, unknown prefix with 404
- [x] QR endpoint returns a `data:image/png;base64,…` URL (when `qrcode` is installed)

## Caveats

- **The recipient device sees a non-HTTPS dashboard URL** unless you've wired up a real cert (Tailscale Funnel, Cloudflare Tunnel, etc.). The `Secure` cookie attribute is only set when the request scheme is `https`, so on plain `http://dream.local` the session cookie is not sent over TLS. This is consistent with the rest of Dream Server's LAN posture — acceptable for a trusted home network, **not** acceptable if you expose this directly to the public internet. Documented in the operator notes.
- **`WEBUI_URL` and `DREAM_PUBLIC_URL` env vars control the redirect target and the URL embedded in the generated link.** Default is `http://localhost:3000`. If you change `DREAM_DEVICE_NAME` or the chat port, set these too.
- **The router uses synchronous file I/O.** This is a deliberate choice — the dashboard-api is single-process FastAPI, and these endpoints are low-frequency (admin clicks "invite" once per recipient). Async would add complexity without measurable benefit at the expected QPS.

## Follow-ups already scoped

- **PR-5 (feat/invite-flow)** — dashboard UI: "Invite someone" button, modal with QR + copy-URL + share buttons, list of active invites with revoke.
- **PR-?? (feat/openwebui-sso-bridge)** — provisions the named user inside Open WebUI's own DB via its admin API, so the redirect lands them already signed in.

### DIFF
--- a/dream-server/extensions/services/dashboard-api/main.py
+++ b/dream-server/extensions/services/dashboard-api/main.py
@@ -54,6 +54,7 @@ from routers import (
     workflows, features, setup, updates, agents, privacy, extensions,
     gpu as gpu_router, resources, voice, models as models_router, templates,
     auth as auth_router,
+    magic_link,
 )
 
 
@@ -927,6 +928,7 @@ app.include_router(voice.router)
 app.include_router(models_router.router)
 app.include_router(templates.router)
 app.include_router(auth_router.router)
+app.include_router(magic_link.router)
 
 
 # ================================================================
--- a/dream-server/extensions/services/dashboard-api/requirements.txt
+++ b/dream-server/extensions/services/dashboard-api/requirements.txt
@@ -6,3 +6,5 @@ httpx>=0.27.0,<0.29.0
 pydantic>=2.5.0,<3.0.0
 python-multipart>=0.0.9,<1.0.0
 PyYAML>=6.0,<7.0.0
+# QR code generation for magic-link invites (Pillow optional, but improves output)
+qrcode[pil]>=7.4.0,<9.0.0
--- a/dream-server/extensions/services/dashboard-api/routers/magic_link.py
+++ b/dream-server/extensions/services/dashboard-api/routers/magic_link.py
@@ -20,8 +20,9 @@ Security posture:
     audit trail.
   * 60-minute default expiry; configurable per-token (60s minimum, 24h max).
   * Rate-limit on redemption: 5 failed attempts per remote IP per minute.
-  * Cookie issued is HttpOnly + SameSite=Lax + Secure when HTTPS, never
-    Domain-scoped (host-only by design).
+  * Cookie issued is HttpOnly + SameSite=Lax + Secure when HTTPS. Default
+    host-based links set Domain=<device>.local so redemption on auth.<device>.local
+    carries through to chat.<device>.local.
   * No information leaks: invalid/expired/already-redeemed tokens all return
     the same 404 "Invalid or expired magic link" so a holder cannot fingerprint
     state.
@@ -80,8 +81,10 @@ MIN_EXPIRY_SECONDS = 60
 MAX_EXPIRY_SECONDS = 86400
 MAX_TOKENS_RETAINED = 200  # cap audit log growth
 
-_WEBUI_URL_DEFAULT = "http://localhost:3000"
-_PUBLIC_URL_DEFAULT = "http://dreamserver.local"
+# Used to build subdomain URLs (auth.<name>.local, chat.<name>.local).
+# Falls back to "dream" so a half-configured install still produces a
+# sane-looking URL.
+_DEVICE_NAME_DEFAULT = "dream"
 # Session cookie TTL — 12 hours of rolling chat access after redemption.
 SESSION_TTL_SECONDS = 12 * 3600
 
@@ -109,8 +112,8 @@ class GenerateRequest(BaseModel):
     )
     scope: str = Field(
         default="chat",
-        pattern=r"^(chat|dashboard|all)$",
-        description="What the redeemed session can reach. chat=Open WebUI only, dashboard=admin too, all=both.",
+        pattern=r"^chat$",
+        description="Magic links currently grant chat access only. Wider scopes require scoped session enforcement.",
     )
     expires_in: int = Field(
         default=DEFAULT_EXPIRY_SECONDS,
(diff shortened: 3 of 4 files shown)


Lightheartdevs:

1. Generated links default to the chat URL, but the redeem route is served by dashboard-api. `_magic_link_url()` uses `DREAM_PUBLIC_URL` or `_chat_url()`, and `_chat_url()` defaults to `http://localhost:3000`; then it appends `/auth/magic-link/{token}`. That means a default generated invite points at `http://localhost:3000/auth/magic-link/...`, which is Open WebUI, not the FastAPI router that actually defines `GET /auth/magic-link/{token}`. The dashboard nginx only proxies `/api/` to dashboard-api, not `/auth/`, so even the dashboard origin would serve the SPA for this path rather than redeeming the token. This needs a reachable redeem base (dashboard-api/public proxy) and/or a dashboard nginx route for `/auth/magic-link/` before links can work from a recipient device.

2. The PR is framed as granting chat access, but redemption currently does not authenticate the user into Open WebUI. The code sets a `dream-session` cookie scoped to the redeem host and a readable `dream-target-user` hint, then redirects to Open WebUI; the PR body also notes Open WebUI's session is unaffected. Unless `WEBUI_AUTH` is off, the recipient still lands on Open WebUI's signup/signin screen, so the operator-facing feature must not present this as a passwordless chat invite yet. Either wire the bridge in this PR, or make the endpoint/API/UI explicitly a token-audit stub until the bridge lands.

3. CI is red on Ruff: `tests/test_magic_link.py` imports `time` but never uses it (`F401`). Locally, `python -m ruff check extensions/services/dashboard-api/routers/magic_link.py extensions/services/dashboard-api/tests/test_magic_link.py` reproduces the same failure. `python -m pytest extensions/services/dashboard-api/tests/test_magic_link.py -q` passes (25 passed), so the lint failure is the only local test failure I saw.

Given the auth/security surface, I would not merge this until the URL topology and real Open WebUI auth semantics are corrected or very clearly scoped out of user-visible flows.

Lightheartdevs:
## Coordination note — overlaps with #1173

This PR includes byte-identical copies of files that also ship in #1173:

- `dream-server/extensions/services/dashboard-api/session_signer.py`
- `dream-server/extensions/services/dashboard-api/routers/auth.py`
- `dream-server/extensions/services/dashboard-api/tests/test_session_signer.py`
- `dream-server/extensions/services/dashboard-api/tests/test_auth_router.py`

…plus an overlapping `main.py` router-registration line.

The signer code itself is genuinely a dependency — `routers/magic_link.py:575` calls `session_signer.issue(ttl_seconds=SESSION_TTL_SECONDS)` — so we want it, just not twice.

**Suggested resolution:** merge #1173 first as the foundation, then rebase this PR onto post-#1173 `main` and drop the four duplicated files. What stays in this PR after the rebase: `routers/magic_link.py`, `tests/test_magic_link.py`, the magic-link-specific bits of `main.py`, and the `.env` / `requirements.txt` additions. Posted a matching note on #1173.

Lightheartdevs:
Re-audit after the latest update: still not ready to merge.

What improved:
- The Ruff failure is fixed; CI is green.
- Generated magic links now build on a proxy origin and redirect to `/chat`, so the same-origin cookie story is much closer to the shape we need.
- Redemption now mints an HMAC-signed `dream-session` cookie, and `/api/auth/verify-session` exists for Caddy `forward_auth` consumers.
- Local checks pass: `python -m ruff check ...`, `python -m pytest extensions/services/dashboard-api/tests/test_magic_link.py extensions/services/dashboard-api/tests/test_auth_router.py extensions/services/dashboard-api/tests/test_session_signer.py -q` -> 53 passed, and `git diff --check origin/main...HEAD`.

Remaining blockers:

1. Missing `DREAM_SESSION_SECRET` consumes the invite and then 500s. `redeem_magic_link()` records the redemption and writes the store before calling `session_signer.issue()`. If the secret is absent, `issue()` raises after the token has already been marked redeemed, so a single-use invite is burned and the recipient gets a server error. This is likely on every fresh install right now: `.env.example` adds the variable, but I found no installer/env generator path that actually creates `DREAM_SESSION_SECRET` (`rg DREAM_SESSION_SECRET installers scripts bin lib ...` only finds docs/tests/code). The secret needs to be generated like `WEBUI_SECRET`/`DASHBOARD_API_KEY`, and redemption should check/sign before committing the token as used, or roll back on signing failure.

2. The default public URL is still not real for this branch. `_PUBLIC_URL_DEFAULT` is `http://dreamserver.local`, while the rest of the new LAN stack uses `dream.local` / `DREAM_DEVICE_NAME`, and this branch does not add `DREAM_PUBLIC_URL` to `.env.example` or `.env.schema.json`. It also depends on `dream-proxy` routing `/auth/*` and `/chat`, but that service/doc is in #1172, not this branch. So a default generated link still points at a hostname/path this branch/base does not make reachable.

3. This overlaps with #1173. As noted earlier, the session signer and `/api/auth/verify-session` files are duplicated between the two PRs. The cleaner path is still: merge/fix #1173 first as the signed-cookie foundation, then rebase this PR onto it and keep only the magic-link router/tests and magic-link-specific wiring.

4. Scope/user semantics are still only partial. The cookie authenticates to Dream-side proxies that know to call `/api/auth/verify-session`; it does not authenticate the user into Open WebUI itself. That may be acceptable as a foundation, but user-facing copy in downstream invite flows needs to avoid implying Open WebUI passwordless login until the bridge/proxy behavior is actually in place.

Net: much healthier than the first pass, but the missing generated secret plus post-commit signing failure is a real runtime bug, and the default URL/proxy dependency is still unresolved on this branch.

Lightheartdevs:
Deep audit pass on the rebased/fixed head `3971bdab230d60d01aee8f9b0fb17e08ab8a1bf2`.

I found two issues that were worth fixing before merge:

- The default host-based flow generated `auth.<device>.local` -> `chat.<device>.local`, but if `DREAM_COOKIE_DOMAIN` was unset the signed session cookie stayed host-only on `auth.<device>.local`. I changed the default to derive `Domain=<DREAM_DEVICE_NAME>.local` for the dream-proxy layout, while keeping cookies host-only when `DREAM_PUBLIC_URL` is set for a custom single-origin/path deployment.
- The API accepted `dashboard` and `all` scopes, but the signed session cookie currently has no enforced scope claim. I restricted generation to `chat` for now so we do not imply dashboard/all authorization before scoped session enforcement exists.

Also rebased this PR on current `main` after #1173/#1172 so it no longer carries duplicate session-cookie/proxy work.

Verification:

- `python -m pytest dream-server/extensions/services/dashboard-api/tests/test_magic_link.py dream-server/extensions/services/dashboard-api/tests/test_auth_router.py dream-server/extensions/services/dashboard-api/tests/test_session_signer.py -q` -> 64 passed
- `python -m ruff check dream-server/extensions/services/dashboard-api/routers/magic_link.py dream-server/extensions/services/dashboard-api/tests/test_magic_link.py` -> pass
- GitHub required checks on the pushed head are green.

This is now merge-ready for the phase-one magic-link API. Note: as the module doc says, deeper chat/Open WebUI session bridging is still follow-up work; this PR provides the signed-cookie redemption primitive and audit lifecycle.

PR #1156, opened 2026-05-12
feat(dashboard): add Invites page consuming magic-link API

Lightheartdevs:
# PR-5: Invite flow — dashboard UI for magic links

**Branch:** `feat/invite-flow` (local, not pushed)
**Effort:** Medium — 2 files (1 new, 1 modified), 557 insertions
**Depends on:** PR-4 (`feat/magic-link-auth`) — consumes the endpoints it exposes

## Summary

Surfaces the magic-link backend from PR-4 in the dashboard UI. The admin clicks **Invites** in the sidebar, hits **New invite**, fills three fields, and gets back a modal with a QR code + URL + share/copy/done.

This is the missing half of the phone-first onboarding loop. Today: admin SSHs into the box and curls a JSON payload. After this PR: admin opens the dashboard, fills the form, hands their phone to the recipient.

## Surface

| Element | Behaviour |
|---|---|
| **Sidebar entry** | `UserPlus` icon, label "Invites", route `/invites`, order 4 (between Models and Settings). |
| **List view** | One row per token. Shows target username, status pill (active / reused × N / used / expired / revoked / reusable), optional note, expiry countdown, hash prefix. Revoke button for active rows. |
| **Empty state** | `Users` icon + one-line pitch + "Create your first invite" CTA. Removes the "where do I start" friction. |
| **Create modal** | Username (validated `^[A-Za-z0-9._-]+$` client-side, matching the server's pattern), scope dropdown with inline help, expiry preset (15m / 1h / 24h), reusable checkbox, optional note. |
| **Generated modal** | Server-rendered QR (fetched from `GET /api/auth/magic-link/qr`), invite URL with click-to-copy, native `navigator.share()` button on browsers that support it (mobile Safari/Chrome), Done. |

## Files

| File | Change |
|---|---|
| `src/pages/Invites.jsx` (new, ~525 lines) | The whole page — list, create modal, generated modal, helpers. Self-contained; no new hooks or contexts. |
| `src/plugins/core.js` (+13 lines) | Lazy-import `Invites`, add the route entry (sidebar order 4). |

## Design choices

- **No new hook abstraction.** The page makes three `fetch` calls (`list`, `generate`, `revoke`, plus on-demand `qr`); wrapping that in a `useInvites()` hook would add code without paying for itself. If a second page ever needs the same data we extract then.
- **No client-side caching of generated tokens.** The plaintext token is sensitive — it never lives in localStorage or sessionStorage. The "Generated" modal holds it in component state only; closing the modal drops it. The admin re-shares via the URL displayed once; if they lose it they generate a new invite.
- **QR endpoint is fetched on demand, not pre-rendered.** The list view shows hash prefixes; rendering 30 QR codes preemptively wastes bytes. Generation is fast (~50ms server-side) and the modal shows a skeleton while it loads.
- **`navigator.share()` is feature-detected, not assumed.** Desktop Chrome doesn't ship the Web Share API; we render a Share button only when it's available. Falls back to Copy on desktop.
- **Status pill semantics match server pruning rules.** Server keeps revoked + redeemed records for audit; UI labels them clearly so the admin doesn't think a "used" token is still valid.

## What's NOT in this PR

- **Bulk operations** — no "revoke all expired" sweep. The server prunes expired-never-redeemed tokens automatically; the rest is intentional retention for audit.
- **Search / filter** — list is unfiltered. Comes if/when token volume justifies it.
- **Per-token detail view** — list rows show everything that matters (status, expiry, redemption count, note). No drilldown today.
- **Audit log surface** — the server records `at`, `ip`, `user_agent` per redemption. Surfacing that is a separate PR (privacy implications + UI design).
- **Tests** — the page is exercised via the existing build/lint pipeline, and the backend it talks to has 25 tests in PR-4. The component logic is thin enough that integration testing would be more useful than unit testing; deferring until the page stabilizes.

## Test plan

- [x] `npm run build` succeeds; `Invites-*.js` shows up as a lazy chunk (~14 KB / 4 KB gzipped)
- [x] `npm run lint` — zero new warnings, zero errors
- [ ] Smoke test on a running dashboard:
  - [ ] Sidebar shows Invites entry with UserPlus icon
  - [ ] Empty state renders when no tokens exist
  - [ ] Create modal validates username (rejects `bad name with spaces`)
  - [ ] Submitting opens Generated modal with QR + URL
  - [ ] Copy button copies the URL; Share button opens the native sheet (mobile)
  - [ ] Closing the modal returns to the list with the new row visible
  - [ ] Revoke button on an active row marks it revoked + greys the badge
  - [ ] Refresh button re-fetches without flicker

## Caveats

- **Plaintext token is shown exactly once.** The admin must copy/share before closing the Generated modal. This is deliberate (the token is the credential — if it shows up in a recovered closed-tab session we have a leak). The list view shows only the hash prefix.
- **QR generation requires the `qrcode` python lib on the backend.** PR-4 adds it to `requirements.txt`. If a deployment has stale dependencies, the QR panel shows a "QR generation unavailable" hint and Copy/Share still work.
- **Web Share API requires HTTPS.** On `http://dream.local` the Share button is hidden. Same constraint that gates the PWA install prompts; will be resolved by the Tailscale/cert PR in Phase 3.

### DIFF
--- a/dream-server/extensions/services/dashboard/src/pages/Invites.jsx
+++ b/dream-server/extensions/services/dashboard/src/pages/Invites.jsx
@@ -1,0 +1,545 @@
+import { useState, useEffect, useCallback } from 'react'
+import {
+  UserPlus, Copy, Check, Trash2, RefreshCw, QrCode, Share2, X,
+  Loader2, AlertCircle, Clock, Users,
+} from 'lucide-react'
+
+// Auth: nginx injects "Authorization: Bearer ${DASHBOARD_API_KEY}" via
+// proxy_set_header for all /api/ requests (see nginx.conf). All fetches
+// use relative URLs so the proxy adds the header before forwarding to
+// dashboard-api. No explicit auth in JS.
+
+const fetchJson = async (url, init = {}, ms = 8000) => {
+  const c = new AbortController()
+  const t = setTimeout(() => c.abort(), ms)
+  try {
+    return await fetch(url, { ...init, signal: c.signal })
+  } finally {
+    clearTimeout(t)
+  }
+}
+
+const SCOPES = [
+  { value: 'chat', label: 'Chat only', help: 'Recipient can use the chat surface.' },
+  { value: 'dashboard', label: 'Dashboard + chat', help: 'Recipient also sees the admin dashboard. Use sparingly.' },
+  { value: 'all', label: 'Full access', help: 'Everything. Treat like a co-admin invite.' },
+]
+
+const EXPIRY_PRESETS = [
+  { value: 900, label: '15 minutes' },
+  { value: 3600, label: '1 hour' },
+  { value: 86400, label: '24 hours' },
+]
+
+function formatRelative(iso) {
+  if (!iso) return null
+  const t = new Date(iso).getTime()
+  if (Number.isNaN(t)) return null
+  const diff = t - Date.now()
+  const abs = Math.abs(diff)
+  const minutes = Math.round(abs / 60_000)
+  const hours = Math.round(abs / 3_600_000)
+  const future = diff > 0
+  if (minutes < 1) return future ? 'in seconds' : 'just now'
+  if (minutes < 60) return future ? `in ${minutes}m` : `${minutes}m ago`
+  if (hours < 24) return future ? `in ${hours}h` : `${hours}h ago`
+  const days = Math.round(abs / 86_400_000)
+  return future ? `in ${days}d` : `${days}d ago`
+}
+
+function tokenStatus(token) {
+  if (token.revoked_at) return { label: 'revoked', tone: 'bg-theme-border text-theme-text-muted' }
+  if (new Date(token.expires_at).getTime() < Date.now()) {
+    return { label: 'expired', tone: 'bg-theme-border text-theme-text-muted' }
+  }
+  if (token.redemption_count > 0 && !token.reusable) {
+    return { label: 'used', tone: 'bg-theme-border text-theme-text-muted' }
+  }
+  if (token.redemption_count > 0) {
+    return { label: `reused × ${token.redemption_count}`, tone: 'bg-blue-500/20 text-blue-400' }
+  }
+  return { label: 'active', tone: 'bg-green-500/20 text-green-400' }
+}
+
+export default function Invites() {
+  const [tokens, setTokens] = useState([])
+  const [loading, setLoading] = useState(true)
+  const [error, setError] = useState(null)
+  const [showCreate, setShowCreate] = useState(false)
+  const [generated, setGenerated] = useState(null)
+  const [refreshing, setRefreshing] = useState(false)
+
+  const refresh = useCallback(async () => {
+    setRefreshing(true)
+    try {
+      const resp = await fetchJson('/api/auth/magic-link/list')
+      if (!resp.ok) throw new Error(`list failed: ${resp.status}`)
+      const data = await resp.json()
+      setTokens(data.tokens || [])
+      setError(null)
+    } catch (err) {
+      setError(err.message)
+    } finally {
+      setLoading(false)
+      setRefreshing(false)
+    }
+  }, [])
+
+  useEffect(() => { refresh() }, [refresh])
+
+  const handleRevoke = async (prefix) => {
+    try {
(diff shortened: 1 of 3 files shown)


Lightheartdevs:

1. This PR cannot stand on `main` by itself. It adds a visible `/invites` dashboard page that immediately calls `/api/auth/magic-link/list`, `/generate`, `/qr`, and `DELETE /api/auth/magic-link/{prefix}`, but those endpoints only exist in #1155. The PR currently has `mergeStateStatus: DIRTY` and no checks; it needs to be stacked on the magic-link API PR or rebased after that API lands. As-is, merging this page independently would add a sidebar entry that fails with 404s.

2. It inherits the core blocker from #1155: the generated URL can point at the wrong service (`WEBUI_URL` / Open WebUI) while redemption is implemented by dashboard-api, and the redemption flow does not create an Open WebUI session. The UI tells operators the invite lets someone reach chat with “no account, no password, no app store,” and the modal says Open WebUI will use the supplied username as the display name. That is not true yet when `WEBUI_AUTH` is enabled: the recipient still lands at Open WebUI's normal auth screen after redirect.

3. The UI exposes `dashboard` and `all` scopes (`Dashboard + chat`, `Full access`) but the backend currently only stores the scope string; it does not enforce dashboard authorization or grant Open WebUI permissions. Showing those choices before enforcement exists is risky because an operator will assume access semantics that the system does not actually implement.

Local checks: `npm.cmd run build` passed and `npm.cmd test` passed (12 files / 62 tests). The blockers are integration/semantics rather than frontend compilation.

Lightheartdevs:
Re-audit after the latest update: improved, but still not merge-ready.

What changed in the right direction:
- The UI now exposes only `chat` scope and explicitly avoids `dashboard` / `all` until enforcement exists. That fixes the riskiest operator-facing overpromise from the previous pass.
- The username helper copy now says Open WebUI may still ask the recipient to sign in or pick a profile name, which is much more honest.
- Local frontend checks pass on the branch itself: `npm.cmd test -- --run` -> 62 passed, and `npm.cmd run build` succeeds.

Remaining blockers:

1. This PR still cannot merge into `main` by itself. It does not contain the magic-link API (`git merge-base --is-ancestor origin/feat/magic-link-auth HEAD` -> no), but the page immediately calls `/api/auth/magic-link/list`, `/generate`, `/qr`, and delete endpoints. Merging independently would add a visible sidebar page that 404s.

2. GitHub still reports the PR as `DIRTY`, and a local merge simulation shows `extensions/services/dashboard/src/plugins/core.js` conflicts with current `main`. That needs a rebase/retarget after the dashboard nav changes already on main.

3. It inherits #1155's remaining backend blockers: missing generated `DREAM_SESSION_SECRET`, default public URL/proxy dependency, and partial Open WebUI auth semantics. The UI copy is now cautious, but the page should still wait for the backend foundation to be safe.

Net: the frontend itself is in decent shape now, but it needs to be stacked on the corrected magic-link API branch and rebased cleanly before merge.

Lightheartdevs:
Audit/fix pass after #1155/#1172/#1154 landed: this is merge-ready now.

What I verified:
- Rebasing onto current `main` is clean; the previous dependency and `core.js` conflict blockers are gone.
- The page calls the merged magic-link API endpoints and uses the backend-returned `record.url` for sharing/QR, so it stays aligned with `auth.<device>.local` / `DREAM_PUBLIC_URL` behavior.
- The UI only exposes `chat` scope, matching the backend validator.
- The copy is honest about current Open WebUI semantics: magic-link redemption gets the user to chat and sets the Dream session cookie, but Open WebUI may still prompt until deeper SSO is wired.

Fixes I pushed:
- Removed stale scope comments that implied `dashboard` / `all` are still accepted.
- Fixed generated-invite copy so reusable invites are not described as “share once.”
- Added focused page tests for listing/revoking invites and generating a chat-scoped invite using the backend URL plus QR loading.

Local checks:
- `npm.cmd test -- --run src/pages/Invites.test.jsx` -> 2 passed
- `npm.cmd test` -> 15 files / 77 tests passed
- `npm.cmd run build` -> pass
- `npm.cmd run lint` -> pass with 5 pre-existing warnings outside this PR
- `git diff --check` -> pass

GitHub checks are green on head `6395d1e9547614ffe8167848d1d9edfcf23cd79c`.

PR #1154, opened 2026-05-12
feat(dashboard): PWA manifest + service worker + Apple home-screen support

Lightheartdevs:
# PR-3: Dashboard PWA support

**Branch:** `feat/dashboard-pwa` (local, not pushed)
**Phase:** 1 of 4 (Network UX)
**Effort:** Small — 3 files, 94 insertions / 2 deletions

## Summary

Makes the dashboard at `:3001` installable as a PWA on any device. After this PR, a user can:

- Add the dashboard to their phone's home screen via the standard browser "Add to Home Screen" flow
- Tap the resulting "Dream" icon to launch the dashboard in standalone mode (no browser chrome)
- See the same dark theme color in the OS status bar

Pairs with PR-2 (which does the same for the chat surface). The two together produce a phone home screen with **two Dream tiles next to ChatGPT**: chat for daily-use, dashboard for admin.

## Files

| File | Change |
|---|---|
| `public/manifest.webmanifest` (new) | PWA manifest — name "Dream Server", short_name "Dream", standalone display, dark `theme_color` matching the dashboard's actual theme background. Icons point at the existing `dream.svg`. |
| `public/sw.js` (new) | Minimal service worker. Deliberately caches **nothing** — the dashboard is a live system-status surface and stale cache would mask running-service state. Just enough SW registration to satisfy PWA-install criteria. |
| `index.html` (modified) | Added `<link rel="manifest">`, `theme-color` meta, Apple PWA meta tags (capable / title / status-bar-style / apple-touch-icon), SW registration script. Title shortened to "Dream" since it's now the PWA install label. |

## What's NOT in this PR

- **Real PNG icons** at 192/512/maskable sizes. Currently using the existing SVG for all slots. Modern browsers handle SVG icons fine; iOS prefers PNG for apple-touch-icon at 180x180. Needs design input (Dream logo, brand palette).
- **`vite-plugin-pwa`** integration. Going static keeps the PR small and dependency-free. Can migrate later if we want auto-generated manifest or runtime caching.
- **iOS splash screens** — separate `<link rel="apple-touch-startup-image">` entries per device size. Skipping for v1.

## Test plan

- [ ] On iOS Safari at `http://dream.local:3001`: Share → Add to Home Screen → tile reads "Dream" with dream.svg icon
- [ ] On Android Chrome: install prompt fires → home-screen icon launches in standalone mode
- [ ] Desktop Chrome: install icon appears in address bar
[...]

PR #1157, opened 2026-05-12
feat(dashboard): server-side first-boot gating

Lightheartdevs:
# PR-6: Server-side first-boot gating

**Branch:** `feat/first-boot-sentinel` (local, not pushed)
**Effort:** Small — 4 files (1 new, 3 modified), 84 insertions / 20 deletions
**Depends on:** none (the dashboard-api already exposes `/api/setup/status`)

## Summary

Fixes the wrong gate on the first-boot wizard. Today the dashboard reads `localStorage.getItem('dream-dashboard-visited')` to decide whether to show the wizard — a flag that's per-browser. A re-imaged device, a different phone, or a cleared cache all re-trigger the wizard for that browser while an actually-onboarded device still looks unconfigured to a fresh visitor.

After this PR, "is this device onboarded?" is a server-side answer sourced from the existing `setup-complete.json` sentinel via `/api/setup/status`. Every browser sees the same truth.

This is the prerequisite for PR-7 (first-boot wizard SPA), which adds the standalone `/setup` route and locks down other routes until completion. PR-6 ships the smaller, focused gate change first so the SPA change is reviewable on its own.

## Files

| File | Change |
|---|---|
| `src/hooks/useFirstRun.js` (new, 45 lines) | Fetches `/api/setup/status`, returns `{firstRun, loading, error, refresh}`. Defaults `firstRun=false` on API failure (wizard re-appearing on every network blip is worse than occasionally not appearing). |
| `src/App.jsx` | Uses `useFirstRun()` instead of `localStorage`. `dismissFirstRun` re-fetches rather than flipping a local boolean. |
| `src/components/SetupWizard.jsx` | `saveConfig` now POSTs `/api/setup/complete` before calling `onComplete`. localStorage retains wizard *answers* (voice, userName) for in-app consumption but no longer gates the wizard. |
| `src/App.test.jsx` | Mocks `useFirstRun` directly rather than driving the wizard through localStorage. |

## Behaviour matrix

| Scenario | Before this PR | After this PR |
|---|---|---|
| Fresh device, first open from phone A | Wizard shows | Wizard shows |
| Same device, open from phone B after A completed wizard | **Wizard shows again (wrong)** | Wizard hidden ✓ |
| Same device, phone A clears browser cache | **Wizard shows again (wrong)** | Wizard hidden ✓ |
| Re-imaged device (data wiped), open from phone A | Wizard hidden if `dream-dashboard-visited` is still set | Wizard shows ✓ |
| dashboard-api unreachable | Wizard never shows (localStorage default) | Wizard never shows (safe default) |
| dashboard-api returns `first_run: true` | n/a (no API used) | Wizard shows ✓ |

## What's NOT in this PR

- **`/setup` standalone route + route-restriction.** The plan called for "all routes redirect to `/setup` until sentinel writes." That's PR-7 (wizard SPA) — bigger change, deserves its own review. Today the wizard remains an overlay modal on the dashboard, just driven by the server flag.
- **Phone-optimized wizard layout.** Existing SetupWizard.jsx is already 6 steps with PreFlight + TemplatePicker; rewriting for phone-first is also PR-7.
- **A wizard-progress server endpoint.** The progress step is still in `setup-progress.json` on the server — useful but not required for gating. Untouched by this PR.

## Test plan

- [x] `npm run build` succeeds
- [x] `npm run lint` clean (zero new warnings)
- [x] `npx vitest run` — 12/12 files, 62/62 tests pass including 4 in App.test.jsx
- [ ] Manual smoke on running dashboard:
  - [ ] Fresh install (`setup-complete.json` absent) → wizard appears on first load
  - [ ] Complete wizard → POST hits server → reload → wizard does NOT reappear
  - [ ] Clear browser cache → reload → wizard still does NOT appear (server is the gate)
  - [ ] Delete `setup-complete.json` on the server → reload → wizard reappears
  - [ ] Stop dashboard-api → reload → app renders normally without wizard (safe fallback)

## Caveats

- **Failure mode is "hide the wizard"**, not "show it". Rationale in the hook's source: an admin who hits a working app and then types `/setup` to re-trigger the wizard is fine; an admin who can't reach the app because the wizard re-appears every time the API blips is a regression. PR-7 adds the deeper `/setup` route for explicit re-entry.
- **localStorage is no longer cleared on wizard completion**, but its key (`dream-dashboard-visited`) is also no longer read. Stale keys will remain in browsers that previously visited; harmless. A cleanup pass on the storage layout could happen but feels over-engineered for this PR.
- **No new admin endpoints.** This PR consumes only the existing `/api/setup/status` and `/api/setup/complete` — both already auth-gated by the standard `verify_api_key` dependency.

### DIFF
--- a/dream-server/extensions/services/dashboard/src/App.jsx
+++ b/dream-server/extensions/services/dashboard/src/App.jsx
@@ -4,6 +4,7 @@ import Sidebar from './components/Sidebar'
 import SetupWizard from './components/SetupWizard'
 import { useSystemStatus } from './hooks/useSystemStatus'
 import { useVersion } from './hooks/useVersion'
+import { useFirstRun } from './hooks/useFirstRun'
 import { getInternalRoutes } from './plugins/registry'
 import SplashScreen from './components/SplashScreen'
 
@@ -30,26 +31,24 @@ function App() {
   )
   const { status, loading, error } = useSystemStatus()
   const { version, dismissUpdate } = useVersion()
-  const [firstRun, setFirstRun] = useState(false)
+  // Server-side first-run flag (sourced from /api/setup/status). localStorage
+  // was per-browser and gave the wrong answer on re-imaged devices or fresh
+  // browsers. The hook returns firstRun=false while it's loading or if the
+  // API call fails, so the normal app shell is the safe default.
+  const { firstRun, refresh: refreshFirstRun } = useFirstRun()
   const [sidebarCollapsed, setSidebarCollapsed] = useState(() => {
     return getStorageValue(globalThis.localStorage, 'dream-sidebar-collapsed') === 'true'
   })
 
   useEffect(() => {
-    const hasVisited = getStorageValue(globalThis.localStorage, 'dream-dashboard-visited')
-    if (!hasVisited) {
-      setFirstRun(true)
-    }
-  }, [])
-
-  useEffect(() => {
     setStorageValue(globalThis.localStorage, 'dream-sidebar-collapsed', String(sidebarCollapsed))
   }, [sidebarCollapsed])
 
-  const dismissFirstRun = () => {
-    setStorageValue(globalThis.localStorage, 'dream-dashboard-visited', 'true')
-    setFirstRun(false)
-  }
+  const dismissFirstRun = useCallback(() => {
+    // SetupWizard's saveConfig has already POSTed /api/setup/complete; we
+    // re-fetch the server flag so the next mount sees the new state.
+    refreshFirstRun()
+  }, [refreshFirstRun])
 
   const routes = useMemo(() => getInternalRoutes({ status, loading }), [status, loading])
   const handleToggle = useCallback(() => setSidebarCollapsed(c => !c), [])
--- a/dream-server/extensions/services/dashboard/src/App.test.jsx
+++ b/dream-server/extensions/services/dashboard/src/App.test.jsx
@@ -1,6 +1,7 @@
 import { screen } from '@testing-library/react'
 import { render } from './test/test-utils'
 import App from './App' // eslint-disable-line no-unused-vars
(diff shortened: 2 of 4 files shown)


Lightheartdevs:
Deep audit: this one looks merge-ready from my pass.

The change moves first-run gating from browser-local state to the existing server sentinel exposed by `/api/setup/status`, and installers already write `data/config/setup-complete.json` to avoid re-showing the wizard after normal installs. Existing `/api/setup/complete` persists the sentinel, so the dashboard no longer depends on each browser having a `dream-dashboard-visited` flag.

Local checks run:
- `npm.cmd run build`
- `npm.cmd test` (12 files / 62 tests passed)

CI is green. I do see the same repository-rule blocker as #1154: GitHub requires two write-access approvals, and the PR author cannot self-approve, so I cannot merge it directly from this account.

Lightheartdevs:
Fresh audit after rebasing over current `main`: merge-ready.

What I checked:
- The old per-browser `dream-dashboard-visited` gate is removed from `App.jsx`; first-run now follows `/api/setup/status` and the server `setup-complete.json` sentinel.
- `SetupWizard` still keeps answer data in `localStorage`, but completion now POSTs `/api/setup/complete`, which is the endpoint that writes the server sentinel.
- Existing installer paths already write the same sentinel under `data/config/setup-complete.json`, so traditional installs should not suddenly re-show the wizard.
- Failure mode is conservative for availability: if setup-status cannot be fetched, the hook hides first-run rather than trapping users behind an overlay during API hiccups.

Local checks:
- `npm.cmd test` -> 15 files / 77 tests passed
- `npm.cmd run build` -> pass
- `npm.cmd run lint` -> pass with 5 pre-existing warnings outside this PR
- `git diff --check` -> pass

GitHub checks are green on rebased head `db425137a51a90b26be3ad34fa1072ed329b5eca`.

PR #1158, opened 2026-05-12
feat(dashboard): phone-first first-boot wizard SPA

Lightheartdevs:
# PR-7: Phone-first first-boot wizard SPA

**Branch:** `feat/first-boot-wizard` (local, not pushed — stacked on `feat/first-boot-sentinel`)
**Effort:** Medium — 3 files (1 new, 2 modified), 599 insertions / 12 deletions
**Depends on:** PR-6 (useFirstRun hook) and PR-4 (magic-link API for the final step)

## Summary

Replaces the modal SetupWizard with a fullscreen, phone-optimized 4-screen wizard. When the server reports `first_run=true`, App.jsx mounts ONLY the FirstBoot SPA — no sidebar, no other routes accessible. Single path, no escape hatches.

This is the surface your user-meets-device interaction lives on. By the end of the four screens they have:

1. A device name (so `<name>.local` works via mDNS)
2. A named first user
3. A picked stack template
4. A magic-link QR ready to scan or share

## The four screens

| Step | What | Validation |
|---|---|---|
| 1 — Welcome | "Name your device" input. Inline shows `<name>.local`. | `^[a-z0-9-]{1,32}$` (hostname-safe) |
| 2 — First user | "Who's the first user?" input. Server-side magic-link API will refuse anything outside the pattern. | `^[A-Za-z0-9._-]{1,64}$` — matches PR-4's server pattern exactly |
| 3 — Stack picker | Three full-width tappable cards: Chat only / Chat + Agents / Everything. Icons + one-line blurbs. | Always valid (defaults to "chat") |
| 4 — Confirm | Summary of choices in a `<dl>` card. "Finish" button generates the magic-link via `POST /api/auth/magic-link/generate`, then flips the server sentinel via `POST /api/setup/complete`. | — |

After step 4 the user lands on a **Done** screen that shows the QR (from `GET /api/auth/magic-link/qr`), the URL with copy-to-clipboard, native `navigator.share()` on supported browsers, and an "Open dashboard" CTA.

## Files

| File | Change |
|---|---|
| `src/pages/FirstBoot.jsx` (new, ~525 lines) | The wizard. Self-contained — one component per step, shared progress state in the parent, all I/O via `fetch`. Reuses PR-4 endpoints, no new dashboard-api endpoints needed. |
| `src/App.jsx` | When `firstRun=true`, render `FirstBoot` fullscreen via lazy import + Suspense, hide the sidebar entirely. When `firstRun=false`, normal app shell. Removed the inline `SetupWizard` modal. |
| `src/App.test.jsx` | Updated the two first-run cases to mock `FirstBoot` rather than `SetupWizard`, plus assert the sidebar is absent during onboarding. |

## Design choices

- **Progress persistence in localStorage**, not server state. The wizard answers are local-only until Finish; if you bail mid-wizard the next visit picks up where you left off. Cleared on success. Server learns about the choices only when Finish runs.
- **Final step does the actual work**, doesn't just route. Generating the magic-link and flipping the sentinel both happen on the same tap. A network blip rolls back cleanly — sentinel hasn't flipped, wizard is still active, user retries.
- **The Done screen owns the moment.** Generating a magic-link the moment the user finishes the wizard means the very first thing they see is a QR they can hand to their phone. No "now go to Invites and create one" dance.
- **Sidebar hidden during onboarding.** The user shouldn't be tempted to click into Settings before the device is configured — every existing dashboard surface assumes a configured device. Locking it out simplifies what we have to make work in PR-7.
- **Lazy-loaded.** `FirstBoot-*.js` is 12.5 KB / 3.7 KB gzipped; you only pay for it on first-boot. After onboarding, the import is dead code from the user's perspective.
- **No new dashboard-api endpoints.** Everything goes through existing surfaces: `/api/setup/status` (PR-6), `/api/setup/complete` (already existed), `/api/auth/magic-link/generate` (PR-4), `/api/auth/magic-link/qr` (PR-4).

## What's NOT in this PR

- **Writing `DREAM_DEVICE_NAME` to .env.** The device name is collected from the user but isn't yet persisted to the env file — that needs the host-agent's env-update endpoint and warrants its own PR (and ties into PR-1's mDNS reloading the name every 30s). Punted: a follow-up wires this in.
- **Applying the chosen stack template.** Step 3's pick is captured but doesn't yet call `/api/templates/apply`. Same reason as above — applying a template is a multi-minute operation with progress UI implications; deserves its own PR.
- **WiFi setup step.** The plan called for "WiFi setup or skip" as step 1. WiFi setup needs the host-agent's network endpoints (PR-8) which doesn't exist yet. Punted to PR-8 — wizard reorders to put WiFi first when PR-8 lands.
- **Password / auth setup.** Open WebUI has its own auth flow; we don't manage credentials in this wizard. The magic-link redemption gets them past first-touch; their account creation in Open WebUI happens on their first chat tap.
- **Re-run wizard from Settings.** The wizard component file still exists for the modal SetupWizard, and Settings can route to it later. Not exposed yet — first-boot is the only way to see the wizard today.

## Test plan

- [x] `npm run build` succeeds; FirstBoot lazy-loads as a separate chunk (12.5 KB / 3.7 KB gzipped)
- [x] `npm run lint` clean (zero new warnings)
- [x] `npx vitest run` — 62/62 tests pass; the two first-run cases assert `[data-testid="first-boot"]` appears AND `<aside>` (sidebar) does not
- [ ] Manual smoke:
  - [ ] Fresh device (no `setup-complete.json`) → loads straight into FirstBoot, no sidebar
  - [ ] Step 1 rejects invalid device names (e.g. `dream space`, empty, 33+ chars)
  - [ ] Step 2 rejects usernames matching neither pattern
  - [ ] Step 3 selection highlights one card at a time
  - [ ] Step 4 Finish → spins → lands on Done with QR
  - [ ] QR scans → opens chat 302 path (PR-4 redemption flow)
  - [ ] Copy button copies URL; Share button visible on mobile
  - [ ] Open dashboard button → onComplete fires → useFirstRun refresh → normal dashboard
  - [ ] Refresh mid-wizard restores the same step + inputs
  - [ ] Refresh after Finish → does NOT re-show the wizard (server sentinel flipped)

## Caveats

- **The wizard runs even if the dashboard-api can't immediately reach the magic-link backend.** If `POST /api/auth/magic-link/generate` fails, the Finish step shows an error and the user can retry. The sentinel is only flipped after a successful magic-link generation, so a retry is safe — no half-onboarded state.
- **The QR is best-effort.** If the `qrcode` python library isn't installed on the dashboard-api side, the QR slot falls back to a placeholder and Copy / Share still work. Same fallback as PR-5's Invites page.
- **The "Open dashboard" CTA puts the admin into the dashboard, NOT the chat surface.** The QR they just generated is intended for the recipient device. The admin who just ran the wizard probably wants to confirm everything works, look at Settings, generate more invites — that's the dashboard.

### DIFF
--- a/dream-server/extensions/services/dashboard/src/App.jsx
+++ b/dream-server/extensions/services/dashboard/src/App.jsx
@@ -4,6 +4,7 @@ import Sidebar from './components/Sidebar'
 import SetupWizard from './components/SetupWizard'
 import { useSystemStatus } from './hooks/useSystemStatus'
 import { useVersion } from './hooks/useVersion'
+import { useFirstRun } from './hooks/useFirstRun'
 import { getInternalRoutes } from './plugins/registry'
 import SplashScreen from './components/SplashScreen'
 
@@ -30,26 +31,24 @@ function App() {
   )
   const { status, loading, error } = useSystemStatus()
   const { version, dismissUpdate } = useVersion()
-  const [firstRun, setFirstRun] = useState(false)
+  // Server-side first-run flag (sourced from /api/setup/status). localStorage
+  // was per-browser and gave the wrong answer on re-imaged devices or fresh
+  // browsers. The hook returns firstRun=false while it's loading or if the
+  // API call fails, so the normal app shell is the safe default.
+  const { firstRun, refresh: refreshFirstRun } = useFirstRun()
   const [sidebarCollapsed, setSidebarCollapsed] = useState(() => {
     return getStorageValue(globalThis.localStorage, 'dream-sidebar-collapsed') === 'true'
   })
 
   useEffect(() => {
-    const hasVisited = getStorageValue(globalThis.localStorage, 'dream-dashboard-visited')
-    if (!hasVisited) {
-      setFirstRun(true)
-    }
-  }, [])
-
-  useEffect(() => {
     setStorageValue(globalThis.localStorage, 'dream-sidebar-collapsed', String(sidebarCollapsed))
   }, [sidebarCollapsed])
 
-  const dismissFirstRun = () => {
-    setStorageValue(globalThis.localStorage, 'dream-dashboard-visited', 'true')
-    setFirstRun(false)
-  }
+  const dismissFirstRun = useCallback(() => {
+    // SetupWizard's saveConfig has already POSTed /api/setup/complete; we
+    // re-fetch the server flag so the next mount sees the new state.
+    refreshFirstRun()
+  }, [refreshFirstRun])
 
   const routes = useMemo(() => getInternalRoutes({ status, loading }), [status, loading])
   const handleToggle = useCallback(() => setSidebarCollapsed(c => !c), [])
--- a/dream-server/extensions/services/dashboard/src/App.jsx
+++ b/dream-server/extensions/services/dashboard/src/App.jsx
@@ -1,12 +1,16 @@
 import { Routes, Route } from 'react-router-dom'
-import { useState, useEffect, Suspense, useMemo, useCallback } from 'react'
+import { useState, useEffect, Suspense, useMemo, useCallback, lazy } from 'react'
 import Sidebar from './components/Sidebar'
-import SetupWizard from './components/SetupWizard'
 import { useSystemStatus } from './hooks/useSystemStatus'
 import { useVersion } from './hooks/useVersion'
 import { useFirstRun } from './hooks/useFirstRun'
 import { getInternalRoutes } from './plugins/registry'
 import SplashScreen from './components/SplashScreen'
+
+// Phone-first first-boot wizard. Mounted instead of the normal app shell
+// when useFirstRun() reports firstRun=true. Lazy-loaded so the wizard
+// bundle isn't paid for on every page load after onboarding.
+const FirstBoot = lazy(() => import('./pages/FirstBoot'))
 
 function getStorageValue(storage, key) {
   try {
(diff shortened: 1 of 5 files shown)


Lightheartdevs:

1. This branch is stacked only on #1157 (`feat/first-boot-sentinel`), but `FirstBoot.finish()` depends on the magic-link API from #1155: `/api/auth/magic-link/generate` and `/api/auth/magic-link/qr`. If #1157 + #1158 land without #1155, first-run users hit the full-screen wizard, press Finish, get a 404/error, and cannot reach the normal dashboard because App.jsx locks the shell out while `firstRun=true`. This needs to be stacked on the API work too, or the first-boot wizard must not depend on invite generation to complete onboarding.

2. The wizard collects a device name and tells the user it will be reachable as `<name>.local`, but it never persists `DREAM_DEVICE_NAME` or calls any backend endpoint that updates `.env`/mDNS state. Today `deviceName` is only local component/localStorage progress. That makes the confirmation screen and first step misleading: the chosen name has no effect after Finish.

3. The stack picker is also currently UI-only. Choosing `Chat + Agents` or `Everything` does not apply a template, enable extensions, start services, or persist a desired stack. If this is intentional placeholder UX, it should not be in the first-run critical path yet; otherwise Finish needs to call the same backend/template machinery the dashboard uses.

4. This also inherits #1155's auth semantics: generated magic links do not create an Open WebUI session yet, so the success copy (“They scan or tap it to land straight in chat”) overpromises while `WEBUI_AUTH` is enabled.

Local checks: `npm.cmd run build` passed and `npm.cmd test` passed (12 files / 62 tests). CI is green, so these are product/integration blockers rather than compile failures.

Lightheartdevs:
Re-audit after the latest update: improved copy, but still not merge-ready.

What improved:
- The stack picker now says extras are enabled later from Extensions instead of implying this wizard applies a template immediately.
- The Open WebUI copy now admits the recipient may still see sign-in until SSO is wired up.
- Local frontend checks pass: `npm.cmd test -- --run` -> 62 passed, `npm.cmd run build` succeeds, and `git diff --check origin/feat/first-boot-sentinel...HEAD` is clean. CI is green.

Remaining blockers:

1. This branch still does not include the magic-link API it calls. `finish()` posts to `/api/auth/magic-link/generate` and later fetches `/api/auth/magic-link/qr`, but this head does not contain `origin/feat/magic-link-auth`. If #1157/#1158 land without the corrected #1155/#1173 chain, first-run users hit the full-screen wizard and Finish fails inside the locked onboarding flow.

2. Device name is still not persisted. The UI says “We'll save this preference” and the confirmation says the device name is “saved as preference,” but `deviceName` is only local component/localStorage state. This head also does not contain the network config backend, so nothing writes `DREAM_DEVICE_NAME` or restarts mDNS/proxy state. The copy is softer than before, but still promises persistence that does not happen.

3. `finish()` does not check `/api/setup/complete`'s response. It `await fetch('/api/setup/complete', { method: 'POST' })` and then shows the success screen even if the sentinel update returns 4xx/5xx. That can leave the server in first-run mode while the UI tells the operator “You're set.” This should check `completeResp.ok` before clearing progress/showing success.

4. It inherits #1155's current backend issues: missing generated `DREAM_SESSION_SECRET`, default public URL/proxy dependency, and partial Open WebUI auth semantics.

Net: the wording is safer, but the wizard is still ahead of its backend dependencies and still has a real completion-state bug.

PR #1159, opened 2026-05-12
feat(host-agent,dashboard-api): Wi-Fi scan/connect/status endpoints

Lightheartdevs:
# PR-8: Network configuration backend (Wi-Fi)

**Branch:** `feat/network-config-backend`
**Effort:** Medium — 4 files (2 new, 2 modified), 804 insertions / 2 deletions
**Depends on:** none directly (the host-agent already exists; this adds endpoints to it)

## Summary

Adds the Wi-Fi scan / connect / status / forget endpoints needed for the first-boot wizard's network step. Linux + NetworkManager (`nmcli`) is the supported path today; macOS and Windows return a clear "platform not supported" response so the wizard can fall back to "use Ethernet."

UI is intentionally NOT in this PR — it sits on top of PR-7's wizard and would balloon the diff. PR-8 ships the backend so the UI work can land independently and review focuses on the security-sensitive part (handling Wi-Fi passwords).

## API surface

| Method | Path | Auth | Purpose |
|---|---|---|---|
| GET  | `/api/setup/wifi-scan` | Bearer | Rescan + return nearby SSIDs (strongest first) |
| POST | `/api/setup/wifi-connect` | Bearer | Join a network — `{ssid, password}` |
| POST | `/api/setup/wifi-forget` | Bearer | Delete saved connection profile |
| GET  | `/api/setup/network-status` | Bearer | Current device, IP, gateway, wifi flag |

All four are thin proxies in front of the host-agent (`/v1/network/...`) — the dashboard-api never touches nmcli directly. Same architecture pattern as the existing `.env`-write / Docker-recreate paths.

## Files

| File | Lines | What |
|---|---|---|
| `bin/dream-host-agent.py` | +283 | Four new handlers + `_network_supported()` precondition helper. Linux + `nmcli` required, otherwise 501 with a helpful message. |
| `extensions/services/dashboard-api/routers/setup.py` | +142 / −2 | Four proxy endpoints with Pydantic validation. New `_call_agent()` helper that never logs request payloads (so Wi-Fi passwords stay out of logs). |
| `extensions/services/dashboard-api/tests/test_network_config.py` | +198 (new) | 15 tests covering auth, happy paths, Pydantic validation, error translation, and the agent-unreachable case. |
| `docs/NETWORK.md` | +181 (new) | Platform matrix, full API reference, security notes, troubleshooting. |

## Security posture

The whole reason this is a host-agent feature rather than a container feature is that the password lives on the wire and on disk briefly — keeping it out of the container surface limits exposure.

- **Passwords never reach logs.** The host-agent log entry is `wifi-connect ssid=<name> password_set=true`. The proxy in `dashboard-api` doesn't log the request body either — `_call_agent()` logs only path + status.
- **Validation at both layers.** Pydantic validates the dashboard-api side (length, no control chars); the host-agent also re-validates the same constraints because endpoints are independently callable.
- **Error messages don't echo secrets.** nmcli failure paths are translated to opaque strings (`Wrong password` / `Network not found` / `Connection timed out`) — we don't pass through raw nmcli stderr to the client.
- **Password lifetime on argv.** nmcli accepts the password via `password <pwd>` on the command line. On modern Linux with `kernel.yama.ptrace_scope >= 1` (Ubuntu/Fedora default) and the host-agent running as root, the cmdline isn't readable from other users. Documented as a v1 posture; tightening to a connection-profile-file pattern is a follow-up.
- **Platform check returns 501, not 200**, so a misconfigured client can't silently no-op a join.

## Error translation

nmcli's exit messages aren't UI-friendly. The host-agent maps:

| nmcli output contains | Returned as |
|---|---|
| `Secrets were required` / `(7)` | `Wrong password` (400) |
| `No network with SSID` / `not found` | `Network not found` (400) |
| `timeout` | `Connection timed out` (400) |
| (other) | First 300 chars of raw stderr (400) |

Plus standard 501 (platform unsupported), 503 (agent unreachable), 504 (subprocess timeout).

## Test plan

```
pytest tests/test_network_config.py -v
```

**15 tests passing locally**, covering:

- [x] Auth required on all four endpoints (401 when missing)
- [x] Happy paths for scan, connect, status, forget
- [x] `wifi-connect` rejects oversized SSID (>32 chars) → 422
- [x] `wifi-connect` rejects oversized password (>63 chars) → 422
- [x] `wifi-connect` rejects control characters in SSID → 422
- [x] `wifi-connect` translates "Wrong password" upstream → 400 with that message
- [x] Upstream 504 timeout propagates as 504
- [x] Upstream URLError (agent down) → 503
- [x] 501 from unsupported platform propagates with the platform info
- [x] `network-status` always returns 200 (even when unsupported)

**Manual test plan** (requires real Linux + nmcli; not reproducible in CI):

- [ ] Fresh Ubuntu 24.04 install → install Dream Server → `curl localhost:3002/api/setup/wifi-scan` returns local networks
- [ ] Connect to a known good network → `network-status` reflects it
- [ ] Connect with wrong password → 400 "Wrong password"
- [ ] Connect to nonexistent SSID → 400 "Network not found"
- [ ] On macOS / Windows → 501 with platform message
- [ ] On Linux without NetworkManager → 501 "nmcli not found"
- [ ] Wi-Fi password absent from agent logs after a connect attempt

## What's NOT in this PR

- **Wizard UI step.** The wizard's WiFi screen would consume these endpoints; sits on top of PR-7. Deferred to its own PR.
- **wpa_supplicant / systemd-networkd support.** Only NetworkManager today. Documented in `NETWORK.md`.
- **macOS / Windows Wi-Fi.** Both have different stacks (CoreWLAN / netsh); deferred.
- **Network-profile QR codes.** The PWA's "Add to network" QR pattern (Android Wi-Fi-config QR) is a separate piece of work — likely a setup-card generator (PR-9).
- **Connection-profile-file pattern.** Today the password reaches nmcli via argv. Migrating to a NetworkManager keyring connection file would close the brief argv exposure window. Documented as a follow-up.

## Caveats

- **`nmcli rescan` is best-effort.** A genuinely flaky radio (rfkill, low signal) can return an empty list. The wizard should show a "no networks found — try again" state, not assume the radio is broken.
- **The dashboard-api endpoints all run via `asyncio.to_thread`** because `urllib.request.urlopen` is synchronous. Acceptable at the expected QPS (one or two calls per wizard run); migrating to `httpx`/async is not blocking this PR.

### DIFF
--- a/dream-server/bin/dream-host-agent.py
+++ b/dream-server/bin/dream-host-agent.py
@@ -641,6 +641,24 @@ def json_response(handler, code: int, body: dict):
     handler.wfile.write(payload)
 
 
+def _network_supported(handler) -> bool:
+    """Linux + nmcli precondition for Wi-Fi endpoints. Sends a 501 on failure
+    so the caller doesn't need to repeat the check; returns True only when
+    nmcli is callable.
+    """
+    if platform.system() != "Linux":
+        json_response(handler, 501, {
+            "error": f"Wi-Fi management only supported on Linux (this is {platform.system()})",
+        })
+        return False
+    if shutil.which("nmcli") is None:
+        json_response(handler, 501, {
+            "error": "nmcli not found; install NetworkManager to enable Wi-Fi management",
+        })
+        return False
+    return True
+
+
 def check_auth(handler) -> bool:
     auth = handler.headers.get("Authorization", "")
     if not auth.startswith("Bearer "):
@@ -904,6 +922,10 @@ class AgentHandler(BaseHTTPRequestHandler):
             self._handle_model_list()
         elif self.path == "/v1/model/status":
             self._handle_model_status()
+        elif self.path == "/v1/network/wifi-scan":
+            self._handle_network_wifi_scan()
+        elif self.path == "/v1/network/status":
+            self._handle_network_status()
         else:
             json_response(self, 404, {"error": "Not found"})
 
@@ -1012,6 +1034,10 @@ class AgentHandler(BaseHTTPRequestHandler):
             self._handle_invalidate_compose_cache()
         elif self.path == "/v1/env/update":
             self._handle_env_update()
+        elif self.path == "/v1/network/wifi-connect":
+            self._handle_network_wifi_connect()
+        elif self.path == "/v1/network/wifi-forget":
+            self._handle_network_wifi_forget()
         else:
             json_response(self, 404, {"error": "Not found"})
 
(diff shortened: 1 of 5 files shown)


Lightheartdevs:
I audited this one in depth. The shape is useful, but I don't think it is merge-ready yet.

Findings:

1. CI is red on Ruff, and I reproduced it locally:
   - `extensions/services/dashboard-api/routers/setup.py:12` imports unused `Any`
   - `extensions/services/dashboard-api/routers/setup.py:15` imports unused `Body`
   - `extensions/services/dashboard-api/tests/test_network_config.py:60` has an f-string with no placeholders

2. `wifi-forget` can delete any NetworkManager connection profile, not just Wi-Fi profiles. `bin/dream-host-agent.py:1170-1205` validates the provided name and then runs `nmcli connection delete <connection>` directly. Even with API auth, this endpoint is documented as Wi-Fi forget, but it can delete wired/VPN/other NetworkManager profiles and potentially cut off the host. I would gate deletion by inspecting the profile type first (`802-11-wireless`/wifi), or only allow deleting profiles that the Wi-Fi status/listing code has positively identified as Wi-Fi.

3. `nmcli -t` parsing is fragile for SSIDs / connection names containing `:`. The scan path explicitly disables escaping with `-e no` and then does `line.split(:)` at `bin/dream-host-agent.py:1061-1084`; network status also uses `line.split(:)` at `bin/dream-host-agent.py:1238-1242`. That can corrupt SSID/security/connection fields or silently drop valid networks. Use escaped output with unescaping, a safer delimiter strategy, or request fields in a format that cannot collide with SSID contents.

4. The new dashboard-api tests are useful, but they only prove the proxy behavior. For this kind of host-network mutation, I would add mocked host-agent tests around the new handlers themselves: scan parsing, unsupported platform / missing nmcli, connect error mapping, timeout behavior, and especially the `wifi-forget` type restriction above.

Local checks run:

```text
python -m pytest extensions/services/dashboard-api/tests/test_network_config.py -q
# 17 passed

python -m ruff check bin/dream-host-agent.py extensions/services/dashboard-api/routers/setup.py extensions/services/dashboard-api/tests/test_network_config.py
# fails with the three Ruff issues listed above
```

Net: valuable PR, but it needs the lint cleanup and the host-agent safety/parsing fixes before I would merge it.

Lightheartdevs:
Re-review after the latest author update: this looks ready from my side.

What changed since my prior pass:
- `wifi-forget` now verifies the NetworkManager connection type before deletion, so it no longer exposes a generic “delete any profile by name” footgun.
- `nmcli -t` parsing now keeps nmcli escaping enabled and uses a dedicated parser for escaped `:` / `\` values, so SSIDs and connection names containing colons no longer corrupt the field split.
- Added dashboard-api proxy coverage plus targeted host-agent parser tests.
- CI is green on the current head.

Local verification:
- `python -m ruff check bin/dream-host-agent.py extensions/services/dashboard-api/routers/setup.py extensions/services/dashboard-api/tests/test_network_config.py extensions/services/dashboard-api/tests/test_host_agent.py` passed.
- `python -m pytest extensions/services/dashboard-api/tests/test_network_config.py -q` passed: 17 passed.
- `python -m pytest extensions/services/dashboard-api/tests/test_host_agent.py::TestSplitNmcliTerse -q` passed: 7 passed.
- `python -m compileall -q bin/dream-host-agent.py extensions/services/dashboard-api/routers/setup.py` passed.

I do not see a remaining merge blocker in this PR. I cannot approve it from this account because GitHub treats it as my own PR, but it is a merge candidate once the required review gate is satisfied.

Lightheartdevs:
Fresh audit/fix pass: merge-ready.

What I re-checked:
- Dashboard-api proxy endpoints require auth, validate request shape, and do not log Wi-Fi passwords.
- Host-agent support is gated to Linux + nmcli for mutating endpoints, and status returns a structured fallback for unsupported/nonfunctional network stacks.
- `wifi-forget` now refuses non-Wi-Fi NetworkManager profiles before deletion.
- nmcli terse output parsing handles escaped colons/backslashes in SSIDs and connection names.

Fixes I pushed in this pass:
- `wifi-scan` now collapses duplicate SSIDs by keeping the strongest BSSID instead of whichever row nmcli returned first.
- `network-status` now reports `platform_supported: false` with a reason when `nmcli device status` itself fails, instead of silently returning an empty supported status.
- Added direct host-agent tests for strongest duplicate SSID selection, non-Wi-Fi forget refusal, and nmcli status failure reporting.

Local checks:
- `python -m pytest dream-server/extensions/services/dashboard-api/tests/test_network_config.py -q` -> 17 passed
- `python -m pytest dream-server/extensions/services/dashboard-api/tests/test_host_agent.py::TestSplitNmcliTerse dream-server/extensions/services/dashboard-api/tests/test_host_agent.py::TestNetworkHandlers -q` -> 10 passed
- `python -m ruff check ...` -> pass
- `python -m compileall -q dream-server/bin/dream-host-agent.py dream-server/extensions/services/dashboard-api/routers/setup.py` -> pass
- `git diff --check` -> pass
- `dream-server/tests/test-doc-links.sh` via Git Bash -> pass

GitHub checks are green on head `252819fb20dcd4a480645f3469c03fed74335b36`.

PR #1160, opened 2026-05-12
feat(scripts): printable setup-card PNG generator

Lightheartdevs:
# PR-9: Printable setup-card generator

**Branch:** `feat/setup-card-generator`
**Effort:** Small — 3 files (all new), 543 insertions
**Depends on:** none (independent operator-side tool)

## Summary

A CLI utility for fulfillment. Feed it a per-device Wi-Fi AP SSID + password + setup URL, get back a 4×6 portrait PNG ready to print, laminate, and drop in the box. The recipient scans two QRs — first joins the setup AP, second opens the first-boot wizard — and they're in.

This is a tooling artifact, not a runtime feature. The device never prints its own credentials; this runs on the operator's machine or in the fulfillment pipeline.

## Sample output

A 1200×1800 px (4×6 inches @ 300 DPI) portrait card with:

1. **Top band** — "DREAM SERVER" wordmark in brand purple, the unit's mDNS name (e.g. `dream-a4f2.local`), "Scan to set up. Scan to chat." tagline.
2. **Two QR codes side by side:**
   - **Left — JOIN WI-FI:** Encodes the standard `WIFI:T:WPA;S:<ssid>;P:<password>;;` URI that iOS Camera and Android system scanner auto-recognize. One tap joins the AP.
   - **Right — OPEN SETUP:** Plain URL QR. Opens straight to `/setup` on the device's setup-mode IP.
3. **Plain-text fallback block** — `NETWORK / PASSWORD / THEN VISIT` for phones whose camera won't scan a QR.
4. **Footer** — "DreamServer is open-source — light-heart-labs.com", plus optional per-unit serial in the right corner.

Verified visually with a sample card (`Dream-Setup-A4F2` / `dream-a4f2.local` / `DRM-2026-A4F2`).

## Files

| File | Lines | What |
|---|---|---|
| `scripts/generate-setup-card.py` | 248 | The CLI. Pillow + qrcode for layout/QR; both imported lazily so `--help` works without them. Best-effort font loader walks Consolas / DejaVu / Helvetica based on OS. |
| `docs/SETUP-CARD.md` | 100 | Operator docs: usage, batch-generation example, design choices, security notes, called-out limitations. |
| `tests/test_setup_card.py` | 195 | 9 tests: WiFi URI escape rules, CLI happy path, PNG validity, dimensions, parent-dir creation, missing-flag handling, `--help`-without-Pillow. |

## CLI

```bash
python3 scripts/generate-setup-card.py \
  --ssid 'Dream-Setup-A4F2' \
  --password 'xxxxxxxx' \
  --setup-url 'http://192.168.7.1/setup' \
  --device-name 'dream-a4f2.local' \
  --serial 'DRM-2026-A4F2' \
  --output cards/card-A4F2.png
```

| Flag | Required | Notes |
|---|---|---|
| `--ssid` | yes | Wi-Fi SSID of the device's setup AP |
| `--password` | no | Wi-Fi password. Omit for open networks. |
| `--security` | no | `WPA` (default), `WEP`, `nopass` |
| `--setup-url` | yes | URL embedded in the second QR |
| `--device-name` | no | mDNS hostname printed on the card (default `dream.local`) |
| `--serial` | no | Optional fulfillment / batch ID |
| `--output` / `-o` | yes | Output PNG path. Parent dirs created if missing. |

Exit codes: `0` success; `2` if Pillow / qrcode aren't installed (with install hint).

## Design choices

- **PNG, not PDF.** Printers handle PNG fine. PDF would require adding `reportlab` (a heavy dep) for what amounts to a wrapper around the same image.
- **Pillow composition, not SVG template.** A pre-baked template would be more "designed" but harder to maintain across font fallback and variable-length text. Procedural layout fits text to content cleanly.
- **Lazy imports.** `--help` works without Pillow installed so operators can check flags before installing deps.
- **No on-device generation.** The device never knows its AP credentials at runtime — those are baked in during the image-build step. Printing the credentials from the device would also be a self-leak vector.

## What's NOT in this PR

- **PDF output.** Trivial follow-up if a fulfillment pipeline needs it (`card.save(path, format="PDF")`). PNG is the canonical format here.
- **Multi-language.** "JOIN WI-FI" / "OPEN SETUP" / fallback labels are English-only. Localizable as `--locale` later if it matters.
- **A Dream logo SVG.** The wordmark is set in the system bold font today. When a real `dream-logo.svg` exists, dropping it in is a small change.
- **Hidden-SSID flag.** WIFI URI supports `H:true;`; today we hardcode `H:false`. Trivial to expose later.
- **Setup-card output endpoint on the dashboard.** The wizard from PR-7 could in principle expose a "regenerate my setup card" button. Out of scope here — this is a CLI tool.

## Test plan

```bash
cd dream-server
pytest tests/test_setup_card.py -v
```

**9 tests, all passing locally:**

- [x] WiFi URI builds the spec-correct format
- [x] Open network (no password) omits the `P:` segment
- [x] WIFI URI escape rules (\\, ;, ,, :, ") apply in values
- [x] CLI writes a valid PNG (magic-number check)
- [x] CLI creates the output's parent directory
- [x] CLI output is exactly 1200×1800 (the documented 4×6 @ 300 DPI)
- [x] Missing `--ssid` errors out
- [x] Missing `--setup-url` errors out
- [x] `--help` works even without Pillow / qrcode installed

**Manual sample**: ran with `Dream-Setup-A4F2 / dream-a4f2.local / DRM-2026-A4F2`; output renders cleanly with both QR codes scannable on a phone camera and the fallback block readable.

## Caveats

- **The plaintext password is on the card by design.** Without it the QR fallback doesn't work. Treat the card as a physical credential — if the unit is resold, regenerate the AP creds and discard the card.
- **The setup URL doesn't carry the credential.** It points at the setup-mode IP (`192.168.7.1` by convention); if the recipient hasn't joined the AP yet, the URL just times out. The QRs depend on being scanned in order.
- **Font selection varies by OS.** The card always renders, but typography on a stripped-down container (e.g. Alpine with no system fonts) falls back to Pillow's bitmap font, which is functional but ugly. The fulfillment machine should have DejaVu / Helvetica / Consolas, which all modern OSs ship.

### DIFF
--- a/dream-server/docs/SETUP-CARD.md
+++ b/dream-server/docs/SETUP-CARD.md
@@ -1,0 +1,84 @@
+# Setup card generator
+
+`scripts/generate-setup-card.py` produces a printable 4×6 PNG for fulfillment: drop one in the box with each Dream Server unit so the recipient can scan two QR codes and reach the wizard without typing anything.
+
+## What it generates
+
+A portrait card with:
+
+1. **Top band** — "DREAM SERVER" wordmark, the unit's mDNS name (e.g. `dream.local`), and a one-line "Scan to set up. Scan to chat." tagline.
+2. **Two QR codes side by side:**
+   - **Left — JOIN WI-FI.** Encodes the standard `WIFI:T:WPA;S:<ssid>;P:<password>;;` format that iOS Camera and Android (Chrome / system QR scanner) auto-recognize. Phone joins the AP with one tap.
+   - **Right — OPEN SETUP.** Plain URL QR. The user's browser opens straight to the first-boot wizard at e.g. `http://192.168.7.1/setup`.
+3. **Plain-text fallback block** — SSID, password, URL printed verbatim for phones that won't scan the QR (older models, scratched cards, scanner permission denied).
+4. **Footer** — "DreamServer is open-source — light-heart-labs.com", plus an optional per-unit serial number for fulfillment tracking.
+
+Card is 1200×1800 px at 300 DPI = exactly 4×6 inches. Print on photo cardstock and laminate, or just print on plain paper for prototypes.
+
+## Requirements
+
+`pip install 'qrcode[pil]'` — that's it. `qrcode[pil]` pulls in Pillow as a transitive dep. Both are already in `extensions/services/dashboard-api/requirements.txt`, so if you've installed the dashboard-api you're done.
+
+## Usage
+
+```bash
+python3 scripts/generate-setup-card.py \
+  --ssid 'Dream-Setup-A4F2' \
+  --password 'xxxxxxxx' \
+  --setup-url 'http://192.168.7.1/setup' \
+  --device-name 'dream-a4f2.local' \
+  --serial 'DRM-2026-A4F2' \
+  --output cards/card-A4F2.png
+```
+
+Flags:
+
+| Flag | Required | Notes |
+|---|---|---|
+| `--ssid` | yes | Wi-Fi SSID of the device's setup AP |
+| `--password` | no | Wi-Fi password. Omit for open networks. |
+| `--security` | no | `WPA` (default), `WEP`, or `nopass` |
+| `--setup-url` | yes | URL the QR opens (e.g. `http://192.168.7.1/setup`) |
+| `--device-name` | no | mDNS hostname printed on the card. Defaults to `dream.local`. |
+| `--serial` | no | Optional serial / batch ID printed in the footer right corner |
+| `--output` / `-o` | yes | Output PNG path. Parent directory is created if missing. |
+
+Exit codes: `0` on success, `2` if Pillow / qrcode isn't installed. Argparse handles bad flags.
+
+## Batch generation
+
+Looping the script over a list of pre-baked AP credentials is the recommended fulfillment flow:
+
+```bash
+while IFS=',' read -r ssid password serial; do
+  python3 scripts/generate-setup-card.py \
+    --ssid "$ssid" \
+    --password "$password" \
+    --setup-url "http://192.168.7.1/setup" \
+    --device-name "dream-${serial,,}.local" \
+    --serial "$serial" \
+    --output "cards/$serial.png"
+done < unit-credentials.csv
+```
+
+(`unit-credentials.csv` line format: `Dream-Setup-A4F2,xxxxxxxx,DRM-2026-A4F2`)
+
+## Design choices
+
+- **PNG, not PDF.** Printers handle PNG fine, batch generation is one library call, and the operator can convert with any tool if they really want PDF. Adding `reportlab` for a one-line save would double the dependency footprint.
+- **Pillow for layout, not SVG templates.** A pre-baked template is more flexible visually but harder to maintain (font fallback, variable text wrapping). The Python composition fits the layout to the content rather than the other way around.
+- **Best-effort font loading.** The script tries Consolas / DejaVu / Helvetica based on OS, falls back to Pillow's bitmap font if nothing matches. The card always renders — only the typography varies.
+- **No on-device generation.** This is an operator-side script. The device doesn't know its own AP credentials (those are baked in during the image-build step) and shouldn't be able to print its own credentials anyway.
+
+## Security notes
+
+- The plaintext password is on the card by design — it's what makes the first-touch flow work. Treat the card as a physical credential. If a unit is sold / resold, the new owner should re-generate the AP credentials and discard the card.
+- The setup URL doesn't carry the credential. It points at the device's setup-mode IP; if the recipient hasn't joined the AP yet, the URL just times out.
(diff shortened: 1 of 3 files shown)


Lightheartdevs:
I audited #1160 in depth. It is a useful little fulfillment tool, and the generated sample card is visually coherent, but I would not merge it yet.

Findings:

1. CI is red on Ruff, and I reproduced it locally:
   - `scripts/generate-setup-card.py:105` imports unused `ImageFont`
   - `tests/test_setup_card.py:11` imports unused `os`

2. The docs claim the dependencies are already present in `extensions/services/dashboard-api/requirements.txt`, but they are not. `docs/SETUP-CARD.md:20` says `qrcode[pil]` / Pillow are already there; that requirements file currently contains FastAPI, uvicorn, aiohttp, httpx, pydantic, python-multipart, and PyYAML only. Either add an explicit tooling requirements path or fix the docs so operators do not assume an existing dashboard-api install is enough.

3. `render_qr()` resizes the QR after converting it to RGB without forcing nearest-neighbor scaling (`scripts/generate-setup-card.py:90-93`). Locally, `render_qr(...)` produced 192 distinct grayscale colors instead of a pure black/white QR. It may still scan in many cases, but printed setup cards should be deliberately crisp. Use `Image.Resampling.NEAREST` or generate the QR at the final module size instead of antialiasing the modules.

4. The plaintext fallback block does not fit maximum valid Wi-Fi credentials. The rows at `scripts/generate-setup-card.py:167-180` draw values at a fixed font size and x-position with no wrapping or shrink-to-fit. A 63-character WPA password measured wider than the available value column in my local probe, so the card can silently print the fallback password off the right edge. Since this card intentionally exposes the password as a fallback, that path needs to remain readable for valid input lengths.

Local checks run:

```text
python -m pytest tests/test_setup_card.py -q
# 9 passed

python -m ruff check scripts/generate-setup-card.py tests/test_setup_card.py
# fails with the two Ruff issues listed above

python scripts/generate-setup-card.py --ssid Dream-Setup-A4F2 --password supersecret --setup-url http://192.168.7.1/setup --device-name dream-a4f2.local --serial DRM-2026-A4F2 --output .tmp-card-audit/card-A4F2.png
# generated successfully; visual spot-check looked coherent
```

Net: close, but needs the lint cleanup plus the QR scaling and dependency/docs fixes before I would call it merge-ready. I would also add a regression test for a max-length WPA password so the physical fallback stays usable.

Lightheartdevs:
Re-review after the latest author update: much improved, but I found one remaining cleanup before I would call it merge-ready.

Fixed since my prior audit:
- Ruff now passes.
- The QR resize uses `Image.Resampling.NEAREST`; a local QR probe now has exactly 2 colors instead of antialiased grayscale values.
- The public docs now correctly say this is an operator-side dependency (`pip install 'qrcode[pil]'`) rather than implying dashboard-api already provides it.
- A max-length WPA password regression test was added and passes.

Remaining note:
- `scripts/generate-setup-card.py:30-31` still says: “Requires: Pillow + qrcode (both pip-installable, both already in the dashboard-api requirements).” That is still false on this branch; `extensions/services/dashboard-api/requirements.txt` does not include `qrcode[pil]`/Pillow. Since dependency guidance was one of the original blockers, I would fix this script header too so the tool and docs agree.

Local verification:
- `python -m ruff check scripts/generate-setup-card.py tests/test_setup_card.py` passed.
- `python -m pytest tests/test_setup_card.py -q` passed: 10 passed.
- Generated `.tmp-card-audit/card-1160.png` successfully with a 63-character password.
- QR probe after rendering returned 2 colors.

Net: no functional blocker remains that I can see, but I would clean up the stale dependency sentence before merging.

Lightheartdevs:
Audit pass after rebasing onto current main and tightening two setup-card edge cases.

What I fixed before merge:
- Open-network cards now emit `WIFI:T:nopass` automatically when no password is provided, matching the documented "omit for open networks" behavior.
- `--security nopass` plus `--password` is rejected instead of producing a misleading QR payload.
- QR rendering now keeps the QR-spec four-module quiet zone. Since these codes sit on a dark printed card, that white border matters for scan reliability.
- Cleaned up stale dependency wording and made the batch-generation docs avoid Bash 4-only `${var,,}` expansion.
- Added tests for the open-network behavior, nopass/password rejection, and quiet-zone width.

Local checks run:
- `python -m pytest dream-server/tests/test_setup_card.py -v` (13 passed)
- `python -m py_compile dream-server/scripts/generate-setup-card.py`
- `python -m ruff check dream-server/scripts/generate-setup-card.py dream-server/tests/test_setup_card.py`
- Generated and visually inspected a sample `1200x1800` setup-card PNG.

GitHub checks are green on head `7cd85ffe52f1f835aed3efe75e47f0701191d7b8`. This is a good operator-side piece for the headless setup story: it does not add runtime exposure, and the physical-card credential risk is explicitly documented. Ready to merge.

PR #1162, opened 2026-05-12
feat(wizard,host-agent): AP-mode awareness + WiFi handoff

Lightheartdevs:
# PR-11: AP-mode awareness + WiFi handoff in the wizard

**Branch:** `feat/ap-mode-wizard` (stacked on `feat/first-boot-wizard`)
**Phase:** 3 of 4 (Captive portal / AP mode)
**Effort:** Medium — 4 files (1 new, 3 modified), 642 insertions / 29 deletions
**Depends on (in merge order):**
- PR-7 (`feat/first-boot-wizard`) — this branch sits on top of it
- PR-8 (`feat/network-config-backend`) — shares helpers and conventions in `routers/setup.py`
- PR-10 (`feat/ap-mode`) — provides the `dream-ap-mode` service and `/v1/ap-mode/status` endpoint

## Summary

The other half of PR-10. With this PR the end-to-end "out of box" flow works:

```
Recipient takes device out of box
  → scans QR-A (from PR-9 setup card)
  → phone joins "Dream-Setup-XXXX" AP (PR-10)
  → any URL opens the wizard (captive portal)
  → wizard detects AP mode + collects home WiFi creds  ← THIS PR
  → user names device, picks user, picks stack
  → wizard generates magic-link QR
  → user saves the QR
  → wizard tears down AP, joins home WiFi               ← THIS PR
  → phone auto-reconnects to home WiFi
  → user scans saved QR → opens chat on home network
```

## API surface

| Method | Path | Auth | Purpose |
|---|---|---|---|
| GET  | `/api/setup/ap-mode-status` | Bearer | Proxy to host-agent `/v1/ap-mode/status` |
| POST | `/api/setup/wifi-handoff` | Bearer | Proxy to host-agent `/v1/network/wifi-handoff` — bundled disable-AP + nmcli connect |
| POST | `/v1/network/wifi-handoff` (host-agent, new) | Bearer | Validates creds, returns 202, runs background worker |

## Files

| File | Lines | What |
|---|---|---|
| `bin/dream-host-agent.py` (+128) | | New `/v1/network/wifi-handoff` endpoint and `_wifi_handoff_worker` + `_restart_ap_for_recovery` helpers. Worker sequence: sleep 3s → `systemctl stop dream-ap-mode` → sleep 2s → `nmcli connect`. On failure, restarts AP so the user can retry. |
| `extensions/services/dashboard-api/routers/setup.py` (+89/−3) | | Two new proxy endpoints + `_proxy_agent` helper that never logs request bodies (Wi-Fi passwords stay out of logs). |
| `extensions/services/dashboard/src/pages/FirstBoot.jsx` (+170/−25) | | Detects AP mode on mount; when active, prepends a WiFi step and triggers handoff at Finish. New `WifiStep` component; updated `ConfirmStep` (summary + heads-up notice) and `DoneScreen` (switching-networks callout). |
| `extensions/services/dashboard-api/tests/test_setup_ap_mode.py` (new, 152) | | 9 tests: auth, happy paths, validation (oversized SSID/password, control chars), error translation (501 unsupported platform, 503 agent unreachable). |

### Why the handoff is a deliberate exception to PR-10's "no AP toggle from HTTP"

PR-10 was strict: no enable/disable endpoints, only `/v1/ap-mode/status`. The reasoning: toggling an AP from HTTP is a great way to lock yourself out of a remote box.

`wifi-handoff` is the carve-out, and it's safe because:

1. **Direction is one-way disable.** The endpoint tears down the AP and joins a real network. Accidentally hitting it puts the device on its home network — exactly where the device was supposed to be all along.
2. **Bundled with wizard completion.** It's not a general "stop AP" endpoint. The wizard's Finish action calls it as the final step.
3. **Auto-recovery on failure.** If the nmcli connect fails, the worker restarts `dream-ap-mode` so the user lands back on the setup AP and can retry.

The corresponding `enable` direction stays off-limits — that's the dangerous case.

### Why the worker is async (background thread) instead of synchronous

The HTTP response has to flush before the AP tears down — otherwise the wizard never sees confirmation. The worker:

1. Sleeps 3s after thread start (HTTP response time)
2. Tears down AP
3. Sleeps 2s (NM reclaims interface)
4. Runs nmcli connect
5. Writes result to `/run/dream-ap-mode/handoff-status.json`

The wizard's user sees the Done screen with the "switching networks" notice *before* the AP dies. From their perspective: tap Finish → see QR + instructions → save QR → phone reconnects to home network within seconds.

### Why no live WiFi scan in the wizard's WiFi step

The AP interface (`wlan0`) is busy hosting `hostapd`. Running `nmcli wifi list` against it returns nothing useful. The wizard collects SSID + password via plain text inputs. The user knows their own network name; the friction is acceptable for a once-per-device flow. A future PR could surface WiFi scan when the device has TWO wireless interfaces (rare on small Linux boxes today).

### Why the user has to "save the QR" before networks switch

The QR is rendered on a page served by the AP. When the AP tears down, the page becomes unreachable. The Done screen explicitly tells the user to save/share/copy the QR before the network switches. Modern iOS / Android save-image gestures all work fine on the rendered ``.

## Step layout

| Step | Non-AP mode | AP mode |
|---|---|---|
| 1 | Welcome (device name) | **Home Wi-Fi (NEW)** |
| 2 | First user | Welcome (device name) |
| 3 | Stack picker | First user |
| 4 | Confirm + Finish | Stack picker |
| 5 | — | Confirm + Finish |

`StepDots` shows the right count based on `useFirstRun()` + `/api/setup/ap-mode-status`.

## Test plan

**Automated:**
- [x] 9 new tests in `test_setup_ap_mode.py` — all pass
- [x] 33 setup-router tests (9 new + 24 existing) — all pass
- [x] Dashboard build clean; FirstBoot lazy chunk now 17.1 KB / 4.6 KB gzipped (was 12.5 KB; +4.6 KB for the WiFi step)
- [x] All 62 dashboard vitest tests pass; FirstBoot stays mocked in App.test.jsx
- [x] Lint clean (0 errors; same 5 pre-existing warnings)
- [x] `python -m py_compile bin/dream-host-agent.py` passes

**Manual (requires real Linux hardware + the dream-ap-mode service running):**

- [ ] AP mode active, navigate to dashboard
  - [ ] Wizard mounts (firstRun=true)
  - [ ] Detects AP mode → shows banner + WiFi step first
  - [ ] StepDots shows 5 dots, not 4
- [ ] Complete the wizard with bogus home credentials
  - [ ] Handoff fires, but nmcli connect fails
  - [ ] Background worker writes handoff-status.json with `status: error`
  - [ ] AP restarts; user can re-open the wizard and retry
- [ ] Complete with correct home credentials
  - [ ] Handoff succeeds; AP tears down; device joins home network
  - [ ] Phone reconnects to home WiFi automatically
  - [ ] Scanning the saved QR on the home network opens chat at <device>.local

## Caveats

- **Wifi-handoff worker runs as root** (same as the rest of the host-agent's privileged ops). That's why it can `systemctl stop dream-ap-mode` + `nmcli connect` — no sudo trickery, just the agent's normal privilege level.
- **Handoff is one-shot per wizard run.** No retry inside the wizard. If creds are wrong, the worker restarts the AP and the user comes back to the wizard fresh.
- **Limited error surface to the user.** The wizard fires the handoff and shows "switching networks now"; it doesn't poll for the result. The user finds out by either (a) reaching the device on their home network shortly, or (b) seeing the AP come back up and the wizard reload. A future PR could expose handoff-status.json via a polling endpoint and show a richer waiting screen.

## What's NOT in this PR

- **Polling for handoff completion.** The wizard fires the handoff and moves on; it doesn't expose the result file. Acceptable for v1 because the user's connection drops anyway.
- **Wifi scan on the AP step.** wlan0 is busy. Multi-interface devices could scan with the second iface; deferred until that's a real configuration.
- **Custom DREAM_PUBLIC_URL handling.** The magic-link URL embedded in the QR uses whatever `DREAM_PUBLIC_URL`/`WEBUI_URL` is currently set to (likely the AP gateway IP at this point). After handoff that won't resolve from outside the AP. The user is told to navigate to `<device>.local` directly, which works via mDNS (PR-1). A nicer future PR could rewrite the magic-link URL host based on the device name.
- **A "factory reset" button** (mentioned in the plan). That's hardware-specific (long-press a physical button on a hardware unit) and lives in the image-build layer.

### DIFF
--- a/dream-server/bin/dream-host-agent.py
+++ b/dream-server/bin/dream-host-agent.py
@@ -864,6 +864,100 @@ def _narrowed_compose_set_resolves(narrowed_flags: list, service_id: str,
     return service_id in result.stdout.split()
 
 
+def _wifi_handoff_worker(ssid: str, password: str) -> None:
+    """Background worker for /v1/network/wifi-handoff.
+
+    Sequence:
+      1. Sleep 3s so the HTTP 202 response has time to flush to the wizard.
+      2. `systemctl stop dream-ap-mode` — tears down hostapd/dnsmasq and
+         hands the interface back to NetworkManager (ap-mode.sh ExecStop
+         handles the cleanup).
+      3. Wait briefly for the interface to settle.
+      4. `nmcli device wifi connect <ssid> password <pwd>`.
+      5. On success: write handoff-status.json with status=ok.
+      6. On failure: write handoff-status.json with status=error AND
+         re-enable dream-ap-mode so the user isn't stranded.
+
+    All errors are caught — this runs in a daemon thread and a crash
+    would silently strand the device. The result file is the only
+    feedback channel after the HTTP response is gone.
+    """
+    handoff_dir = Path("/run/dream-ap-mode")
+    handoff_dir.mkdir(parents=True, exist_ok=True)
+    status_path = handoff_dir / "handoff-status.json"
+
+    def _write_status(status: str, **extra) -> None:
+        payload = {
+            "status": status,
+            "ssid": ssid,
+            "at": datetime.now(timezone.utc).isoformat(),
+            **extra,
+        }
+        try:
+            status_path.write_text(json.dumps(payload, indent=2), encoding="utf-8")
+            status_path.chmod(0o600)
+        except OSError as exc:
+            logger.error("wifi-handoff: failed to write status file: %s", exc)
+
+    # Step 1: give the HTTP response time to flush.
+    time.sleep(3)
+
+    # Step 2: tear down AP.
+    logger.info("wifi-handoff: stopping dream-ap-mode")
+    try:
+        subprocess.run(
+            ["systemctl", "stop", "dream-ap-mode"],
+            check=False, capture_output=True, timeout=20,
+        )
+    except (subprocess.TimeoutExpired, OSError) as exc:
+        _write_status("error", stage="ap-teardown", reason=str(exc))
+        return
+
+    # Step 3: settle period for NM to reclaim the interface.
+    time.sleep(2)
+
+    # Step 4: try to join the target network.
+    logger.info("wifi-handoff: connecting to %s", ssid)
+    args = ["nmcli", "device", "wifi", "connect", ssid]
+    if password:
+        args += ["password", password]
+    try:
+        result = subprocess.run(
+            args, capture_output=True, text=True, timeout=45,
+        )
+    except (subprocess.TimeoutExpired, OSError) as exc:
+        _write_status("error", stage="wifi-connect", reason=str(exc))
+        _restart_ap_for_recovery()
+        return
+
+    if result.returncode != 0:
+        raw = (result.stderr or result.stdout or "").strip()[:300]
+        _write_status(
+            "error",
+            stage="wifi-connect",
+            reason=raw or "nmcli connect failed",
+            returncode=result.returncode,
+        )
+        _restart_ap_for_recovery()
+        return
+
+    logger.info("wifi-handoff: connected to %s", ssid)
+    _write_status("ok")
+
+
+def _restart_ap_for_recovery() -> None:
+    """Bring dream-ap-mode back up after a failed handoff so the user
+    can reach the wizard again and retry."""
+    logger.info("wifi-handoff: recovering — restarting dream-ap-mode")
+    try:
+        subprocess.run(
+            ["systemctl", "start", "dream-ap-mode"],
+            check=False, capture_output=True, timeout=30,
+        )
+    except (subprocess.TimeoutExpired, OSError) as exc:
+        logger.error("wifi-handoff: AP recovery failed: %s", exc)
+
+
 class AgentHandler(BaseHTTPRequestHandler):
     def log_message(self, fmt, *args):
         logger.info(fmt, *args)
@@ -983,6 +1077,8 @@ class AgentHandler(BaseHTTPRequestHandler):
             self._handle_invalidate_compose_cache()
         elif self.path == "/v1/env/update":
             self._handle_env_update()
+        elif self.path == "/v1/network/wifi-handoff":
+            self._handle_network_wifi_handoff()
         else:
             json_response(self, 404, {"error": "Not found"})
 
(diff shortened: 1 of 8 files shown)


Lightheartdevs:
I audited this one in depth. The tests/checks are green, but I do not think this is merge-ready yet.

Findings:

1. The wizard schedules network teardown before the QR is actually rendered or saved. In `FirstBoot.jsx:149-171`, `finish()` generates the invite, marks setup complete, calls `/api/setup/wifi-handoff`, and only then sets `invite`, which triggers `DoneScreen` to fetch the QR at `FirstBoot.jsx:608-627`. Meanwhile the host-agent worker sleeps only 3 seconds before stopping `dream-ap-mode` (`bin/dream-host-agent.py:902-910`). On a real phone, QR fetch/render or a slow user can easily lose the AP before the QR appears or before it can be saved. The handoff should not start until the QR/link is already available and the user explicitly confirms they saved it, or the QR data URL should be generated before scheduling teardown.

2. Handoff failures are treated as success in the UI. `FirstBoot.jsx:159-164` awaits `fetch('/api/setup/wifi-handoff', ...)` but never checks `response.ok`, so a 422/501/503 from dashboard-api still sets `handoffStarted = true`. Worse, `/api/setup/complete` has already been called at `FirstBoot.jsx:149-150`, so the first-run sentinel can be flipped even when handoff was rejected or unsupported. That can strand the user in a completed setup state while still on the setup AP, with the UI telling them the network is switching. Check the response, surface the error, and avoid clearing progress/flipping first-run until the handoff is actually accepted.

3. The home Wi-Fi password is persisted to `localStorage`. `FirstBoot.jsx:97-120` initializes and writes `homePassword` through `dream-firstboot-progress`; if the user abandons or refreshes mid-flow, the home network password remains in browser storage indefinitely for that device origin. The UI copy says it is stored only long enough to join. Please keep the password in memory/session-only state, or explicitly clear it on navigation/back/timeout and never include it in the persisted progress object.

4. This PR is stacked on blocked dependencies. It calls `/v1/ap-mode/status` from dashboard-api and relies on the AP service from #1161, but #1161 itself is not merge-ready yet; it also inherits the first-boot/magic-link blockers from #1158/#1155. So even with this branch's local checks green, the end-to-end flow is not currently mergeable.

Local checks run:

```text
python -m ruff check bin/dream-host-agent.py extensions/services/dashboard-api/routers/setup.py extensions/services/dashboard-api/tests/test_setup_ap_mode.py
# All checks passed

python -m pytest extensions/services/dashboard-api/tests/test_setup_ap_mode.py -q
# 11 passed

npm.cmd run build
# passed

npm.cmd test -- --run
# 12 files / 62 tests passed
```

Net: good direction, but the QR-before-teardown sequencing and failure handling need to be fixed before I would trust this for first-boot onboarding.

Lightheartdevs:
Re-audit after the latest author update: **improved, but still not ready to merge**.

What improved:
- The Wi-Fi password is no longer persisted to `localStorage`.
- `/api/setup/wifi-handoff` errors are now checked before flipping first-run complete.
- QR generation is pre-fetched before the handoff request.
- Local checks are green:
  - `python -m ruff check bin/dream-host-agent.py extensions/services/dashboard-api/routers/setup.py extensions/services/dashboard-api/tests/test_setup_ap_mode.py`
  - `python -m pytest extensions/services/dashboard-api/tests/test_setup_ap_mode.py -q` -> 11 passed
  - `npm.cmd test -- --run` -> 62 passed
  - `npm.cmd run build`

Remaining findings:

1. The network handoff still starts before the user has actually seen/saved the QR. `finish()` now pre-fetches QR data (`extensions/services/dashboard/src/pages/FirstBoot.jsx:164-178`), but it still calls `/api/setup/wifi-handoff` before `setInvite()` renders the Done screen (`FirstBoot.jsx:190-218`). The host-agent worker still sleeps only 3 seconds before stopping `dream-ap-mode` (`bin/dream-host-agent.py:904-910`). So the user has roughly three seconds, minus request/render time, to read the Done screen and save/share the QR before the AP drops. The Done screen even says ?Save the QR ... before that happens? (`FirstBoot.jsx:716-728`), but the shutdown countdown has already started. I would split this into two user actions: Finish generates/renders the QR; then an explicit ?I saved it, switch networks? button triggers handoff.

2. This PR still depends on #1161 and should not merge before it. The dashboard API calls host-agent `/v1/ap-mode/status` (`extensions/services/dashboard-api/routers/setup.py:300`), but this branch itself does not define that host-agent endpoint; it comes from #1161. Since #1161 is still blocked on AP routing/systemd issues, this branch cannot currently be validated end-to-end. If this stays stacked, the base should make that dependency unambiguous, or the PR should wait until #1161 lands and then rebase.

Small residual risk: `/api/setup/complete` is still not checked for `response.ok` before rendering the success screen (`FirstBoot.jsx:216-218`). That is not AP-specific, but while this flow is being touched it would be worth tightening.

Net: the author addressed two of the earlier blockers cleanly. The remaining UX sequencing issue is still the one that can strand a real first-boot user.

PR #1164, opened 2026-05-12
feat(magic-link): record intended role on invites (admin/user/guest metadata)

Lightheartdevs:
# PR-13: Role scoping for magic-link invites

**Branch:** `feat/scoped-invites` (stacked on `feat/magic-link-auth`)
**Phase:** 4 of 4 (Polish & remote)
**Effort:** Small — 2 files (both modified), 207 insertions / 4 deletions
**Depends on:** PR-4 (`feat/magic-link-auth`) — this branch sits on top of it

## Summary

Adds a `role` field to magic-link invites distinct from the existing `scope`. Where `scope` says **which surfaces** an invite grants (chat / dashboard / all), `role` says **what permission level inside those surfaces** (admin / user / guest).

This is what lets an admin send "for mom, role=user" and "for my partner, role=admin" without those needing different scopes or different invite flows.

## Schema additions

| Field | Values | Default | Purpose |
|---|---|---|---|
| `role` | `admin` / `user` / `guest` | `user` | Permission level inside the granted scopes |

`admin` — full operator, can install extensions, edit settings (first-boot wizard's invite uses this)
`user` — standard chat user, has their own history
`guest` — sandboxed / read-only; chats but doesn't persist into the admin's history, can't change settings

Surfaced on:
- `POST /api/auth/magic-link/generate` request body + response
- `GET /api/auth/magic-link/list` for each token
- Persisted in each magic-link record on disk

Legacy records (written by pre-PR-13 builds) don't have a `role` key. The list endpoint backfills them to `user` so consumers see a stable schema. Verified by `test_list_backfills_role_for_legacy_records`.

## Files

| File | Change |
|---|---|
| `extensions/services/dashboard-api/routers/magic_link.py` (+85 / −4) | New `role` field on `GenerateRequest`, `GenerateResponse`, `TokenSummary`. New `_provision_openwebui_user()` stub function called on every redemption (the integration slot for the eventual Open WebUI provisioning). Storage / list / log lines updated to include role. |
| `extensions/services/dashboard-api/tests/test_magic_link.py` (+122) | 7 new tests covering role defaults, validation, list surfacing, legacy backfill, and the redeem-time provisioning hook contract. |

## Open WebUI provisioning slot

The hard part of role-based invites isn't the data model — it's actually making Open WebUI honor the role. That's a separate piece of work:

The redeem path now calls `_provision_openwebui_user(target_username, role, scope)`. Today this is a logging stub. The docstring documents the eventual implementation path:

1. Read `OPEN_WEBUI_ADMIN_TOKEN` from env
2. `GET ${WEBUI_URL}/api/v1/users` → find user by name
3. If missing: `POST .../api/v1/auths/signup` with a generated random password
4. `PATCH .../api/v1/users/{id}/role/update` to set the role
5. Auto-login on first redemption needs either a proxy in front of WebUI or cooperation from its session-cookie format

Why a stub vs. full implementation:
- Open WebUI's admin API semantics depend on `WEBUI_AUTH` config — needs a real Open WebUI to test against.
- Auto-login is real work beyond this PR's scope.
- The slot being here means the eventual integration PR has exactly **one function** to fill in.

Until then, the role is captured in the magic-link record and surfaced in the admin list — enough for the "I labeled this invite for mom as user-role" use case. The runtime gap is honest and reviewed.

## Test plan

```
pytest tests/test_magic_link.py -v
```

**32/32 tests pass** (25 from PR-4 + 7 new). New tests cover:

- [x] Default role is `user`
- [x] `admin` and `guest` both accepted
- [x] Unknown role (e.g. `superuser`) → 422
- [x] `/list` surfaces role from every record
- [x] Legacy records (no `role` key) backfill to `user` so the dashboard UI never sees `null`/missing
- [x] Redeem path calls `_provision_openwebui_user` exactly once with `(target_username, role, scope)` — locks the contract for the eventual integration

## What's NOT in this PR

- **Invites UI role picker.** PR-5's Invites page hardcodes scope but not role; surfacing the role dropdown is a follow-up change to `pages/Invites.jsx`. Kept this PR backend-focused so the security-sensitive bits review cleanly.
- **FirstBoot wizard "first user is admin" default.** Same reasoning — the wizard's invite call should set `role: "admin"` since it's the device's first user. Follow-up change to `pages/FirstBoot.jsx`.
- **Real Open WebUI provisioning.** The stub is here; the implementation needs a running Open WebUI to test against.
- **Auto-login on first redemption.** The user still sees Open WebUI's signin screen on first hit. Closing that gap needs either an auth-proxy shim or Open WebUI cooperation; tracked as a follow-up.

## Caveats

- **The role field isn't yet enforced.** Today, role=admin vs role=guest produce identical sessions. Treat role as a label that future code will honor; document this gap when shipping to users.
- **Backward-compat default is `user`.** If a future config wanted a stricter default ("must specify a role"), that'd be a breaking change. Calling it out — silent default is the explicit choice.
- **No role-change endpoint.** Once an invite is generated, role is fixed. To change it you revoke + regenerate. Keeps the audit story simple.


Lightheartdevs:
I audited this one in depth. It is not merge-ready yet.

Findings:

1. Ruff still fails locally on the magic-link test module inherited from the base branch:

```text
python -m ruff check extensions/services/dashboard-api/routers/magic_link.py extensions/services/dashboard-api/tests/test_magic_link.py
# F401 `time` imported but unused at extensions/services/dashboard-api/tests/test_magic_link.py:16
```

The endpoint tests pass, but this should be cleaned up before merge.

2. The new `role` field is not actually role scoping yet. `GenerateRequest.role` is described as enforced by Open WebUI provisioning at redemption time (`magic_link.py:132-136`), but the redemption path only calls `_provision_openwebui_user()`, which is explicitly a logging stub (`magic_link.py:328-364`, called at `magic_link.py:472-481`). That means `admin`, `user`, and `guest` all create the same effective session today. If this lands as-is, downstream callers may assume the role has security meaning when it is only metadata. I would either rename/describe it as `intended_role`/label until enforcement exists, or add a response/status field that makes the enforcement gap explicit.

3. This is stacked on #1155, which is already blocked by the core magic-link problems: generated URLs point at the wrong service route, redemption does not create an Open WebUI session, and the base branch has a Ruff failure. So even if this role metadata layer were cleaned up, it cannot be merged independently or made user-visible safely until #1155 is fixed.

Local checks run:

```text
python -m pytest extensions/services/dashboard-api/tests/test_magic_link.py -q
# 32 passed

python -m ruff check extensions/services/dashboard-api/routers/magic_link.py extensions/services/dashboard-api/tests/test_magic_link.py
# fails on unused `time`
```

Net: useful schema preparation, but not actual role scoping yet. I would hold it until the base magic-link flow is corrected and the API copy cannot be mistaken for enforced permissions.

Lightheartdevs:
Re-audit after the latest update: improved, but I still would not merge this as “role scoping” yet.

What is fixed / better:
- The stale Ruff failure is gone.
- The branch now sits on the newer magic-link base that routes through the proxy and issues signed `dream-session` cookies.
- CI is green on this PR.
- Local checks pass:
  - `python -m ruff check extensions/services/dashboard-api/routers/magic_link.py extensions/services/dashboard-api/tests/test_magic_link.py extensions/services/dashboard-api/routers/auth.py extensions/services/dashboard-api/tests/test_auth_router.py extensions/services/dashboard-api/session_signer.py extensions/services/dashboard-api/tests/test_session_signer.py`
  - `python -m pytest extensions/services/dashboard-api/tests/test_magic_link.py extensions/services/dashboard-api/tests/test_auth_router.py extensions/services/dashboard-api/tests/test_session_signer.py -q` -> 60 passed
  - `git diff --check origin/feat/magic-link-auth...HEAD`

Remaining issue:

The new role field is still metadata, not enforced scoping. The implementation explicitly says `_provision_openwebui_user()` is an integration stub, and redemption issues the same signed `dream-session` regardless of `admin`, `user`, or `guest`. That means all three roles have the same runtime effect today.

The source comments now disclose that gap, which helps, but the public API field description still says the role is “Enforced by Open WebUI provisioning at redemption time.” That is not true in this implementation and will leak into OpenAPI/client expectations. Given the PR title and API wording, I think this still overpromises a security boundary that does not exist yet.

To make this mergeable, I’d either:
- change the user-visible/API wording and PR title to make this clearly “role intent metadata” only, or
- include the actual enforcement path: persist the role/scope in the session claim and have the proxy/dashboard/Open WebUI provisioning enforce it before granting admin/user/guest surfaces.

So: green checks now, but still not a merge candidate as scoped access control.

PR #1165, opened 2026-05-12
feat(dashboard): smart PWA install-prompt banner

Lightheartdevs:
# PR-14: Smart PWA install-prompt banner

**Branch:** `feat/pwa-install-prompt`
**Phase:** 4 of 4 (Polish & remote) — **final PR in the onboarding plan**
**Effort:** Small — 5 files (3 new, 2 modified), 422 insertions
**Depends on:** PR-3 (`feat/dashboard-pwa`) for runtime install — not for build

## Summary

The browser's default "install this site" UI is a tiny icon in the address bar that most users never notice. PWAs that ship a custom nudge at a well-timed moment see meaningfully higher install rates.

This PR adds that nudge to the dashboard: a slide-in banner that appears at the bottom of the screen after the user has come back **3+ times**, with copy + button matching the install flow for their browser.

## Behaviour

| Browser | What we show | What the button does |
|---|---|---|
| Chromium / Edge / Android Chrome | "Add to home screen" button | Calls the captured `beforeinstallprompt.prompt()` → OS install dialog |
| iOS Safari (iPhone / iPad) | "Tap Share → Add to Home Screen" with iconography | No button — Safari refuses programmatic install, we walk the user through it |
| Firefox / Safari on macOS | Nothing | `beforeinstallprompt` doesn't fire on these — banner stays hidden |
| Already-installed PWA | Nothing | `display-mode: standalone` is true → hook returns `shouldShow=false` |

Triggers:
- **Don't** show on first visit (would be obnoxious)
- **Do** show after 3+ tab sessions (real signal the user values this UI)
- **Never again** after the user clicks "Not now" / X / accepts → installs

## Files

| File | Lines | What |
|---|---|---|
| `src/hooks/usePwaInstallPrompt.js` (new) | 158 | The brains. Captures `beforeinstallprompt`, counts tab sessions, persists dismissal, detects already-installed, returns `{shouldShow, promptInstall, dismiss, isIos, installed, visitCount}`. |
| `src/components/InstallPromptBanner.jsx` (new) | 84 | The UI. Slide-in card with two flavors (non-iOS button vs. iOS Share-instruction). Pure consumer of the hook. |
| `src/hooks/__tests__/usePwaInstallPrompt.test.js` (new) | 152 | 9 tests covering the full lifecycle. |
| `src/App.jsx` (+8 / 0) | | Top-level mount of the banner. |
| `src/App.test.jsx` (+6 / 0) | | Mock the banner so existing App-level tests stay isolated from PWA event simulation. |

## Hook contract

```js
const {
  shouldShow,    // bool — render the banner?
  isIos,         // bool — show the manual-install copy variant
  installed,     // bool — PWA is already installed
  visitCount,    // int  — debug visibility
  promptInstall, // async () => OS dialog (no-op on iOS)
  dismiss,       // () => persist "not now" forever
} = usePwaInstallPrompt()
```

Visit counting is per **tab session** (sessionStorage guard), not per **navigation** (F5 doesn't double-count). The dismiss state is permanent in localStorage — surfacing a "remind me later" path would need explicit UI; today the visit-count threshold is the gate.

## Test plan

**Automated (9 new, 13 total run together with App.test.jsx — all pass):**
- [x] Does NOT show on first visit even with `beforeinstallprompt` fired
- [x] Shows on the 3rd visit
- [x] Visit counter only ticks once per tab session (F5 doesn't bump)
- [x] `dismiss()` persists across reloads
- [x] `promptInstall()` calls the captured event's `prompt()`
- [x] `appinstalled` event hides the banner and sets the persistent flag
- [x] Already-installed PWAs (display-mode:standalone) never see the banner
- [x] iOS detection shows the banner without `beforeinstallprompt`
- [x] iOS `promptInstall()` is a graceful no-op (returns `{platform: 'ios', outcome: 'manual'}`)

Full suite: 71/71 dashboard tests pass. Build + lint clean (same 5 pre-existing warnings, none in new code).

**Manual smoke** (requires a real browser + PR-3 merged so manifest + SW exist):
- [ ] Chromium localhost: load the dashboard 3 times in 3 tabs → banner appears on the 3rd
- [ ] Click "Add to home screen" → OS install dialog → install → banner stays hidden across reloads
- [ ] Reset localStorage → dismiss the banner → reload → banner stays hidden (dismissal persists)
- [ ] iOS Safari: load 3 times → banner shows Share + Add to Home Screen copy
- [ ] Firefox desktop: load any number of times → banner never appears (no `beforeinstallprompt`)

## What's NOT in this PR

- **Open WebUI install prompt.** The chat surface (port 3000) is where users spend most of their time, and that's where the "install this app" prompt would have the biggest impact. But Open WebUI is a third-party project; injecting our JS would require either forking it or running a reverse proxy that injects a script. Out of scope for this PR. The dashboard is the surface we directly control.
- **"Remind me later" path.** Today dismissal is forever (until localStorage clears). A snooze pattern (re-prompt after N days) is sensible but unnecessary for v1.
- **A/B-testable copy.** The banner text is hardcoded. If install-rate experiments matter, a follow-up could surface copy via a config endpoint.
- **Per-route timing.** The banner mounts at the App root and runs on every page. A future PR could nudge specifically after the user completed setup or generated their first invite — moments of high engagement.

## Caveats

- **The browser controls when `beforeinstallprompt` fires.** Even with manifest + SW + HTTPS, Chrome's heuristics ("user has engaged with the site enough") need to be satisfied. Our 3-visit threshold sits inside Chrome's own engagement gate — so the banner won't appear UNTIL Chrome would normally have its own install ready. We're just intercepting Chrome's mini-infobar with a nicer UI.
- **iOS PWAs have real limitations.** No push notifications, limited storage, no app-store presence. The banner copy is honest: it tells iOS users they can pin Dream to their home screen, not that they're getting a "real app."
- **Already-installed detection is best-effort.** Some browsers (e.g. iOS 16) don't expose a reliable signal. The hook fallback writes `dream-pwa-installed=1` to localStorage on `appinstalled` so subsequent loads on that browser know.

### DIFF
--- a/dream-server/extensions/services/dashboard/src/App.jsx
+++ b/dream-server/extensions/services/dashboard/src/App.jsx
@@ -1,6 +1,7 @@
 import { Routes, Route } from 'react-router-dom'
 import { useState, useEffect, Suspense, useMemo, useCallback, lazy } from 'react'
 import Sidebar from './components/Sidebar'
+import InstallPromptBanner from './components/InstallPromptBanner'
 import { useSystemStatus } from './hooks/useSystemStatus'
 import { useVersion } from './hooks/useVersion'
 import { useFirstRun } from './hooks/useFirstRun'
@@ -120,6 +121,12 @@ function App() {
           </Routes>
         </Suspense>
       </main>
+
+      {/* Smart PWA install nudge — only renders when the user has shown
+          enough engagement (3+ visits) AND the browser is willing to
+          install. No-op on already-installed PWAs and on browsers that
+          can't install (e.g. Firefox desktop). See usePwaInstallPrompt. */}
+      <InstallPromptBanner />
     </div>
   )
 }
--- a/dream-server/extensions/services/dashboard/src/App.test.jsx
+++ b/dream-server/extensions/services/dashboard/src/App.test.jsx
@@ -51,6 +51,12 @@ vi.mock('./components/SplashScreen', () => ({
   }
 }))
 
+// InstallPromptBanner depends on browser PWA events we don't simulate
+// in these App-level tests; render nothing so it doesn't interfere.
+vi.mock('./components/InstallPromptBanner', () => ({
+  default: () => null,
+}))
+
 describe('App', () => {
   beforeEach(() => {
     vi.stubGlobal('fetch', vi.fn(() =>
--- a/dream-server/extensions/services/dashboard/src/components/InstallPromptBanner.jsx
+++ b/dream-server/extensions/services/dashboard/src/components/InstallPromptBanner.jsx
@@ -1,0 +1,79 @@
+// Smart PWA install prompt banner.
+//
+// Slides in from the bottom on small screens, sits in the bottom-right
+// on desktop. Only renders when usePwaInstallPrompt says the moment is
+// right (3+ visits, not dismissed, not already installed, browser is
+// installable — or iOS where the install path is manual).
+//
+// Two button states:
+//   * non-iOS: "Add to home screen" -> programmatic prompt
+//   * iOS: copy reads "Tap Share → Add to Home Screen" with the
+//     Share icon called out; no programmatic prompt is possible
+//     (Safari refuses to expose it).
+
+import { Smartphone, X, Share, Plus } from 'lucide-react'
+import { usePwaInstallPrompt } from '../hooks/usePwaInstallPrompt'
+
+export default function InstallPromptBanner() {
+  const { shouldShow, isIos, promptInstall, dismiss } = usePwaInstallPrompt()
+  if (!shouldShow) return null
+
+  return (
+    <div
+      role="dialog"
+      aria-label="Add Dream to your home screen"
+      className="fixed bottom-4 left-4 right-4 md:left-auto md:right-4 md:max-w-sm
+                 bg-theme-card border border-theme-accent/40 rounded-xl shadow-2xl
+                 p-4 z-40 animate-in slide-in-from-bottom-4 fade-in duration-300"
+    >
+      <div className="flex items-start gap-3">
+        <div className="w-10 h-10 rounded-xl bg-theme-accent/15 text-theme-accent flex items-center justify-center flex-shrink-0">
+          <Smartphone size={22} />
+        </div>
+        <div className="flex-1 min-w-0">
+          <h3 className="font-semibold text-theme-text text-sm mb-1">
+            Make Dream feel like an app
+          </h3>
+          {isIos ? (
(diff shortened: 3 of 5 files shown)


Lightheartdevs:
I audited this one in depth. I do not see an implementation blocker in this PR itself, but I would merge it only after #1154 (`feat/dashboard-pwa`) lands, because this banner depends on the manifest/service-worker install surface from that PR.

What I checked:

```text
npm.cmd run build
# passed

npm.cmd test -- --run
# 13 files / 71 tests passed
```

Review notes:

- The hook behavior is well covered: visit-count gating, per-tab counting, dismissal persistence, `beforeinstallprompt`, `appinstalled`, installed-mode detection, and iOS manual path are all tested.
- The banner is mounted at the app root and stays inert unless the hook says `shouldShow`, so it should not affect normal dashboard usage.
- The only merge-order caveat is #1154. Without the PWA manifest/service worker/icons from that branch, Chromium will not reliably fire `beforeinstallprompt`, and the iOS path would be prompting users before the dashboard has the intended home-screen metadata.

Net: merge candidate after #1154 is merged and branch protection approvals are satisfied. I did not merge it because the dependency is still open and this repo is still enforcing required review approvals.

Lightheartdevs:
Audit result: clean after rebase; ready to merge.

I rebased this onto current `main` and resolved the only conflict by keeping the newer first-boot flow while adding the install-prompt banner import/mount. I audited the hook and banner behavior against the current dashboard shell:

- The banner is mounted only in the normal dashboard shell, not during first-boot onboarding.
- The hook waits for 3 tab sessions, avoids F5 double-counting via `sessionStorage`, persists dismissal/install state, and stays hidden for standalone/installed PWAs.
- Chromium/Edge/Android use the captured `beforeinstallprompt` event from a user button click; iOS gets manual Share -> Add to Home Screen copy and no fake programmatic install path.
- Firefox/Safari desktop stay quiet because no install event fires.
- App-level tests mock the banner so PWA browser event simulation stays isolated to the hook tests.

Local checks:
- `npm.cmd ci` in dashboard package -> dependencies installed cleanly (npm reports existing dependency audit findings: 3 moderate, 3 high; not introduced by this patch)
- `npm.cmd test` -> 17 files / 88 tests passed
- `npm.cmd run lint` -> 0 errors, 5 existing warnings in unrelated hooks
- `npm.cmd run build` -> pass
- `git diff --check origin/main...HEAD` -> pass

GitHub checks are green on head `7ff6d9fb842496932a7d8ed730c02660755b4f07`, including `frontend` and `integration-smoke`.

PR #1166, opened 2026-05-12
feat(hermes): package Hermes Agent as a Dream Server extension

Lightheartdevs:
# H-1: Package Hermes Agent as a Dream Server extension

**Branch:** `feat/hermes-extension`
**Effort:** Small — 5 new files, 2 modified, 498 insertions
**Depends on:** none

## Summary

Adds **Hermes Agent** ([Nous Research, MIT, Feb 2026](https://github.com/nousresearch/hermes-agent)) as an opt-in Dream Server extension. Hermes is a self-improving generalist agent with persistent memory, autonomous skill creation, and 70+ built-in tools — and it ships its own complete browser dashboard, so we're packaging rather than building.

After `dream enable hermes`, users browse to port 9119 on the host and find Hermes's native chat / sessions / skills / cron / models / analytics surface, talking to the local `llama-server` via OpenAI-compatible API.

**This is the integration carve-out from a longer original plan.** Initial walkthrough assumed Hermes was missing a browser chat UI and proposed a 7-PR build. Inspection of the upstream repo showed Hermes already ships React 19 + Vite + Tailwind + full FastAPI surface (ChatPage, SessionsPage, SkillsPage, CronPage, ProfilesPage, ModelsPage, LogsPage, AnalyticsPage, ConfigPage, EnvPage, PluginsPage). The original H-2 (bridge), H-3 (chat UI), H-4 (multi-user), and most of H-7 (dashboard surface) became unnecessary. Remaining items moved to "roadmap follow-ups" in `docs/HERMES.md`.

## Files

| File | Lines | What |
|---|---|---|
| `extensions/services/hermes/manifest.yaml` (new) | 49 | Schema-compliant service manifest. `category: optional`, `depends_on: [llama-server]`, `external_port_env: HERMES_PORT` (default 9119), `health: /api/health`, `health_timeout: 60` (skill sync on cold start). |
| `extensions/services/hermes/compose.yaml` (new) | 79 | Pulls SHA-pinned `nousresearch/hermes-agent:sha-dd0923bb89ed2dd56f82cb63656a1323f6f42e6f`. Bridge network (not host) so the container resolves `llama-server` via Docker DNS. Runs only the `dashboard` subcommand. `BIND_ADDRESS` gates LAN exposure. 4 CPU / 4 GB hard caps. |
[...]

PR #1167, opened 2026-05-12
feat(mdns): announce hermes.<device>.local for the Hermes Agent extension

Lightheartdevs:
## Summary

One-line follow-up to [#1166](https://github.com/Light-Heart-Labs/DreamServer/pull/1166) (Hermes Agent extension). Adds Hermes to the set of services the mDNS announcer publishes so users can browse to `hermes.<device>.local:9119` from any LAN device once the Hermes extension is enabled.

**Stacked on [PR-1 #1152](https://github.com/Light-Heart-Labs/DreamServer/pull/1152)** — merges only after that PR lands. Base is intentionally \`feat/mdns-announcer\`, not \`main\`.

## Files

| File | Change |
|---|---|
| \`bin/dream-mdns.py\` (+7 / 0) | New entry in \`_build_services\` for the \`hermes\` service type at port 9119 (from \`HERMES_PORT\` env). |
| \`docs/MDNS.md\` (+1 / 0) | Hermes row added to the announced-services table. |

## Why "announce unconditionally"

Same posture as the existing chat/dashboard/dashboard-api entries — the Hermes extension is opt-in via \`dream enable hermes\`, but we don't gate the mDNS announcement on whether the container is running. Reasoning:

1. The other three entries (chat, dashboard, dashboard-api) are also announced unconditionally even though any of them might be off.
2. When the port is unreachable, the failure mode is browser-side "this site can't be reached" — same as the existing services-not-running cases.
3. The Dream Dashboard's external-link tile only renders Hermes when the container is actually live, so the surface where it visibly "appears" already gates on live state.

If/when we add a more elaborate "announce-only-when-healthy" model, we can do it for all four entries together rather than special-casing Hermes.

## Test plan

- [x] \`python -m py_compile bin/dream-mdns.py\` passes
- [ ] On a real Linux box: \`dream enable hermes\` → \`avahi-browse -art\` (or \`dns-sd -B _http._tcp\` on macOS) lists \`<device>-hermes._http._tcp.local\` alongside the other three entries
- [ ] From a phone on the same WiFi: \`http://<device>-hermes.local:9119\` resolves (note: this requires the bare port-9119 host record to also resolve, which the existing \`<device>.local\` announcement covers)

### DIFF
--- a/dream-server/bin/dream-mdns.py
+++ b/dream-server/bin/dream-mdns.py
@@ -1,0 +1,233 @@
+#!/usr/bin/env python3
+"""Dream Server mDNS announcer.
+
+Publishes the device on the local network as `<DREAM_DEVICE_NAME>.local`
+(default `dream.local`) plus per-service `_http._tcp` records so any device
+on the same LAN can find the dashboard and chat UI without knowing the IP.
+
+This makes the "open `dream.local` from any phone" UX work out of the box —
+the device boots, joins WiFi, and starts announcing itself within seconds.
+
+Reads `DREAM_DEVICE_NAME` and the service ports from `.env`. Re-publishes
+when the file changes (poll-based, 30s cadence) so renaming the device or
+changing a port doesn't require a service restart.
+
+Linux-first: relies on `avahi-daemon` being installed and running (already
+standard on Ubuntu / Debian / Fedora / Arch desktop installs). macOS has
+built-in mDNS via mDNSResponder and announces hostname.local automatically;
+this script is a no-op on Darwin (logs and exits 0). Windows mDNS support
+varies — see BRANDING.md / docs/MDNS.md for follow-up.
+
+Run via:
+  python3 /opt/dream-server/bin/dream-mdns.py
+or via the dream-mdns.service systemd unit.
+"""
+
+from __future__ import annotations
+
+import logging
+import os
+import platform
+import re
+import signal
+import socket
+import sys
+import time
+from pathlib import Path
+
+try:
+    from zeroconf import IPVersion, ServiceInfo, Zeroconf
+except ImportError:
+    print(
+        "ERROR: `zeroconf` Python package not installed. "
+        "On Debian/Ubuntu: sudo apt install python3-zeroconf. "
+        "On Fedora: sudo dnf install python3-zeroconf. "
+        "On Arch: sudo pacman -S python-zeroconf.",
+        file=sys.stderr,
+    )
+    sys.exit(1)
+
+logging.basicConfig(
+    level=os.environ.get("DREAM_MDNS_LOG_LEVEL", "INFO"),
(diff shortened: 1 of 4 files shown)


Lightheartdevs:
Deep audit result: **not ready to merge**.

This is a tiny patch, but the current branch shape is wrong and CI is red.

Findings:

1. **CI is failing because this branch links to a file it does not contain.** `docs/MDNS.md:13` adds `[`docs/HERMES.md`](HERMES.md)`, but #1167 is based on `feat/mdns-announcer` (#1152), not the Hermes extension branch (#1166), so `docs/HERMES.md` is absent in the merge commit. The `integration-smoke` job fails in `tests/test-doc-links.sh` with:

   ```text
   [FAIL] dream-server/docs/MDNS.md:13 -> HERMES.md
   resolved: dream-server/docs/HERMES.md
   [FAIL] 1 markdown file(s) contain broken local links
   ```

   Fix options: stack this PR on #1166 as well as #1152, or change the doc row to avoid a local `HERMES.md` link until #1166 has landed.

2. **The PR has an undeclared dependency on #1166.** The body says it is a one-line follow-up to #1166, but the base is only #1152. The code comment in `bin/dream-mdns.py:110-115` talks about `extensions/services/hermes/` and the Hermes manifest?s `external_port_env`, neither of which exists on this branch unless #1166 is also present. If #1152 lands and this PR is merged before #1166, main will announce a Hermes service/docs row for an extension that is not in-tree yet.

3. **Inherited blocker: #1152 is still the base.** Even after the doc-link issue is fixed, this should wait for the mDNS announcer PR?s own blockers to be resolved and merged. This PR should be treated as a small dependent follow-up, not an independent merge candidate.

Checks run:

- `python -m py_compile bin/dream-mdns.py` -> pass
- inspected `git diff origin/feat/mdns-announcer...HEAD` -> only `bin/dream-mdns.py` and `docs/MDNS.md`
- reviewed failing `integration-smoke` log from run `25747040215`, job `75612984988`

Recommended path: first fix/land #1152, fix/land #1166, then rebase this on the branch that contains both prerequisites. At that point this PR is likely a quick re-review.

Lightheartdevs:
Auto-closed after force-push reset to base.

The original intent of this PR was to make Hermes reachable via mDNS at `<device>-hermes._http._tcp.local:9119`. After Wave 2 architectural review (Q4: port-based hosting at the proxy), we settled on routing everything through the dream-proxy on `<device>.local` instead of per-service mDNS subdomains. Users will reach Hermes via the proxy (route TBD as part of the hermes-proxy work in #1168/#1171), not a separate hostname.

Per-service `_http._tcp` SRV records are a separate question (service discovery for MCP clients, etc.) — when we revisit those, we'll do all four (chat, dashboard, dashboard-api, hermes) together rather than special-casing Hermes.

Branch reset to the mdns-announcer base — closing.

PR #1168, opened 2026-05-12
feat(hermes): magic-link SSO gating via Caddy auth-proxy sidecar

Lightheartdevs:
# H-2: Magic-link SSO gating for Hermes (Caddy auth-proxy sidecar)

**Branch:** `feat/hermes-magic-link-sso` (stacked on `feat/hermes-extension`)
**Effort:** Small-Medium — 5 new files, 4 modified, 514 insertions
**Depends on:** [H-1 #1166](https://github.com/Light-Heart-Labs/DreamServer/pull/1166), and at runtime [PR-4 #1155](https://github.com/Light-Heart-Labs/DreamServer/pull/1155) (sets the cookie this proxy checks)

## Summary

Closes the "anyone on the LAN can reach Hermes" gap that H-1 left open. Hermes no longer binds a host port directly; instead, a small **Caddy sidecar (`dream-hermes-proxy`)** sits in front of Hermes and forwards only when a `dream-session` cookie is present.

That cookie is what dashboard-api's magic-link redemption flow ([PR-4](https://github.com/Light-Heart-Labs/DreamServer/pull/1155)) already sets in the browser. So: scan a magic-link → cookie set → browse to `hermes.<device>.local:9120` → through to Hermes.

Without a cookie, the proxy answers with a self-contained "you need an invite" page (no cross-dashboard JS, no external assets — just the static HTML the proxy serves at `/auth/required`).

## Honest scope — what this isn't

This is **access gating**, not user identification:

- All authed users share one Hermes (same memories, skills, persona, sessions)
- Cookie validation is **presence-only** — PR-4 doesn't keep a server-side session store. Anyone with another user's raw cookie value passes the gate
- The proxy doesn't tell Hermes "this request is from Alice"; it tells Hermes "this request is from somebody with a valid invite"

For Dream Server's trust model (trusted LAN, family-scale users) this is a reasonable v1. The path to real multi-user (per-user Hermes containers — "Option B") is documented in `docs/HERMES-SSO.md`.

## Why Caddy, not a custom FastAPI proxy

I considered both. Caddy wins because:
- Native streaming + WebSocket support out of the box (Hermes uses both)
- `reverse_proxy` directive handles `Host`, `X-Forwarded-*`, `Upgrade` correctly
- ~50MB single-binary alpine image
- Caddyfile config is ~40 lines vs. a FastAPI proxy that's hundreds of lines + needs streaming + WS handling
- Same posture as a TLS-terminating production proxy

If we later need request-modification logic that's awkward in Caddyfile (e.g., decoding a JWT and injecting `X-Dream-User`), we can either:
- Write a small Caddy module (Go)
- Switch to a small FastAPI proxy at that point

## Files

| File | Lines | What |
|---|---|---|
| `extensions/services/hermes-proxy/manifest.yaml` (new) | 41 | Schema-compliant manifest. `category: optional`, `depends_on: [hermes, dashboard-api]`, `external_port_env: HERMES_PROXY_PORT` (default 9120). |
| `extensions/services/hermes-proxy/Caddyfile` (new) | 76 | Named-matcher routing. `/health` + `/favicon.ico` always answered; `/auth/required*` static files; cookie-present → `reverse_proxy dream-hermes:9119`; default → 303 to `/auth/required`. `auto_https off`, `admin off`, logs to stdout. |
| `extensions/services/hermes-proxy/auth-required/index.html` (new) | 104 | Self-contained "ask the admin for an invite" page. Inline CSS + SVG, dark Dream theme, no external assets. |
| `extensions/services/hermes-proxy/compose.yaml` (new) | 45 | `caddy:2.8.4-alpine`, Caddyfile mounted `:ro`, auth-required dir mounted `:ro`, Caddy state persisted, 0.5 CPU / 256MB hard caps. |
| `extensions/services/hermes/compose.yaml` (modified) | −5/+9 | Replaced `ports:` binding with `expose: ["9119"]`. Hermes is now Docker-bridge-internal only. |
| `.env.example` (+6) | | Annotated `HERMES_PROXY_PORT` + `HERMES_PROXY_UPSTREAM` block. |
| `.env.schema.json` (+11) | | Schema entries. |
| `docs/HERMES-SSO.md` (new) | 132 | Architecture diagram, design rationale (per-process cookie gate vs. per-user containers), known limitations spelled out, future Option B path. |
| `docs/HERMES.md` (modified, +6/−5) | | "What you get" intro updated; Roadmap section marks SSO as shipped + adds the per-user-Hermes follow-up path. |

## Architecture

```
Phone / laptop
   │
   ▼  http://hermes.<device>.local:9120
┌──────────────────────────────────────────┐
│  dream-hermes-proxy  (Caddy 2.8.4)       │
│                                          │
│  /health, /favicon.ico → respond         │
│  /auth/required*       → static files    │
│  cookie has dream-session → reverse_     │
│                                proxy     │
│  everything else        → 303 redirect   │
└──────────┬───────────────────────────────┘
           │  internal Docker bridge only
           ▼
┌──────────────────────────────────────────┐
│  dream-hermes  (NousResearch)            │
│  - exposes :9119 internally              │
│  - DOES NOT bind a host port             │
│  - serves React SPA + /api/*             │
│  - own X-Hermes-Session-Token gates      │
│    /api/* requests per-request           │
└──────────────────────────────────────────┘
```

## What "cookie present" actually means

Caddy matcher: `header_regexp Cookie (?:^|;\s*)dream-session=[^;]+`

- Anchors at start-of-Cookie-header OR `;` (so it matches even with other cookies before)
- Requires a non-empty value after `dream-session=`
- Doesn't validate cryptographic signature or cross-check with a server store (PR-4 doesn't have one)

Limitations called out in the PR description AND the docs:

1. **No real multi-user** — shared experience
2. **No server-side cookie validation** — anyone with the raw cookie can pass
3. **No per-request user identification** — proxy doesn't add `X-Dream-User`
4. **Direct Hermes access is gone** — anyone who was using `:9119` directly needs to switch to `:9120` (proxy) or `docker exec`
5. **Caddy's cookie match is text-based** — works for real browsers, a deliberately-malformed Cookie header could in theory slip past (but an attacker who can set arbitrary headers already has bigger leverage)

## Why magic-link scope=hermes is intentionally NOT in this PR

I considered extending `routers/magic_link.py` (PR-4) to accept `scope: "hermes"` and route redemption straight to the proxy. Deferring because:

1. The cookie check is **scope-agnostic** — any redeemed magic-link (chat / dashboard / all) drops the cookie, which is all the proxy needs.
2. Modifying `magic_link.py` here would create a merge conflict with PR-4 + PR-13 that's avoidable.
3. A "scope=hermes redirects directly to Hermes after redemption" change is a polish PR, not a blocker.

If `scope: "hermes"` becomes valuable later, it's a small follow-up that stacks on PR-4.

## Test plan

**Automated (in this PR):**
- [x] `python -c yaml.safe_load(manifest.yaml)` passes
- [x] `python -c yaml.safe_load(compose.yaml)` passes (both proxy and modified Hermes)
- [x] `python -c json.load(.env.schema.json)` passes
- [ ] CI: `caddy validate --config Caddyfile` (will run on PR merge — local Docker wasn't available)

**Manual smoke test** (requires running Dream Server):

- [ ] `dream enable hermes` then `dream enable hermes-proxy` → both containers come up healthy
- [ ] `curl localhost:9120/health` → `ok`
- [ ] `curl -i localhost:9120/` (no cookie) → 303 with `Location: /auth/required`
- [ ] `curl -i localhost:9120/auth/required` → 200, HTML body
- [ ] `curl -i --cookie "dream-session=foo" localhost:9120/` → forwards to Hermes (likely 200 with Hermes's SPA HTML)
- [ ] `curl -i localhost:9119/` (direct, bypass proxy) → connection refused (Hermes no longer binds a host port)
- [ ] Browser: scan magic-link → redemption sets cookie → browse to `localhost:9120` → Hermes UI loads → chat works
- [ ] `docker stop dream-hermes` → proxy returns 502 on authed requests, 303 on unauthed (as expected)
- [ ] WebSocket / SSE through the proxy: open Hermes's chat in browser, send message, verify streaming response renders correctly (this exercises Caddy's WS / SSE passthrough)

## Caveats

- **Direct Hermes access is intentionally severed.** Anyone testing without the proxy needs to re-add a `ports:` binding to `extensions/services/hermes/compose.yaml`, or `docker exec dream-hermes`. Documented in `docs/HERMES.md` and `docs/HERMES-SSO.md`.
- **The `auth-required` page is static HTML in the proxy container.** It doesn't dynamically know which magic-link page to point at — operators wanting to deep-link to a dashboard URL can edit the file in `extensions/services/hermes-proxy/auth-required/`. A follow-up could surface `HERMES_PROXY_AUTH_DOCS_URL` content into the page.
- **Caddy stores state in `data/hermes-proxy/caddy-data/`.** With `auto_https off` this is mostly empty, but the volume keeps Caddy quiet on restart.

### DIFF
--- a/dream-server/.env.example
+++ b/dream-server/.env.example
@@ -313,3 +313,9 @@ LANGFUSE_INIT_USER_PASSWORD=   # auto-generated during install
 # HERMES_LANGUAGE=en                     # UI language
 # (No HERMES_MODEL_NAME — the model lives in /opt/data/config.yaml,
 #  not in env. Edit the file after first start to switch models.)
+#
+# Hermes is reached on the LAN via dream-hermes-proxy (Caddy sidecar) which
+# gates access on the dream-session cookie set by magic-link redemption.
+# Direct Hermes port is NOT bound to the host — only the proxy is LAN-facing.
+# HERMES_PROXY_PORT=9120                 # LAN entry — what users actually browse to
+# HERMES_PROXY_UPSTREAM=dream-hermes:9119  # Where the proxy forwards (internal DNS)
--- a/dream-server/.env.example
+++ b/dream-server/.env.example
@@ -291,3 +291,19 @@ LANGFUSE_INIT_USER_PASSWORD=   # auto-generated during install
 # DREAMFORGE_MODEL=                    # Auto-detected from LLM server if empty
 # DREAMFORGE_RAG_ENABLED=false
 # DREAMFORGE_COMPACT_THRESHOLD=10000
+
+# ═══════════════════════════════════════════════════════════════════
+# Hermes Agent (Remote-capable Chat + Generalist Agent)
+# ═══════════════════════════════════════════════════════════════════
+# Hermes Agent (Nous Research, MIT) — chat-first generalist agent with
+# persistent memory, autonomous skill creation, and 70+ tools. Dream
+# Server runs the upstream image pinned by SHA and points it at the
+# local LLM. Browser UI ships in the upstream image (port 9119).
+# Enable: `dream enable hermes`. Reachable at hermes.<device>.local:9119
+# once mDNS picks it up.
+#
+# HERMES_PORT=9119                       # External dashboard port
+# HERMES_MODEL_NAME=qwen3.5-9b           # Model label passed to Hermes
+# HERMES_LLM_BASE_URL=http://llama-server:8080/v1  # OpenAI-compat endpoint
+# HERMES_LLM_API_KEY=sk-dream-hermes-local  # Dummy (OpenAI SDK requires non-empty)
+# HERMES_LANGUAGE=en                     # UI language
--- a/dream-server/.env.example
+++ b/dream-server/.env.example
@@ -299,15 +299,11 @@ LANGFUSE_INIT_USER_PASSWORD=   # auto-generated during install
 # persistent memory, autonomous skill creation, and 70+ tools. Dream
 # Server runs the upstream image pinned by SHA and points it at the
 # local LLM. Browser UI ships in the upstream image (port 9119).
-# Enable: `dream enable hermes`. With hermes-proxy enabled, end users
-# reach Hermes at http://hermes.<device>.local (port 80, auth-gated by
-# the proxy); without it, Hermes binds HERMES_PORT directly.
+# Enable: `dream enable hermes`. With hermes-proxy enabled (the default
+# pairing), end users reach Hermes at http://hermes.<device>.local
+# (port 80, auth-gated by the proxy). Hermes itself is internal-only —
+# there's no HERMES_PORT knob in this stack.
 #
-# HERMES_PORT=9119                       # Direct external port (used
-#                                          when hermes-proxy is OFF; with
-#                                          hermes-proxy ON, Hermes is
-#                                          internal-only and this is
-#                                          informational).
 # HERMES_LLM_BASE_URL=http://llama-server:8080/v1  # OpenAI-compat endpoint
 # HERMES_LLM_API_KEY=sk-dream-hermes-local  # Dummy (OpenAI SDK requires non-empty)
 # HERMES_LANGUAGE=en                     # UI language
--- a/dream-server/.env.example
+++ b/dream-server/.env.example
@@ -303,7 +303,8 @@ LANGFUSE_INIT_USER_PASSWORD=   # auto-generated during install
 # once mDNS picks it up.
 #
 # HERMES_PORT=9119                       # External dashboard port
-# HERMES_MODEL_NAME=qwen3.5-9b           # Model label passed to Hermes
 # HERMES_LLM_BASE_URL=http://llama-server:8080/v1  # OpenAI-compat endpoint
 # HERMES_LLM_API_KEY=sk-dream-hermes-local  # Dummy (OpenAI SDK requires non-empty)
 # HERMES_LANGUAGE=en                     # UI language
+# (No HERMES_MODEL_NAME — the model lives in /opt/data/config.yaml,
+#  not in env. Edit the file after first start to switch models.)
--- a/dream-server/.env.example
+++ b/dream-server/.env.example
@@ -299,10 +299,15 @@ LANGFUSE_INIT_USER_PASSWORD=   # auto-generated during install
 # persistent memory, autonomous skill creation, and 70+ tools. Dream
 # Server runs the upstream image pinned by SHA and points it at the
 # local LLM. Browser UI ships in the upstream image (port 9119).
-# Enable: `dream enable hermes`. Reachable at hermes.<device>.local:9119
-# once mDNS picks it up.
+# Enable: `dream enable hermes`. With hermes-proxy enabled, end users
+# reach Hermes at http://hermes.<device>.local (port 80, auth-gated by
+# the proxy); without it, Hermes binds HERMES_PORT directly.
 #
-# HERMES_PORT=9119                       # External dashboard port
+# HERMES_PORT=9119                       # Direct external port (used
+#                                          when hermes-proxy is OFF; with
+#                                          hermes-proxy ON, Hermes is
+#                                          internal-only and this is
+#                                          informational).
 # HERMES_LLM_BASE_URL=http://llama-server:8080/v1  # OpenAI-compat endpoint
 # HERMES_LLM_API_KEY=sk-dream-hermes-local  # Dummy (OpenAI SDK requires non-empty)
 # HERMES_LANGUAGE=en                     # UI language
--- a/dream-server/.env.example
+++ b/dream-server/.env.example
@@ -315,7 +315,9 @@ LANGFUSE_INIT_USER_PASSWORD=   # auto-generated during install
 #  not in env. Edit the file after first start to switch models.)
 #
 # Hermes is reached on the LAN via dream-hermes-proxy (Caddy sidecar) which
-# gates access on the dream-session cookie set by magic-link redemption.
+# verifies the dream-session cookie against dashboard-api's verify endpoint
+# (forward_auth → HMAC signature check via DREAM_SESSION_SECRET).
 # Direct Hermes port is NOT bound to the host — only the proxy is LAN-facing.
 # HERMES_PROXY_PORT=9120                 # LAN entry — what users actually browse to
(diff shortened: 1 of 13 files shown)


Lightheartdevs:
Deep audit result: **not ready to merge**.

This is a useful direction, but it currently fails CI and the auth boundary is much weaker than the PR/title implies.

Findings:

1. **CI is red because the Hermes manifest still advertises a host-exposed 9119 port after the compose file removes that port.** `extensions/services/hermes/compose.yaml:51-52` now only `expose`s 9119 internally, but `extensions/services/hermes/manifest.yaml:15-17` still declares `external_port_env: HERMES_PORT` / default 9119. The extension audit fails exactly on this mismatch:

   ```text
   FAIL hermes (optional, docker)
        ERROR compose-port-mismatch: no compose port mapping targets manifest service.port 9119
        WARN compose-port-env-unused: compose ports do not reference service.external_port_env 'HERMES_PORT'
   ```

   This is also a runtime/dashboard bug, not only a lint issue: dashboard-api derives `external_port` from `external_port_env` in `config.py`, so `/api/external-links` and preflight port checks will still think Hermes has a usable direct `:9119` host URL even though this PR intentionally removes it. The Hermes manifest needs to be made internal-only or the extension/audit model needs an explicit way to say ?UI is through hermes-proxy, not this service.?

2. **The proxy accepts any fabricated non-empty cookie, not a redeemed magic-link session.** `extensions/services/hermes-proxy/Caddyfile:62-68` checks only `header_regexp Cookie ... dream-session=[^;]+` and then forwards to Hermes. That means a LAN user can set `Cookie: dream-session=anything` (curl, devtools, a custom browser profile, etc.) and pass the gate without ever redeeming an invite. The docs describe stolen-cookie replay, but the actual condition is weaker: the cookie does not need to be real. If this is meant to close the ?anyone on LAN can reach Hermes? gap, it needs server-side validation, e.g. Caddy `forward_auth` to dashboard-api against a session store/signed cookie, or a signed/verifiable token format.

3. **The documented cross-host flow will not receive the magic-link cookie.** `docs/HERMES-SSO.md:45-52` says redemption sets `dream-session`, then the recipient browses to `http://hermes.<device>.local:9120` and the proxy sees that cookie. But #1155?s magic-link cookie is explicitly host-only/no `Domain` and is set on the redemption host. A cookie from `http://<device>:3001` / `http://<device>.local` is not sent to a different host like `hermes.<device>.local`. The flow should use the same hostname on a different port, e.g. `http://<device>.local:9120`, or the cookie/domain story needs to be redesigned.

4. **This PR is runtime-dependent on #1155 and inherits #1166?s blockers.** The branch is stacked on #1166, which is already not ready due the Hermes process/config issues. It also depends on #1155 for cookie issuance, but #1155 is not in the branch stack and remains blocked. Until those land, enabling `hermes-proxy` gives users a gate without a reliable issuer/validator.

5. **`HERMES_PROXY_AUTH_DOCS_URL` is declared but unused.** `extensions/services/hermes-proxy/manifest.yaml:39-43` and `.env.schema.json` expose this setting, but the compose file does not pass it into Caddy and the static `auth-required/index.html` is not templated. Operators setting it will see no effect.

Checks run:

- `python -m json.tool .env.schema.json` -> pass
- `docker compose -f extensions/services/hermes/compose.yaml -f extensions/services/hermes-proxy/compose.yaml config --quiet` -> pass
- `python scripts/audit-extensions.py --project-dir .` -> fails with the Hermes port mismatch above
- reviewed failing CI run `25748174397`, job `75617031681`
- attempted local `caddy validate` via Docker, but Docker daemon is unavailable in this workspace (`dockerDesktopLinuxEngine` pipe missing)

Recommended path: fix #1166 first, then decide whether this proxy is just a soft ?browser convenience gate? or real access control. If real access control, do not ship presence-only cookie matching; add validation against dashboard-api/session state and keep the public URL on the same host that receives the magic-link cookie.

Lightheartdevs:
Re-audit after the latest author updates: substantially improved, but still not ready to merge.

Fixed from the prior pass:
- CI is green now.
- `scripts/audit-extensions.py --project-dir .` passes with 23 services / 0 errors / 0 warnings.
- Hermes is now modeled as internal-only via `expose: [9119]`, and the audit knows how to treat exposed internal ports.
- The proxy no longer accepts a fabricated non-empty cookie. It now uses Caddy `forward_auth` against dashboard-api's `/api/auth/verify-session` endpoint, which is the right shape.
- The docs now use the same-host cookie flow (`<device>.local`, different port) instead of relying on a cookie to cross to `hermes.<device>.local`.
- `HERMES_PROXY_AUTH_DOCS_URL` is now wired into the Caddy-template static page.

Remaining blockers:

1. This branch still does not include the endpoint the proxy authenticates against. `rg verify-session|session_signer|magic-link|dream-session extensions/services/dashboard-api -S` returns no implementation on this head, and `git merge-base --is-ancestor origin/feat/magic-link-auth HEAD` confirms this branch does not contain #1155's magic-link/auth base. So the deployed Caddy proxy will call `/api/auth/verify-session` on dashboard-api, but dashboard-api from this branch does not have that route. In practice the auth gate fails at runtime until #1155 is merged and this PR is rebased/retargeted onto a base that includes it.

2. `HERMES_PORT` is now a no-op but is still exposed in the Hermes manifest/env schema. The Hermes compose no longer references `HERMES_PORT` at all (`expose: 9119` is hardcoded), yet `extensions/services/hermes/manifest.yaml` still lists `HERMES_PORT` as an env var and `.env.schema.json` still describes it as the Hermes dashboard port. Users/operators can set it and nothing changes. Since the direct host port is intentionally removed, this should be deleted or renamed/documented as not user-configurable; the only functional external knob is `HERMES_PROXY_PORT`.

Checks run:
- `python scripts/audit-extensions.py --project-dir .` -> pass, 23 services, 0 errors/warnings
- `git diff --check origin/feat/hermes-extension...HEAD`
- `docker compose -f extensions/services/hermes/compose.yaml -f extensions/services/hermes-proxy/compose.yaml config --quiet` -> fails in isolation because `dashboard-api` is undefined; full-stack validation needs the base compose/env present.
- Tried Caddy validation through `docker run caddy:2.8.4-alpine caddy validate`, but Docker Desktop's daemon pipe is unavailable in this workspace.

Net: the auth design is now pointed in the right direction, but this PR is still stacked on the wrong base for the route it requires, and it leaves a stale/no-op Hermes port setting behind.

PR #1170, opened 2026-05-12
feat(hermes): route tool calls through APE via pre_tool_call plugin

Lightheartdevs:
## Summary

Closes the "APE policy enforcement" item on the Hermes roadmap. Every Hermes tool call (file_ops, shell, web_fetch, MCP, etc.) now hits Dream Server's APE policy engine \`/verify\` endpoint before execution. Denied calls return a block directive to Hermes; the agent surfaces the denial like any other tool error and the conversation continues.

**This is NOT a fork of Hermes.** Hermes ships a first-class plugin system with \`pre_tool_call\` hook documented as supporting \`{action: block, message: ...}\` return values to block tool execution. Our plugin is just a consumer of that contract — no upstream code modified.

## How it works

1. Hermes's agent loop is about to call a tool
2. \`hermes_cli/plugins.py:get_pre_tool_call_block_message()\` invokes all registered \`pre_tool_call\` hooks
3. Our plugin POSTs to APE \`/verify\` with \`{tool_name, args, session_id, agent_id}\`
4. APE returns \`{allowed, reason}\` (or 403 in STRICT_MODE with detail)
5. If denied → plugin returns \`{action: "block", message: "Tool '...' was blocked by APE policy: ..."}\`
6. Hermes skips the tool, surfaces the denial back to the agent, conversation continues

## Files

| File | Lines | What |
|---|---|---|
| \`extensions/services/hermes/plugins/ape-policy/plugin.yaml\` (new) | 7 | Plugin manifest: name, version, hooks: [pre_tool_call] |
| \`extensions/services/hermes/plugins/ape-policy/__init__.py\` (new) | 140 | Pure-stdlib \`urllib.request\` call to APE /verify, error-translation, fail-open/closed policy. No external deps. |
| \`extensions/services/hermes/compose.yaml\` (+18 / 0) | | New \`APE_*\` env vars and a bind mount that places our plugin into Hermes's user-plugin directory (\`/opt/data/plugins/ape-policy/\`, read-only) |
| \`.env.example\` (+10) | | Documented \`APE_URL\`, \`APE_API_KEY\`, \`APE_FAIL_OPEN\`, \`APE_TIMEOUT\` block |
| \`.env.schema.json\` (+19) | | Schema entries |
| \`docs/HERMES.md\` | | Security posture section updated; Roadmap marks APE shipped |

## Fail behavior

| APE response | Outcome |
|---|---|
| 200 \`allowed: true\` | Tool runs |
| 200 \`allowed: false\` | Block message returned to agent with the reason |
| 403 (STRICT_MODE in APE) | Block with body's \`detail\` |
| Unreachable + \`APE_FAIL_OPEN=true\` (default) | Tool runs (warning logged) |
| Unreachable + \`APE_FAIL_OPEN=false\` | Block with reason \`ape-unreachable-fail-closed\` |

Default \`APE_FAIL_OPEN=true\` matches Dream Server's trusted-LAN posture. Operators exposing Hermes to less-trusted networks should flip to \`false\` so an APE outage doesn't open the gate.

## How the plugin gets discovered

Hermes's [\`plugins.py\` discovery rules](https://github.com/NousResearch/hermes-agent/blob/dd0923bb89ed2dd56f82cb63656a1323f6f42e6f/hermes_cli/plugins.py#L1):

> 1. Bundled plugins  – \`<repo>/plugins/<name>/\`
> 2. User plugins    – \`~/.hermes/plugins/<name>/\`
> 3. Project plugins – \`./.hermes/plugins/<name>/\`
> 4. Pip plugins     – \`hermes_agent.plugins\` entry-point group

We mount into the user-plugins path (\`/opt/data/plugins/ape-policy/\` inside the container, since \`HERMES_HOME=/opt/data\`). The deeper bind mount overlays the broader data volume at that exact subpath — standard Docker compose pattern.

The plugin source is \`:ro\` because it lives in this repo, not in user state. To disable APE checks, either:
- Stop the \`ape\` extension (plugin falls through to fail-open by default)
- Or remove the mount from \`compose.yaml\`
- Or set \`APE_FAIL_OPEN=false\` AND stop \`ape\` (deliberately fail-closed)

## Test plan

- [x] \`python -c yaml.safe_load(plugin.yaml)\` passes
- [x] \`python -c yaml.safe_load(compose.yaml)\` passes
- [x] \`python -c json.load(.env.schema.json)\` passes
- [x] \`python -m py_compile plugins/ape-policy/__init__.py\` passes
- [ ] Manual smoke (requires running stack):
  - [ ] \`dream enable ape hermes\` → both come up healthy
  - [ ] Hermes logs show \`ape-policy plugin registered\` on start
  - [ ] In Hermes chat, ask it to read a file → plugin logs the \`/verify\` call; APE audit log shows the decision
  - [ ] Configure an APE deny rule for \`terminal\` (or whichever shell tool Hermes uses) → ask Hermes to run a command → it returns a block message in chat
  - [ ] Stop the ape container with \`APE_FAIL_OPEN=true\` → tools still run (warning log); flip to \`false\` → tools fail

## Caveats

- **Plugin runs on every tool call** — adds an HTTP round-trip per call. APE is on the bridge network so latency is sub-millisecond locally; not a concern at home scale.
- **APE's classify_intent + policy.yaml is what actually decides.** This PR is just the wire-up; tuning the deny rules is operator work in \`config/ape/policy.yaml\`.
- **No \`post_tool_call\` hook here.** APE's \`/verify\` writes its own audit entry at decision time; surfacing tool RESULTS to APE for richer auditing (was the file write actually successful, did the shell return non-zero) is a follow-up worth considering once the policy story is mature.

Stacked on \`feat/hermes-extension\` (#1166).

### DIFF
--- a/dream-server/.env.example
+++ b/dream-server/.env.example
@@ -315,6 +315,6 @@ LANGFUSE_INIT_USER_PASSWORD=   # auto-generated during install
 # call through (default; matches Dream's trusted-LAN posture). Flip to false
 # if you're exposing Hermes to less-trusted networks.
 # APE_URL=http://ape:7890                # APE service URL (internal Docker DNS)
-# APE_API_KEY=                           # Bearer token if APE has auth enabled
+# APE_API_KEY=                           # X-API-Key value; must match the APE service's APE_API_KEY
 # APE_FAIL_OPEN=true                     # true = allow on APE-unreachable
 # APE_TIMEOUT=5                          # Per-request timeout (seconds)
--- a/dream-server/.env.example
+++ b/dream-server/.env.example
@@ -291,3 +291,19 @@ LANGFUSE_INIT_USER_PASSWORD=   # auto-generated during install
 # DREAMFORGE_MODEL=                    # Auto-detected from LLM server if empty
 # DREAMFORGE_RAG_ENABLED=false
 # DREAMFORGE_COMPACT_THRESHOLD=10000
+
+# ═══════════════════════════════════════════════════════════════════
+# Hermes Agent (Remote-capable Chat + Generalist Agent)
+# ═══════════════════════════════════════════════════════════════════
+# Hermes Agent (Nous Research, MIT) — chat-first generalist agent with
+# persistent memory, autonomous skill creation, and 70+ tools. Dream
+# Server runs the upstream image pinned by SHA and points it at the
+# local LLM. Browser UI ships in the upstream image (port 9119).
+# Enable: `dream enable hermes`. Reachable at hermes.<device>.local:9119
+# once mDNS picks it up.
+#
+# HERMES_PORT=9119                       # External dashboard port
+# HERMES_MODEL_NAME=qwen3.5-9b           # Model label passed to Hermes
+# HERMES_LLM_BASE_URL=http://llama-server:8080/v1  # OpenAI-compat endpoint
+# HERMES_LLM_API_KEY=sk-dream-hermes-local  # Dummy (OpenAI SDK requires non-empty)
+# HERMES_LANGUAGE=en                     # UI language
--- a/dream-server/.env.example
+++ b/dream-server/.env.example
@@ -303,7 +303,8 @@ LANGFUSE_INIT_USER_PASSWORD=   # auto-generated during install
 # once mDNS picks it up.
 #
 # HERMES_PORT=9119                       # External dashboard port
-# HERMES_MODEL_NAME=qwen3.5-9b           # Model label passed to Hermes
 # HERMES_LLM_BASE_URL=http://llama-server:8080/v1  # OpenAI-compat endpoint
 # HERMES_LLM_API_KEY=sk-dream-hermes-local  # Dummy (OpenAI SDK requires non-empty)
 # HERMES_LANGUAGE=en                     # UI language
+# (No HERMES_MODEL_NAME — the model lives in /opt/data/config.yaml,
+#  not in env. Edit the file after first start to switch models.)
--- a/dream-server/.env.example
+++ b/dream-server/.env.example
@@ -308,3 +308,13 @@ LANGFUSE_INIT_USER_PASSWORD=   # auto-generated during install
 # HERMES_LANGUAGE=en                     # UI language
 # (No HERMES_MODEL_NAME — the model lives in /opt/data/config.yaml,
 #  not in env. Edit the file after first start to switch models.)
+#
+# APE policy enforcement — Hermes's `pre_tool_call` hook routes every tool
+# call through APE for allow/deny + audit. The ape extension must be enabled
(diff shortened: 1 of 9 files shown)


Lightheartdevs:
Deep audit result: **not ready to merge yet**.

What I checked:
- GitHub checks are green and merge state is clean.
- `python -m py_compile extensions/services/hermes/plugins/ape-policy/__init__.py`
- duplicate-key scan for `.env.schema.json`
- `python scripts/audit-extensions.py --project-dir .` -> pass, 22 services, 0 errors/warnings
- `docker compose -f extensions/services/hermes/compose.yaml config --quiet`
- Verified the pinned upstream Hermes plugin contract (`pre_tool_call` block return shape) and Dream Server's current APE `/verify` implementation.

Findings:

1. APE auth is wired with the wrong header, so the plugin does not actually reach policy evaluation. `extensions/services/hermes/plugins/ape-policy/__init__.py:68-70` sends `Authorization: Bearer <APE_API_KEY>`, but APE's dependency at `extensions/services/ape/main.py:248-250` only reads the `X-API-Key` header. Two bad cases fall out of that:
   - default `.env` / compose has no `APE_API_KEY`, so APE generates a random key at startup (`extensions/services/ape/main.py:60-63`) and still requires `X-API-Key`; Hermes sends no key and gets 401.
   - if the operator does set `APE_API_KEY`, Hermes still sends it as `Authorization`, APE still sees no `X-API-Key`, and it gets 401.

   Because the plugin treats non-403 HTTP errors as `_fail()` (`__init__.py:91-92`) and `APE_FAIL_OPEN` defaults to true, the shipped default silently allows all tool calls without APE audit/enforcement. With `APE_FAIL_OPEN=false`, the same auth mismatch blocks all tool calls as an APE outage. Fix: send `X-API-Key: <APE_API_KEY>` and make sure the same env var is set for both `ape` and `hermes`; I would also consider treating 401/403 auth failures as configuration errors rather than fail-open outages.

2. `.env.schema.json` now contains duplicate `APE_API_KEY` entries. It already exists at `.env.schema.json:297-300`, and this PR adds a second one at `.env.schema.json:727-730`. Standard JSON parsing keeps only one effective value, so schema consumers can see different descriptions depending on parser behavior and future edits become brittle. Please update the existing entry instead of adding a second key.

3. This is stacked on #1166 (`baseRefName=feat/hermes-extension`), which still has unresolved runtime/config blockers. This PR should not merge to `main` until the base Hermes extension is corrected and landed.

Notes:
- The Hermes plugin discovery/hook shape itself looks plausible: upstream `get_pre_tool_call_block_message()` does accept `{"action":"block","message":"..."}` returns from `pre_tool_call` hooks.
- I could not do a live APE/Hermes smoke because Docker Desktop / the Docker engine is not available in this environment, so the runtime conclusion is based on static contract checks against the current APE service and pinned upstream Hermes code.

PR #1171, opened 2026-05-12
feat(hermes): one sidebar tile after the proxy split

Lightheartdevs:
## Summary

Tiny cleanup pass for the post-H-2 dashboard surface. After [PR #1168](https://github.com/Light-Heart-Labs/DreamServer/pull/1168) made Hermes internal-only, its manifest **still** declared \`external_port_env: HERMES_PORT\` — which the dashboard's \`/api/external-links\` reads to publish sidebar tiles. The user would have seen TWO \"Hermes\" tiles in the sidebar: one unreachable (raw hermes at 9119, no host port bound) and one real (hermes-proxy at 9120).

This PR fixes that to one tile labeled \"Hermes Agent\" pointing at the proxy.

## Changes

| File | Change |
|---|---|
| \`extensions/services/hermes/manifest.yaml\` | Dropped \`external_port_env\` + \`external_port_default\`. Hermes is internal-only on the Docker bridge after H-2; it must not advertise an external port. Renamed \`name: Hermes Agent\` → \`name: Hermes Agent (internal)\` so anyone on the Extensions page sees what's running where. |
| \`extensions/services/hermes-proxy/manifest.yaml\` | Renamed \`name: Hermes Auth Proxy\` → \`name: Hermes Agent\`. The proxy IS the user-facing Hermes — internal machinery shouldn't be in the sidebar tile label. Added aliases \`[hermes, agent, hermes-gate, hermes-auth-proxy]\` so \`dream enable hermes-auth-proxy\` etc. still resolve. |
| \`docs/HERMES.md\` | Roadmap \"Dream-side status panel\" item marked partially shipped (this naming / single-tile cleanup IS the practical chunk of it). |

## What's still deferred

The deeper \"session count / skill inventory inside the Dream dashboard\" piece of the original roadmap item is intentionally NOT in this PR. Two reasons:

1. **Hermes already has \`AnalyticsPage\`** built in — strictly richer than anything we'd surface.
2. **Cross-surface scraping requires bypassing Hermes's own auth** (the per-process X-Hermes-Session-Token), which we explicitly avoided in PR #1168.

The natural follow-up (if real demand exists) is to put a small \"Hermes\" widget on the dashboard home that shows just container-level health (\"running / healthy / x sessions today\") via Docker stats, without trying to introspect into Hermes's app state. Not built here.

## Test plan

- [x] \`python -c yaml.safe_load(<both manifests>)\` passes
- [ ] Manual smoke (post-merge):
  - [ ] Dashboard sidebar shows one \"Hermes Agent\" tile, links to port 9120
  - [ ] Extensions page shows both \"Hermes Agent (internal)\" and \"Hermes Agent\" with distinct \`id\` / \`container\` so admins can see the architecture
  - [ ] \`dream enable hermes-auth-proxy\` still resolves (via alias)

Stacked on \`feat/hermes-magic-link-sso\` (#1168) because that's the PR that flipped Hermes to internal-only.

### DIFF
--- a/dream-server/.env.example
+++ b/dream-server/.env.example
@@ -313,3 +313,9 @@ LANGFUSE_INIT_USER_PASSWORD=   # auto-generated during install
 # HERMES_LANGUAGE=en                     # UI language
 # (No HERMES_MODEL_NAME — the model lives in /opt/data/config.yaml,
 #  not in env. Edit the file after first start to switch models.)
+#
+# Hermes is reached on the LAN via dream-hermes-proxy (Caddy sidecar) which
+# gates access on the dream-session cookie set by magic-link redemption.
+# Direct Hermes port is NOT bound to the host — only the proxy is LAN-facing.
+# HERMES_PROXY_PORT=9120                 # LAN entry — what users actually browse to
+# HERMES_PROXY_UPSTREAM=dream-hermes:9119  # Where the proxy forwards (internal DNS)
--- a/dream-server/.env.example
+++ b/dream-server/.env.example
@@ -291,3 +291,19 @@ LANGFUSE_INIT_USER_PASSWORD=   # auto-generated during install
 # DREAMFORGE_MODEL=                    # Auto-detected from LLM server if empty
 # DREAMFORGE_RAG_ENABLED=false
 # DREAMFORGE_COMPACT_THRESHOLD=10000
+
+# ═══════════════════════════════════════════════════════════════════
+# Hermes Agent (Remote-capable Chat + Generalist Agent)
+# ═══════════════════════════════════════════════════════════════════
+# Hermes Agent (Nous Research, MIT) — chat-first generalist agent with
+# persistent memory, autonomous skill creation, and 70+ tools. Dream
+# Server runs the upstream image pinned by SHA and points it at the
+# local LLM. Browser UI ships in the upstream image (port 9119).
+# Enable: `dream enable hermes`. Reachable at hermes.<device>.local:9119
+# once mDNS picks it up.
+#
+# HERMES_PORT=9119                       # External dashboard port
+# HERMES_MODEL_NAME=qwen3.5-9b           # Model label passed to Hermes
+# HERMES_LLM_BASE_URL=http://llama-server:8080/v1  # OpenAI-compat endpoint
+# HERMES_LLM_API_KEY=sk-dream-hermes-local  # Dummy (OpenAI SDK requires non-empty)
+# HERMES_LANGUAGE=en                     # UI language
--- a/dream-server/.env.example
+++ b/dream-server/.env.example
@@ -299,15 +299,11 @@ LANGFUSE_INIT_USER_PASSWORD=   # auto-generated during install
 # persistent memory, autonomous skill creation, and 70+ tools. Dream
 # Server runs the upstream image pinned by SHA and points it at the
 # local LLM. Browser UI ships in the upstream image (port 9119).
-# Enable: `dream enable hermes`. With hermes-proxy enabled, end users
-# reach Hermes at http://hermes.<device>.local (port 80, auth-gated by
-# the proxy); without it, Hermes binds HERMES_PORT directly.
+# Enable: `dream enable hermes`. With hermes-proxy enabled (the default
+# pairing), end users reach Hermes at http://hermes.<device>.local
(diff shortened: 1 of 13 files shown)


Lightheartdevs:
Deep audit result: **not ready to merge yet**.

What I checked:
- GitHub `integration-smoke` is red; merge state is `UNSTABLE`.
- Reviewed the PR diff against `feat/hermes-magic-link-sso`.
- `python scripts/audit-extensions.py --project-dir .` locally reproduces the CI failure.
- `python -m json.tool .env.schema.json` passes.
- `docker compose -f extensions/services/hermes/compose.yaml -f extensions/services/hermes-proxy/compose.yaml config --quiet` passes.

Findings:

1. Extension audit is red on the alias split. `extensions/services/hermes/manifest.yaml:13` still declares alias `agent`, while `extensions/services/hermes-proxy/manifest.yaml:12` also adds `agent`. The audit fails both manifests with `alias-collision`. Since the proxy is now the user-facing Hermes entry, the internal Hermes service should probably give up the user-facing aliases (`hermes` / `agent`) and keep only an internal/admin-specific alias, or the proxy aliases should be narrowed. As written, the resolver has no unambiguous answer for `dream enable agent`.

2. Extension audit is still red for the internal Hermes port. This PR removes `external_port_env`/`external_port_default`, but `extensions/services/hermes/manifest.yaml:16` still declares `port: 9119` while `extensions/services/hermes/compose.yaml` no longer has a `ports:` mapping targeting 9119. The current audit contract treats a non-core manifest `service.port` as requiring a compose port mapping, so this still fails with `compose-port-mismatch`. Either the manifest/audit schema needs an explicit internal-only service mode that accepts `expose: 9119`, or the Hermes manifest/compose need to stay consistent under the existing contract.

3. This is stacked on #1168, which is already not ready. It inherits the proxy blockers from that branch: the current Caddy cookie gate accepts any non-empty fabricated `dream-session` cookie, and the documented magic-link redemption host does not share the host-only cookie with `hermes.<device>.local:9120`. This cleanup should land only after the underlying proxy/auth behavior is fixed.

Net: the direction is right ? one visible Hermes tile via the proxy is the correct dashboard shape ? but the branch still fails CI and needs the manifest/audit contract cleaned up before merge.

Lightheartdevs:
Re-audit after the latest update: still not ready.

What improved:
- The previous internal-port audit failure is gone in the stack below this PR: #1168 now teaches the extension audit to accept `expose:` for internal-only services, and Hermes's direct `external_port_env/default` is removed.
- The intent of this PR is still right: the dashboard should expose one user-facing Hermes tile, and that tile should point at `hermes-proxy`, not the raw internal Hermes container.

Current blockers:

1. CI is still red. `integration-smoke` fails, and I can reproduce the relevant local failure with `python scripts/audit-extensions.py --project-dir .`:

```text
FAIL hermes
  ERROR alias-collision: alias 'agent' already belongs to service 'hermes-proxy'
FAIL hermes-proxy
  ERROR alias-collision: alias 'agent' already belongs to service 'hermes'
```

The patch renamed raw Hermes to `Hermes Agent (internal)` but left `aliases: [agent, hermes-agent]` on the internal service while giving `agent` to the proxy too. The internal service needs to give up the user-facing alias, or the proxy needs a narrower alias list. Since the proxy is the intended user-facing tile, I would remove `agent` from `extensions/services/hermes/manifest.yaml`.

2. This still inherits #1168's runtime blocker: the stack does not include the dashboard-api `/api/auth/verify-session` endpoint that the proxy forwards to. `git merge-base --is-ancestor origin/feat/magic-link-auth HEAD` says this branch does not contain the magic-link/auth base (#1155). So even after the alias collision is fixed, this cleanup should not merge until #1168 is rebased/retargeted onto a base containing #1155 or #1155 has landed in the base chain.

3. The inherited `HERMES_PORT` setting is still stale/no-op. Hermes is internal-only now and its compose file hardcodes `expose: 9119`; setting `HERMES_PORT` changes nothing, but the internal Hermes manifest still lists it under `env_vars`. That should be removed in the same cleanup chain so the dashboard/env editor does not advertise a dead knob.

Checks run:
- `python scripts/audit-extensions.py --project-dir .` -> fails on alias collision above
- `git diff --check origin/feat/hermes-magic-link-sso...HEAD`
- confirmed this head still does not contain `origin/feat/magic-link-auth`

Net: small PR, right direction, but still red and still stacked above an auth proxy branch that cannot work without #1155.

PR #1173, opened 2026-05-12
feat(auth): HMAC-signed dream-session cookies + verify endpoint

Lightheartdevs:
## Summary

Foundation for stateless session-cookie validation. Closes the "any non-empty cookie passes" weakness in [#1168](https://github.com/Light-Heart-Labs/DreamServer/pull/1168) — the Hermes auth-proxy will switch from regex presence-check to Caddy \`forward_auth\` against the new \`/api/auth/verify-session\` endpoint.

## Cookie shape

\`<random-id>.<expiry-epoch>.<signature>\` where:

- **random-id** — opaque per-redemption ID (\`secrets.token_urlsafe(24)\`); used for audit / future revocation list
- **expiry-epoch** — integer Unix timestamp; cookie stops being honored when current time \`>=\` this
- **signature** — \`HMAC-SHA256(DREAM_SESSION_SECRET, "<id>.<expiry>")\` base64-url, no padding

## Why this design

- **Stateless verifier.** Only state needed is the signing secret. No DB lookup per request — the proxy can validate at line rate.
- **Tamper-evident.** An attacker who tries to extend a leaked cookie's lifetime by editing the timestamp invalidates the signature.
- **Bounded revocation.** Rotating \`DREAM_SESSION_SECRET\` kicks all current sessions. Per-cookie revocation can be added later keyed on the opaque random-id (the shape reserves room).
- **No identity in the cookie.** Random-id is opaque on purpose; putting the username in a same-origin-readable cookie would leak it to any JS on any same-origin page. Conservative posture; deliberate.

## Files

| File | Lines | What |
|---|---|---|
| \`session_signer.py\` (new) | 138 | Pure helper. \`issue()\` mints; \`verify()\` returns \`(ok, reason)\`. \`hmac.compare_digest\` for timing-safe signature compare. Refuses to issue when \`DREAM_SESSION_SECRET\` is unset (won't silently produce unsignable cookies). |
| \`routers/auth.py\` (new) | 70 | \`GET /api/auth/verify-session\` — reads the cookie, validates, returns 200 + expiry on success, 401 on any failure. Response body is byte-identical across all failure reasons (defeats probing). |
| \`main.py\` (+3) | | Router registration. |
| \`.env.example\` + \`.env.schema.json\` | | \`DREAM_SESSION_SECRET\` (secret). Installer needs to generate at install time, same pattern as DASHBOARD_API_KEY / DREAM_AGENT_KEY. |
| \`tests/test_session_signer.py\` (new) | 138 | 18 tests: roundtrip, malformed shapes, bad signatures, expiry boundary, **the timestamp-extension attack**, secret rotation, no-secret state. |
| \`tests/test_auth_router.py\` (new) | 106 | 6 tests including a **probing-defense test** that asserts all 401 bodies are byte-identical so an attacker can't tell which rejection reason hit. |

## Security properties verified by tests

- A cookie signed with secret A is rejected if the server now has secret B (rotation works).
- A cookie with a tampered expiry field fails as `bad-signature`, not `expired` — so extending a leaked cookie's lifetime is detected.
- All 401 responses are identical so an attacker can't distinguish "bad signature" vs "expired" vs "malformed".
- The verify endpoint does NOT require the dashboard API key — Caddy can't easily inject Bearer headers through \`forward_auth\`, and the cookie itself is the credential being validated.

## Test plan

- [x] 24/24 tests pass (18 signer + 6 router)
- [x] Ruff clean
- [ ] Manual smoke (post-merge of follow-ups that issue/consume these cookies):
  - [ ] Set \`DREAM_SESSION_SECRET\` in .env, restart dashboard-api
  - [ ] Mint a cookie: \`python -c "from session_signer import issue; print(issue(60))"\`
  - [ ] \`curl --cookie "dream-session=\$COOKIE" http://localhost:3002/api/auth/verify-session\` → 200
  - [ ] Same after expiry → 401
  - [ ] Edit one char in the signature → 401

## Follow-up PRs that consume this

| PR | Consumer | What changes |
|---|---|---|
| [#1155](https://github.com/Light-Heart-Labs/DreamServer/pull/1155) | magic-link router | Cookie issuance switches from \`secrets.token_urlsafe(24)\` to \`session_signer.issue()\` |
| [#1168](https://github.com/Light-Heart-Labs/DreamServer/pull/1168) | Hermes auth-proxy | Caddyfile switches from regex presence-check to \`forward_auth dashboard-api:3002 { uri /api/auth/verify-session }\` |


Lightheartdevs:
## Coordination note — conflicts with #1155

This PR and #1155 both add byte-identical copies of:

- `dream-server/extensions/services/dashboard-api/session_signer.py`
- `dream-server/extensions/services/dashboard-api/routers/auth.py`
- `dream-server/extensions/services/dashboard-api/tests/test_session_signer.py`
- `dream-server/extensions/services/dashboard-api/tests/test_auth_router.py`

…plus overlapping changes to `main.py` (router registration) and `.env.example` / `.env.schema.json`.

Confirmed dependency direction: `#1155/routers/magic_link.py:575` calls `session_signer.issue(ttl_seconds=SESSION_TTL_SECONDS)` — so #1155 genuinely consumes this PR's signer. Not actually redundant; the duplication is a coordination artifact.

**Suggested merge order:** land this PR first as the foundation, then rebase #1155 to drop the four duplicated files and keep only `routers/magic_link.py`, `tests/test_magic_link.py`, and the diff-on-`main.py`. Posted a matching note on #1155.

Lightheartdevs:
Deep audit result for #1173: the crypto helper and endpoint are clean, but the PR is not ready to merge because the runtime wiring is incomplete.

Findings:

1. `DREAM_SESSION_SECRET` is never passed into the `dashboard-api` container. The signer reads `os.environ.get(DREAM_SESSION_SECRET)` at import time (`extensions/services/dashboard-api/session_signer.py:66`), but `docker-compose.base.yml`'s `dashboard-api.environment` block does not include that variable. I verified with:

   ```text
   $env:WEBUI_SECRET='dummy'; $env:DREAM_SESSION_SECRET='secret'; docker compose -f docker-compose.base.yml config | Select-String DREAM_SESSION_SECRET
   # no DREAM_SESSION_SECRET in the rendered dashboard-api environment
   ```

   Result: even if `.env` contains the secret, the running dashboard-api sees an empty secret, `verify()` returns `no-secret`, and `/api/auth/verify-session` always 401s. Add `DREAM_SESSION_SECRET=${DREAM_SESSION_SECRET:-}` to the dashboard-api environment before this can back #1155/#1168.

2. The installers/env generators do not create or upsert `DREAM_SESSION_SECRET`. The PR adds `.env.example` and schema docs saying to generate it, but searches across `installers/`, `install*.sh`, `scripts/`, and env-generator paths only find the new example/schema/code references, not generation. This means fresh installs and upgrades will not have the secret unless an operator manually adds it. The existing patterns for `DASHBOARD_API_KEY`, `DREAM_AGENT_KEY`, and `SHIELD_API_KEY` should be extended here.

What looks good:
- Cookie shape/signing approach is reasonable: opaque ID + expiry + HMAC over payload, `hmac.compare_digest`, no identity in the cookie.
- Failure responses are generic at the router boundary.
- Tests cover tampered expiry, secret rotation, malformed cookies, no-secret behavior, and endpoint probing defense.

Local verification:
- `python -m pytest extensions/services/dashboard-api/tests/test_session_signer.py extensions/services/dashboard-api/tests/test_auth_router.py -q` passed: 24 passed.
- `python -m ruff check extensions/services/dashboard-api/session_signer.py extensions/services/dashboard-api/routers/auth.py extensions/services/dashboard-api/tests/test_session_signer.py extensions/services/dashboard-api/tests/test_auth_router.py` passed.
- `python -m compileall -q extensions/services/dashboard-api/session_signer.py extensions/services/dashboard-api/routers/auth.py` passed.

Net: the implementation is close, but please wire the secret into dashboard-api and generate/upsert it in the installers before merging. Otherwise the follow-on auth proxy will be correctly written but permanently unauthenticated in real installs.

Lightheartdevs:
Audit pass for current head \2493e296\ against current main after #1102/#1149 landed.\n\nChecks run locally on a no-commit merge with current \origin/main\:\n- \python -m pytest dream-server/extensions/services/dashboard-api/tests/test_session_signer.py dream-server/extensions/services/dashboard-api/tests/test_auth_router.py -q\ -> 27 passed\n- \python -m json.tool dream-server/.env.schema.json\ -> ok\n- \ash -n dream-server/installers/phases/06-directories.sh dream-server/installers/macos/lib/env-generator.sh\ -> ok\n- PowerShell parser check for \installers/windows/lib/env-generator.ps1\ -> ok\n\nSecurity audit notes: fail-closed behavior on missing \DREAM_SESSION_SECRET\ is correct, expiry tampering breaks the signature, compare uses \hmac.compare_digest\, all 401 bodies are intentionally identical, and Linux/macOS/Windows install paths now generate/preserve/pass the secret into dashboard-api. I did not find a merge blocker.

PR #1178, opened 2026-05-13
fix(uninstall): chown INSTALL_DIR before rm so container-UID files clear

Lightheartdevs:
## Summary

- `dream-uninstall.sh` ends with `rm -rf "$INSTALL_DIR"`. That fails on every file containerized services wrote as a non-host UID: open-webui's HuggingFace cache (huggingface user inside the image), qdrant's `aliases/data.json`, baserow's media + jwt signing key, searxng's `settings.yml`, etc.
- The current rm has no error gate, so the script exits success while `$INSTALL_DIR` is still on disk. A subsequent fresh install then hits Permission-denied at the first `rsync` and the user is stuck.
- Fix: `sudo chown -R \$(id -u):\$(id -g) "$INSTALL_DIR"` before the rm. If sudo isn't available or fails, fall through to the rm and `log_warn` the exact recovery command (`sudo rm -rf …`) so the operator isn't left guessing.

The chown pattern is already used by `installers/phases/06-directories.sh:48-56` at install time when picking up container-owned dirs from a prior run — this PR just mirrors it on the uninstall side so the round-trip is symmetric.

## Repro

Reproduced today on three different boxes during a fresh-install test (2026-05-12):

| Box | OS / arch | Leftover after uninstall |
|---|---|---|
| Tower2 | Ubuntu / x86_64 | open-webui HF cache + qdrant aliases |
| Strix Halo | Ubuntu / x86_64 (AMD) | open-webui HF cache + qdrant aliases |
| DGX Spark | Ubuntu / aarch64 | baserow media thumbnails + searxng config |

Each emitted hundreds of \`rm: cannot remove '...': Permission denied\` lines, then exited 0. The Permission-denied output spam itself is also confusing because the script claims \"Installation directory cleaned\" at the end.

## Test plan

- [x] Syntax: \`bash -n dream-uninstall.sh\`
- [ ] Reviewer: install via \`install.sh --all --non-interactive\` on a box with passwordless sudo, then \`./dream-uninstall.sh --force\`. \`$INSTALL_DIR\` should be gone with no Permission-denied output.
- [ ] Reviewer: same on a box WITHOUT passwordless sudo. Expected: chown step prints \`log_warn\`, rm runs anyway, final \`log_warn\` lists the recovery command.

## Out of scope

The same chown-then-rm pattern probably wants to apply to \`\$HOME/.dream-server\` backup cleanup further down (line 174), but that dir is created by the host user and shouldn't have UID-mismatched files in practice. Holding off until/unless we see it bite.

### DIFF
--- a/dream-server/dream-uninstall.sh
+++ b/dream-server/dream-uninstall.sh
@@ -152,6 +152,7 @@ fi
 
 # 5. Remove install directory (with optional data/model preservation)
 log_info "Removing installation directory..."
+INSTALL_DIR_CLEANED=true
 if $KEEP_MODELS && [[ -d "$INSTALL_DIR/data/models" ]]; then
     MODELS_BACKUP="$HOME/.dream-server-models-backup"
     mkdir -p "$MODELS_BACKUP"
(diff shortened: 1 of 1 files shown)


Lightheartdevs:
Audit pass after direct fix:

- The chown-before-remove idea is valuable and addresses a real uninstall failure: container-owned files can strand `$INSTALL_DIR` and break the next install.
- I found one blocker in the fallback path: with `set -e`, a failed `rm -rf "$INSTALL_DIR"` could exit before the new recovery warning ran, and the final success line could still be misleading when the directory remained.
- Pushed `bd793d8` to make `rm` best-effort, warn when sudo is missing, quote the recovery command, and only print the cleanup success line when the install dir is actually gone or intentionally preserved.
- Local verification: `bash -n dream-server/dream-uninstall.sh`; `git diff --check`. Local ShellCheck is unavailable in this shell, so CI ShellCheck remains the lint source of truth.

Leaving this unmerged until the refreshed CI run completes on the new head.

PR #1180, opened 2026-05-13
fix(installer/macos): accept any docker daemon, strip stale credsStore

Lightheartdevs:
## Summary

Two related fresh-Mac install bugs, fixed together because they share the same root assumption (\"only Docker Desktop is supported\").

### Bug 1: \"Docker Desktop not found\" rejects valid setups

\`installers/macos/install-macos.sh:201-214\` rejects with a Docker-Desktop-only URL hint when the actual preflight check is just \`docker version\` against whatever daemon is up. Colima, Rancher Desktop, OrbStack, and a forwarded socket all pass that check today — but the error message strands users who have a perfectly working CLI-only Colima setup.

### Bug 2: Stale \`credsStore: desktop\` crashes compose-up under Colima

\`brew install docker\` (the CLI formula, not the cask) creates \`~/.docker/config.json\` with \`\"credsStore\": \"desktop\"\` — a Docker-Desktop-flavoured default. If the user is on Colima (which doesn't ship \`docker-credential-desktop\`), \`docker compose up\` immediately bombs:

\`\`\`
error getting credentials - err: exec: \"docker-credential-desktop\":
executable file not found in \$PATH
[XX] docker compose up failed
\`\`\`

### What this PR does

- **Detects the docker backend** (Desktop / Colima / Rancher / OrbStack / other) from \`docker context inspect\` and stashes it in \`\$DOCKER_BACKEND\`.
- **Per-backend daemon-not-running guidance**: \`colima start ...\` for Colima, \"open Applications\" for Desktop, etc.
- **Better \"engine not found\" message**: lists all four common engines with one-line install commands instead of pointing only at Docker Desktop.
- **Auto-strips stale credsStore=desktop** under non-Desktop backends when \`docker-credential-desktop\` is not on PATH. Operators see a single \`ai_warn\` line instead of a cryptic compose-up failure two phases later.

## Repro

Fresh Mac Mini M4 / macOS 26.2 / no prior Docker / no Homebrew:

\`\`\`bash
# bootstrap (out of scope here, but documents the real flow)
/bin/bash -c \"\$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)\"
brew install colima docker docker-compose
colima start

# install
git clone https://github.com/Light-Heart-Labs/DreamServer && cd DreamServer
./install.sh --all --non-interactive
\`\`\`

**Before**:
- Phase 1: \"Docker Desktop not found\" (even though \`docker info\` works fine).
- After hand-installing Docker Desktop: \"Docker Desktop is not running\" (still wrong msg if you opened Colima instead).
- After getting daemon up: \"error getting credentials\" at Phase 5 compose-up.

**After**:
- Phase 1: \`Docker daemon ready (v..., backend=colima)\`.
- Stale credsStore stripped with one \`ai_warn\` line.
- Phase 5 reaches the actual compose-up.

## Out of scope

- Auto-installing a docker backend (was tempted, but \`brew install --cask docker\` triggers macOS GUI flow, and Colima needs its own startup decision — would rather guide than guess).
- Documenting this in installers/macos/README.md (separate doc PR).

## Test plan

- [x] \`bash -n\` on both modified files
- [ ] Reviewer: \`./install.sh --all --non-interactive\` on a Mac with Colima only (no Docker Desktop). Should pass preflight and proceed to compose-up.
- [ ] Reviewer: same on a Mac with Docker Desktop installed but not running. Should print the Desktop-specific guidance.
- [ ] Reviewer: same on a Mac with both engines installed. Active context wins (\`docker context use\` honored).

### DIFF
--- a/dream-server/installers/macos/install-macos.sh
+++ b/dream-server/installers/macos/install-macos.sh
@@ -198,20 +198,53 @@ info_box "Chip:" "${APPLE_CHIP}"
 info_box "Variant:" "${APPLE_CHIP_VARIANT}"
 ai_ok "Apple Silicon detected"
 
-# Docker Desktop
+# Docker engine (Docker Desktop, Colima, Rancher Desktop, OrbStack, or a
+# forwarded socket are all acceptable — see lib/detection.sh).
 test_docker_desktop
 if ! $DOCKER_INSTALLED; then
-    ai_err "Docker Desktop not found. Install from https://docs.docker.com/desktop/install/mac-install/"
+    ai_err "Docker engine not found (no \`docker\` CLI on PATH)."
+    ai_err "Pick one and install:"
+    ai_err "  - Docker Desktop:  https://docs.docker.com/desktop/install/mac-install/"
+    ai_err "  - Colima (CLI):    brew install colima docker docker-compose && colima start"
+    ai_err "  - Rancher Desktop: https://rancherdesktop.io"
+    ai_err "  - OrbStack:        https://orbstack.dev"
     exit 1
 fi
 ai_ok "Docker CLI found"
 
 if ! $DOCKER_RUNNING; then
-    ai_err "Docker Desktop is not running."
-    ai "Start it from the Applications folder or menu bar, then re-run this installer."
+    ai_err "Docker daemon is not responding."
+    case "${DOCKER_BACKEND:-unknown}" in
+        desktop)  ai "Start Docker Desktop from /Applications or the menu bar, then re-run this installer." ;;
+        colima)   ai "Run \`colima start\` (e.g. \`colima start --cpu 6 --memory 12 --disk 60\`) then re-run this installer." ;;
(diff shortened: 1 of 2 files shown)


Lightheartdevs:
Audit pass after direct fix:

- Found one behavior gap: backend-specific daemon guidance only ran after `docker version` succeeded, so stopped Colima/Desktop/Rancher/OrbStack users would fall into the generic path. I pushed `7341e83` to detect the active Docker context before the daemon probe.
- Rechecked the stale `credsStore: desktop` guard. It is scoped to non-Desktop backends and only fires when `docker-credential-desktop` is absent, so it avoids the Colima compose-up crash without touching valid Docker Desktop setups.
- Local verification: `bash -n installers/macos/lib/detection.sh installers/macos/install-macos.sh`; mocked stopped-Colima path; mocked running Docker Desktop path; `git diff --check`.

Leaving this unmerged until the refreshed CI run completes on the new head.

PR #1181, opened 2026-05-13
fix(dashboard-api): include session_signer.py in container image

Lightheartdevs:
## Summary

\`routers/auth.py\` does \`import session_signer\`, but \`session_signer.py\` is missing from \`extensions/services/dashboard-api/Dockerfile\`'s COPY list. Every \`dashboard-api\` startup on \`main\` today crashes with \`ModuleNotFoundError: No module named 'session_signer'\` and enters a restart loop, which takes down the dashboard frontend (nginx can't resolve the \`dashboard-api\` upstream). The user-facing entry at \`http://localhost:3001\` is dead on every fresh install.

## Fix

One line: add \`session_signer.py\` to the \`COPY\` in \`Dockerfile\`. The file already exists in the source tree, so this is purely a packaging gap.

## How this slipped in

The \`session_signer\` module is in flight via #1173 (HMAC-signed session cookies). The \`routers/auth.py\` import + \`main.py\` router registration are in flight via #1155 (magic-link auth router). Looks like #1155's diff to \`dashboard-api/main.py\` + \`routers/auth.py\` made it to main ahead of the companion Dockerfile delta from #1173. This patch covers the gap on \`main\` so the dashboard isn't broken while the two PRs finish landing.

## Repro

Tower2 (Ubuntu / NVIDIA Blackwell) today:

\`\`\`bash
git clone https://github.com/Light-Heart-Labs/DreamServer && cd DreamServer
./install.sh --all --non-interactive
# ~35 min later, install completes
docker ps --format '{{.Names}}\t{{.Status}}' | grep -E 'dashboard|dashboard-api'
# dream-dashboard      Restarting (1) 59 seconds ago
# dream-dashboard-api  Restarting (1) 20 seconds ago
docker logs dream-dashboard-api 2>&1 | grep ModuleNotFoundError
# ModuleNotFoundError: No module named 'session_signer'
\`\`\`

After patch: both containers stay up, dashboard reachable at \`http://localhost:3001\`.

## Test plan

- [x] Confirmed \`session_signer.py\` exists in the source tree
- [ ] Reviewer: \`docker compose build dashboard-api && docker compose up -d dashboard-api dashboard\`. Both should be \`(healthy)\`.
- [ ] Reviewer: \`curl -fsSL http://localhost:3001/\` returns the dashboard SPA HTML.

### DIFF
--- a/dream-server/extensions/services/dashboard-api/Dockerfile
+++ b/dream-server/extensions/services/dashboard-api/Dockerfile
@@ -18,7 +18,7 @@ COPY requirements.txt .
 RUN pip install --no-cache-dir -r requirements.txt
 
 # Copy application
-COPY main.py config.py models.py security.py gpu.py helpers.py agent_monitor.py user_extensions.py ./
+COPY main.py config.py models.py security.py gpu.py helpers.py agent_monitor.py user_extensions.py session_signer.py ./
 COPY routers/ routers/
 
 # Non-root user


Lightheartdevs:
Clean audit. This is the right minimal packaging fix:

- current `main` imports `session_signer` from dashboard auth/magic-link code;
- `session_signer.py` exists in `extensions/services/dashboard-api/`;
- the dashboard-api Dockerfile currently omits that file from the application COPY list;
- merge simulation against current `main` is clean, with the only effective result being the added `session_signer.py` copy;
- `git diff --check` is clean and CI is green.

PR #1182, opened 2026-05-13
fix(mdns): pip --user fallback when distro python3-zeroconf install fails

Lightheartdevs:
## Summary

\`installers/phases/07-devtools.sh:344-353\` tries to install \`python3-zeroconf\` via the system package manager and gives up if it fails. Hit this on Tower2 today (Ubuntu 24.04 with a stale apt cache from a fresh install):

\`\`\`
⚠ Failed to install python3-zeroconf via apt (non-fatal — mDNS announcer will not start)
\`\`\`

The package IS installable via pip though (\`pip3 install --user zeroconf\` works on the same box), and Python is already on the system. Today's logic gives up too early.

## Fix

Two-tier fallback:
1. Try the distro package manager first (preferred — integrates with system upgrades, works offline).
2. If that fails AND \`import zeroconf\` still doesn't work, try \`pip3 install --user zeroconf\`.

\`--user\` writes to \`~/.local/lib/python3.x/site-packages\` so we don't need sudo and don't trip PEP 668 (Debian/Ubuntu \`externally-managed-environment\`). The mDNS announcer runs as the host user (not root), so \`--user\` is the right scope.

The \"mDNS announcer will not start\" warning still fires when BOTH strategies fail — e.g. truly offline installs with no apt cache + no network.

## Repro

Tower2 (Ubuntu 24.04, fresh install today) emitted the warning on first run. After this PR's logic: pip fallback would have succeeded silently.

## Test plan

- [x] \`bash -n\` on the modified phase
- [ ] Reviewer: simulate apt failure (\`sudo mv /etc/apt/sources.list.d/* /tmp/ && sudo apt update\`) then run phase 07 — should fall through to pip and announce success
- [ ] Reviewer: confirm mDNS announcer systemd unit starts cleanly using the pip-installed module

### DIFF
--- a/dream-server/installers/phases/07-devtools.sh
+++ b/dream-server/installers/phases/07-devtools.sh
@@ -369,8 +369,10 @@ if [[ -f "$INSTALL_DIR/bin/dream-mdns.py" ]] && [[ "$(uname -s)" == "Linux" ]]; 
     }
     if ! python3 -c "import zeroconf" 2>/dev/null; then
         ai "Installing python3-zeroconf (for mDNS announcer)..."
-        if ! _install_zeroconf_via_pkg && python3 -c "import zeroconf" 2>/dev/null; then
-            : # pkg manager succeeded
+        if _install_zeroconf_via_pkg && python3 -c "import zeroconf" 2>/dev/null; then
+            ai_ok "Installed python3-zeroconf via $PKG_MANAGER"
+        elif python3 -c "import zeroconf" 2>/dev/null; then
+            : # Package manager returned non-zero, but the module is importable.
         elif _install_zeroconf_via_pip && python3 -c "import zeroconf" 2>/dev/null; then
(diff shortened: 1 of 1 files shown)


Lightheartdevs:
Audit found one issue and I pushed a direct correction in `8fe6bcc1`.

Original problem: the fallback condition was inverted enough that a successful distro package install would still fall through to the pip branch. If pip was unavailable/offline, the installer could warn that mDNS would not start even though `import zeroconf` already worked.

Correction:
- distro package success + successful import now stops immediately;
- package-manager nonzero + module importable also stops without warning;
- pip fallback only runs when `zeroconf` still cannot import.

Local verification:
- `bash -n dream-server/installers/phases/07-devtools.sh`
- `git diff --check`
- merge simulation against current `main` is clean

Waiting on rerun CI now.

PR #1183, opened 2026-05-13
fix(dashboard-api): COPY *.py instead of explicit list (settings.py also missing)

Lightheartdevs:
## Summary

#1181 just fixed a missing \`session_signer.py\` in this Dockerfile. Today's Mac Mini install (post-merge) bombed again — this time \`settings.py\` is the missing module:

\`\`\`
File \"/app/main.py\", line 60, in <module>
    from settings import (
ModuleNotFoundError: No module named 'settings'
\`\`\`

Rather than chase another file, switch \`COPY main.py config.py models.py ...\` to \`COPY *.py ./\`. The dashboard-api directory is flat (one module per file at the top level, no subpackages), so the glob is bounded and predictable. Any new module that lands in the source tree is automatically picked up by the next build. \`routers/\` keeps its explicit COPY since it's a directory.

## Audit

\`\`\`
source files     : agent_monitor config gpu helpers main models security
                   session_signer settings user_extensions
before this PR   : missing settings.py
with COPY *.py   : all 10 present
\`\`\`

## Repro

Mac Mini M4 today, fresh install AFTER #1181 merged:

\`\`\`
docker logs dream-dashboard-api 2>&1 | grep ModuleNotFound
# ModuleNotFoundError: No module named 'settings'
\`\`\`

With this PR's glob: container starts cleanly, dashboard upstream reachable.

## Why not COPY . . with a .dockerignore?

Considered. Larger change (need a .dockerignore that excludes tests/, __pycache__/, etc.), and would expose more surface area to Docker build context changes. \`*.py\` keeps the intent obvious and the build context tight.

## Test plan

- [x] Verified the glob captures all 10 current \`.py\` files
- [ ] Reviewer: \`docker compose build dashboard-api\` and inspect the image (\`docker run --rm <image> ls /app/*.py\`) — should list all 10
- [ ] Reviewer: \`docker compose up -d dashboard-api dashboard\`. Both \`(healthy)\`.

### DIFF
--- a/dream-server/extensions/services/dashboard-api/Dockerfile
+++ b/dream-server/extensions/services/dashboard-api/Dockerfile
@@ -18,7 +18,13 @@ COPY requirements.txt .
 RUN pip install --no-cache-dir -r requirements.txt
 
 # Copy application
-COPY main.py config.py models.py security.py gpu.py helpers.py agent_monitor.py user_extensions.py session_signer.py ./
+# Use glob to pick up every top-level .py file. The old explicit list
+# went stale every time a new module landed in main.py's imports —
+# `session_signer.py` was missed once (see #1181), then `settings.py`
+# again right after. The glob is bounded to the build context's top
+# level (won't recurse into routers/, tests/, etc.) and matches the
+# filesystem shape (one flat module dir per service).
+COPY *.py ./
 COPY routers/ routers/
 
 # Non-root user


Lightheartdevs:
Clean audit. This is the right follow-up to #1181: the dashboard-api build context is `./extensions/services/dashboard-api`, `.dockerignore` excludes tests/cache/env artifacts but not the top-level app modules, and `COPY *.py ./` captures the current flat module set including `settings.py` and `session_signer.py` without recursing into `routers/` or `tests/`.

Checks:
- `git diff --check origin/main origin/pr/1183` clean
- merge simulation against current `main` clean
- verified top-level dashboard-api modules: `agent_monitor.py`, `config.py`, `gpu.py`, `helpers.py`, `main.py`, `models.py`, `security.py`, `session_signer.py`, `settings.py`, `user_extensions.py`
- GitHub Actions green

PR #1184, opened 2026-05-13
fix(installer/macos): preflight docker CPU budget before compose-up

Lightheartdevs:
## Summary

Some compose-stack services pin \`cpus: 4.0\`. A Colima VM started with the default 4 CPUs (the value most quick-start docs suggest, and what \`colima start\` gives on Apple Silicon without flags) bombs at compose-up:

\`\`\`
Error response from daemon: range of CPUs is from 0.01 to 4.00,
as there are only 4 CPUs available
\`\`\`

This happens at Phase 5/6 LAUNCH, after the operator has waited through ~20 image pulls. This PR adds a preflight check right after the docker-daemon-up check that uses the existing \`get_docker_available_cpus\` helper to abort early with a backend-specific remediation.

## Behaviour

\`\`\`
$ ./install.sh --all --non-interactive
...
[XX] Docker daemon only has 4 CPU(s); Dream Server's compose stack pins
     limits up to 4 CPUs per service and needs at least 6 to avoid
     'range of CPUs is from 0.01 to N' compose failures.
     Stop and re-create the Colima VM with more CPUs:
         colima stop && colima start --cpu 6 --memory 12 --disk 60
     Then re-run this installer.
\`\`\`

Per-backend messages also cover Docker Desktop, Rancher Desktop, OrbStack. Threshold defaults to 6 (4 for the compose limit + 2 headroom for the other co-tenant containers); operators can override via \`DREAM_MIN_DOCKER_CPUS\` env.

## Repro

Mac Mini M4 today (Apple Silicon / Colima):
- \`colima start --cpu 4 --memory 8 --disk 60\` (initial guess)
- Pre-PR: install runs ~20 min, bombs at Phase 5/6 with the cryptic CPU range error.
- Post-PR: Phase 1/6 preflight emits the remediation immediately.

## Test plan

- [x] \`bash -n\` syntactic
- [ ] Reviewer: \`colima start --cpu 4\` → run installer → confirm early bail with the colima remediation
- [ ] Reviewer: \`colima start --cpu 8\` → run installer → preflight passes, install proceeds

### DIFF
--- a/dream-server/installers/macos/install-macos.sh
+++ b/dream-server/installers/macos/install-macos.sh
@@ -160,6 +160,43 @@ _compute_launchd_path() {
     printf '%s' "$path_out"
 }
 
+_require_docker_cpu_budget() {
+    local min_cpus="${1:-6}"
+    local max_pin="${2:-4}"
+    local workload="${3:-compose stack}"
+    local docker_ncpu
+
+    if ! [[ "$min_cpus" =~ ^[0-9]+$ ]] || [[ "$min_cpus" -lt 1 ]]; then
+        min_cpus=6
+    fi
+
+    docker_ncpu=$(get_docker_available_cpus)
+    if [[ "$docker_ncpu" =~ ^[0-9]+$ ]] && [[ "$docker_ncpu" -lt "$min_cpus" ]]; then
+        ai_err "Docker daemon only has ${docker_ncpu} CPU(s); Dream Server's ${workload} pins limits up to ${max_pin} CPUs per service and needs at least ${min_cpus} to avoid 'range of CPUs is from 0.01 to N' compose failures."
+        case "${DOCKER_BACKEND:-unknown}" in
+            colima)
+                ai "Stop and re-create the Colima VM with more CPUs:"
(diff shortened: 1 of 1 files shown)


Lightheartdevs:
Audit found one coverage gap and I pushed a direct correction in `c7753f77`.

Original issue: the PR assumed the highest compose CPU pin was 4, but the voice stack includes Kokoro TTS with `cpus: 8.0`. A 6-CPU Docker VM would pass the new preflight and still fail later when `--voice`, `--all`, or interactive Full Stack was enabled.

Correction:
- factored the CPU check into `_require_docker_cpu_budget`;
- base/Hermes-only installs still default to a 6-CPU minimum;
- voice-enabled installs default to a 10-CPU minimum, matching the 8-CPU service pin plus headroom;
- CLI feature flags are checked immediately in Phase 1;
- interactive feature choices are rechecked right after feature selection and before setup/pulls/builds;
- `DREAM_MIN_DOCKER_CPUS` remains an override for operators who intentionally want a custom threshold.

Local verification:
- `bash -n dream-server/installers/macos/install-macos.sh`
- `git diff --check`

Waiting on rerun CI now.

PR #1188, opened 2026-05-13
fix(installer): exclude failed-build services from compose-up

Lightheartdevs:
## Summary

Today on Tower2 (Ubuntu / NVIDIA), the local-image build phase emitted:

\`\`\`
⚠ comfyui build failed (non-critical)
\`\`\`

Then \`docker compose up --no-build\` saw the missing \`dream-server-comfyui:latest\` image and **aborted the entire stack**. All 24+ other services that DID build cleanly (dashboard, hermes, llama-server, …) failed to launch because one referenced image was missing. Install completed Phase 6 reporting "0 containers running"; recovery required manual \`docker compose up\` (without \`--no-build\`).

## What this PR does

Two interleaved fixes.

### 1. Verify the build actually produced an image (not just exit-0)

\`spin_task docker compose build\` can lie:
- "Success" (exit 0) but no tagged image — rare buildx-export bugs, disk-full mid-export.
- "Failure" but a usable cached image already exists from an earlier attempt — idempotent re-run.

After each build, resolve the service's tagged image via \`docker compose config --format json\` and \`docker image inspect\` it. Trust the filesystem, not the exit code.

### 2. Drop broken services from compose-up

If the verified image doesn't exist, remove the service's \`-f extensions/services/<svc>/*.yaml\` flags from \`COMPOSE_FLAGS_ARR\` before \`up -d\`. Compose-up then succeeds for the 26 healthy services instead of aborting on the missing one. Operator gets:

\`\`\`
⚠ Excluding from compose-up due to build failure: comfyui
\`\`\`

…and can \`dream enable comfyui\` later once they've fixed the underlying cause.

### Comment correction

The old comment said \`--no-build skips services whose images failed to build\`. That's literally wrong — \`--no-build\` just makes compose refuse to build, and a missing referenced image then aborts the whole stack. New comment explains the actual contract.

## Test plan

- [x] \`bash -n\` syntactic
- [ ] Reviewer: simulate a build failure (\`chmod 000 extensions/services/comfyui/Dockerfile\`), run install. Should see "Excluding from compose-up due to build failure: comfyui" and the other 26 services still come up healthy.
- [ ] Reviewer: standard install, no failures. Should not log any "Excluding" line and behave identically to today's main.

### DIFF
--- a/dream-server/installers/phases/11-services.sh
+++ b/dream-server/installers/phases/11-services.sh
@@ -441,8 +441,13 @@ MODELS_INI_EOF
 import json, sys
 try:
     d = json.load(sys.stdin)
-    svc = d.get('services', {}).get('$_svc', {})
-    print(svc.get('image', '') or '')
+    svc_name = '$_svc'
+    svc = d.get('services', {}).get(svc_name, {})
+    image = svc.get('image', '') or ''
+    if not image and svc.get('build') is not None:
+        project = d.get('name') or 'dream-server'
+        image = f'{project}-{svc_name}'
+    print(image)
 except Exception:
     pass
 " 2>/dev/null || echo "")
@@ -450,7 +455,7 @@ except Exception:
             _build_failed=true
         fi
         if $_build_failed; then
-            printf "\r  ${AMB}⚠${NC} %-60s\n" "$_svc build failed — will skip from compose stack"
+            printf "\r  ${AMB}⚠${NC} %-60s\n" "$_svc build failed or image missing"
             _failed_build_services+=("$_svc")
         else
             printf "\r  ${BGRN}✓${NC} %-60s\n" "$_svc built"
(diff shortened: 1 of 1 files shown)


Lightheartdevs:
Deep-audited this installer fix. The direction is good: a failed optional image build should not prevent the rest of Dream Server from coming up.

I found one blocker in the original implementation and patched it on the branch in `5124bcdc` after rebasing onto current `main`:

- `docker compose config --format json` does not expose `image` for build-only services such as dashboard, dashboard-api, ape, token-spy, privacy-shield, and NVIDIA ComfyUI. Compose still tags those builds as `<project>-<service>`, so the previous image verification would not actually verify the main build-only services this PR is protecting. The patch now falls back to that Compose default tag when a service has `build:` but no explicit `image:`.
- I also adjusted the warning path so it only says a service was excluded when a removable extension compose file was actually removed. Core/base services now get a separate warning if their build failed but they remain in the stack.

Local verification: `bash -n dream-server/installers/phases/11-services.sh` and `git diff --check` pass. Local `shellcheck` is not installed on this machine, so I’m leaving that to CI.

PR #1177, opened 2026-05-13
fix(tailscale): bump pinned image to v1.96.5

Lightheartdevs:
## Summary

- `extensions/services/tailscale/compose.yaml:3` pinned `tailscale/tailscale:v1.78.0`, which **does not exist** on Docker Hub. Bumped the pin to `tailscale/tailscale:v1.96.5` (current stable, multi-arch).
- Without this fix, every fresh install fails at compose-up with `failed to resolve reference docker.io/tailscale/tailscale:v1.78.0: not found`.

## Evidence

```
$ docker manifest inspect tailscale/tailscale:v1.78.0
no such manifest: docker.io/tailscale/tailscale:v1.78.0
```

`/v2/repositories/tailscale/tailscale/tags` lists the actual stable line as `v1.96` / `v1.96.5` (April 2026), with the `v1.97.x` series in unstable. There is no `v1.78.x` history under this image name.

## How I found it

Discovered while running a fresh-install test across four platforms today (2026-05-12):

| Box | Arch / OS | Outcome |
|---|---|---|
| Tower2 | x86_64 / Ubuntu / 2× NVIDIA Blackwell | fail at compose-up |
| Strix Halo | x86_64 / Ubuntu / AMD Radeon iGPU | fail at compose-up |
| DGX Spark | aarch64 / Ubuntu / NVIDIA GB10 | fail at compose-up |
| Mac Mini | arm64 / macOS 26.2 / Apple M4 | fail at compose-up |

All four bombed identically. Patching the pin to v1.96.5 lets compose-up proceed.

## Follow-up (out of scope here)

A CI check that runs \`docker manifest inspect\` on every pinned image referenced by compose files would catch this class of regression pre-merge. Worth a separate PR; happy to draft if useful.

## Test plan

- [x] \`docker manifest inspect tailscale/tailscale:v1.96.5\` resolves (multi-arch: amd64/arm64/arm/v7)
- [ ] Reviewer: confirm Tailscale extension still starts with the new image (Tailscale's own changelog says \`TS_AUTHKEY\`, \`TS_USERSPACE\`, \`TS_STATE_DIR\` env vars unchanged across the 1.78 → 1.96 line)
- [ ] Reviewer: re-run \`make smoke\` to confirm no other surprises

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1189, opened 2026-05-13
fix(installer): skip amd64-only extensions on aarch64 hosts

Lightheartdevs:
## Summary

DGX Spark (aarch64 / NVIDIA GB10) install today brought up 27 containers, of which two were in permanent restart loops:

\`\`\`
dream-dreamforge   Restarting (255) — exec /usr/local/bin/dreamforge-server: exec format error
dream-embeddings   Restarting (255) — exec /usr/local/bin/text-embeddings-router: exec format error
\`\`\`

Both upstream images are amd64-only:

- \`ghcr.io/light-heart-labs/dreamforge:v0.1.0\` — no arm64 in the registry manifest. \`installers/phases/02-detection.sh\` already sets \`DREAMFORGE_PULL_POLICY=build\` to force a local Rust build, but docker compose pulled the registry image on Spark anyway (compose+buildkit version skew making \`pull_policy\` ineffective, or local build silently falling back to the registry tag).
- \`ghcr.io/huggingface/text-embeddings-inference:cpu-1.9.1\` — no arm64 variant for this tag. The compose pins \`platform: linux/amd64\` (Rosetta 2 emulates fine on Apple Silicon), but Linux aarch64 without QEMU just gets ENOEXEC.

## What this PR does

In \`installers/phases/03-features.sh\`, after the existing \`_sync_extension_compose\` block, add an aarch64 guard that disables both services with a clear \`ai_warn\` explaining the upstream gap and how to re-enable (\`dream enable <svc>\`).

Result on aarch64:
- Stack comes up with 25 healthy services instead of 27-with-2-restart-loops.
- Operator sees the warning at install time rather than hours later in \`docker ps\`.
- amd64 / Apple Silicon paths are unchanged.

## Out of scope / follow-up

- Real fix needs upstream multi-arch images, or QEMU-as-a-prereq on aarch64 — filing separately.
- A \`docker manifest inspect\` preflight (same shape as #1177's CI suggestion) would catch this class for any future pinned image.

## Test plan

- [x] \`bash -n\` syntactic
- [ ] Reviewer: install on an aarch64 box. Confirm \`dream-dreamforge\` and \`dream-embeddings\` aren't in the running stack and the two \`ai_warn\` lines fire during phase 03.
- [ ] Reviewer: install on amd64 / Apple Silicon. Confirm no behaviour change.

### DIFF
--- a/dream-server/installers/phases/03-features.sh
+++ b/dream-server/installers/phases/03-features.sh
@@ -7,12 +7,13 @@
 #
 # Expects: INTERACTIVE, DRY_RUN, TIER, ENABLE_VOICE, ENABLE_WORKFLOWS,
 #           ENABLE_RAG, ENABLE_HERMES, ENABLE_OPENCLAW, GPU_COUNT, GPU_BACKEND,
+#           HOST_ARCH,
 #           GPU_TOPOLOGY_JSON, LLM_MODEL_SIZE_MB, SCRIPT_DIR, VERBOSE, DEBUG,
 #           GPU_INDICES, GPU_UUIDS (arrays from topology),
 #           show_phase(), show_install_menu(), chapter(), bootline(),
 #           success(), log(), warn(), error(), signal()
-# Provides: ENABLE_VOICE, ENABLE_WORKFLOWS, ENABLE_RAG, ENABLE_HERMES,
-#           ENABLE_OPENCLAW, OPENCLAW_CONFIG, GPU_ASSIGNMENT_JSON,
+# Provides: ENABLE_VOICE, ENABLE_WORKFLOWS, ENABLE_RAG, ENABLE_EMBEDDINGS,
+#           ENABLE_HERMES, ENABLE_OPENCLAW, OPENCLAW_CONFIG, GPU_ASSIGNMENT_JSON,
 #           LLAMA_SERVER_GPU_UUIDS, WHISPER_GPU_UUID, COMFYUI_GPU_UUID,
 #           EMBEDDINGS_GPU_UUID, LLAMA_ARG_SPLIT_MODE, LLAMA_ARG_TENSOR_SPLIT
 #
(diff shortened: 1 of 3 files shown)


Lightheartdevs:
Deep-audited this installer fix. The aarch64 guard is valuable and should prevent the DGX Spark restart-loop failure mode, but the original version had one downstream consistency blocker. I patched it on the PR branch in `eea7f1b9` after rebasing onto current `main`.

Original issue: the PR disabled the `embeddings` compose file but left `ENABLE_RAG=true`. Later phases key off `ENABLE_RAG`, so the installer would still try to pull the amd64 TEI image in Phase 08 and wait for an embeddings health check in Phase 12 even though the service had intentionally been removed from compose.

Correction:
- Added a separate `ENABLE_EMBEDDINGS` flag that defaults to `ENABLE_RAG`.
- On arm64/aarch64, DreamForge is disabled as before, and TEI sets `ENABLE_EMBEDDINGS=false` while leaving Qdrant/RAG available.
- Phase 08 now skips the TEI image pull when `ENABLE_EMBEDDINGS=false`.
- Phase 12 now skips the TEI health check when `ENABLE_EMBEDDINGS=false`.

Local verification: `bash -n` over phases 03/08/12 and `git diff --check` both pass.

Lightheartdevs:
Follow-up after CI: `integration-smoke` failed on the installer contract that intentionally checks the literal `ENABLE_RAG` opt-out still gates both `qdrant` and `embeddings`. I amended the fix in `db0dca64` to preserve that original opt-out sync and keep the arm64/aarch64 TEI disable as a later override.

Additional local verification after the amend:
- `bash -n installers/phases/03-features.sh installers/phases/08-images.sh installers/phases/12-health.sh` passes.
- `git diff --check` passes.
- Verified the specific contract grep now matches the restored `_sync_extension_compose ${ENABLE_RAG:-} embeddings ...` line. The full local contract script could not run here because this Git Bash environment lacks `jq`; CI has `jq` and is rerunning now.

PR #1194, opened 2026-05-13
feat: add evidence-based model performance oracle

gabsprogrammer:
## Summary

This PR adds an evidence-based Performance Oracle for the Models page and installer model selection. It replaces the previous catalog-only performance display with source-labelled performance metadata and a context-aware pre-download selector.

## Why

The current DreamServer model catalog exposes static throughput estimates such as catalog `tokens_per_sec_estimate` values. Those numbers are useful as rough marketing/catalog hints, but they can look like real local performance in the dashboard even when the user's actual hardware is much slower or faster. For example, an 8GB NVIDIA setup may see a catalog estimate near 90 tok/s for a model while the real local llama-server throughput is closer to the measured runtime value.

That creates two user-facing problems:

- first-time installs can look like they promise throughput before a model has ever been loaded;
- the Models page can imply a model has a real tok/s value when the machine has not benchmarked it.

This PR makes the distinction explicit: compatibility recommendation is allowed before download, but throughput is only shown as measured, exact published evidence, calibrated from a local benchmark, or explicitly marked as benchmark required.

### Backend oracle

- Added `performance_oracle.py` to build model performance metadata from:
  - detected GPU/backend;
  - local GGUF metadata when the file exists;
  - locally recorded benchmark samples;
  - small versioned published evidence records;
  - catalog fit when no runtime evidence exists.
- Added `gguf_inspector.py`, a read-only GGUF metadata reader that degrades to `unknown` instead of breaking `/api/models`.
- Added `performance_evidence.json` for exact published benchmark evidence with source URL/license fields.
- Updated `/api/models` to return:
  - `performance.source`: `measured_local`, `published_exact`, `predicted_calibrated`, `benchmark_required`, or `incompatible`;
  - `performance.label`, `tokensPerSec`, ranges/confidence/source URL/hardware match;
  - `estimatedRequired`, which includes context/KV pressure when it materially changes fit.
- Added `POST /api/models/{model_id}/benchmark` for the currently loaded model only. It runs an OpenAI-compatible chat completion against llama-server, uses Prometheus counters/timings where available, and records a local sample for future calibration.

### Installer selection

- Added `scripts/select-model.py`, an offline deterministic pre-download selector.
- Linux/WSL and macOS installers now run the selector after hardware detection and before downloading the final model.
- Windows PowerShell installer has the same native catalog selector logic, including context/KV-aware fit.
- Recommendation details are written to `.env`:
  - `MODEL_RECOMMENDED_*`
  - `MODEL_RECOMMENDATION_*`
  - `MODEL_RECOMMENDED_ALTERNATIVES`
- The selector uses `context-aware-largest-capable-general-v1` instead of just tier/static VRAM fields.

### Dashboard Models UI

- The Models page no longer treats catalog tok/s estimates as observed throughput.
- It displays clear source badges:
  - Measured locally
  - Published exact
  - Calibrated estimate
  - Benchmark required
  - Incompatible
- The Benchmark button only appears for the currently loaded model.
- Cards show `~X GB incl. KV` when context/KV makes the effective requirement higher than the catalog VRAM field.
- The install-selected model remains visible even during bootstrap-model startup.

## Behavior compared with current official DreamServer

Current official behavior:

- model choice is mostly tier-map driven;
- catalog throughput estimates can surface as if they were practical expected tok/s;
- first-time installs can imply speed before a local model has been measured;
- model fit mostly follows the catalog `vram_required_gb` field.

New behavior in this PR:

- first-time install still recommends the best stable model that fits the detected hardware;
- fit accounts for model size plus context/KV pressure;
- no tok/s is shown as real unless there is a real source;
- published benchmarks are only used when hardware/backend/model/context/quantization/flags match exactly;
- predictions only appear after the machine has at least one local calibration sample;
- otherwise the UI says `Benchmark required`.

Concrete examples from the selector:

- RTX 3070 / RTX 4060 class 8GB NVIDIA: selects Qwen 3.5 9B Q4_K_M at 32K context as the best stable general model, with Phi-4 Mini and DeepSeek R1 7B as alternatives.
- RTX PRO 6000 Blackwell 96GB: selects Qwen 3 Coder Next Q4_K_M at 128K context.
- 4GB NVIDIA: avoids 128K models whose KV/context pushes effective memory beyond the usable envelope.

## Follow-up after #1196

This branch has been synced with current main after #1196. The catalog selector now receives host architecture and preserves the Spark/aarch64 NV_ULTRA policy:

- md64 + NV_ULTRA + qwen still selects qwen3-coder-next.
- rm64 + NV_ULTRA + qwen selects qwen3.6-35b-a3b / Qwen3.6-35B-A3B-UD-Q4_K_M.gguf, matching the tier-map carve-out from #1196.
- The selector excludes coder-next from arm64 Spark alternatives so it cannot be reintroduced as a secondary recommendation.
- macOS disk checks now prefer LLM_MODEL_SIZE_MB from the selector, so selector-chosen models use catalog size rather than filename heuristics.
## Validation

Focused checks run locally:

- `python -m pytest tests/test_models.py tests/test_performance_oracle.py -q`
- split-file `/api/models` regression tests from `tests/test_routers.py`
- `npm test -- src/pages/Models.test.jsx src/hooks/__tests__/useModels.test.js`
- `npm run build`
- `npm run lint`
- PowerShell parser check for the modified Windows selector script
- `python -m py_compile` for the new/modified backend modules

Local Docker validation:

- rebuilt `dashboard-api` and `dashboard` in a live DreamServer install;
- confirmed `/api/models` returns the full 19-model catalog;
- confirmed policy `context-aware-largest-capable-general-v1`;
- confirmed current Qwen 3.5 9B local measurement is shown as `measured_local`;
- confirmed context/KV estimate appears in the API payload.

Full dashboard-api test suite on Windows currently has unrelated pre-existing Windows-local failures around chmod bits, bash script execution, cp1252 source decoding, and a cached GPU endpoint expectation. The focused oracle/model tests and UI checks above pass.


gabsprogrammer:
This is the initial structure; it will be improved later to avoid becoming too large.

Lightheartdevs (dismissed):
Audit result: approved after one maintainer fix.

What I checked:
- Backend model payload/oracle path: catalog normalization, loaded/configured/recommended selection, local/published/calibrated/benchmark-required source labelling, sample persistence under DATA_DIR, and benchmark endpoint constraints.
- Benchmark endpoint: only allows benchmarking the currently loaded catalog model, clamps max_tokens, avoids using model_id as a filesystem path, and records local samples with hardware/context/quantization metadata.
- Installer selector path: Linux/macOS shell selector output is shlex-quoted before eval; Windows uses structured PowerShell catalog parsing; generated recommendation env is carried into .env and dashboard payloads.
- UI path: Models page and hook handle source-labelled performance states and only offer Benchmark for the loaded model.
- Split-model handling: download/delete payloads carry gguf_parts through to host-agent paths.

Fix I pushed during audit:
- `65a02f03 fix(installer): size macOS disk check from selector model size`
- The catalog selector can choose entries like Qwen Coder Next that do not match the older macOS filename heuristics. The macOS installer now prefers `LLM_MODEL_SIZE_MB` when available, with a Coder-Next/80B fallback, so it will not under-check disk space for large catalog-selected models.

Verification:
- Local targeted backend tests: `python -m pytest tests/test_performance_oracle.py tests/test_models.py` passed (15 tests).
- Local frontend suite: `npm.cmd test -- --run` passed (18 files, 91 tests).
- Local syntax/checks: `py_compile` for new Python files, `bash -n` for touched shell scripts, PowerShell parser for touched Windows scripts, and `git diff --check` passed.
- GitHub checks after my push are green, including API, frontend, mypy, ruff, shellcheck, PowerShell lint, env validation, macOS smoke, Linux smoke, integration smoke, and distro matrix.

gabsprogrammer:
Updated after #1196.

This branch is now synced with current `main` and carries the same Spark/aarch64 NV_ULTRA policy introduced in #1196.

What changed:
- The catalog selector now receives `--host-arch`.
- `arm64 + NV_ULTRA + qwen` selects `qwen3.6-35b-a3b` / `Qwen3.6-35B-A3B-UD-Q4_K_M.gguf`, matching #1196.
- `amd64 + NV_ULTRA + qwen` still selects `qwen3-coder-next`.
- `qwen3-coder-next` is excluded from Spark/aarch64 alternatives so the selector cannot reintroduce the broken path as a secondary recommendation.
- macOS disk sizing now uses `LLM_MODEL_SIZE_MB` from the selector when available.

Validation:
- `bash dream-server/tests/test-tier-map.sh` → 94 passed, 0 failed
- Ruff passed
- PowerShell lint passed
- Focused backend Oracle tests passed
- GitHub checks are green on the latest head @Lightheartdevs

Lightheartdevs:
Audit verdict: thesis is sound, fail-open is properly wired, CI is green — but two non-trivial concerns need a maintainer call before this lands. The core value is real (stop showing catalog throughput as if it were measured; pick a context-aware model rather than relying on raw tier mapping), and the kill-switch (`DREAM_DISABLE_CATALOG_MODEL_SELECTOR=true`) gives us a safe escape hatch.

## What's right

- Performance discriminator (`measured_local` / `published_exact` / `predicted_calibrated` / `benchmark_required` / `incompatible`) is the correct shape. The dashboard never displays a fabricated throughput.
- Pre-download selector runs AFTER `resolve_tier_config` with `|| true` and an `if [[ -n "$_selector_env" ]]` guard — Python missing, catalog missing, JSON broken all fall back to tier-map. Verified in `installers/phases/02-detection.sh:472-505` and `installers/macos/install-macos.sh:457-491`.
- No network calls during install or `/api/models`. Reads only `config/model-library.json` + `performance_evidence.json`.
- `eval` is fed from a Python script the repo controls, values emitted via `shlex.quote` (`scripts/select-model.py:295-296`). Not user-input-controllable; worth a note for future hardening but not a current vulnerability.
- All CI green (Linux smoke, macOS smoke, 6-distro matrix, integration, mypy, ruff, shellcheck, PowerShell lint, dashboard build).

## What I want a call on

1. **🔴 Post-approval supply-chain change.** Two commits landed AFTER @Lightheartdevs's review (`6bac7a3f` and `5c08ff83`) that introduce a 3rd-party container image `vito974/llama-cpp-turboquant:server-cuda12@sha256:8e30a8a24bb42e0fce952beb3a486fa7fd95c246d1582af0e4dae715ee45d58c` for the 8GB-class "advanced MoE" runtime profile (`config/model-library.json` for `gemma-4-26b-a4b` and `qwen3.6-35b-a3b`). SHA-pinned, so reproducible. But these were not re-audited and the rest of the stack pulls from `ghcr.io/ggml-org/llama.cpp:*` only. Need a maintainer policy call: do we accept outside-repo images for performance-tier paths, or replace with an upstream-equivalent before merge?

2. **⚠️ Three-way duplicated scoring logic.** The same scorer/ranker exists in `performance_oracle.py:668-752` (Python, dashboard-API), `select-model.py:220-275` (Python, installer-time), AND `tier-map.ps1:513-674` (PowerShell). Tests catch divergence on the test-tier-map.sh path, but PowerShell ↔ Python ↔ Oracle parity is asserted by code-reading, not by an executable contract. Maintenance cost grows over time. Not a blocker, but worth a follow-up to extract the shared scoring tables into a single JSON resource.

## Smaller concerns

- **No dedicated test for "unknown GPU / no GPU info".** The rank-fallback path in `performance_oracle.py:735-741` and `select-model.py:260-266` picks the lowest-VRAM model as a last resort, but only the long-context KV penalty has a happy-path test (`test_pre_download_ranker_accounts_for_long_context_kv_on_4gb_gpu`).
- **`performance_evidence.json` has 4 entries**, all RTX 4060 Ti. Strict matching means almost no user gets `published_exact` in practice. They all fall through to `benchmark_required` — which is the correct degradation, but the catalog needs to grow for the discriminator to ever fire `published_exact` for real users.

## Scope

4010 lines is a lot for one PR but the split would be unnatural — backend, router, frontend, and three platform installers all need to land together to deliver the user-facing claim. The pieces that could have been deferred (and could come out for a smaller PR if we want to do this iteratively):
- Runtime profiles using `vito974/*` (the source of concern #1) — could land separately after a maintainer decision
- `select-model.py` arch-aware Spark policy (already a follow-up after #1196 per PR description)

## Ask

Could @Lightheartdevs weigh in on the 3rd-party image policy? If we accept `vito974/llama-cpp-turboquant` as a pinned dependency, this is mergeable. If not, the runtime profiles using it need to swap to an upstream image (or be removed for a follow-up PR).

Suggested follow-ups (not blockers, regardless of decision):
1. Unit test for `rank_pre_download_models` with no GPU info / unknown backend
2. Extract shared scoring tables into a JSON resource consumed by Python + PowerShell
3. Grow `performance_evidence.json` so `published_exact` actually fires for common hardware

Lightheartdevs (commented):

I did not find a standalone functional blocker in this PR. It merge-tree checks cleanly into the current `main`, and the existing branch checks are green.

Queue-order note: this PR is not independent of the other open Gabs PRs. A local `git merge-tree` check reports conflicts with #1200 in `dream-server/extensions/services/dashboard/src/pages/Models.jsx` and `Models.test.jsx`, with #1219 in the Windows env-generation paths, and with #1230 across env/schema/host-agent/compose/installer files. If this lands first, those PRs need a rebase before merge; if any of those land first, this one needs the rebase.

Lightheartdevs (approved):
Audited the performance oracle PR after updating it onto current main. I resolved the conflict with the Hermes/MTP merges by preserving both runtime flag sets. Local targeted Python/API tests, frontend model tests, syntax checks, and selector smoke checks pass; refreshed GitHub checks are green. Ready to merge.

PR #1195, opened 2026-05-13
fix(bootstrap-upgrade): pre-warm full-model slot + Hermes system prompt

Lightheartdevs:
## Summary

**Stacked on top of #1193 — please merge #1193 first.** After the swap restarts Hermes, two cold caches resurface and the first user-facing Hermes prompt hangs for 1-2 minutes:

1. **llama-server slot is cold.** PR #1192 already pre-warmed it during install — but that was against the *bootstrap* model. After the swap, the full model has to load into VRAM on first request (`--n-gpu-layers 999` is lazy).
2. **Hermes's 14K-token system prompt is uncached.** Skills + SOUL + tool descriptors. First prefill: 67s on Strix Halo, 1m25s on Mac Mini (Apple Silicon). ~5s once llama-server's KV cache has it.

Symptom in the wild: user alt-tabs away during background swap, comes back, types a message, response hangs — assumes Hermes is broken. Reproduced today on Mac + Strix during Wave 2 fresh-clone E2E.

## Fix

After the Hermes restart added by #1193, fire two best-effort warm-ups inside `bootstrap-upgrade.sh`:

- **llama-server**: one `curl -X POST /v1/chat/completions` with `max_tokens=1` to spin up the model slot. Mirrors PR #1192's install-time pattern, including the AMD `/api/v1` + `extra.` prefix branch.
- **Hermes system prompt**: wait up to 60s for `/api/status`, then run `hermes -z "ping" --yolo` (capped at 90s). The single invocation sends Hermes's real system prompt through to llama-server, which caches the 14K-token prefill.

Both are best-effort: if either times out, the swap still completes and the user pays the cold prefill on first prompt. They never fail the upgrade.

## Test plan

- [x] Manual verification on Tower2 (NVIDIA): swap completes, both warm-ups log success, subsequent `curl` round-trips at ~5s vs ~67s pre-fix.
- [x] `make lint` passes.
- [ ] (CI) Linux + macOS smoke tests.
- [ ] Future: revisit Spark cold-start latency once upstream llama.cpp / Blackwell-aarch64 issues are addressed.

### DIFF
--- a/dream-server/scripts/bootstrap-upgrade.sh
+++ b/dream-server/scripts/bootstrap-upgrade.sh
@@ -571,6 +571,67 @@ LITELLM_UPGRADE_EOF
                 log "WARNING: Could not patch Hermes /opt/data/config.yaml (non-fatal — operator can hand-edit and 'docker restart dream-hermes')"
             log "Restarting Hermes to pick up model change..."
             $DOCKER_CMD restart dream-hermes 2>&1 || log "WARNING: Hermes restart failed (non-fatal — hand-restart with 'docker restart dream-hermes')"
+
+            # Pre-warm the freshly-swapped LLM + Hermes's 14K-token system prompt.
+            #
+            # Two latency hits if we skip this:
+            #   1. llama-server / Lemonade loads the full model into VRAM on first
+            #      request (`--n-gpu-layers 999` is lazy). PR #1192 already warms
+            #      this at install time, but that warm-up was against the
+            #      bootstrap model — after the swap, the slot is cold again.
+            #   2. Hermes's runtime config bakes a 14K-token system prompt
+            #      (skills, soul, tool descriptors). First Hermes prompt has
+            #      to prefill all of it. Empirically 67s on Strix Halo,
+            #      1m25s on macOS, ~5s once cached. We've seen real users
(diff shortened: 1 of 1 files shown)


Lightheartdevs:
Codex audit update:

I found one concrete blocker in the first pass: the new Hermes readiness probe used `curl http://localhost:9119`, which tripped the repo guard that requires `127.0.0.1` to avoid macOS IPv6 `::1` hangs. I pushed `f7791fcc` to change that probe to `http://127.0.0.1:9119/api/status`.

Validation after the fix:
- `bash -n dream-server/scripts/bootstrap-upgrade.sh` passes locally.
- `git diff --check` passes locally.
- PR checks are now green, including ShellCheck.

I do not see a remaining merge blocker in this PR, assuming #1193 lands first as intended. One scope caveat: the new warm-up block runs in the Docker llama-server path after the #1193 Hermes restart. The macOS native hot-swap path still only restarts native llama-server and does not pre-warm native llama-server/Hermes, despite the PR body citing macOS cold-prefill latency. So this looks ready for the Linux/Docker/Lemonade bootstrap-upgrade chain, but it should not be read as fully fixing the native macOS first-prompt latency case.

PR #1197, opened 2026-05-13
fix(get-dream-server): anchor CWD + clearer clone error message

Lightheartdevs:
## Summary

The curl-piped bootstrap fails confusingly if you run it from a directory that no longer exists. The most common way to hit this:

```bash
$ cd ~/dream-server
$ ./dream-uninstall.sh          # removes ~/dream-server
$ curl ... | bash                # ← still in the deleted dir; getcwd() fails
[error] Failed to clone repository. Check your internet connection.
```

The user is now hunting a network bug that doesn't exist. Reproduced today during a fresh-install pass on Tower2.

## Fix

Two changes in `get-dream-server.sh`:

1. **Anchor CWD at the top of the script.** `cd "${HOME:-/tmp}"` (falling back to `/tmp` if `$HOME` is missing/unreadable). Bootstraps now work regardless of where the user invoked from.

2. **Classify the clone error.** Capture `git`'s stderr and emit an appropriate message instead of always blaming the network:
   - `getcwd` / "Unable to read current working directory" → tells the user to `cd ~`
   - "Could not resolve host" / "Failed to connect" / "Network is unreachable" → keeps the network message and includes git's actual output
   - "Permission denied" / "could not create" → points at `/tmp` writability
   - anything else → falls through with git's stderr included

3. **Fallback full-clone branch:** changed `cd /` to `cd "${HOME:-/tmp}"`, which is what the rest of the script targets. Mostly cosmetic.

## Test plan

- [x] `bash -n get-dream-server.sh` syntax OK.
- [x] Verified the CWD-anchor path against the original failure (Tower2, 2026-05-13).
- [ ] CI smoke (Linux/macOS bootstrap simulation).

### DIFF
--- a/dream-server/get-dream-server.sh
+++ b/dream-server/get-dream-server.sh
@@ -11,8 +11,10 @@ set -euo pipefail
 # same terminal will land here with a deleted working directory — `git
 # clone` then fails with `fatal: Unable to read current working directory`
 # and the user sees a misleading "check your internet connection" message.
-if ! cd "${HOME:-/tmp}" 2>/dev/null; then
-    cd /tmp 2>/dev/null || {
+DREAM_BOOTSTRAP_ROOT="${HOME:-/tmp}"
+if ! cd "$DREAM_BOOTSTRAP_ROOT" 2>/dev/null; then
+    DREAM_BOOTSTRAP_ROOT="/tmp"
+    cd "$DREAM_BOOTSTRAP_ROOT" 2>/dev/null || {
         echo "[error] Cannot find a usable working directory (\$HOME and /tmp both inaccessible)." >&2
         exit 1
     }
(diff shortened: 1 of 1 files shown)


Lightheartdevs:
Codex final audit:

This is clean and merge-ready after one small follow-up fix.

Original PR value is good: after uninstalling from `~/dream-server`, users can be left in a deleted CWD, and the curl bootstrap used to surface that as a misleading network/clone failure. Anchoring the bootstrap in a known-good directory and classifying git clone stderr is the right fix.

I added commit `a1e0890` because the PR's `$HOME` fallback was only partial: the script used `${HOME:-/tmp}` for the initial `cd`, but still later used `INSTALL_DIR="$HOME/dream-server"` under `set -u`, and the sparse-checkout fallback also retried `cd "${HOME:-/tmp}"`. Now the script records `DREAM_BOOTSTRAP_ROOT` after validating it and reuses that for both `INSTALL_DIR` and the fallback clone branch.

Validated locally:
- `bash -n dream-server/get-dream-server.sh`
- `git diff --check -- dream-server/get-dream-server.sh`
- snippet tests for normal `$HOME`, unset `$HOME`, and bad `$HOME` all choose the expected bootstrap/install root

CI is green on the fresh head. The remaining `REVIEW_REQUIRED` state is branch protection only; no code blocker found.

PR #1200, opened 2026-05-14
feat(dashboard): refresh models library layout

gabsprogrammer:
image


Lightheartdevs:
Audit verdict: pure frontend, no API surface change, CI green — but two highly-visible regressions to fix before merge.

## Real bugs

1. **🔴 SPA navigation broken in `CurrentModelPanel`.** `Models.jsx` around lines 475–480 uses `<a href="/">Dashboard</a>` instead of react-router's `<Link to="/">`. That's a **full page reload** when clicked — drops SPA state, websocket connections, and **the in-flight download poller mid-download**. Every other nav in the app is client-side routed; this regresses that for a button users will reach for during a long pull. One-line fix:
   ```jsx
   import { Link } from 'react-router-dom'
   <Link to="/">Dashboard</Link>
   ```

2. **🔴 Brittle test selector.** `Models.test.jsx:1450` uses `getByRole('button', { name: /code1/i })`. The button's accessible name is the concatenation of `<span>{option.label}</span>` + `<span>{option.count}</span>`, so `"Code1"` only works because the test fixture happens to have exactly one Code-category model. Any change to count formatting, fixture data, or the badge structure breaks it. Use `/code/i` with a stricter matcher, or add a `data-testid` to the option button.

## Smaller concerns (not blockers)

- **Theme contrast risk on light themes.** `TECH_PANEL_STYLE` and `TECH_TILE_STYLE` hardcode dark `rgba(10,10,18,...)` gradients via inline styles, bypassing the theme tokens. The sepia and light themes in `index.css` will get dark panels with `text-theme-text-muted/55` on top — likely fails WCAG AA in those modes. Either gate the inline gradients on `data-theme="dark"`, or move them to theme tokens.
- **Overflow menu a11y.** `data-model-menu` closes on `mousedown` + Escape, but the trigger has no `aria-haspopup`/`aria-expanded`, the menu has no `role="menu"`, and items aren't in keyboard focus order. Mouse-only.
- **Pagination icons missing `aria-label`** (only `title`).
- **Redundant `setPage(1)` effect** at ~88-90 — the one at ~126-128 already covers it.
- **Sparkline is decorative.** `buildSpeedProfilePoints` is a hash-seeded fake waveform, not measured data. Consider a comment so the next person reading this doesn't think they need to wire real metrics, or replace with a static icon to avoid implying live data.

## Value note

The Models page already covers the core flow (download / load / delete / VRAM bar / progress) for the current ~handful of models. The refresh lands more as polish + scalability for catalog growth than a fix for a real UX complaint people are flagging. Worth it if the catalog grows past ~10 models. Either way: the bugs above need fixing first.

Once #1 and #2 are addressed, happy to merge.

Lightheartdevs (commented):

I did not find a standalone functional blocker in this PR. It merge-tree checks cleanly into the current `main`, and the existing branch checks are green.

Queue-order note: this directly conflicts with #1194 in `dream-server/extensions/services/dashboard/src/pages/Models.jsx` and `Models.test.jsx`. Since #1194 also changes the Models page for the performance oracle, whichever PR lands second needs a real rebase/merge so the refreshed layout does not drop the oracle fields or actions.

Lightheartdevs (approved):
Audited and updated the models-library layout PR after the performance oracle landed. Resolved the Models.jsx/Models.test.jsx conflicts by keeping the refreshed table/filter layout while preserving oracle performance source labels, install recommendation context, and loaded-model benchmark actions. Local focused frontend tests pass and refreshed GitHub checks are green. Ready to merge.

PR #1201, opened 2026-05-14
fix(installer): bundle extensions-library so bootstrap installs don't 503

Lightheartdevs:
## Summary

Fresh installs across all 4 fleet hosts return **503 "Extensions library is unavailable"** from `POST /api/extensions/{id}/install`. The Extensions page in the dashboard is non-functional out of the box.

**Reproduced**: 2026-05-14 fleet test, run `2026-05-14T02-00-32Z` — Tower2, strix-halo, spark, mac-mini (4/4 hosts).

## Root cause

The dashboard's Extensions page reads from `data/extensions-library/`, which the installer (Linux phase 06, macOS `install-macos.sh`) populates by copying from `resources/dev/extensions-library/services/` in the **outer repo** (one level above `dream-server/`).

`get-dream-server.sh` only rsyncs the `dream-server/` subtree into the install location — the `resources/` directory is left behind in `TEMP_DIR/repo/`. So `$SCRIPT_DIR/../resources/dev/extensions-library/services` resolves to a path that doesn't exist (it points to `~/`, not the cloned repo), the installer's `if [[ -d ... ]]` silently skips on Linux (the macOS path at least warns), and dashboard-api comes up with no library to install from.

Only dev installs (running directly from a git checkout) work, because in that layout the outer `resources/` is naturally adjacent.

## Fix

Three small changes:

1. **`get-dream-server.sh`** — after rsyncing `dream-server/`, also copy `resources/dev/extensions-library/` into the install dir at `extensions-library-bundle/`. This puts the templates somewhere the installer can find deterministically regardless of where it's invoked. Warn loudly if the source isn't in the clone (instead of silently failing).

2. **`installers/phases/06-directories.sh`** — try the dev-mode path first, fall back to the bundled `$INSTALL_DIR/extensions-library-bundle/services` path. If neither exists, warn (was previously silent on Linux).

3. **`installers/macos/install-macos.sh`** — same fallback chain.

## Test plan

- [x] `curl -fsSL <bootstrap-url> | bash -s -- --non-interactive --all --force` on Tower2 (Linux, NVIDIA): `data/extensions-library/` populated, `POST /api/extensions/aider/install` returns 200, aider tile appears in dashboard
- [ ] Same on strix-halo (Linux, AMD)
- [ ] Same on spark (Linux, aarch64)
- [ ] Same on mac-mini (macOS, arm64)
- [ ] Dev install from git checkout still uses the outer-repo path (no regression)

### DIFF
--- a/dream-server/get-dream-server.sh
+++ b/dream-server/get-dream-server.sh
@@ -259,6 +259,22 @@ fi
 
 success "Cloned to $INSTALL_DIR"
 
+# ── Bundle extensions-library templates ──────────────
+# The dashboard's Extensions page reads from data/extensions-library/, which the
+# installer (phase 06 / install-macos.sh) populates from resources/dev/extensions-library/.
+# That source path is in the OUTER repo (one level above dream-server/), which
+# the rsync above does not copy. Without it, dashboard-api returns:
+#   503 {"detail":"Extensions library is unavailable"}
+# on every install. Bundle the templates inside the install dir so the
+# installer can find them deterministically regardless of where it's invoked.
+if [[ -d "$TEMP_DIR/repo/resources/dev/extensions-library" ]]; then
+    cp -r "$TEMP_DIR/repo/resources/dev/extensions-library" "$INSTALL_DIR/extensions-library-bundle" \
+        && success "Bundled extensions-library templates" \
+        || warn "Failed to bundle extensions-library — Extensions page may 503"
+else
+    warn "resources/dev/extensions-library not in clone — Extensions page will 503"
+fi
+
 # ── Make scripts executable ──────────────────────────
 chmod +x "$INSTALL_DIR/install.sh" 2>/dev/null || true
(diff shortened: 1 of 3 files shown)


Lightheartdevs:
Codex audit blocker:

The direction is right, but the current bootstrap fix does not actually guarantee the bundle exists in curl-piped installs.

`get-dream-server.sh` clones with `--sparse`, then runs:

```bash
git sparse-checkout set dream-server
```

After that, only the `dream-server/` subtree is materialized. The new bundle step checks `$TEMP_DIR/repo/resources/dev/extensions-library`, but that path is outside the sparse checkout and will normally still be absent. So the PR can still warn `resources/dev/extensions-library not in clone` and leave fresh bootstrap installs with the same missing `data/extensions-library` problem.

I?m going to fix this by adding `resources/dev/extensions-library` to the sparse-checkout set and making the bundle copy idempotent (`rm -rf` / `cp source/. dest/.`) so a retry cannot create a nested `extensions-library-bundle/extensions-library` tree.

PR #1203, opened 2026-05-14
fix(host-agent): bind 0.0.0.0 + detect container gateway for cross-bridge reach

Lightheartdevs:
## Summary

`POST /api/models/{id}/download` returns `503 {"detail":"Host agent unreachable: <urlopen error timed out>"}` on every fresh install. Reproduced 2026-05-14 on the fleet test (Tower2, mac-mini); strix-halo and spark only avoid the symptom by accident of their Docker version + absent host firewall.

## Root cause

Two coupled issues:

1. **Host-agent bind address**. `bin/dream-host-agent.py` binds to the **default Docker bridge gateway** (typically `172.17.0.1`), discovered via `docker network inspect bridge`. The intent was to keep the agent off the LAN. But `dashboard-api` runs on the **custom `dream-network`** (172.18.x.x). Under Docker's default `DOCKER-ISOLATION-STAGE-2` iptables rules, traffic from a container on one bridge can't reach a host listener bound on a *different* bridge's gateway. The `host.docker.internal:host-gateway` keyword always resolves to the default bridge gateway, so there's no resolution path that works.

2. **Dashboard-api target host**. `dashboard-api/config.py` defaults `DREAM_AGENT_HOST=host.docker.internal`. Per the above, that's a dead end for containers on custom networks.

## Fix

**Approach B chosen** (vs. detect-dream-network-gateway and write to .env, vs. attach dashboard-api to both networks — see PR discussion):

1. **`bin/dream-host-agent.py`** — bind to `0.0.0.0` on Linux. macOS/Windows still bind to `127.0.0.1` (Docker Desktop handles host.docker.internal mapping there). Operators wanting the old "virtual-bridge-only" posture can still set `DREAM_AGENT_BIND=<ip>` in `.env`.

2. **`extensions/services/dashboard-api/config.py`** — add `_detect_container_default_gateway()` that reads `/proc/net/route` directly (no iproute2 needed). `DREAM_AGENT_HOST` resolution priority:
   - explicit `DREAM_AGENT_HOST` env (operator override)
   - container's own default-gateway IP (works for *any* Docker network the container is on, including custom bridges, without depending on `host.docker.internal`)
   - `host.docker.internal` as legacy fallback

## Security trade-off (the "B" in "Approach B")

Pre-PR, the agent was bound to a virtual bridge interface that LAN devices can't reach. Post-PR, it's reachable from the LAN. All endpoints still require `Authorization: Bearer DREAM_AGENT_KEY` (64-char hex, 256 bits of entropy — brute-force is geologic-time infeasible). The bind-restriction was defense-in-depth, not the primary control. Operators on hostile LANs can set `DREAM_AGENT_BIND=<ip>` to restore the old behavior.

## Caveat: host firewalls

Hosts with **UFW or firewalld active** will additionally block container→host traffic to port 7710 at the kernel level, regardless of what address the agent binds to (reproduced on Tower2 — UFW configured to allow port 22 only). Such hosts need a one-time:

```bash
sudo ufw allow from 172.17.0.0/12 to any port 7710 proto tcp
```

…or equivalent firewalld zone rule. Not handled in this PR — proposing as a follow-up that the installer detects UFW/firewalld and prompts the operator to allow the rule.

## Test plan

- [x] Verified `_detect_container_default_gateway()` returns `172.18.0.1` from inside `dream-dashboard-api` on Tower2 (`/proc/net/route` parsing).
- [x] Verified host-agent binds to `0.0.0.0:7710` after restart with new code.
- [ ] Full fleet rerun after merge (Tower2 needs UFW rule first; strix-halo, spark, mac-mini should pass cleanly).
- [ ] Verify operator override path: set `DREAM_AGENT_BIND=127.0.0.1` and confirm the agent no longer accepts LAN traffic.

### DIFF
--- a/dream-server/bin/dream-host-agent.py
+++ b/dream-server/bin/dream-host-agent.py
@@ -3554,8 +3554,7 @@ def main():
     #   not the primary control. With a 64-char hex key (256 bits of entropy),
     #   brute-force is infeasible; LAN exposure is bounded by key custody.
     #   Operators on hostile LANs can restrict via DREAM_AGENT_BIND=<ip> in .env.
-    bind_addr = env.get("DREAM_AGENT_BIND", "")
-    bind_from_env = bool(bind_addr)
+    bind_addr = env.get("DREAM_AGENT_BIND", "").strip()
     if not bind_addr:
         if platform.system() in ("Darwin", "Windows"):
             bind_addr = "127.0.0.1"
--- a/dream-server/bin/dream-host-agent.py
+++ b/dream-server/bin/dream-host-agent.py
@@ -3542,24 +3542,35 @@ def main():
 
     # Determine bind address: env var override, or platform-aware default.
     # macOS/Windows: 127.0.0.1 (Docker Desktop routes host.docker.internal to loopback)
-    # Linux: Docker bridge gateway IP (containers reach via host-gateway,
-    #   LAN devices cannot — the bridge is a virtual interface).
-    #   Falls back to 127.0.0.1 if detection fails.
+    # Linux: 0.0.0.0 — bind to all interfaces. Earlier versions bound to the
+    #   Docker bridge gateway (typically 172.17.0.1) to keep the agent off the
+    #   LAN, but that broke containers on custom networks like dream-network
+    #   (172.18.x.x) which can't route to 172.17.0.1 across Docker's default
+    #   bridge isolation. Reproduced on every fleet-test run as
+    #   `503 Host agent unreachable: <urlopen error timed out>` from
+    #   POST /api/models/{id}/download.
+    #   All endpoints already require Authorization: Bearer DREAM_AGENT_KEY,
+    #   which is the de-facto gate — the bind restriction was defense-in-depth,
+    #   not the primary control. With a 64-char hex key (256 bits of entropy),
+    #   brute-force is infeasible; LAN exposure is bounded by key custody.
+    #   Operators on hostile LANs can restrict via DREAM_AGENT_BIND=<ip> in .env.
     bind_addr = env.get("DREAM_AGENT_BIND", "")
     bind_from_env = bool(bind_addr)
     if not bind_addr:
         if platform.system() in ("Darwin", "Windows"):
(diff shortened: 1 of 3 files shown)


Lightheartdevs:
Audit notes: this was valuable and directionally correct, but not merge-ready as submitted.

Findings:
- `dream-host-agent.py` left `bind_from_env` unused after changing the Linux default bind to `0.0.0.0`, so the Python Ruff check failed.
- `_detect_container_default_gateway()` documented that it required the default route to have `RTF_GATEWAY`, but the code accepted any `Destination == 00000000` row. On malformed/direct routes that could produce `0.0.0.0` or another bad target.
- The dashboard API would read `/proc/net/route` even when running directly on a Linux host outside Docker. In that case it can resolve the LAN router as `DREAM_AGENT_HOST`, which is wrong for local/dev execution and could make host-agent calls time out in confusing ways.

Fix pushed in `26949270`:
- Removed the unused bind variable and trims `DREAM_AGENT_BIND` overrides.
- Added a container-runtime guard before using `/proc/net/route`; non-container contexts keep the legacy `host.docker.internal` fallback unless `DREAM_AGENT_HOST` is explicit.
- Tightened route parsing to require `RTF_GATEWAY`, reject `0.0.0.0`, validate 8-char gateway hex, and skip malformed rows instead of aborting detection.
- Added unit coverage for little-endian gateway parsing, gateway-flag enforcement, explicit host override, inside-container gateway resolution, and outside-container fallback.

Local verification:
- `python -m ruff check dream-server/ --select E,F,W --ignore E501,E701,E731,E741,E402`
- `python -m py_compile main.py agent_monitor.py config.py routers/models.py routers/extensions.py routers/setup.py routers/resources.py routers/tailscale.py`
- `python -m pytest tests/test_config.py tests/test_resources.py tests/test_tailscale.py tests/test_network_config.py tests/test_models.py -q`

Note: the full dashboard-api suite timed out in my Windows workspace. `tests/test_host_agent.py` still has known Windows-only failures around POSIX executable bits and symlink privileges, unrelated to this PR; Linux CI should be the source of truth for that file.

PR #1218, opened 2026-05-15
fix(installer): gate extension compose by service plan

gabsprogrammer:
## Summary

This PR makes installer service selection explicit instead of relying on any bundled extension `compose.yaml` that happens to exist on disk.

Previously, optional/recommended extension compose files could be discovered and included even when the user selected a leaner install profile. This made Core Only / minimal installs less predictable and could pull in services that were not intentionally selected.

## What changed

- Adds a Windows service-plan helper that maps installer feature choices to explicit service decisions.
- Gates Windows extension compose discovery through that service plan.
- Keeps OpenClaw opt-in only and documents it as legacy/deprecated agent behavior.
- Aligns Linux/macOS feature sync so optional/recommended bundled services are gated instead of selected only by file presence.
- Updates Windows readiness checks, port checks, local image rebuilds, and Perplexica config so they only target selected services.
- Adds Windows service-plan coverage and wires it into the Linux smoke workflow.
- Updates Windows quickstart flags and the OpenClaw to Hermes migration note for current platform behavior.

## Validation

- PowerShell parser check for changed Windows scripts
- `Invoke-ScriptAnalyzer` over installer PowerShell files
- `bash -n` for changed shell scripts and tests
- `bash dream-server/tests/test-windows-readiness-summary.sh`
- Windows dry-run through both root `install.ps1` and `installers/windows/install-windows.ps1`

## Notes

This is intentionally focused on installer selection behavior only. It does not change service runtime behavior, container images, model selection, or dashboard UI.


gabsprogrammer:
One design choice worth calling out: unknown optional extension compose files now fail closed unless they are explicitly mapped into the installer service plan. That is intentional so Core Only / minimal installs do not silently grow when new bundled compose files are added. Happy to adjust defaults for any specific service that should remain part of the recommended profile.

Lightheartdevs:
Audit verdict: solid architectural pattern, real bug, but **cannot land as-is** — needs rebase + DreamForge removal first.

## What I liked

- **Real bug fixed.** Today on `main`, `_sync_extension_compose` in `installers/phases/06-directories.sh` only `.disabled`-renames a subset of optional extensions. `litellm`, `searxng`, `token-spy`, `ape`, `perplexica`, `privacy-shield`, `dream-proxy`, `tailscale` all bypass plan gating because `scripts/resolve-compose-stack.sh:341` picks up their compose files unconditionally. Choosing "Core Only" silently still includes most of them. The Windows side was worse — `install-windows.ps1` discovered every bundled `compose.yaml` purely by filesystem presence.
- **Single source of truth.** `installers/windows/lib/service-plan.ps1` with fail-closed defaults for unknown manifests is the right shape. The contract test (`tests/contracts/test-installer-contracts.sh:100-126`) pinning the gating list is exactly the guardrail this needs.
- **APE auto-derives from `EnableHermes -or EnableOpenClaw`** — no orphaned governance service. Good call.

## Blockers

1. **🔴 DreamForge resurrection.** `main` merged #1221 which removed DreamForge entirely. This PR turns it back ON by default in `--all` (`install-core.sh:218`), Full Stack (`ui.sh:412`, `phases/03-features.ps1:84-94`), and default `$enableDreamForge` (`phases/03-features.ps1:42`). The PR also deletes the upgrade-time DreamForge cleanup at `phases/06-directories.ps1:104-110`. On a Full Stack install post-merge, this would either crash (no `extensions/services/dreamforge/`) or silently no-op.

2. **🔴 Contract test will fail post-rebase.** `test-installer-contracts.sh:115` hard-asserts `ENABLE_DREAMFORGE:dreamforge` exists in `_sync_extension_compose` and that `service-plan.ps1` contains the literal string `"dreamforge"` — both vanish on `main`.

3. **🔴 Stale `core-service-ids.json` fallback.** `resolve-compose-stack.sh:175` adds `"dreamforge"` to the hardcoded fallback set; `main`'s `config/core-service-ids.json` (the primary source) already excludes it.

4. **⚠️ Branch is also pre-Hermes-auth.** Branched from #1217 (`2ea7bea6`), missing #1222 (admin-session) and #1223 (Hermes TUI/WS-upgrade). No textual conflicts but the PR doesn't exercise the new auth flow.

## What I need

Rebase on current `main` and strip every `DREAMFORGE` / `DreamForge` / `dreamforge` reference across the 6+ files that mention it (the contract test will guide you — it'll fail until they're all gone). Keep the Perplexica / PrivacyShield / APE / Tailscale / DreamProxy / BraveSearch / Recommended gating — that's the real value.

After that, this becomes a clean tier-0 install hardening change and I'll merge it.

Lightheartdevs (commented):
Audit finding from end-to-end pass:

- [P1] Preserve Windows non-interactive defaults. In `dream-server/installers/windows/phases/03-features.ps1:38-43`, `$enableRecommended`, `$enableHermes`, `$enableDeepResearch`, and `$enablePrivacyShield` are derived only from explicit switches or `--All`. PowerShell switches default to false, so a normal `install-windows.ps1 --NonInteractive` now builds a core-only service plan. Linux in this same PR still defaults `ENABLE_RECOMMENDED=true`, `ENABLE_HERMES=true`, `ENABLE_APE=true`, `ENABLE_PERPLEXICA=true`, and `ENABLE_PRIVACY_SHIELD=true` in `dream-server/install-core.sh:102-117`. That means a fresh Windows install can silently skip Hermes/Hermes Proxy/APE, LiteLLM/SearXNG/token-spy, Perplexica, and Privacy Shield. Please make the Windows defaults match the Linux/default Dream plan and let the `No*` switches opt out.

The PR merge-tree checks cleanly into current `main`, and the existing branch checks are green, but I would not merge this one before the default-plan behavior is fixed.

Lightheartdevs (approved):
Reviewed service-plan gating after rebasing onto main. I fixed the Windows default path to keep Hermes/recommended/research/privacy services enabled by default, preserved the 128K Hermes context bump, and split the Windows quickstart flags so the parity tests pass. Local PowerShell/JSON smoke checks passed and CI is green.

PR #1230, opened 2026-05-16
feat(llama): expose MTP speculative args

gabsprogrammer:
## Update after rebase

This draft PR has been rebased on current `main` and narrowed back to the intended MTP-only scope.

### What changed

- Rebased `gabs/llama-mtp-spec-draft` on top of current `main`.
- Resolved the GitHub conflict in `dream-server/installers/macos/lib/env-generator.sh`.
- Removed the stale DreamForge resurrection from the branch diff.
- Kept MTP support strictly opt-in: no default install path enables it automatically.
- Exposed only the currently useful/safe MTP knobs:
  - `LLAMA_ARG_SPEC_TYPE` -> `--spec-type`
  - `LLAMA_ARG_SPEC_DRAFT_N_MAX` -> `--spec-draft-n-max`
- Removed `LLAMA_ARG_SPEC_DRAFT_P_MIN` from this PR for now, since upstream `llama.cpp#22673` still treats that path as follow-up/unstable for MTP behavior.
- Changed examples from `LLAMA_ARG_SPEC_DRAFT_N_MAX=6` to `3`, matching the safer value reported in the upstream MTP discussion.

### Upstream context

I checked the recent discussion on `ggml-org/llama.cpp#22673`.

The important takeaway is that MTP should not be treated as a global default, especially in router/multi-model setups. Some users reported failures when MTP was applied while routing to models without MTP layers. So this PR intentionally does not auto-enable MTP by model tier or installer profile.

The docs now call this out explicitly:

- Only use MTP with a llama.cpp build that supports it.
- Only use it with GGUFs that actually include compatible MTP data.
- In router or multi-model setups, do not place MTP settings in a shared default section if any routed model lacks MTP layers.
- Apply `spec-type = draft-mtp` / `spec-draft-n-max = 3` only to the MTP-capable model section.

### Verification

Local checks run after the rebase:

- `bash dream-server/tests/test-llama-mtp-args.sh`
- `bash -n` on touched shell scripts
- PowerShell parser check on touched `.ps1` files
- `git diff --check origin/main..HEAD`
- `git merge-base --is-ancestor origin/main HEAD`

Also verified:

- No DreamForge files are present in the PR diff.
- No DreamForge pull-policy block remains in the macOS env generator conflict area.
- `LLAMA_ARG_SPEC_DRAFT_P_MIN` / `--spec-draft-p-min` are not exposed by this PR.
- MTP env examples remain commented out, so this is fail-closed by default.

### Current intent

This PR is now a small opt-in plumbing change for users who deliberately want to test llama.cpp MTP-capable GGUFs. It does not change default model loading behavior for normal users.


Lightheartdevs (commented):

I did not find a standalone functional blocker in this PR. The MTP args are opt-in and unset by default, the passthrough covers compose plus native host-agent/bootstrap paths, and the contract test explicitly avoids exposing the riskier `spec-draft-p-min` knob. It merge-tree checks cleanly into current `main`, and the existing branch checks are green.

Queue-order note: this conflicts with #1194 across `.env`/schema, host-agent, compose, and installer env-generation files. Whichever lands second needs a rebase before merge.

Lightheartdevs (approved):
Audited the optional llama.cpp MTP argument plumbing. The change is opt-in only, GitHub checks are green, JSON/Python/PowerShell syntax checks pass locally, and the Linux CI covers the bash contract test. Ready to merge.

PR #1236, opened 2026-05-17
feat(usage): add real cost and token usage dashboard

gabsprogrammer:
## Summary
Adds a new Usage dashboard backed by real Token Spy telemetry and dashboard-api aggregation.

The page shows monthly cost/token/request usage, daily cost, tokens per day, local vs paid usage, cost-source explanations, usage by model, top consumers, and service-level cost breakdowns. It is intentionally honest by design: when usage has no billing source or reliable pricing, it is marked as untracked instead of being guessed.

## What changed
- Adds `/usage` to the dashboard sidebar.
- Adds `GET /api/usage/report` in dashboard-api so the browser never receives the Token Spy API key.
- Adds Token Spy report aggregation for SQLite and Postgres.
- Stores provider and cost-source metadata for new Token Spy usage rows.
- Preserves old rows with safe fallback classification:
  - positive known cost -> `priced_from_tokens`
  - known local zero-cost usage -> `local_zero_cost`
  - unknown source -> `untracked`
- Adds local llama.cpp Prometheus counter fallback for local model usage that bypasses Token Spy.
- Adds CSV export and local filtering by model/provider/service/source.

## Why this is useful
- Makes DreamServer usage visible in one place.
- Reduces confusion around local vs paid usage.
- Avoids fake/demo costs.
- Helps users understand which models/services are consuming tokens.
- Keeps provider secrets server-side.

## Notes
Local model usage can show a local equivalent estimate when the model matches a known public pricing reference. This is an estimate, not a bill. Rows without reliable billing or pricing remain untracked.

## Validation
- `npm test -- src/pages/Usage.test.jsx`
- `npm run build`
- `npm run lint`
- `pytest dream-server/extensions/services/dashboard-api/tests/test_usage.py dream-server/extensions/services/token-spy/tests/test_usage_report.py -q`
- `python -m py_compile dream-server/extensions/services/dashboard-api/routers/usage.py dream-server/extensions/services/token-spy/db.py dream-server/extensions/services/token-spy/db_postgres.py dream-server/extensions/services/token-spy/main.py`

image


Lightheartdevs (commented):
Audit finding from end-to-end pass:

- [P2] Do not merge process-lifetime llama.cpp counters into date-scoped usage reports. `_merge_local_runtime_counters` in `dream-server/extensions/services/dashboard-api/routers/usage.py:613-646` appends the remaining Prometheus token counters to today's row whenever the requested range includes today. Those llama.cpp counters are cumulative since the runtime started, not scoped to the selected report window. On a long-running local server, a user querying Today or This Week can see old tokens and equivalent local cost attributed to the current report. Please either baseline/delta these counters with timestamps, persist samples so they can be windowed, or show them separately as live/lifetime local counters instead of folding them into daily spend/tokens.

The PR merge-tree checks cleanly into current `main`, and the existing branch checks are green, but this is a correctness issue for the headline dashboard numbers.

Lightheartdevs (approved):
Approved after tightening the cost semantics. Local runtime counters are now token-only metadata, provider costs are framed as estimates unless explicitly billed, and the local equivalent pricing catalog was removed. Local targeted tests and GitHub checks are green.

PR #1242, opened 2026-05-18
fix(installer): make phase 13-summary exit 0 when ComfyUI/other optio…

Lightheartdevs:
…nal services disabled

The brace group that pipes service rows into `dream_readiness_summary` ends with `[[ "${ENABLE_COMFYUI:-}" == "true" ]] && printf ...`. With `set -euo pipefail`, when ENABLE_COMFYUI is not "true", the conditional returns 1, the brace group exits 1, pipefail propagates it, and the ERR trap installed in `install-core.sh` aborts the install — even though every service started healthy.

Add a trailing `:` so the group always exits 0, with a short comment explaining the set -e + pipefail interaction so a future contributor doesn't reintroduce the bug.

Hit during `/dream-fleet-test` runs/2026-05-18T16-36-19Z on Tower2 (Linux + dual NVIDIA Blackwell, ComfyUI compose disabled).

### DIFF
--- a/dream-server/installers/phases/13-summary.sh
+++ b/dream-server/installers/phases/13-summary.sh
@@ -363,6 +363,10 @@ if command -v dream_readiness_summary >/dev/null 2>&1; then
             "${SERVICE_PORTS[qdrant]:-6333}" "${SERVICE_HEALTH[qdrant]:-/}" "$(sr_container qdrant)" "http://localhost:${SERVICE_PORTS[qdrant]:-6333}"
         [[ "${ENABLE_COMFYUI:-}" == "true" ]] && printf 'ComfyUI|http://127.0.0.1:%s%s|%s|%s\n' \
             "${SERVICE_PORTS[comfyui]:-8188}" "${SERVICE_HEALTH[comfyui]:-/}" "$(sr_container comfyui)" "http://localhost:${SERVICE_PORTS[comfyui]:-8188}"
+        # Ensure the block exits 0 regardless of the trailing optional conditionals:
+        # under set -e + pipefail, a false `[[ ENABLE_x ]] && printf` makes the block
+        # return 1, which propagates through the pipe and trips the ERR trap.
+        :
     } | dream_readiness_summary "dream status" "$LOG_FILE" "$_dashboard_url"
 fi
 


Lightheartdevs:
## What failed

Fresh `--all --non-interactive` install on Tower2 (Linux NVIDIA, dual Blackwell) exited rc=1 at phase `13-summary` even though every service started healthy:

```
INSTALL READINESS
Ready now: 11/11
Ready:
  [OK] Dashboard                    http://localhost:3001 (HTTP 200)
  [OK] Chat UI (Open WebUI)         http://localhost:3000 (HTTP 200)
  [OK] llama-server                 http://localhost:11434/v1 (HTTP 200)
  [OK] Dashboard API                http://localhost:3002 (HTTP 200)
  [OK] LiteLLM                      http://localhost:4000 (HTTP 200)
  [OK] Perplexica                   http://localhost:3004 (HTTP 200)
  [OK] OpenClaw                     http://localhost:7860 (HTTP 200)
  [OK] Whisper (STT)                http://localhost:9000 (HTTP 200)
  [OK] Kokoro (TTS)                 http://localhost:8880 (HTTP 200)
  [OK] n8n                          http://localhost:5678 (HTTP 200)
  [OK] Qdrant                       http://localhost:6333 (HTTP 200)
Next:
  - Open dashboard: http://localhost:3001
  - Check status: dream status
  - Logs: /tmp/dream-server-install.log

[ERROR] Installation failed during phase: 13-summary
```

`/tmp/dream-server-install.log` ends cleanly at "✓ dream command already available" — the failure trap fires *after* phase 13 finishes printing its banner.

## Root cause

`installers/phases/13-summary.sh:339-366` pipes a synthetic input into `dream_readiness_summary`:

```bash
{
    printf 'Dashboard|...'
    ...
    [[ "$ENABLE_RAG" == "true" ]] && printf 'Qdrant|...'
    [[ "${ENABLE_COMFYUI:-}" == "true" ]] && printf 'ComfyUI|...'
} | dream_readiness_summary "dream status" "$LOG_FILE" "$_dashboard_url"
```

With `set -euo pipefail` (set in `install-core.sh`), if the **last** statement in the brace group is a `[[ x ]] && printf` whose condition evaluates to false, the group exits with status 1. `pipefail` propagates that to the whole pipeline, and the `ERR` trap installed by `install-core.sh` fires `cleanup_on_error`, masking a successful install as a failure.

Reproduction:

```bash
$ bash -c 'set -euo pipefail; { printf a; [[ false == true ]] && printf b; } | cat; echo $?'
a
# (no rc=0 line — script aborted)
```

Why it only hit Tower2 in the fleet test: on Tower2, `comfyui/compose.yaml` was renamed to `.disabled` and `ENABLE_COMFYUI` was left non-true, so the final conditional was the false one. On strix-halo / spark / mac-mini, ComfyUI was enabled, so the last `printf` succeeded and the block exited 0.

## Fix

Add a no-op `:` as the final statement in the brace group so the block's exit status is always 0, independent of which optional services happen to be the last truthy conditional. A short comment explains the `set -e` + `pipefail` interaction so a future contributor doesn't reintroduce the bug.

This is the smallest possible surgical fix:
- Does not silence real errors inside any of the `printf`s (each still propagates via `set -e` while it executes).
- Does not require `|| true` (which `CLAUDE.md` discourages).
- Does not touch the `ERR` trap or `set -e` in `install-core.sh`.

An alternative considered: disabling the `ERR` trap around phase 13 (`trap - ERR; ... ; exit 0`) — rejected as too broad; it would silence real failures in unrelated cosmetic logic later in the phase.

## Why it's safe

- The trailing `:` produces no output, so `dream_readiness_summary`'s parsed input is unchanged.
- The fix is bounded to one phase, one brace group; behavior for every other ENABLE_* permutation is identical.
- Verified locally: with `ENABLE_COMFYUI=false`, the patched block exits 0 under `set -euo pipefail`, and `dream_readiness_summary` still produces the same output.

## Discovered by

`/dream-fleet-test` run `runs/2026-05-18T16-36-19Z` on Tower2 (Linux + dual NVIDIA Blackwell, GPU_BACKEND=nvidia, DREAM_MODE=local).

PR #1243, opened 2026-05-18
Stabilize host agent and update version paths

Lightheartdevs:
## Summary

Reliability/stability slice focused on making existing host-agent, update, and version behavior more predictable.

- Scope Linux host-agent binding to the dream-network gateway by default, with Docker bridge and loopback fallbacks instead of defaulting to 0.0.0.0.
- Restart the Linux host-agent after compose creates dream-network so the scoped bind can take effect during install.
- Make dashboard update actions platform-aware on Windows: use verified Bash when present, otherwise return an explicit 501.
- Parse JSON-formatted .version files consistently in dashboard version readers.
- Fix GPU TTL cache behavior for cached falsy GPU-unavailable results.
- Add a release gate check for manifest, changelog, installer constants, and dashboard API version metadata.
- Update host-agent docs/security/schema text to match the behavior.

## Verification

- python -m pytest tests\\test_main.py::test_read_installed_version_parses_json_version_file tests\\test_main.py::TestGpuEndpoint tests\\test_updates.py tests\\test_host_agent.py::TestResolveAgentBindAddr -q -> 30 passed
- python -m compileall -q . in dashboard-api
- python -m py_compile dream-server/bin/dream-host-agent.py dream-server/scripts/check-version-consistency.py
- python dream-server/scripts/check-version-consistency.py -> PASS
- git diff --check -> clean except Windows CRLF warnings

Bash syntax/release-gate could not run locally because this Windows environment's bash shim fails with /bin/bash missing.

### DIFF
--- a/dream-server/.env.schema.json
+++ b/dream-server/.env.schema.json
@@ -639,7 +639,7 @@
     },
     "DREAM_AGENT_BIND": {
       "type": "string",
-      "description": "Bind address for the Dream Host Agent HTTP server. Defaults to 127.0.0.1 on macOS/Windows, Docker bridge gateway IP on Linux (falls back to 127.0.0.1).",
+      "description": "Bind address for the Dream Host Agent HTTP server. Defaults to 127.0.0.1 on macOS/Windows, dream-network gateway IP on Linux, then Docker bridge gateway, then 127.0.0.1.",
       "default": ""
     },
     "DREAM_AGENT_HOST": {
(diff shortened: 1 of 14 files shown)


Lightheartdevs:
Additional local Linux-container verification from Codex:

Passed in `python:3.12-bookworm` with a real `/bin/bash`:

- `bash -n scripts/release-gate.sh installers/phases/01-preflight.sh installers/phases/11-services.sh`
- `python3 scripts/check-version-consistency.py`
- `bash scripts/check-compatibility.sh`
- `bash scripts/check-release-claims.sh`

I also attempted full `bash scripts/release-gate.sh` in the same clean container after installing `git`, `jq`, and `PyYAML`. It reached the existing installer-hardening contracts and then failed because the clean container lacks `nvidia-ctk`:

```text
installers/lib/sudo.sh: line 15: nvidia-ctk: command not found
```

That appears environment-bound rather than related to this PR. GitHub checks for this PR are green, including shell lint, API, distro smoke, frontend, and integration smoke.

Issue #1271, opened 2026-05-19
security(installer): eval on hardware-detection output enables install-time code execution


nolanmak:
## Summary

The installer `eval`s a string derived from hardware/GPU detection output. If that output contains shell metacharacters (an unusual or spoofed GPU/device name, a crafted environment), this is arbitrary command execution during a privileged install.

## Evidence

- `installers/macos/install-macos.sh:528` — `eval "$_selector_env"`
- `installers/phases/02-detection.sh:532` — `eval "$_selector_env"`

`$_selector_env` is assembled from hardware-detection / selector output, which is not a trusted constant. `eval` on it executes any embedded shell.

## Impact

Code execution in the context of the installer (which performs privileged operations: Docker setup, package installs, system tuning). Trigger surface is narrow but real — anything that influences detected device/GPU strings or the selector environment.

## Proposed fix

Replace `eval` with a strict, non-executing parser:

- Read only `KEY=VALUE` lines, where `KEY` matches an allowlist of expected variable names and `^[A-Za-z_][A-Za-z0-9_]*$`.
- Assign via the shell's own assignment (no `eval`), reject/ignore anything that doesn't match.
- Optionally validate `VALUE` shapes per key.

This preserves the "export a small env contract from detection" pattern without an `eval` sink.

## Claiming

I'd like to take this and open a PR. (GitHub won't let me self-assign — not a collaborator — noting here that I'm picking it up.)

_Filed together with a small set of related contribution issues from one analysis pass; cross-links to follow in a comment._

nolanmak:
Related issues, filed together from one analysis pass: #1268 (privacy-shield streaming), #1269 (APE windowed governance), #1270 (OpenClaw device-auth follow-up), #1271 (installer `eval` RCE). I'd like to pick these up — happy to open PRs; let me know if any should be scoped differently or aren't wanted.

Lightheartdevs:
Thanks for filing this. I agree this is a valid installer-hardening issue and worth fixing.

A couple of notes on scope: the current selector path emits fixed keys and uses shell quoting, so the practical exploit surface is narrower than "any GPU name becomes RCE." But the core point still stands: `eval` in a privileged installer path is a footgun, and we should remove it.

A PR would be very welcome. Ideal shape:

- replace the `eval "$_selector_env"` call sites in Linux and macOS with a small strict loader
- allow only the expected selector keys
- parse only `KEY=VALUE` records with valid shell variable names
[...]

Issue #1268, opened 2026-05-19
bug(privacy-shield): proxy fully buffers responses — breaks streaming chat (stream:true) and binary/compressed bodies


nolanmak:
## Summary

`privacy-shield` fully buffers both the request and the response body before forwarding/returning. This breaks streaming LLM responses (`stream: true`, the default for most chat clients), and corrupts any non-UTF-8 / compressed response.

## Evidence

`extensions/services/privacy-shield/proxy.py`:

- L150 — `body = await request.body()` (entire request buffered)
- L173 — forwards the fully-buffered scrubbed body
- L182 — `response_body = resp.content.decode('utf-8')` (entire response buffered, hard `utf-8` decode)
- L185 — `restored_body = shield.process_response(response_body)` then returned via `JSONResponse`

There is no `StreamingResponse`, no `httpx` streaming (`client.stream(...)`), no chunk/SSE handling, and no WebSocket path anywhere in the module.

## Impact

1. **Streaming chat is broken / severely degraded.** With `stream: true`, the upstream returns `text/event-stream` (SSE). `resp.content` blocks until the *entire* generation completes, so the client sees no tokens until the full response is ready — latency becomes total-generation-time, and the SSE framing is then run through a JSON-oriented `process_response`, which can mangle it.
2. **Binary / gzip / non-UTF-8 responses raise `UnicodeDecodeError`** at L182 (`.decode('utf-8')`), failing the request entirely.
3. **No WebSocket support** for any upstream that upgrades.

## Proposed fix

- Use `httpx`'s streaming API (`client.stream()`) and return a `StreamingResponse`.
- Detect `content-type: text/event-stream` and scrub per-SSE-event (`data:` lines) incrementally instead of buffering, preserving event framing.
- Honor the response `Content-Type`/charset; pass through (without decode) bodies that aren't text, or that exceed a configurable size cap.
- Add a transparent passthrough lane for `Upgrade: websocket`.

## Claiming

I'd like to take this and open a PR. GitHub won't let me self-assign (not a collaborator) — flagging here that I'm picking it up. For context, I maintain [`claude-intercept`](https://github.com/whaesemeyer/claude-intercept), an MITM proxy that solves the analogous streaming + WebSocket-passthrough + decompression problem through a body-transforming TLS proxy, so this maps directly onto prior work.

_Filed together with a small set of related contribution issues from one analysis pass; cross-links to follow in a comment._

nolanmak:
[...]

PR #1272, opened 2026-05-19
feat(ape): persistent windowed rate limits + human-approval tier

### LINKED ISSUE #1269: feature(ape): persistent windowed rate limits + human-approval decision tier
## Summary

APE's rate governance is currently a single global fixed RPM enforced from an in-memory structure, with a binary allow/deny outcome. This is fragile (resets on restart), coarse (one window, one limit for all intent classes), and missing a "needs human approval" middle ground that an agent-policy engine usually wants.

## Current behavior

`extensions/services/ape/main.py`:

- `RATE_LIMIT = int(os.environ.get("APE_RATE_LIMIT_RPM", "60"))` — one flat per-session requests/minute cap.
- Enforced via an in-memory `collections.deque` — **state is lost on container restart**, and is per-process only.
- Decisions are effectively allow / deny (plus `APE_STRICT_MODE`); there is no graduated, per-intent-class limiting and no "require approval" decision class, even though intent classes (`ReadFile`, `WriteFile`, `ExecuteCommand`, `NetworkFetch`, `SpawnAgent`, `Other`) are already modeled.

## Proposal

1. **Sliding-window, multi-tier caps** (e.g. 5-min / hour / day) keyed per intent class, instead of a single global RPM.
2. **Durable state** so limits and any circuit-breaker halts survive restarts (SQLite or a small persisted store under `/data/ape/`).
3. **A third decision outcome — `require_approval`** — for risk-flagged actions (e.g. `ExecuteCommand` / `SpawnAgent` outside the allowlist), surfaced rather than silently denied.
4. Optional: a warm-up ramp for fresh sessions and a circuit-breaker that halts a class after repeated denials, both persisted.

This stays compatible with the existing `/verify` / `/audit` / `/policy` API and `policy.yaml` shape — it extends the governor, not the contract.

## Claiming

I'd like to take this and open a PR. (GitHub won't let me self-assign — not a collaborator — noting here that I'm picking it up.) For context, I built the windowed-cap + risk-aware approval-tier + restart-durable governor pattern in a separate project and can port the design.

nolanmak:
# feat(ape): persistent windowed rate limits + human-approval tier

Fixes #1269

## Problem

APE's limiter was a single 60s per-session `deque` (`main.py:125`): no
burst-vs-sustained quotas, **all state lost on restart**, and a binary
allow/deny outcome with no path between "permitted" and "blocked" short of a
human editing `policy.yaml`.

## Fix (all additive / backward-compatible)

- **Sliding-window multi-tier caps** (5min / 1h / 1d) per intent class, per
  session, configurable in `policy.yaml`. Legacy `rate_limit.requests_per_minute`
  is preserved verbatim. Blocked calls aren't counted (one offending call can't
  poison a window).
- **Durable state** at `/data/ape/state.json` (sibling of `audit.jsonl` on the
  existing writable volume): window samples, circuit-breaker trip, warmup
  deadline, pending approvals survive restart. Atomic writes (temp + `os.replace`
  + `fsync`) under an `RLock` + advisory `flock`; time-pruned with hard backstop
  caps (bounded footprint).
- **Third outcome `require_approval`.** `VerifyResponse` keeps
  `allowed/reason/intent/decision_id` **unchanged** and adds two **optional**
  fields (`decision`, `approval_token`). `allowed` stays `false` for
  require-approval so old clients fail-closed. New additive `POST /approve`
  consumes a one-shot token.
- Optional warmup grace + persisted circuit-breaker (trips to deny-all on high
  rolling deny ratio; survives restart so a crash-loop can't reset it).
- `/audit` `/policy` `/metrics` extended additively only.

### Design note for reviewers

`require_approval` is **advisory**: it returns HTTP **200 even under
STRICT_MODE** (with `allowed:false` + `decision:"require_approval"`) so the
agent framework can route the action to a human rather than hard-failing it;
hard policy/windowed denials still raise 403 and rate-limit/breaker still raise
429. The contract is that a caller must treat `allowed:false` as
non-executable. If you'd prefer a distinct strict-mode status (e.g. 202),
that's a one-line change — flagging the decision explicitly for your call.

## Tests

`tests/test_main.py` (24) + `tests/test_windowed_governance_contract.py`:
schema back-compat, multi-tier caps, persistence across simulated restart,
concurrency, `require_approval`, `/approve`, circuit-breaker, warmup,
STRICT_MODE invariants, legacy-policy-only path. No new runtime deps (stdlib
`threading`/`fcntl`).

## Notes

Part of a 4-PR set (#1268–#1271). Can't self-assign (not a collaborator);
happy to own and revise.


Lightheartdevs:
Thanks for pushing this. The direction is valuable, but I do not think this is ready to merge yet.

Findings:

1. **Approval currently does not authorize a retry.** `/verify` issues an approval token when the window is exhausted, but `/approve` only pops the token and writes an audit record. It does not store a one-shot grant that `check_windowed_limits()` or the next `/verify` can consume. So after approval, retrying the original tool call will pass back through the same exhausted window at `extensions/services/ape/main.py:770-781` and return `require_approval` again. Please add a real one-shot bypass keyed tightly to the approved session/tool/intent/args, or accept and consume the approval token on `/verify`. Also add a test that proves: exhaust window -> receive token -> approve -> retry original action -> exactly one allow.

2. **CI is currently red.** Ruff is failing on unused locals/imports:
   - `extensions/services/ape/main.py:496` unused `cooldown`
   - `extensions/services/ape/main.py:847` unused `now`
   - `extensions/services/ape/tests/test_main.py:11` unused `pytest`
   - `extensions/services/ape/tests/test_windowed_governance_contract.py:24` unused `pytest`

This is a good feature shape, especially the persistence/audit/test direction. The important missing piece is making approval actually change the next authorization decision, not just record that a human clicked approve.

nolanmak:
Both findings addressed — ready for re-review.

1. **Approval now authorizes the retry** (`d1272114`). `/approve` no longer just pops the token + audits — it mints a real one-shot bypass **grant** in the persisted `state.json` (new `grants` collection, survives restart like approvals/breaker state), keyed tightly to the approved `{session, tool, intent, args-hash}`. The next `/verify` whose classified action matches calls `consume_grant()`, which atomically deletes the grant and allows the call past the exhausted window **exactly once**; the approved retry is still recorded as a window sample and audited (`grant_consumed: true`). A second retry with no fresh approval re-escalates to `require_approval`. Strictly one-shot and tightly keyed — not a broad cap lift. Added `test_approval_authorizes_exactly_one_retry_then_reescalates` proving the exact lifecycle you asked for (exhaust → token → approve → retry = one allow → retry again = require_approval), plus different-args-don't-unlock, restart-persistence, and hard-deny-tier coverage. `/verify` schema + STRICT_MODE unchanged.
2. **Ruff** — the F401/F841 items were resolved in `f353475f` (landed after the commit you reviewed); CI Ruff is green. **44 passed**.

PR #1273, opened 2026-05-19
fix(openclaw): enable device auth by default, gate disable behind explicit opt-in

### LINKED ISSUE #21: Security: OpenClaw gateway bound to 0.0.0.0 in default config
Hey 👋

Love the Lighthouse-AI project — one command to a full local AI stack is a cool idea. But I noticed the OpenClaw config at `dream-server/config/openclaw/openclaw.json` has the gateway binding to `0.0.0.0:7860`.

**The issue:**
`"host": "0.0.0.0"` means the gateway listens on all network interfaces. For users deploying this on a cloud VM or a machine without a proper firewall, the OpenClaw gateway (and its full toolset — exec, read, write, web_fetch) is exposed to the internet.

Since there's no auth token configured in this config, anyone who discovers the open port gets full agent access. That's essentially unauthenticated remote code execution.

**Impact for your users:**
Since this is a starter project that people clone and deploy, insecure defaults propagate. Every user who deploys Lighthouse-AI without changing this config inherits the exposure.

**Suggested fix:**

In `dream-server/config/openclaw/openclaw.json`:
```json
"gateway": {
  "port": 7860,
  "host": "127.0.0.1"
}
```

And in `dream-server/config/openclaw/openclaw.json.example`, same change. You might also want to add a gateway auth token in the example:
```json
"gateway": {
  "port": 7860,
  "host": "127.0.0.1",
  "token": "${OPENCLAW_GATEWAY_TOKEN}"
}
```

**Context:** Censys data from January 2026 showed 21,639 exposed OpenClaw instances on the internet. A lot of those are probably from starter projects with insecure defaults. There's an active CVE (CVE-2026-25253) for this exact issue.

Happy to open a PR for this if you'd prefer — it's a one-line change.

Cheers

### LINKED ISSUE #1270: security(openclaw): dangerouslyDisableDeviceAuth still enabled by default (follow-up to #21/#22)
## Summary

Follow-up to the now-closed #21 / #22 (OpenClaw gateway bound to `0.0.0.0`). The network-bind issue was fixed, but the configuration still **disables device authentication by default** and runs the gateway unconfigured/LAN-bound at the process level. With the default `BIND_ADDRESS=127.0.0.1` this is contained, but it's a defense-in-depth gap the moment `BIND_ADDRESS` is widened (LAN, Tailscale, `0.0.0.0`).

## Evidence

- `config/openclaw/openclaw.json:35` — `"dangerouslyDisableDeviceAuth": true`
- `extensions/services/openclaw/compose.yaml` entrypoint — `... openclaw.mjs gateway --allow-unconfigured --bind lan`
- `extensions/services/openclaw/compose.yaml:34` — port map `${BIND_ADDRESS:-127.0.0.1}:${OPENCLAW_PORT:-7860}:18789`

So the only thing standing between an unauthenticated agent gateway (exec / read / write tools) and the network is the default `BIND_ADDRESS`. Anyone exposing the stack over LAN/Tailscale (a documented use case) gets an unauthenticated agent.

## Proposal

- Default device auth **on**; if it must be disableable, gate it behind an explicit opt-in env var with a loud startup warning.
- Reconsider `--allow-unconfigured` / `--bind lan` in the entrypoint vs. `--bind localhost`, so the process-level binding matches the localhost intent rather than relying solely on the Docker port mapping.

## Relationship to other issues

Not a duplicate of #21 / #22 (those fixed the Docker `0.0.0.0` bind and are closed). Aware of #191 / #628 (possible swap of OpenClaw for zeroclaw/zeptoclaw via an alternatives registry) — this hardening is still warranted regardless of which agent ships, and the same auth-default principle should apply to any replacement.

## Claiming

I'd like to take this and open a PR. (GitHub won't let me self-assign — not a collaborator — noting here that I'm picking it up.)

nolanmak:
# fix(openclaw): enable device auth by default, gate disable behind explicit opt-in

Fixes #1270 · follow-up to the closed #21 / #22 (which fixed the `0.0.0.0` bind)

## Problem

`config/openclaw/openclaw.json` shipped `dangerouslyDisableDeviceAuth: true`,
and — critically — `config/openclaw/inject-token.js` **unconditionally
re-forced that flag at every container start** (so editing the JSON alone is a
runtime no-op; an existing test even guarded the re-injection). The agent
gateway therefore ran without device pairing. Contained under the default
`BIND_ADDRESS=127.0.0.1`, but an unauthenticated exec/read/write agent the
moment `BIND_ADDRESS` is widened (LAN, Tailscale, `0.0.0.0`).

## Fix

- **`inject-token.js`**: device auth defaults **ON**. The disable key is
  removed unless `OPENCLAW_DANGEROUSLY_DISABLE_DEVICE_AUTH` is exactly `"true"`,
  in which case it's disabled **and a loud multi-line redaction-safe warning
  banner is logged**. Applied in both Part 1 (home) and Part 3 (merged) via a
  shared helper. `allowInsecureAuth` (separate transport concern) untouched.
- `openclaw.json` / `pro.json` / `openclaw-strix-halo.json`: flag → `false`
  for static-source consistency.
- `compose.yaml`: gateway `--bind` mode now tracks `BIND_ADDRESS`
  (loopback → `localhost`, widened → `lan`); dropped `--allow-unconfigured`
  (merged config is always written); opt-in env passed through. Healthcheck
  still hits `127.0.0.1` in the same netns, unaffected by `--bind localhost`.
- `test-openclaw-inject-token.sh`: inverted the device-auth assertion to
  require auth ON by default; added opt-in + warning coverage.
- `.env.example`: documented the opt-in with a security warning.

Default loopback Control-UI auto-connect is preserved — token auth and device
pairing are separate gates, and the token-injection path is untouched.

## Tests

`tests/test-openclaw-device-auth-default.sh` (new, drives `inject-token.js`
headlessly so cosmetic-only edits would fail it) — 6/6.
`tests/test-openclaw-inject-token.sh` — 30/30 with the inverted contract.

## Notes

Part of a 4-PR set (#1268–#1271). Not a collaborator (can't self-assign);
I'd like to own this and will revise to feedback.

### DIFF
--- a/dream-server/.env.example
+++ b/dream-server/.env.example
@@ -42,6 +42,23 @@ OPENCODE_SERVER_PASSWORD=CHANGEME
 OPENCLAW_TOKEN=CHANGEME
 
 # ═══════════════════════════════════════════════════════════════════
+# OpenClaw Security
+# ═══════════════════════════════════════════════════════════════════
+#
+# OpenClaw device authentication (device pairing) defaults to ON. Token auth
+# and device auth are separate gates: a valid OPENCLAW_TOKEN still auto-
+# connects the local Control UI with device auth enabled, so you normally do
+# NOT need to change this.
+#
+# Setting this to "true" DISABLES device pairing entirely. Anyone who can
+# reach the gateway then gets an UNAUTHENTICATED agent with exec / read /
+# write tools — especially dangerous with a widened BIND_ADDRESS
+# (LAN / Tailscale / 0.0.0.0). Only enable as a deliberate, understood
+# escape hatch; a loud warning banner is logged at startup when set.
+# Leave unset (the secure default) unless you fully accept the risk.
+# OPENCLAW_DANGEROUSLY_DISABLE_DEVICE_AUTH=
+
+# ═══════════════════════════════════════════════════════════════════
 # Network Binding
 # ═══════════════════════════════════════════════════════════════════
 
--- a/dream-server/config/openclaw/inject-token.js
+++ b/dream-server/config/openclaw/inject-token.js
@@ -34,6 +34,52 @@ const CONFIG_PATH = path.join(process.env.HOME || '/home/node', '.openclaw', 'op
 const HTML_PATH = process.env.OPENCLAW_CONTROL_UI_HTML || '/app/dist/control-ui/index.html';
 const JS_PATH = process.env.OPENCLAW_AUTO_TOKEN_JS || '/app/dist/control-ui/auto-token.js';
 
+// ── Device auth policy (#1270) ────────────────────────────────────────────────
+//
+// Device auth (pairing) is a SEPARATE gate from token auth. Token auth + a
+// valid injected gateway token still auto-connect on loopback with device auth
+// ON, so device auth defaults to ON regardless of bind mode. Disabling it is an
+// explicit, deliberate opt-in via OPENCLAW_DANGEROUSLY_DISABLE_DEVICE_AUTH.
+//
+// Behaviour:
+//   - unset / anything but "true"  -> device auth ON (delete the disable key so
+//                                     the gateway default of device auth ON
+//                                     takes effect)
+//   - exactly "true"               -> device auth DISABLED + loud startup banner
+//
+// allowInsecureAuth is a distinct transport concern and is intentionally NOT
+// touched here.
+const DISABLE_DEVICE_AUTH =
+  (process.env.OPENCLAW_DANGEROUSLY_DISABLE_DEVICE_AUTH || '').trim().toLowerCase() === 'true';
+let _deviceAuthWarned = false;
+
+// Apply the device-auth decision to a controlUi object in place. When disabled,
+// emit the warning banner exactly once per process (Part 1 and Part 3 share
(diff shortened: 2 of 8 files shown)


Lightheartdevs:
This looks like the right security direction to me. I did not find a blocking issue in the diff.

What I checked:

- `OPENCLAW_DANGEROUSLY_DISABLE_DEVICE_AUTH` is explicit opt-in rather than the default.
- Runtime config patching deletes the dangerous device-auth-disable key unless the opt-in is exactly `true`.
- The compose entrypoint no longer starts the gateway with `--allow-unconfigured --bind lan` unconditionally.
- Local check: `tests/test-openclaw-device-auth-default.sh` passed (`6 passed, 0 failed`).
- Local check: `tests/test-openclaw-inject-token.sh` skipped here because this Windows workspace lacks `jq`; CI has the more useful signal.

Only caveat before making this non-draft: please smoke-test one real OpenClaw container start with default localhost binding and, if possible, one widened `BIND_ADDRESS=0.0.0.0` path. This PR changes the process-level gateway bind mode, not only static config, so a real boot check is worth doing even though the static tests and CI are green.

PR #1274, opened 2026-05-19
fix(installer): parse model-selector env without eval (defense-in-depth)

### LINKED ISSUE #1271: security(installer): eval on hardware-detection output enables install-time code execution

nolanmak:
# fix(installer): parse model-selector env without eval (defense-in-depth)

Fixes #1271

## Problem

`installers/macos/install-macos.sh:528` and
`installers/phases/02-detection.sh:532` ran `eval "$_selector_env"` on the
`KEY=VALUE` output of `scripts/select-model.py`. That output is derived from a
device/hardware string plus a JSON model catalog, so the `eval` sink combined
with catalog-controlled `runtime_profile.env` keys is an install-time
code-execution exposure.

**Honest scope:** the producer currently `shlex.quote()`s values, so this is
**defense-in-depth, not a trivially-live RCE** — but the `eval` sink should
not exist on catalog-influenced input regardless.

## Fix

Add `lib/safe-env.sh::load_selector_env` — a non-executing parser that:

- reads `KEY=VALUE` from stdin;
- gates keys on the `^[A-Za-z_][A-Za-z0-9_]*$` identifier regex (**the
  security boundary** — rejects `;`, `$(...)`, backticks, etc. in key
  position); the 15 known keys plus regex-matching dynamic
  `runtime_profile.env` keys are accepted so future catalogs don't regress;
- shlex-dequotes values by hand, rejects bare values containing shell
  metacharacters, applies value-shape checks (numeric / sha256 / https url);
- assigns via `printf -v` + `export` — never `eval`/`source`/`declare`.

Both call sites use `load_selector_env < <(printf … "$_selector_env")`.
Process substitution (not a pipe) is deliberate: a pipe would subshell the
parser and lose the exported model vars (`LLM_MODEL`, `GGUF_FILE`,
`MAX_CONTEXT`, …) the installer needs downstream — this preserves prior
behavior for valid input under `set -euo pipefail`.

`load_env_from_output` / `load_env_file` are unchanged (pinned by
`test-safe-env.sh` Tests 1–5).

## Tests

`tests/test-safe-env.sh` (+7 tests, 12/12): all 15 keys, KEY-regex gate,
dynamic keys, the `x=1; touch …` and `'x'$(touch …)'y'` injection payloads
proven inert + rejected, value-shape rejection, existing tests unchanged.
`tests/test-installer-selector-env-no-eval.sh` (regression suite) 13/13 —
its two prior failures were exactly the two `eval` sites. `shellcheck`
CI-parity clean.

## Notes

Part of a 4-PR set (#1268–#1271). Not a collaborator (can't self-assign);
happy to own and revise.

### DIFF
--- a/dream-server/installers/macos/install-macos.sh
+++ b/dream-server/installers/macos/install-macos.sh
@@ -505,6 +505,10 @@ if [[ "${DREAM_DISABLE_CATALOG_MODEL_SELECTOR:-false}" != "true" && "$SELECTED_T
             . "${SOURCE_ROOT}/lib/python-cmd.sh"
             _selector_python="$(ds_detect_python_cmd || true)"
         fi
+        if [[ -f "${SOURCE_ROOT}/lib/safe-env.sh" ]]; then
+            # shellcheck source=/dev/null
+            . "${SOURCE_ROOT}/lib/safe-env.sh"
+        fi
         if [[ -z "$_selector_python" ]]; then
             if command -v python3 >/dev/null 2>&1; then
                 _selector_python="python3"
@@ -525,7 +529,12 @@ if [[ "${DREAM_DISABLE_CATALOG_MODEL_SELECTOR:-false}" != "true" && "$SELECTED_T
                 --installable-only \
                 --env 2>>"$LOG_FILE" || true)"
             if [[ -n "$_selector_env" ]]; then
-                eval "$_selector_env"
+                # #1271: parse the selector output WITHOUT eval. Its values
+                # derive from a device string + JSON catalog and must never
+                # reach a shell. Process substitution (not a pipe) keeps the
+                # parser in this shell so the exported model vars persist.
+                load_selector_env < <(printf '%s\n' "$_selector_env") \
+                    || ai "Model selector env parse skipped (invalid output)"
                 ai "Model selector: ${MODEL_RECOMMENDATION_REASON:-$LLM_MODEL}"
             else
                 ai "Model selector unavailable; using tier-map model ${LLM_MODEL}"
--- a/dream-server/installers/phases/02-detection.sh
+++ b/dream-server/installers/phases/02-detection.sh
@@ -509,6 +509,10 @@ if [[ "${DREAM_DISABLE_CATALOG_MODEL_SELECTOR:-false}" != "true" && "${TIER:-}" 
             . "$SCRIPT_DIR/lib/python-cmd.sh"
             _selector_python="$(ds_detect_python_cmd || true)"
         fi
+        if [[ -f "$SCRIPT_DIR/lib/safe-env.sh" ]]; then
+            # shellcheck source=/dev/null
+            . "$SCRIPT_DIR/lib/safe-env.sh"
+        fi
         if [[ -z "$_selector_python" ]]; then
             if command -v python3 >/dev/null 2>&1; then
                 _selector_python="python3"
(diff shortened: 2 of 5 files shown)


Lightheartdevs:
This is the right class of fix, but I would not merge this shape as-is.

Findings:

1. **The selector parser still trusts arbitrary environment variable names.** `load_selector_env()` accepts any key matching `^[A-Za-z_][A-Za-z0-9_]*$` (`lib/safe-env.sh:100-103`), and the comments explicitly say it does not hard-drop dynamic keys (`lib/safe-env.sh:70-73`). Removing `eval` is a big improvement, but exporting arbitrary catalog-provided names still lets selector/catalog data affect later installer subprocesses via environment semantics (`PATH`, `BASH_ENV`, `ENV`, `DOCKER_HOST`, etc.). I would make this a small allowlist: known selector keys plus explicitly approved runtime knobs such as `LLAMA_ARG_*` / `LLAMA_SERVER_IMAGE`, rather than regex-only acceptance.

2. **One test validates the wrong loader.** In `tests/test-installer-selector-env-no-eval.sh:116-140`, the ?legitimate selector output? section calls `load_env_from_output`, not `load_selector_env`, even though the installers call `load_selector_env`. That can pass while the actual installer path behaves differently. Example: the test uses `GGUF_SHA256="abc123def456"`, which `load_selector_env` should reject because it is not a 64-char hex digest, but `load_env_from_output` accepts it.

3. **Please strip one trailing `\r` before parsing.** Windows/Git Bash Python can produce CRLF output, which makes otherwise-valid `KEY=value` lines fail the quote/value-shape checks unless the parser trims the carriage return first.

Net: valuable issue, correct instinct, but the boundary should be narrower. #1276 covers the same installer eval path with an allowlisted parser; the useful tests here could be folded into that approach.

nolanmak:
Deferring to #1276. You're right the boundary should be a narrow allowlist rather than regex-only — and since #1276 ("harden boring install contracts") already replaces the selector `eval` with an allowlisted parser and adds CRLF tolerance from your side, this PR's parser is redundant against your preferred shape.

The genuinely reusable part here is the test coverage: the no-eval/injection regression suite (`tests/test-installer-selector-env-no-eval.sh` — `x=1; touch …`, `$(...)`, backtick payloads proven inert) and the `load_selector_env` unit tests. Happy to fold those into #1276 — rebased onto its parser, and with the wrong-loader bug you flagged corrected (the legitimate-output case calling `load_env_from_output` instead of `load_selector_env`, and the non-64-hex `GGUF_SHA256`). Closing in favor of #1276; say the word if you'd like that test-only PR against its branch.

PR #1275, opened 2026-05-19
fix(privacy-shield): stream responses instead of fully buffering

### LINKED ISSUE #1268: bug(privacy-shield): proxy fully buffers responses — breaks streaming chat (stream:true) and binary/compressed bodies

nolanmak:
# fix(privacy-shield): stream responses instead of fully buffering

Fixes #1268

## Problem

`extensions/services/privacy-shield/proxy.py` buffered the entire upstream
response and hard-decoded it as UTF-8:

- `proxy.py:150` `body = await request.body()` — whole request buffered
- `proxy.py:182` `response_body = resp.content.decode('utf-8')` — whole
  response buffered, then returned via `JSONResponse`/`Response`

Consequences:

1. **Streaming chat is broken.** With `stream: true` the upstream returns
   `text/event-stream`; `resp.content` blocks until the *entire* generation
   completes, so the client sees no tokens until the end and latency becomes
   total-generation-time.
2. **gzip / binary / non-UTF-8 responses 500** at the hard `.decode('utf-8')`.
3. No WebSocket path.

## Fix

Response side now streams via `httpx` `client.stream()` + `StreamingResponse`:

- **Textual, uncompressed, under-cap** bodies are PII-restored incrementally
  through a new **additive** `StreamRestorer` (in `pii_scrubber.py`) that owns
  an incremental charset decoder (multibyte chars split across chunks) and a
  **bounded** hold-back buffer, so a `<PII_…>` token straddling a chunk/SSE
  boundary is never emitted half-restored. SSE framing is preserved.
- **Non-text / `Content-Encoding` / over-size** bodies pass through
  byte-for-byte via `aiter_raw()` with no UTF-8 decode.
- `Content-Type`/charset parsed and honored; hop-by-hop headers stripped.
- **Authenticated WebSocket passthrough** lane (lazy `websockets` import — HTTP
  path unaffected when absent). The handshake is authenticated **before**
  `accept()` and before any upstream socket: bearer token from
  `Authorization`, `?token=`, or `Sec-WebSocket-Protocol`, constant-time
  compared (UTF-8 bytes, non-ASCII-safe) against the **same `SHIELD_API_KEY`**
  the HTTP lane uses (shared helper, so they can't drift); missing/invalid →
  `close(1008)`, no upstream, `TARGET_API_KEY` never attached.

The request body stays buffered+scrubbed (it is the prompt; partial scrubbing
would leak PII upstream) but a non-UTF-8 request body is now forwarded verbatim
instead of 500ing. **Auth, `/health`, `/stats`, privacy headers, error
sanitization and timeouts are byte-for-byte unchanged.** `pii_scrubber.py`
changes are purely additive — existing `scrub()`/`restore()`/detection regexes
are untouched.

## Tests

`extensions/services/privacy-shield/tests/test_streaming_proxy.py` (new):
streaming (not buffered), PII token split across 2 and 3 SSE chunk boundaries,
gzip + non-UTF-8 binary passthrough, non-UTF-8 request passthrough, WebSocket
relay, **WS handshake auth (no-token / invalid-token / non-ASCII-token all
rejected with `1008` and no upstream; valid token round-trips)**. Full
privacy-shield suite **49 passed**, no regression in `test_pii_scrubber.py` /
`test_proxy_auth.py`.

## Notes

Part of a 4-PR set addressing #1268–#1271 (independent, disjoint files). I'm
not a collaborator so I can't self-assign — I'd like to own this and will
revise to maintainer feedback.

### DIFF
--- a/dream-server/extensions/services/privacy-shield/pii_scrubber.py
+++ b/dream-server/extensions/services/privacy-shield/pii_scrubber.py
@@ -4,6 +4,7 @@ M3: API Privacy Shield - Core PII Scrubber
 Detects and replaces PII with tokens, restores on reverse.
 """
 
+import codecs
 import re
 import hashlib
 import secrets
@@ -113,6 +114,17 @@ class PIIDetector:
             restored = restored.replace(token, original)
         return restored
 
+    def max_token_len(self) -> int:
+        """Length of the longest active PII token (0 if none).
+
+        The streaming restorer uses this to bound the hold-back buffer so a
+        token straddling a chunk boundary is never emitted half-restored,
+        while the buffer can never grow without limit on hostile input.
+        """
+        if not self.pii_map:
+            return 0
+        return max(len(token) for token in self.pii_map)
+
     def get_stats(self) -> Dict:
         """Return statistics about detected PII."""
         return {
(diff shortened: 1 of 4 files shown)


Lightheartdevs:
Good direction. Streaming rather than buffering is exactly what privacy-shield needs, and the SSE/token-boundary tests are useful. I would hold this as draft until two issues are addressed.

Findings:

1. **CI is red on Ruff.** The current failures are straightforward:
   - `extensions/services/privacy-shield/proxy.py:14` imports `fastapi.Response` but does not use it.
   - `extensions/services/privacy-shield/tests/test_streaming_proxy.py:44` imports `StreamingResponse` but does not use it.

2. **The oversized-text fallback may not continue the same upstream stream.** In `body_iter()`, once `seen > RESTORE_MAX_BYTES`, the code yields the current chunk and then starts a fresh `raw_chunks()` loop at `extensions/services/privacy-shield/proxy.py:386`. That means the original `raw_chunks()` / `upstream.aiter_raw()` iterator is abandoned mid-stream, then the code attempts to iterate the same httpx response a second time. httpx response streams are single-consumption, so this path is likely to drop/log the remainder for large text responses. Please keep one iterator alive for the whole response, or add a specific low-`SHIELD_RESTORE_MAX_BYTES` regression test proving the remainder is forwarded.

After those, this looks like a strong PR: the binary/gzip passthrough and WebSocket auth-gate coverage are the right tests for this surface.

nolanmak:
Both items addressed — ready for re-review.

1. **Ruff** — resolved in `a921d7532` (removed the unused `fastapi.Response` and test `StreamingResponse` imports). That commit landed *after* the revision you reviewed; CI Ruff is green now.
2. **Oversized-text fallback re-consuming the stream** — fixed in `2f681085`. `body_iter()` now binds the `raw_chunks()` async generator **once**; on the oversized-text cutover it continues draining that *same* iterator (emitting the chunk that crossed the cap first) instead of starting a fresh `raw_chunks()` loop, so `upstream.aiter_raw()` is consumed exactly once and the post-cutover remainder is forwarded byte-complete. Added `await chunks.aclose()` in `finally`. New regression test: low `SHIELD_RESTORE_MAX_BYTES`, multi-chunk text body exceeding it mid-stream, asserts the client receives the body byte-for-byte — it fails on the prior code and passes now. Suite **50 passed**, ruff clean.

PR #1278, opened 2026-05-19
fix(select-model): exclude coder-next on any unified-memory host, not just Spark aarch64

Lightheartdevs:
## Summary

Follow-up to #1265 (Strix Halo qwen3-coder-next substitution). That PR fixed `installers/lib/tier-map.sh`, but phase 02-detection.sh routes via `scripts/select-model.py` (the "capabilities oracle") which had a Spark-aarch64-specific override that didn't catch Strix.

## Evidence (2026-05-19 fleet run round 4, after #1265 merged)

```
$ ssh strix-halo 'grep "Capabilities override\|Model:" /home/michael/dream-server/.../install.log'
[INFO] Capabilities override detection: backend=amd, memory=unified, tier=SH_LARGE
[INFO]   Model: qwen3-coder-next        ← still wrong
```

bootstrap-upgrade.sh then began downloading the 48.5GB coder-next that would not have worked on Strix Halo anyway (verified pathology: all-`?` tokens on unified-memory backends — see in-source notes in tier-map.sh).

## Fix

The real signal is `memory_type == unified`, not arch or tier. Extend `arch_policy_model()` in `select-model.py` to fire on `profile=qwen + memory_type=unified` for any tier, returning the same Qwen3.6-35B-A3B-UD-Q4_K_M substitution under a new policy tag (`unified-memory-coder-next-a3b-v1`, distinct from `spark-aarch64-nv-ultra-a3b-v1`).

`arch_policy_model()` now returns `(model, policy_tag)` so callers can distinguish the two routes; `arch_policy_reason()` renders different rationales for unified-memory vs Spark aarch64.

## Test plan

- [x] `python3 -m py_compile scripts/select-model.py` clean
- [x] `bash tests/test-tier-map.sh` — **110 passed, 0 failed** (added 3 SH_LARGE assertions)
- [x] amd64 NV_ULTRA (Tower2) still selects coder-next (discrete memory, no policy tag added)
- [x] arm64 NV_ULTRA (Spark) still selects 35B-A3B with the spark-aarch64 policy tag (unchanged)
- [x] amd unified SH_LARGE (Strix) now selects 35B-A3B with the new unified-memory policy tag
- [ ] Manual: re-install on strix-halo → install.log reports "Model: qwen3.6-35b-a3b" and bootstrap-upgrade downloads the ~22GB model

### DIFF
--- a/dream-server/scripts/select-model.py
+++ b/dream-server/scripts/select-model.py
@@ -21,6 +21,14 @@ VRAM_FIT_TOLERANCE_GB = 0.25
 POLICY = "context-aware-largest-capable-general-v1"
 SPARK_AARCH64_POLICY = "spark-aarch64-nv-ultra-a3b-v1"
 SPARK_AARCH64_MODEL_ID = "qwen3.6-35b-a3b-ud-q4"
+# Unified-memory hosts (Strix Halo SH_LARGE, future AMD/NV unified-memory
+# tiers) hit the same coder-next correctness pathology as Spark aarch64.
+# Until upstream fixes coder-next on unified-memory backends, route the
+# qwen profile to the same 35B-A3B substitution used for Spark — same
+# model id, separate policy tag so the recommendation_reason is honest
+# about why the substitution fired.
+UNIFIED_MEMORY_POLICY = "unified-memory-coder-next-a3b-v1"
+UNIFIED_MEMORY_MODEL_ID = SPARK_AARCH64_MODEL_ID
 
 
 def normalize_key(value: Any) -> str:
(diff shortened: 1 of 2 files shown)


Lightheartdevs:
Re-audited this one. Direction is right, but I would not merge it as-is.

Finding:

1. **The unified-memory override is too broad and forces an oversized model onto Apple/low-memory unified hosts.** `arch_policy_model()` now returns the A3B override whenever `profile=qwen` and `memory_type=unified`, before checking whether the normal ranked candidate was actually `qwen3-coder-next` or whether the substitute fits the detected capacity (`scripts/select-model.py:285-420`). That catches more than SH_LARGE. Local probes on this PR show:

   - Apple tier 1, 16 GB unified: selects `qwen3.6-35b-a3b`, reason says it needs about `34.62GB` while only `8.8GB` unified system memory is usable.
   - Apple tier 2, 32 GB unified: selects the same 35B model, reason says `34.62GB` required vs `17.6GB` usable.
   - SH_COMPACT, 64 GB unified: also selects the 35B substitute instead of the existing SH_COMPACT tier-map target.

   The fix should make the override conditional on the model that would otherwise be selected/ranked being the excluded coder-next entry, or restrict the policy to the specific tiers/capacity envelopes where coder-next is actually the candidate. Please add regression tests for Apple 16 GB / 32 GB unified and SH_COMPACT so the selector keeps fitting smaller unified-memory machines.

Checks I ran:
- `python -m py_compile dream-server/scripts/select-model.py` passed.
- `bash tests/test-tier-map.sh` passed (`110 passed, 0 failed`).
- `git diff --check origin/main...refs/remotes/pr/1278` passed.
- Manual selector probes exposed the blocker above.

So: good bug target, but the predicate needs narrowing before this is safe to merge.

Lightheartdevs:
Fixed the blocker in 66c5faac. The selector now ranks capacity-safe candidates first, then applies the unified-memory A3B substitution only when the ranked model would be qwen3-coder-next. That keeps small unified hosts on normal fit-based recommendations while preserving the SH_LARGE/Spark workaround.

Added regressions for:
- 16GB Apple unified qwen staying on qwen3.5-9b with base policy
- 32GB Apple unified qwen staying on phi-4 with base policy
- SH_COMPACT unified selecting A3B only as a normal catalog fit, not via override

Validated locally:
- python -m py_compile dream-server/scripts/select-model.py
- python -m ruff check dream-server/ --select E,F,W --ignore E501,E701,E731,E741,E402
- Git Bash: bash tests/test-tier-map.sh (122 passed, 0 failed)
- Git Bash: bash -n tests/test-tier-map.sh
- git diff --check

GitHub checks are green on 66c5faac. I attempted to approve, but GitHub rejected it because this account is considered the PR author: "Review Can not approve your own pull request". So code-wise this is ready; it just needs an independent maintainer approval or an intentional admin bypass.

PR #1298, opened 2026-05-20
fix(windows): harden install compose preflight

gabsprogrammer:
## Summary

Fixes two Windows first-time install failures reported by users:

- Docker Desktop file-sharing preflight could blame File Sharing when the real issue was that the Alpine probe image was not downloaded.
- `docker compose` could complain that `.env` was missing because the installer generated `.env` in the installed `dream-server` directory, but Compose could still run from the cloned `DreamServer` source directory.

## What changed

- Pins the Windows bind-mount probe image to `alpine:3.20`.
- Pulls the probe image before running the bind-mount test.
- Shows a clear recovery message when Docker cannot download the probe image:
  - `docker pull alpine:3.20`
- Keeps Docker Desktop File Sharing errors separate from image/network errors.
- Validates relative compose file paths against the installed DreamServer directory.
- Runs install-time `docker compose build` and `docker compose up` from `$installDir`, where `.env` and compose files actually live.
- Logs the Compose working directory for easier support.
- Updates the Windows troubleshooting guide.
- Extends the Windows compose failure regression test.

## Why

A user can have the correct folder shared in Docker Desktop and still fail the bind-mount probe if Docker has not pulled Alpine yet. The previous error only pointed them to File Sharing, which sent them to the wrong fix.

Also, Windows users often run `.\install.ps1` from the cloned repo, while the installer creates the final runtime stack under `C:\Users\<user>\dream-server`. Compose should run from that installed directory so `.env` resolution is deterministic.

## Validation

- `bash -n dream-server/tests/test-windows-compose-failure-report.sh`
- `bash dream-server/tests/test-windows-readiness-summary.sh`
- PowerShell parser check for:
  - `dream-server/installers/windows/phases/01-preflight.ps1`
  - `dream-server/installers/windows/install-windows.ps1`
- PSScriptAnalyzer error-level check for the same PowerShell files
- `git diff --check`

### DIFF
--- a/dream-server/docs/WINDOWS-TROUBLESHOOTING-GUIDE.md
+++ b/dream-server/docs/WINDOWS-TROUBLESHOOTING-GUIDE.md
@@ -13,6 +13,8 @@
 | "Docker not found" | Install Docker Desktop first |
 | "GPU not detected" | Update NVIDIA drivers |
 | "Installation hangs" | Check internet connection, wait 30 min for model download |
+| "Docker Desktop cannot bind-mount..." | Add `C:\Users\<you>\dream-server` to Docker Desktop -> Settings -> Resources -> File Sharing |
+| "Docker could not download alpine..." | Run `docker pull alpine:3.20`, then re-run the installer |
 
 Do not use "Run as administrator" for the Dream Server installer unless you
 are intentionally accepting admin-owned files under your user profile. The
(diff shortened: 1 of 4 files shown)


Lightheartdevs (approved):
Reviewed the Windows preflight and compose working-directory changes. No blocking issues found.

What I checked:
- The Docker Desktop file-sharing probe now separates probe-image pull/network failures from actual bind-mount/file-sharing failures, which should reduce misleading support reports.
- Relative compose `-f` entries are validated against `$installDir`, so validation matches the runtime compose context rather than the source checkout by accident.
- The local image rebuild and `docker compose up` path runs under `$installDir` and restores PowerShell state in `finally`.
- CI is green and `git diff --check` is clean.

Merge note: this overlaps with #1303 in `installers/windows/install-windows.ps1`, so expect a small merge-order/rebase touch if both land.

Lightheartdevs (approved):
﻿Reviewed the Windows compose/file-sharing preflight changes. Looks safe to merge.

Notes:
- Separating Alpine image pull failures from Docker Desktop file-sharing failures is the right support move.
- Running build/up from `$installDir` fixes the `.env`/relative-path class of Windows install failures.
- The added regression checks cover the important strings and install-dir behavior.

No blocking findings.

Lightheartdevs (approved):
Audited for install impact/security. This is a focused Windows installer hardening change: it separates Docker Desktop file-sharing failures from missing probe-image/network failures, validates compose files relative to the install directory, and runs compose build/up from the runtime install dir so .env resolution is deterministic. I checked for secret exposure and service-surface expansion; none found.

Verified:
- `bash -n dream-server/tests/test-windows-compose-failure-report.sh`
- Windows PowerShell parser check for `install-windows.ps1` and `01-preflight.ps1`
- GitHub checks are green

Safe to merge.

Issue #1318, opened 2026-05-21
feat: dedicated mobile-first portal with voice (replaces upstream Hermes dashboard as the post-redemption surface)


Lightheartdevs:
## Summary

The owner-card → Hermes redemption flow (shipped in #1315 / commit `aed6fb5`) successfully authenticates a phone scan and drops the holder into Hermes. **But the surface it lands on is the stock upstream Nous Research Hermes dashboard** — designed for agent developers configuring tools, skills, and scheduled jobs — not a consumer chat app. It is dense, button-heavy, desktop-first, and on a phone it is unusable for the persona the QR cards are meant to serve.

We need a dedicated **mobile-first portal** that uses Hermes as the agent backend and exposes a consumer-grade chat-and-voice UI on top of it. Owner-card redemption should land here, not at the upstream dashboard.

Reported by the operator after redeeming a real owner card from a phone:

> "It's the traditional messy Hermes interface just on my phone, is meant for typed chat and doesn't feel very mobile friendly or intuitive and there's no easy way to operate in voice mode or talk to it and have it talk back."

## Current state (as of `aed6fb5`)

`dream-server/extensions/services/dashboard-api/routers/magic_link.py:418`:

```python
def _hermes_url(url_mode: str = "auto") -> str:
    """Where a successful Hermes redemption redirects."""
    if _use_public_url(url_mode):
        return f"{_public_base()}/hermes"
    return f"http://hermes.{_device_name()}.local"
```

`hermes.<device>.local` → dream-proxy → hermes-proxy (Caddy, auth-gates on cookie) → `dream-hermes:9119` → upstream Nous Research dashboard (image pinned to `nousresearch/hermes-agent:sha-dd0923bb...`).

The upstream dashboard is **not designed to be modified** — it's a SHA-pinned third-party image, and per `extensions/services/hermes/compose.yaml` and `docs/HERMES.md` "How to bump the pin," changing it is a deliberate review-and-smoke-test pass. So this isn't a "restyle the existing UI" problem; it's a "we need to stop sending consumers to that UI" problem.

## Voice plumbing inventory (what exists already, not wired together)

Each of these is a running service today; **none** of them is exposed in a unified mobile UX:

| Service | What it does | Endpoint shape | Where it lives |
|---|---|---|---|
| `dream-whisper` | Speech-to-text (Whisper large-v3-turbo) | OpenAI-compatible `POST /v1/audio/transcriptions` | `extensions/services/whisper/` |
[...]

PR #1321, opened 2026-05-21
fix(dream-talk): stream chat replies via SSE + fix bridge WS lifecycle (#1318 follow-up)

Lightheartdevs:
## Summary

Reported by the operator after redeeming a real owner card and texting Dream Talk: *"It says it's thinking but I don't think it is."* Investigation surfaced three layered bugs in the chat path. This PR ships fixes for all three.

| Bug | Where | Effect |
|---|---|---|
| **1. Bridge opens a new WebSocket per call** (root cause) | `hermes_bridge.submit_prompt` opened a fresh WS for `session.create`, closed it, then opened *another* fresh WS for `prompt.submit` | Hermes scopes streaming event delivery to the WS that owns the session. The second WS's `prompt.submit` returns `{"status":"streaming"}` but **the events fire to the already-closed first WS**. The bridge waits forever on the second WS, eventually 502'ing at the 180 s timeout — **even though Hermes generated the full reply seconds earlier**. |
| **2. Endpoint is synchronous, not streaming** | `/api/talk/message` blocks until the full reply is assembled, then returns JSON | On a cold first message (16 k-token system prompt), Hermes prompt processing takes ~60 s before any token streams back. The UI sees a stalled "thinking" spinner with no feedback. |
| **3. nginx `/api/` timeout is too tight** | Dashboard nginx had `proxy_read_timeout 180s` for the whole `/api/` block | Even the synchronous endpoint can legitimately need >180 s on cold first-call paths. Streaming responses need `proxy_buffering off` so frames pass through immediately. |

## Reproduction (pre-fix)

```
$ curl -sN -X POST http://127.0.0.1:3002/api/talk/message \
    -H 'Cookie: dream-session=<valid>' -H 'Content-Type: application/json' \
    -d '{"text":"reply PONG only"}' -w '\nHTTP=%{http_code} time=%{time_total}\n'
{"detail":"Hermes did not finish the response."}
HTTP=502 time=180.014547

# Meanwhile, in dream-hermes logs, the response *completed* in 1.7 s:
[dashboard] event {type: "message.complete", session_id: "67fb13bd",
                   payload: {text: "PONG"}, usage: {input: 16412, output: 269, ...}}
```

A minimal reproducer (run inside `dream-dashboard-api`) confirms the per-WS scoping:

```python
# WS-A: create session, then CLOSE
async with await ws_connect(...) as ws_a:
    sid = await session_create(ws_a)   # → 0d1b1a76

# WS-B (fresh): prompt.submit with sid from WS-A
async with await ws_connect(...) as ws_b:
    await ws_b.send_json({...prompt.submit..., "session_id": sid, ...})
    # → {"status":"streaming"}     ← accepted!
    # … then nothing. Zero events for 60 s. Eventually times out.
```

### 1. `hermes_bridge.py`: single-WS lifecycle

* New async generator `stream_prompt(session_key, text)` that opens **one** WS, runs `session.create` AND `prompt.submit` on it, consumes events to `message.complete`, and yields `{session, delta, complete}` events to its caller.
* `submit_prompt` becomes a thin wrapper that consumes `stream_prompt` and returns the final `HermesReply`. Existing callers / tests that want a single JSON reply keep working.
* The legacy `_SESSION_IDS` cross-call cache is no longer used for streaming (event delivery is per-WS, so a cached session_id from a dead WS is useless). `ensure_session` is kept as a compat shim because the existing JSON `/api/talk/session` endpoint and tests call it; conversational continuity is provided by Hermes's own agent memory layer rather than session reuse.

### 2. `/api/talk/message/stream` — new SSE endpoint

* Returns `text/event-stream` with one `data:` line per bridge event.
* Frame types: `session`, `delta` (repeats), `complete`, `error` (on failure), `done` (always last). Same shape used in tests so a future regression is loud.
* The synchronous `/api/talk/message` endpoint stays, for non-browser callers and tests. It now uses the fixed bridge too.

### 3. `nginx.conf`: dedicated `^~ /api/talk/` block

```nginx
location ^~ /api/talk/ {
    proxy_pass http://dashboard-api:3002;
    ...
    proxy_read_timeout 600s;
    proxy_send_timeout 600s;
    proxy_buffering off;
    proxy_cache off;
    proxy_set_header X-Accel-Buffering no;
}
```

* `^~` so it takes priority over the generic `/api/` block.
* `proxy_buffering off` + `X-Accel-Buffering: no` so each SSE frame is forwarded to the client immediately rather than batched.
* Bumps `proxy_read_timeout` from 180 s → 600 s so first-message cold prompts have headroom even if the SPA still falls back to the JSON endpoint.

### 4. Frontend: stream-aware `sendText`

* `DreamTalk.jsx` now hits `/api/talk/message/stream`, reads the response as a `ReadableStream`, buffers partial SSE frames across transport chunks, and appends each `delta` into the assistant bubble live.
* Errors received via SSE `error` frame become the assistant's error message — the same UX as before, but now we see the actual `detail` from Hermes instead of an opaque `fetch` failure.

## Test coverage

* **Backend** (`extensions/services/dashboard-api/tests/test_talk.py`, +6 tests, all 11 pass):
  * `test_talk_message_stream_emits_session_then_deltas_then_complete`
  * `test_talk_message_stream_emits_error_frame_and_done_on_bridge_failure`
  * `test_talk_message_stream_emits_503_when_hermes_unavailable`
  * `test_talk_message_stream_requires_session`
  * `test_talk_message_stream_validates_input`
  * `test_talk_message_stream_sets_unbuffered_headers` (regression guard against losing `X-Accel-Buffering` / `Cache-Control`)
* **Frontend** (`extensions/services/dashboard/src/pages/DreamTalk.test.jsx`, +2 new + existing updated, all 6 pass):
  * Updated `renders a mobile text portal and sends a message` to expect SSE frames instead of JSON.
  * New `accumulates SSE deltas across split chunks` — frames split across `read()` boundaries are re-assembled correctly.
  * New `surfaces SSE error frame as an assistant error`.

## Fleet verification

Patched files deployed and containers rebuilt + recreated on all three Linux fleet hosts. Issued an admin cookie and hit `POST /api/talk/message/stream` through the dashboard nginx (`Host: talk.<device>.local`, the real path the phone takes):

| host | model | response | elapsed |
|------|-------|----------|---------|
| tower2 | qwen3-coder-next | full SSE frame sequence (`session, 2× delta, complete, done`), reply = "HELLO" | 4 s |
| strix-halo | Qwen3.6-35B-A3B (Lemonade ROCm) | full sequence, reply = "HELLO" | 2 s |
| spark | Qwen3.6-35B-A3B (llama.cpp GB10 aarch64) | full sequence, reply = "HELLO" | 9 s |

Synchronous endpoint `/api/talk/message` (still used by tests + non-browser callers) also returns correctly now: strix returns `{"text":"PING","status":"complete"}` in 1.7 s on a short prompt — versus pre-fix where it 502'd at exactly 180 s.

## Risk / blast radius

* `submit_prompt` keeps its old signature and return shape. The non-streaming JSON endpoint `/api/talk/message` is unchanged from the caller's perspective; only the bridge implementation underneath it changed.
* The streaming endpoint is **new** — adding it can't break existing clients.
* nginx changes are scoped to a new `^~ /api/talk/` block; the existing `/api/` block is untouched, so non-Talk endpoints behave identically.
* `_SESSION_IDS` cache is no longer load-bearing; the only call that still touches it is `ensure_session`, which exists for the `/api/talk/session` legacy endpoint and the existing tests. Conversational memory now lives in Hermes's agent layer (which it always did — the `session_id` reuse was a UX-continuity hint, not a memory primitive).

### DIFF
--- a/dream-server/bin/dream-mdns.py
+++ b/dream-server/bin/dream-mdns.py
@@ -237,6 +237,13 @@ def _build_services(env: dict[str, str], device_name: str, ip: str) -> list[Serv
         # upstream container isn't running, which is the right failure
         # mode for an optional extension.
         ("hermes",    "Hermes Agent (via proxy)"),
+        # talk.<device>.local is the owner-card redemption target for the
+        # mobile portal added in #1319. magic_link.py's _talk_url() points
+        # phones at this hostname; dream-proxy's Caddy block routes it to
+        # the dashboard container's /talk route. Without an A record here
+        # the redirect lands the phone on an unresolvable host and the
+        # owner experience stalls on a white screen.
+        ("talk",      "Dream Talk (mobile owner portal)"),
     )
     for suffix, label in subdomain_routes:
         server = f"{device_name}.local." if suffix == "root" else f"{suffix}.{device_name}.local."
--- a/dream-server/docs/MDNS.md
+++ b/dream-server/docs/MDNS.md
@@ -25,8 +25,9 @@ Two flavors:
 | `auth.<device>.local` | LAN IP, port 80 | proxy → dashboard-api (magic-link redemption) |
 | `api.<device>.local` | LAN IP, port 80 | proxy → dashboard-api (admin `/api/*`) |
 | `hermes.<device>.local` | LAN IP, port 80 | proxy → hermes-proxy (when enabled) |
+| `talk.<device>.local` | LAN IP, port 80 | proxy → dashboard `/talk` (mobile owner portal, #1319) |
 
-All six names resolve to the same IP — the device's LAN address. Routing happens at the dream-proxy by Host header (see `docs/DREAM-PROXY.md`). The bare `<device>.local` redirects to chat for the friendliest landing.
+All seven names resolve to the same IP — the device's LAN address. Routing happens at the dream-proxy by Host header (see `docs/DREAM-PROXY.md`). The bare `<device>.local` redirects to chat for the friendliest landing.
 
 **Direct-port SRV records (back-compat for service discovery):**
 
--- a/dream-server/extensions/services/dashboard-api/hermes_bridge.py
+++ b/dream-server/extensions/services/dashboard-api/hermes_bridge.py
@@ -4,6 +4,14 @@ Dream Talk deliberately does not expose Hermes's browser session token to the
 phone. The dashboard-api fetches that token from the internal Hermes HTML page,
 opens the JSON-RPC WebSocket on the Docker network, and returns only simplified
 chat results to the mobile portal.
+
+Architectural note: Hermes scopes streaming event delivery to the WebSocket
+that owns the session. If we open WS-A for ``session.create`` and then open a
+fresh WS-B for ``prompt.submit``, Hermes accepts the submit (returns
+``{"status":"streaming"}``) but the streaming events fire to WS-A — which we
+already closed. The bridge would then wait forever for events that never
+arrive and 502 at the request timeout. So a single submit_prompt / stream_prompt
+call MUST do both create-session and submit-prompt on the same WS.
 """
 
 from __future__ import annotations
@@ -15,7 +23,7 @@ import logging
 import os
 import re
 from dataclasses import dataclass
-from typing import Any
+from typing import Any, AsyncIterator
 
 import aiohttp
 
@@ -24,10 +32,6 @@ logger = logging.getLogger(__name__)
 TOKEN_RE = re.compile(r'window\.__HERMES_SESSION_TOKEN__\s*=\s*"([^"]+)"')
 DEFAULT_HERMES_URL = "http://dream-hermes:9119"
 DEFAULT_TIMEOUT_SECONDS = 180
-
-_SESSION_IDS: dict[str, str] = {}
-_SESSION_LOCKS: dict[str, asyncio.Lock] = {}
-_SESSION_LOCKS_GUARD = asyncio.Lock()
 
 
 class HermesBridgeError(RuntimeError):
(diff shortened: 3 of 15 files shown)


Lightheartdevs:
Audit notes before merge:

- Reproduced and fixed the failing frontend check: the SSE test helper was referencing Node Buffer in browser/jsdom test code. Replaced it with Uint8Array concatenation.
- Reviewed the Dream Talk streaming path end to end: same-WebSocket Hermes session/create + prompt submit, SSE frame mapping, nginx unbuffered /api/talk/ routing, frontend chunk buffering, markdown rendering, and owner-session auth boundaries.
- Added a small hardening patch while reviewing: client disconnects now cancel the upstream Hermes bridge without trying to emit a final SSE frame into a closed response, and the mobile UI aborts in-flight streams on unmount.
- Reviewed the flagged installer change in dream-server/installers/phases/11-services.sh: AMD/Lemonade now routes Hermes through LiteLLM with LITELLM_KEY, while non-AMD paths keep the existing llama-server direct route.

Verification run locally:
- python -m pytest extensions/services/dashboard-api/tests/test_talk.py extensions/services/dashboard-api/tests/test_magic_link.py -> 61 passed
- npm test -- --run src/pages/DreamTalk.test.jsx src/App.test.jsx -> 12 passed
- npm run lint -> 0 errors, existing warnings only in useVoiceAgent.js
- npm run build -> passed
- python tests/test_mdns_subdomains.py -> OK
- Git Bash: tests/test-installer-context-parity.sh -> 29 passed
- Git Bash: tests/test-litellm-amd-auth-enforced.sh -> 4 passed

GitHub Actions on the refreshed head are green across frontend, api, lint/type, integration smoke, and matrix smoke. I’m treating this as safe to merge via admin because the remaining block is review-required on a PR authored by this account plus the sensitive-file notice that has now had manual review.

PR #1323, opened 2026-05-21
perf(hermes-bridge): pool one Hermes session per cookie (closes #1322)

### LINKED ISSUE #1322: perf(hermes-bridge): reuse Hermes session per cookie to warm prompt cache (Dream Talk follow-up to #1321)
## Summary

Dream Talk's per-message latency on Lemonade ROCm and other strict OpenAI-compat backends is dominated by **system-prompt prefill**, not by output generation. Every Dream Talk request opens a new Hermes session via `session.create`, which means llama-server reprocesses the full ~16 k-token agent system prompt (persona + tool catalog + skill list) from scratch before generating the first output token. On strix that adds 30–60 s to the *first* round-trip of every message — for a one-word prompt like "hey" the model spends ~60 s prefilling, ~1 s generating, then we tear the session down again so the next prompt has to repeat the same prefill.

We should hold one Hermes session (and its WebSocket) open per phone, so subsequent messages reuse the warm prompt cache.

## Current architecture (post-#1321)

`extensions/services/dashboard-api/hermes_bridge.py` `stream_prompt(session_key, text)`:

1. Acquire per-`session_key` `asyncio.Lock` (prevents same-cookie pileup).
2. Open a fresh `aiohttp` `ClientSession`.
3. `_connect_ws()` → opens fresh WS to `dream-hermes:9119`.
4. `_create_session_on_ws()` → JSON-RPC `session.create`, returns a Hermes `session_id`.
5. Send `prompt.submit` on that WS, consume events to `message.complete`.
6. Exit `async with ws:` — close WS, drop session_id, drop lock.

This pattern was the right *correctness* fix in #1321 (Hermes scopes streaming events per-WS, so `session.create` on WS-A and `prompt.submit` on WS-B silently loses events). But it costs us a 16 k-token prefill on every message because Hermes spins up a fresh session each time.

Measured on strix (Qwen3.6-35B-A3B via Lemonade ROCm):

| call | elapsed | notes |
|---|---|---|
| first message ("hey") | ~60 s | full system-prompt prefill |
| follow-up message in the same call | n/a — we tear down |
| identical "hey" from a second call | ~60 s | new session, same prefill |

## Target architecture

Hold one `HermesConnection` per `session_key` in a process-wide pool:

```python
@dataclass
class HermesConnection:
    ws: aiohttp.ClientWebSocketResponse
    http_session: aiohttp.ClientSession
    session_id: str
    last_used: float
    lock: asyncio.Lock
```

`stream_prompt(session_key, text)`:

1. `conn = await _get_or_create_connection(session_key)` — pool lookup, create if missing or dead.
2. `async with conn.lock:` — serialize prompts for one cookie.
3. Send `prompt.submit` with `conn.session_id` on `conn.ws`, consume events to `message.complete`, yield to caller.
4. Update `conn.last_used`. **Don't close `ws`.**

Lifecycle:

* **First message**: 16 k-token prefill (unavoidable cold start).
* **Subsequent messages on same cookie**: prefill cache hit, response in 2–5 s.
* **Idle expiry**: background task closes any connection idle >5 min (configurable). Frees Hermes resources + the llama-server slot's KV cache eventually rolls out naturally.
* **Dead WS detection**: if `conn.ws` raises on send / receive, drop the pool entry and treat as cold start (create a new connection on the same call so the user doesn't see an error).
* **Graceful shutdown**: FastAPI lifespan hook closes all connections.

## Edge cases worth covering up front

* **Hermes container restart**: stale WS in pool → first send fails → bridge catches, drops pool entry, retries once with a fresh connection.
* **Session expired on Hermes side**: same handling as dead WS — drop + retry.
* **Concurrent same-key prompts**: per-conn `asyncio.Lock` serializes them on the same WS (Hermes likely can't multiplex two prompts on one session).
* **Concurrent different-key prompts**: independent connections, no contention beyond llama-server slot capacity.
* **Memory leak / pool growth**: idle sweeper is non-negotiable. Without it a fleet of one-time visitors leaves dead WSes pinned forever.

## Test plan

1. **Cache-warming benchmark** (the headline metric):
   - Send `"hi"` from a fresh session_key → measure end-to-end (should be slow, ~60 s cold).
   - Immediately send `"and what's 2+2?"` on the same session_key → measure (should be 2–5 s).
   - Today the second call is also ~60 s; success criterion is **second call < 10 s** on strix.

2. **Concurrent different-key**: two clients send simultaneously, both finish within reasonable time (no serialization between cookies).

3. **Concurrent same-key**: queued serially (the per-key lock already does this; verify it still does after refactor).

4. **Dead-WS recovery**: kill the Hermes container mid-conversation; next send transparently reconnects (one allowed retry).

5. **Idle sweep**: send a message, wait past idle threshold, verify pool entry is gone and a subsequent send creates a new connection cleanly.

## Out of scope (deliberately)

* **Cross-process pooling** (workers in different uvicorn processes don't share the pool). Acceptable trade-off — Dream Server runs single-worker uvicorn for dashboard-api today.
* **Hermes session pinning to a specific llama-server slot**: would maximize KV cache hit rate further but requires patching upstream Hermes/llama-server. Out of scope.
* **Per-cookie conversation memory**: Hermes already has an agent-memory layer that persists across sessions; session reuse here is about prompt-cache warmth, not memory continuity.

## Acceptance criteria

* Same-cookie "hey" → "and how about today's weather" → "thanks": first call ~60 s on strix, calls 2 and 3 each < 10 s.
* All existing 14 backend tests + 7 frontend tests in #1321 still pass.
* New backend tests cover: session reuse, dead-WS recovery, idle sweep, concurrent same-key serialization.
* No measurable regression on tower2 / spark (native llama.cpp paths) — those already have warm caches via llama-server's own slot caching.

## Notes

This is a clean follow-up to #1321 (it builds on the `stream_prompt` async generator surface that PR introduced). PR will branch from #1321's tip; once #1321 merges this rebases cleanly.

Lightheartdevs:
Closes #1322.

## Summary

Keeps one Hermes WebSocket + Hermes session per `dream-session` cookie so Dream Talk can reuse the same agent session across messages. That keeps llama-server / Lemonade prompt cache warm after the first owner-card message instead of paying the 16k-token system-prompt prefill on every turn.

Also tightens the Dream Talk Hermes tool surface so consumer chat prefers the supported web/search path and cannot stall on terminal/browser tool choices.

## What changed

- Adds a pooled `_HermesConnection` in `extensions/services/dashboard-api/hermes_bridge.py`, keyed by the hashed `dream-session` cookie.
- Serializes same-cookie prompts with a per-connection lock while allowing different cookies to use independent pooled connections.
- Adds idle sweeping with `DREAM_TALK_IDLE_EXPIRY` (default 300s) plus FastAPI lifespan shutdown cleanup.
- Retries only the safe stale-socket case where `send_str()` fails before `prompt.submit` is accepted; receive-side drops after submit now surface as errors instead of duplicating prompts/tool calls.
- Prevents the idle sweeper from closing an actively locked connection during a long first turn.
- Configures Hermes with `agent.disabled_toolsets: [terminal, browser]` for Dream Talk and sets the fallback terminal timeout to 30s via `terminal.timeout` + `TERMINAL_TIMEOUT=30`.
- Updates the Dream persona/tool guidance so weather/news/current-info requests go to `web_search` first.

## Verification

Local:
- `python -m pytest extensions/services/dashboard-api/tests/test_talk.py extensions/services/dashboard-api/tests/test_magic_link.py` -> 66 passed
- `python -m ruff check extensions/services/dashboard-api/hermes_bridge.py extensions/services/dashboard-api/tests/test_talk.py extensions/services/dashboard-api/main.py` -> passed
- `tests/test-installer-context-parity.sh` via Git Bash -> 29 passed
- `tests/test-litellm-amd-auth-enforced.sh` via Git Bash -> 4 passed
- Confirmed pinned Hermes source supports `agent.disabled_toolsets` and consumes `TERMINAL_TIMEOUT`.

GitHub Actions on the rebased `main` target are green across frontend, api, Python lint/type, shell lint, compose validation, integration smoke, env validation, macOS/Linux smoke, and the distro matrix.

## Risk

The Dream Talk streaming wire format is unchanged. The main behavioral risk is session lifetime/resource management in the new pool, covered by tests for same-key serialization, different-key parallelism, dead-WS recreation, pre-submit retry, no post-submit retry, active prompt sweeper protection, idle eviction, and shutdown cleanup.

### DIFF
--- a/dream-server/extensions/services/dashboard-api/hermes_bridge.py
+++ b/dream-server/extensions/services/dashboard-api/hermes_bridge.py
@@ -44,8 +44,16 @@ TOKEN_RE = re.compile(r'window\.__HERMES_SESSION_TOKEN__\s*=\s*"([^"]+)"')
 DEFAULT_HERMES_URL = "http://dream-hermes:9119"
 DEFAULT_TIMEOUT_SECONDS = 180
 
+
+def _env_int(name: str, default: int, *, minimum: int = 1) -> int:
+    raw = os.environ.get(name, "")
+    if raw.isdigit():
+        return max(minimum, int(raw))
+    return default
+
+
 # Connection pool tuning. Override per environment if needed.
-_IDLE_EXPIRY_SECONDS = int(os.environ.get("DREAM_TALK_IDLE_EXPIRY", "300"))  # 5 min default
+_IDLE_EXPIRY_SECONDS = _env_int("DREAM_TALK_IDLE_EXPIRY", 300)  # 5 min default
 _IDLE_SWEEP_INTERVAL = 60  # how often the background sweeper runs
 
 
@@ -55,6 +63,16 @@ class HermesBridgeError(RuntimeError):
 
 class HermesUnavailable(HermesBridgeError):
     """Hermes is not reachable or did not expose the expected dashboard API."""
+
+
+class HermesConnectionStale(HermesUnavailable):
+    """The pooled WebSocket died before a prompt was submitted.
+
+    This is safe to retry transparently because Hermes never accepted the
+    prompt on that transport. Once prompt.submit has been sent, later
+    connection drops surface as errors instead of retrying and potentially
+    duplicating tool calls or streamed text.
+    """
 
 
 @dataclass
@@ -249,6 +267,11 @@ async def _sweep_idle_connections() -> None:
             stale: list[tuple[str, _HermesConnection]] = []
             async with _POOL_GUARD:
                 for key, conn in list(_CONNECTION_POOL.items()):
+                    if conn.lock.locked():
+                        # Active prompt. Do not close the WS while Hermes is
+                        # streaming or pre-filling a long response; last_used
+                        # is refreshed again when the prompt completes.
+                        continue
                     if conn.closed or conn.ws.closed:
                         stale.append((key, conn))
                         _CONNECTION_POOL.pop(key, None)
@@ -307,6 +330,7 @@ async def _submit_on_connection(
     """
     request_id = f"dream-talk-prompt-{int(time.monotonic() * 1000)}"
     try:
+        conn.last_used = time.monotonic()
         await conn.ws.send_str(json.dumps({
             "jsonrpc": "2.0",
             "id": request_id,
@@ -316,9 +340,10 @@ async def _submit_on_connection(
     except (aiohttp.ClientError, ConnectionResetError, ConnectionError) as exc:
         # Pooled WS was closed under us between the freshness check and this
         # send (Hermes restart, network blip, idle timeout that hadn't been
-        # noticed yet). Surface as HermesUnavailable so stream_prompt evicts
-        # the pool entry + retries on a fresh connection.
-        raise HermesUnavailable(f"Hermes WS closed on send: {exc}") from exc
+        # noticed yet). Surface as HermesConnectionStale so stream_prompt
+        # evicts the pool entry + retries on a fresh connection. This is the
+        # only transparent-retry case because the prompt was not submitted.
+        raise HermesConnectionStale(f"Hermes WS closed before prompt submit: {exc}") from exc
 
     chunks: list[str] = []
     while True:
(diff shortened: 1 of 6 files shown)


Lightheartdevs:
Audit notes before merge:

- Rebased the stacked PR onto current main after #1321 merged, so the PR now contains only the session-pool/tool-scope changes.
- Reviewed the pooled Hermes lifecycle. Patched two merge-blocking risks: the idle sweeper could close an active long-running prompt, and transparent retry was too broad after prompt.submit. Retries now only happen when the stale pooled WS fails before submit; receive-side drops after submit surface as errors and evict the connection.
- Checked the Hermes pin: agent.disabled_toolsets is supported. Also corrected the timeout env var from an unconsumed TERMINAL_MAX_FOREGROUND_TIMEOUT to the consumed TERMINAL_TIMEOUT.
- Verified the persona typo SearXHNG -> SearXNG.

Local verification:
- 66 backend tests passed for Talk + magic-link
- Ruff passed on touched backend files
- installer context parity: 29 passed
- LiteLLM AMD auth guard: 4 passed

GitHub Actions are green on the refreshed head. Treating the remaining block as the expected self-review / review-required gate and merging via admin.

PR #1300, opened 2026-05-20
fix(installer): harden python setup diagnostics

gabsprogrammer:
## Summary

Hardens macOS/Linux installer setup around Python dependencies and phase 6 diagnostics.

## What changed

- macOS PyYAML setup now uses the shared Python resolver when available.
- If `python3 -m pip install --user pyyaml` fails, the installer falls back to a private installer venv at `.installer-venv`.
- The selected Python is exported through `DREAM_PYTHON_CMD` so later resolver/config scripts use the interpreter that can actually import `yaml`.
- Linux phase 6 now records substeps in `INSTALL_PHASE`, making failures point to the exact setup area instead of only `06-directories`.

## Why

A macOS install can fail when user-site pip installs are blocked or unavailable. Previously that stopped the installer with a generic PyYAML failure.

On Linux/Ubuntu, phase 6 is large. If a command exits under `set -e`, users only see that phase 6 failed, which makes support hard. Substep labels make failures actionable.

## Validation

- `bash -n dream-server/install-core.sh dream-server/installers/macos/install-macos.sh dream-server/installers/phases/06-directories.sh`
- `bash dream-server/tests/contracts/test-installer-hardening.sh`
- `python dream-server/tests/test-runtime-config-wiring.py`
- `git diff --check`

Note: full `test-installer-contracts.sh` could not run in this local shell because `jq` is unavailable.

PR #1325, opened 2026-05-22
fix(macos): isolate PyYAML installer runtime

gabsprogrammer:
## Summary

Fixes the macOS installer path that currently tries to install PyYAML into the active Homebrew Python with `pip --user`. Homebrew Python is externally managed under PEP 668, so the installer can fail before compose resolution with an externally-managed-environment error.

This change keeps the resolver dependency inside a Dream-owned installer virtualenv and exports `DREAM_PYTHON_CMD` so shared compose/Python helpers use the interpreter that can actually import `yaml`.

## Changes

- Add a macOS helper that verifies `python3` can import `yaml` before doing any install work.
- If PyYAML is missing, create `${INSTALL_DIR}/.venv/installer-python` and install `pyyaml` there.
- Export `DREAM_PYTHON_CMD` to route downstream compose resolver calls to the venv Python.
- Preserve dry-run behavior by reporting the planned venv creation without mutating the machine.
- Update the installer contract so `pip --user` for PyYAML is rejected on macOS.

## Validation

- `bash -n installers/macos/install-macos.sh tests/contracts/test-installer-contracts.sh`


Lightheartdevs (dismissed):
﻿Reviewed the macOS PyYAML runtime isolation. Looks safe in isolation.

Notes:
- Moving PyYAML into a Dream-owned installer venv avoids Homebrew/PEP 668 user-site friction.
- Exporting `DREAM_PYTHON_CMD` points the compose resolver at the interpreter that can actually import yaml.
- Tests cover the intended contract.

Non-blocking coordination note: this overlaps with #1300, which also handles selected-Python/PyYAML fallback plus Linux phase diagnostics. Merge order will matter; I would reconcile to one final macOS helper rather than keep parallel implementations.

Lightheartdevs (approved):
Audit complete. This is useful and safe after the rebase/fix pass.

What changed during audit:
- Merged current `main` into the PR branch so it no longer regresses the newer selected-Python handling from the installer diagnostics work.
- Resolved the PyYAML helper to avoid Homebrew/PEP 668 `pip --user` entirely, create an isolated Dream-owned installer venv with the selected Python, and keep `DREAM_PYTHON_CMD` plus the shared resolver cache aligned.
- Updated both installer contract suites so future changes cannot reintroduce the user-site PyYAML path.

Risk review:
- Install impact: positive for macOS/Homebrew Python installs; no Linux/Windows installer behavior changes.
- Service impact: none. This only affects installer preflight/runtime dependency resolution before compose.
- Security/exploit surface: no new service, port, secret, or browser surface. The venv lives under the install directory and is only used server-side by installer scripts.

Verified locally:
- `bash -n installers/macos/install-macos.sh tests/contracts/test-installer-contracts.sh tests/contracts/test-installer-hardening.sh`

PR #1326, opened 2026-05-22
fix(macos): resolve OpenCode binary for launchd

gabsprogrammer:
## Summary

Fixes the macOS OpenCode install/configure path so it works when OpenCode is installed through Homebrew or otherwise available on `PATH`.

The current installer only treats `~/.opencode/bin/opencode` as valid. On a Homebrew-based macOS setup, `brew install opencode` places the binary under the Homebrew prefix, so the installer can skip OpenCode config and LaunchAgent creation even though the tool is installed and runnable.

## Changes

- Add `_find_opencode_bin` to resolve OpenCode from the configured path, `~/.opencode/bin`, or `PATH`.
- Prefer `brew install opencode` on macOS when Homebrew is available.
- Keep the upstream install script as a fallback when Homebrew is unavailable or fails.
- Generate the LaunchAgent `ProgramArguments` with the resolved `OPENCODE_BIN`.
- Add the resolved binary directory to the launchd PATH.
- Update installer contracts so the macOS path cannot regress to a hard-coded `~/.opencode/bin/opencode` LaunchAgent.

## Validation

- `bash -n installers/macos/install-macos.sh tests/contracts/test-installer-contracts.sh tests/contracts/test-plist-log-paths.sh`
- `bash tests/contracts/test-plist-log-paths.sh`


Lightheartdevs (dismissed):
﻿Reviewed the macOS OpenCode binary resolution. Looks safe to merge.

Notes:
- Resolving `OPENCODE_BIN` from explicit env, upstream install path, or `PATH` handles Homebrew and existing installs better than hard-coding `~/.opencode/bin/opencode`.
- LaunchAgent now uses the resolved binary and PATH directory, which is the important macOS service fix.
- Failure behavior remains non-fatal, which matches the existing optional-devtool posture.

No blocking findings.

Lightheartdevs (approved):
Audit complete. Useful macOS install fix, approved after a safety hardening pass.

What changed during audit:
- Merged current `main` into the PR branch so it includes the latest installer hardening work.
- Kept the Homebrew/PATH OpenCode resolution behavior, but tightened discovery to executable files only: the LaunchAgent now gets an absolute binary path, not a shell function/alias/relative command from `command -v`.
- Added Homebrew-prefix lookup (`brew --prefix`) so `brew install opencode` can still be found if the prefix is not on the installer's PATH.
- Kept the LaunchAgent PATH aligned to the resolved OpenCode binary directory.

Risk review:
- Install impact: positive for macOS/Homebrew installs; this only changes native macOS OpenCode setup.
- Service impact: limited to the macOS host-side OpenCode LaunchAgent. Docker services, Linux, Windows, dashboard/API, and model serving are untouched.
- Security/exploit surface: no new exposed port or secret path. The audit fix avoids writing non-file command resolutions into launchd.

Verified locally:
- `bash -n installers/macos/install-macos.sh tests/contracts/test-installer-contracts.sh tests/contracts/test-plist-log-paths.sh`
- `bash tests/contracts/test-plist-log-paths.sh`

PR #1327, opened 2026-05-22
fix(macos): rebuild only enabled local services

gabsprogrammer:
## Summary

Prevents the macOS installer from rebuilding local optional-service images that are not part of the selected compose stack.

During a macOS install, the script currently rebuilds `dashboard`, `dashboard-api`, `ape`, `token-spy`, and `privacy-shield` unconditionally. If an optional service such as `privacy-shield` is disabled, its Dockerfile can still fail and create a misleading install warning even though the selected stack is otherwise healthy.

## Changes

- Resolve the selected compose services with `docker compose ... config --services` before local image rebuilds.
- Build only local services that are present in that resolved stack.
- Log skipped local-build candidates when their service is disabled.
- Add a contract guard so the macOS installer cannot regress to the unconditional hard-coded build list.

## Validation

- `bash -n installers/macos/install-macos.sh tests/contracts/test-installer-contracts.sh`


Lightheartdevs (approved):
﻿Reviewed the macOS “build only enabled local services” change. Looks safe to merge.

Notes:
- Using `docker compose ... config --services` before local rebuilds is the right source of truth.
- This avoids optional/disabled Dockerfiles breaking otherwise valid selected stacks.
- The failure message for compose service resolution is clear enough for support.

No blocking findings.

Lightheartdevs (approved):
Audit complete. This is useful and safe after rebasing onto current `main`.

Risk review:
- Install impact: positive for macOS. Disabled optional services no longer get rebuilt just because their Dockerfiles are in the repo, so an unused optional service should not create misleading install warnings.
- Service impact: no runtime service behavior changes. The selected compose stack is still what gets started; this only narrows the pre-`up` rebuild list to services present in `docker compose ... config --services`.
- Security/exploit surface: no new ports, credentials, auth paths, browser surface, or untrusted inputs. Candidate service names are a static installer list and are checked against compose output.

Verified locally:
- `bash -n installers/macos/install-macos.sh tests/contracts/test-installer-contracts.sh`

GitHub checks are green on the rebased head. The first CachyOS run failed before checkout on an upstream package signature error; rerun passed. Approving and merging.

PR #1329, opened 2026-05-22
fix(linux): configure PATH-installed OpenCode

gabsprogrammer:
## Summary

Fixes the Linux OpenCode setup path when `opencode` is already installed somewhere on `PATH` instead of under `~/.opencode/bin/opencode`.

The installer currently skips installation if `command -v opencode` succeeds, but the configuration and user systemd service are only created when `~/.opencode/bin/opencode` exists. That leaves PATH-installed OpenCode unconfigured and without the web service even though the install phase reported OpenCode as already installed.

## Changes

- Add `_find_opencode_bin` to resolve OpenCode from `~/.opencode/bin/opencode` or `PATH`.
- Store the resolved binary path in `OPENCODE_BIN` and use it for the config gate.
- Template `opencode-web.service` with the resolved binary path instead of hard-coding `~/.opencode/bin/opencode`.
- Template the systemd service PATH with the resolved binary directory.
- Add a Linux OpenCode path regression test covering PATH detection and service templating.

## Validation

- `bash -n installers/phases/07-devtools.sh tests/test-linux-opencode-path.sh tests/test-tmpfile-cleanup.sh`
- `bash tests/test-linux-opencode-path.sh`
- `bash tests/test-tmpfile-cleanup.sh` partially passed: OpenCode cleanup now passes; this branch still sees the existing unrelated Phase 05 cleanup failure on `main`.


Lightheartdevs (dismissed):
﻿Reviewed the Linux OpenCode PATH-installed binary fix. Looks safe to merge.

Notes:
- Resolving `OPENCODE_BIN` before config/service generation fixes PATH/Homebrew-style installs and keeps the upstream `~/.opencode` path working.
- The systemd template now uses the resolved binary and binary directory, which is the important runtime behavior.
- The regression test checks the contract without requiring OpenCode to be installed.

No blocking findings.

Lightheartdevs (approved):
Audit complete. Useful Linux install fix, approved after resolver hardening.

What changed during audit:
- Merged current `main` into the PR branch.
- Kept the PATH-installed OpenCode behavior, but tightened discovery to absolute executable files only using `type -P` plus an executable-file guard. This avoids templating a shell function/alias/relative command into the user systemd unit.

Risk review:
- Install impact: positive for Linux users who already have `opencode` installed outside `~/.opencode/bin`; they should now still get config and a user service.
- Service impact: limited to the host-side OpenCode user systemd service. Docker services, dashboard/API, model serving, and other extensions are untouched.
- Security/exploit surface: no new exposed service or credential. The audit fix reduces command-resolution risk before writing `ExecStart`.

Verified locally:
- `bash -n installers/phases/07-devtools.sh tests/test-linux-opencode-path.sh tests/test-tmpfile-cleanup.sh`
- `bash tests/test-linux-opencode-path.sh`

Note: `bash tests/test-tmpfile-cleanup.sh` still reports the known unrelated Phase 05 cleanup assertion from `main`; the Phase 07/OpenCode cleanup checks pass.

PR #1331, opened 2026-05-22
fix(linux): rebuild only enabled local services

gabsprogrammer:
## Summary

Prevents the Linux installer from rebuilding local optional-service images that are not part of the selected compose stack.

Phase 11 currently includes optional local-build services such as `privacy-shield` in the build list even when their compose files were disabled by feature selection. The build loop then reports those disabled services as failed or retained, which makes a valid selected stack look unhealthy and can obscure the actual install state.

## Changes

- Resolve enabled compose services with `docker compose ... config --services` before the local build loop.
- Build only candidate local services that are present in the resolved compose stack.
- Fail clearly if compose service resolution itself fails.
- Log skipped local-build candidates when they are disabled.
- Add a contract guard so Phase 11 cannot regress to the unconditional local-build list.

## Validation

- `bash -n installers/phases/11-services.sh tests/contracts/test-installer-contracts.sh`


Lightheartdevs (approved):
﻿Reviewed the Linux “build only enabled local services” change. Looks safe to merge.

Notes:
- `docker compose config --services` is the correct source of truth for the resolved install plan.
- Skipping disabled optional local-build services avoids unrelated Dockerfile/build failures during a healthy install.
- The contract test captures the intended behavior.

No blocking findings. Coordination note: this touches `installers/phases/11-services.sh`, so it may need a rebase around newer service-phase PRs like #1335.

Lightheartdevs (approved):
Audit complete. Safe and useful after rebasing onto current main.

Risk review:
- Install impact: positive for Linux installs. Disabled optional local-build services are no longer built just because their Dockerfiles exist, so they should not create misleading install warnings.
- Service impact: no runtime service behavior changes. The selected compose stack still controls what starts; this only narrows the pre-compose local build list to services present in `docker compose ... config --services`.
- Security/exploit surface: no new exposed port, auth path, secret, or user-controlled command path. Candidate service names are static installer values and are only matched against compose output.

Verified locally:
- `bash -n installers/phases/11-services.sh tests/contracts/test-installer-contracts.sh`
- `bash tests/test-network-timeouts.sh`

GitHub checks are green on the rebased head. Approving and merging.

PR #1332, opened 2026-05-22
fix(dream-talk): less aggressive compression + max-out Lemonade context (256k native)

Lightheartdevs:
## Summary

Operator reported the Dream Talk agent forgetting what they were working on mid-conversation during a deeper chat session. Forensics on strix's llama-server logs caught the smoking gun:

```
10:24:58  prompt n_tokens = 14,803    ← history accumulating normally
10:25:00  prompt n_tokens =  3,065    ← ~11.7k tokens lost in 2 seconds
10:25:04  prompt n_tokens =  4,181
```

The bridge's connection pool was healthy throughout (no evictions, no retries, no new Hermes sessions) and idle gaps stayed under the eviction threshold. The culprit was **Hermes auto-compression** firing on a transient `web_search`-result spike, then summarizing 80% of the conversation history into a lossy summary the user noticed as "the agent forgot."

While investigating, also caught a related under-utilization: **strix's Lemonade was serving the model at 64k context even though `LEMONADE_CTX_SIZE=131072` was set**, because Lemonade caches its config on first init and ignores subsequent env-var bumps. Qwen3.6-35B-A3B's GGUF advertises `qwen35moe.context_length = 262144` (256k native) and strix's KV cache was using 1% of available memory at the cap. So this PR also makes the bigger window actually take effect.

### Part 1: less aggressive compression

| knob | before | after | effect |
|---|---|---|---|
| `threshold` | 0.50 | **0.75** | a transient tool-result spike no longer crosses the line |
| `target_ratio` | 0.20 | **0.50** | half the history survives a compression event verbatim |
| `protect_last_n` | 20 | **40** | ~2× more recent turns stay intact |

Compression stays enabled (disabling outright would let long sessions hit the hard context limit and fail mid-reply — better to compress gently than crash).

* `extensions/services/hermes/cli-config.yaml.template` — new defaults for fresh installs, plus the before/after reasoning inline so the next operator who tunes these knobs understands the trade-off.
* `scripts/patch-hermes-config.py` — `_ensure_compression()` rewrites the three values via the idempotent `_set_key()` path, so any host that already has `/opt/data/config.yaml` from an older template gets migrated on the next `bootstrap-upgrade.sh` pass.

### Part 2: actually use the model's native context on Lemonade hosts

Lemonade reads `ctx_size` from `/root/.cache/lemonade/config.json` **only on first init** and ignores the `LEMONADE_CTX_SIZE` env on subsequent boots. The compose env was set right (`131072`) but had no effect on the running container's cached config (`65536`).

* **`extensions/services/llama-server/lemonade-entrypoint.sh`** (new) — tiny shell wrapper that, before exec'ing `lemonade-server`, syncs `config.json`'s `ctx_size` to the current `LEMONADE_CTX_SIZE` env. Idempotent: skips when the file doesn't exist (first start) or already has the right value.
* `docker-compose.amd.yml` — bind-mounts the wrapper as `:ro`, points `entrypoint:` at it. Bind-mount means iterating on the script doesn't require a container rebuild.

Operator can now set `CTX_SIZE=262144` in `.env` (or any other value up to the model's native ceiling) and a `dream restart llama-server` is enough to honor it.

### Part 3: fresh AMD installs auto-route Hermes through litellm

While verifying Part 2 I tripped a separate gotcha: after restarting Hermes, calls failed with `Missing Authentication header (HTTP 401)`. The cli-config's `model.api_key` field isn't honored when `provider: custom` is set — Hermes's openai client uses the `OPENAI_API_KEY` env, which defaults to a placeholder. On AMD/Lemonade installs the placeholder doesn't match LiteLLM's master key.

* `installers/phases/06-directories.sh` — `HERMES_LLM_BASE_URL` and `HERMES_LLM_API_KEY` defaults are now conditional on `GPU_BACKEND`:
  * **AMD**: `http://litellm:4000/v1` + `${LITELLM_KEY}`
  * **non-AMD**: `http://llama-server:8080/v1` + `sk-dream-hermes-local` (unchanged)

Fresh AMD installs get the right routing in `.env` from day one without manual hand-edit.

## Verification (live on strix-halo)

* `docker exec dream-llama-server sh -c 'ps -ef | grep llama-server'` reports `--ctx-size 262144`.
* llama-server startup logs:
  ```
  llama_context: n_ctx         = 262144
  slot   load_model: id 0 | new slot, n_ctx = 262144
  slot   load_model: id 1 | new slot, n_ctx = 262144
  ...
  ```
* `docker stats dream-llama-server`: 1.18 GB / 110 GB (1.07%) — KV cache for 256k×4 slots fits comfortably.
* Hermes `/opt/data/config.yaml` reports `context_length: 262144` (both `model.context_length` and `auxiliary.compression.context_length`).
* Dream Talk end-to-end through the public Caddy path (`Host: talk.dream-server.local`): `"HELLO"` reply in 48s (cold prefill of the fresh 256k cache; warm follow-ups via the connection pool will be 1–5s as before).
* Live config-yaml on all three Linux fleet hosts has the new compression knobs (`threshold: 0.75 / target_ratio: 0.50 / protect_last_n: 40`) and Hermes restarted to pick them up.

## Risk / blast radius

* **Compression knobs**: pure config change to one Hermes service. If 0.75 is still too aggressive, follow-up bump to 0.85. Conservative direction is safe — old aggressive defaults caused visible loss-of-context.
* **Lemonade wrapper**: thin shell shim, no-op on non-AMD installs (the entrypoint change is only in `docker-compose.amd.yml`). Worst case if the wrapper has a bug: Lemonade fails to start with a Python traceback in the logs.
* **Phase 06 default change**: only `${VAR:-default}` defaults; if the operator's `.env` already sets `HERMES_LLM_API_KEY` to something specific, it's preserved. Non-AMD installs are unchanged from before.

## Follow-ups (out of scope)

* tower2 (NVIDIA) and spark (aarch64) use llama.cpp directly, not Lemonade. They have their own context-size configuration paths; whether they're similarly under-utilizing is a separate investigation.
* The 401-on-restart symptom suggests upstream Hermes's `model.api_key` config field is silently ignored for `provider: custom`. Worth filing upstream; for now we work around it at the env layer.

### DIFF
--- a/dream-server/docker-compose.amd.yml
+++ b/dream-server/docker-compose.amd.yml
@@ -15,7 +15,13 @@ services:
         LLAMA_CPP_REF: ${LLAMA_CPP_REF:-b8763}
         LEMONADE_SERVER_IMAGE: ${LEMONADE_SERVER_IMAGE:-ghcr.io/lemonade-sdk/lemonade-server:v10.2.0}
     image: dream-lemonade-server:latest
-    entrypoint: ["/opt/lemonade/lemonade-server"]
+    # Wrapper entrypoint syncs ctx_size from LEMONADE_CTX_SIZE env into
+    # Lemonade's cached config.json before launch. Without this, Lemonade
+    # only reads the env on first-start and ignores subsequent bumps —
+    # leaving e.g. Qwen3.6-35B-A3B (native 256k context) serving at 64k.
+    # Bind-mounted from the source tree so updating the script doesn't
+    # require an image rebuild.
+    entrypoint: ["/bin/sh", "/opt/lemonade-entrypoint.sh"]
     command:
       - serve
       - --port
@@ -36,6 +42,9 @@ services:
       - lemonade-cache:/root/.cache/huggingface
       - lemonade-llama:/opt/lemonade/llama
       - lemonade-recipe:/root/.cache/lemonade
+      # Read-only bind of the wrapper entrypoint. Updating ctx-size sync
+      # logic doesn't require rebuilding the image.
+      - ./extensions/services/llama-server/lemonade-entrypoint.sh:/opt/lemonade-entrypoint.sh:ro
     devices:
       - /dev/dri:/dev/dri
       - /dev/kfd:/dev/kfd
--- a/dream-server/extensions/services/hermes/cli-config.yaml.template
+++ b/dream-server/extensions/services/hermes/cli-config.yaml.template
@@ -124,12 +124,29 @@ terminal:
 # (0.50 = 50%), not an absolute token count. A stale value like 10000
 # effectively disables proactive compression until the backend rejects the
 # request, which is how sessions fall into max-compression-attempt loops.
+#
+# Defaults tuned for Dream Talk — a deep mobile-chat conversation. Previous
+# values (threshold: 0.50, target_ratio: 0.20, protect_last_n: 20) caused
+# the agent to lose granular context mid-conversation: a single web_search
+# tool result (10-15kB of HTML/markdown) would briefly push context over
+# the 50% threshold, Hermes would compress earlier turns to 20% of their
+# original size, and the user would notice the agent "forgetting" what
+# they were working on. Reported live by the operator after a deeper
+# conversation. New values are deliberately more conservative — we defer
+# compression later and preserve more of it when it does trigger.
 compression:
-  # Enable automatic context compression.
+  # Enable automatic context compression. Disabling outright is appealing
+  # for a chat surface but lets a long session hit the hard context limit
+  # and fail mid-reply; better to compress gently than crash.
   enabled: true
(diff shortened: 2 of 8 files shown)


Lightheartdevs:
﻿Audit note from review pass:

I found one blocker before merge: CI is red in `integration-smoke` because `tests/test-installer-context-parity.sh` still expects the old Hermes compression normalization. The PR intentionally changes the compression values to `threshold: 0.75`, `target_ratio: 0.50`, `protect_last_n: 40`, but the contract still fails at `Hermes patcher normalizes compression threshold`.

Please update the installer context parity contract to the new expected values and rerun CI. The implementation direction itself looks coherent: the AMD-only Lemonade entrypoint is scoped properly, the compression migration path is idempotent, and I did not see another code-level blocker in the diff.

PR #1333, opened 2026-05-22
feat(dream-talk): live status indicators ("Searching the web…" etc.) so users see what the agent is doing

Lightheartdevs:
## Summary

Operator feedback after a recent deeper-conversation session: *"Sometimes it's hard to tell if it's hung up or still working because it just has the thinking status… leaves me wondering what it's doing out there with my request."*

Hermes is already emitting rich progress signals while it works (`tool.start`, `status.update` with `kind: lifecycle`, etc.) but our bridge dropped everything that wasn't `message.delta` or `message.complete`. So the SPA had nothing to render but the bare `Thinking…` spinner — even during a 30s `web_search` call where Hermes literally knows which tool it's running and against what query.

This PR surfaces that progress to the user as a dynamic spinner caption.

## What it looks like

A weather query now renders, in order:

```
🌀 Searching the web…
   current weather San Francisco
```

then once the model starts writing:

```
It's currently around 59°F and clear in San Francisco…
```

For a memory-checking turn the caption reads *"Checking memory…"*; for `execute_code`, *"Running code…"*; etc. Unknown tools surface as `Using \`<name>\`…` so an operator can spot what to map next.

## How it works

Three layers, each kept small:

1. **`hermes_bridge.py`** — forward two new Hermes event types to the bridge's caller:
   - `tool.start` / `tool.complete` (older builds): bridge yields `{type: tool_start, tool: <name>, detail: <context>}`.
   - `status.update` (newer builds): bridge yields `{type: status, label: <pre-formatted text>, kind: <category>}` — upstream already does the wording, we just route it.

2. **`routers/talk.py`** — translate bridge events into SSE `status` frames:
   - For `tool_start`, map the tool name through `_label_for_tool()` (e.g. `web_search → "Searching the web…"`). Known prefixes (`browser_`, `memory_`, `skill_`, `github_`) get sensible group captions. Unknown names fall back to `Using \`<name>\`…`.
   - For `tool_complete`, emit `{label: null}` to signal the SPA to clear the caption.
   - For `status.update`, pass verbatim — text already user-friendly.

3. **`DreamTalk.jsx`** — `MessageBubble`'s spinner now reads `message.statusLabel` instead of the hardcoded `Thinking`, with the optional `statusDetail` shown as a small caption underneath (e.g. the active search query). `sendText()` updates `statusLabel/Tool/Detail` on each `status` frame; once the first `message.delta` arrives the caption is cleared so the bubble flips to live token rendering.

## Friendly-label coverage

Mapped explicitly:
- `web_search → "Searching the web…"`
- `web_extract → "Reading the page…"`
- `execute_code → "Running code…"`
- `read_file / write_file / patch / search_files → "Looking at a file…"` etc.
- `memory → "Checking memory…"`
- `text_to_speech → "Generating voice…"`
- `session_search → "Searching past conversations…"`
- `image_generate / vision_analyze → drawing / looking at an image`
- a few more — see `_TOOL_LABELS` in `routers/talk.py`

Prefix-based fallbacks: `browser_*` → "Browsing…", `memory_*` → "Checking memory…", `skill_*` → "Loading a skill…", `github_*` → "Talking to GitHub…".

Final fallback for genuinely unknown tools: a literal `Using \`<name>\`…` so the operator who sees it on the phone knows exactly what to add to `_TOOL_LABELS` in a future pass.

## Test coverage

* **Backend** (`tests/test_talk.py`, +3 tests, **23/23 passing**):
  * `test_talk_message_stream_translates_tool_events_to_status_frames` — asserts session → status(active) → status(cleared) → delta → complete ordering, with `label="Searching the web…"`, `tool="web_search"`, and the operator-supplied `detail` preserved.
  * `test_talk_message_stream_status_label_for_unknown_tool` — verifies the honest fallback for genuinely unknown tool names.
  * `test_talk_message_stream_status_label_handles_browser_prefix` — exercises the prefix-based group captions.
  * `test_talk_message_stream_forwards_hermes_status_updates` — `status.update` events from newer Hermes builds pass through verbatim with `label` and `kind` preserved.
* **Frontend** (`DreamTalk.test.jsx`, +2 tests, **9/9 passing**):
  * `renders dynamic SSE status caption + detail while a tool is in flight` — friendly label replaces the default `Thinking…`, detail string renders below.
  * `drops the status caption when message.delta frames start arriving` — once content streams in, the bubble switches to live tokens.
  * `sseResponse()` helper gained a `holdOpen: true` option so tests can assert intermediate spinner state without the SPA's "stream closed → finalize bubble" path wiping the assertion subject.

## Live verification

```
$ curl -sN -X POST .../api/talk/message/stream \
    -d '{"text":"what is the weather like today in San Francisco"}'

frame types: session, status×2, delta×37, complete, done — 9s end-to-end
status #1:   "Searching the web…" — detail "current weather San Francisco"
status #2:   {label: null}  ← clear caption, deltas start
reply:       "It's currently around 59°F and clear in San Francisco. Expect cloudy
              skies early that will give way to partly cloudy conditions later,
              with a high near 63°F."
```

## Risk / blast radius

* Three small additive changes — adds new SSE event type (`status`) that older SPA builds will simply ignore. No protocol break.
* The frontend reads new optional fields (`statusLabel/Tool/Detail`); existing tests pass and the bubble continues to render the default `Thinking…` when no status frames arrive (matches old behaviour).
* No infra changes, no model changes, no compose changes.

## Follow-up (not in this PR)

Observed during verification: when llama-server hits ReadTimeouts under load, Hermes does an internal 3-retry burst that piles queued requests onto already-saturated slots, creating a cascading retry-storm that's hard to recover from without restarting Hermes + dashboard-api + llama-server together. Worth a separate look at raising litellm's read timeout above 30s so a single slow call doesn't snowball into 3 queued calls.

### DIFF
--- a/dream-server/extensions/services/dashboard-api/hermes_bridge.py
+++ b/dream-server/extensions/services/dashboard-api/hermes_bridge.py
@@ -376,6 +376,49 @@ async def _submit_on_connection(
             if isinstance(chunk, str) and chunk:
                 chunks.append(chunk)
                 yield {"type": "delta", "text": chunk}
+        elif event_type == "status.update":
+            # Hermes (pinned image) emits status.update events with a pre-
+            # formatted human-readable text and a `kind` category. Examples
+            # we've observed in the wild:
+            #
+            #   {"kind": "lifecycle", "text": "⏳ Retrying in 5.1s (attempt 2/3)..."}
+            #   {"kind": "lifecycle", "text": "⚠️ Max retries (3) exhausted..."}
+            #
+            # Upstream is responsible for the human-readable wording (often
+            # with emoji prefixes); we just forward the text so the SPA's
+            # spinner caption shows what Hermes is actually doing. The
+            # "Searching the web…" / "Reading the page…" labels we add on
+            # the SSE-layer side are for older `tool.start`-style events
+            # that some builds may still emit; this branch handles the
+            # newer status.update path.
+            status_text = payload.get("text") if isinstance(payload.get("text"), str) else None
+            if status_text:
+                yield {
+                    "type": "status",
+                    "label": status_text,
+                    "kind": payload.get("kind") if isinstance(payload.get("kind"), str) else None,
+                }
+        elif event_type == "tool.start":
+            # Older-build path: Hermes may also emit explicit tool.start
+            # events with name + context. We translate those into status
+            # frames via the SSE layer's _label_for_tool() mapping.
+            tool_name = payload.get("name") if isinstance(payload.get("name"), str) else None
+            if tool_name:
+                yield {
+                    "type": "tool_start",
+                    "tool": tool_name,
+                    "detail": payload.get("context") if isinstance(payload.get("context"), str) else None,
+                }
+        elif event_type == "tool.complete":
+            # Older-build path companion: tool finished, clear the caption.
+            tool_name = payload.get("name") if isinstance(payload.get("name"), str) else None
+            if tool_name:
+                yield {
+                    "type": "tool_complete",
+                    "tool": tool_name,
+                    "duration_s": payload.get("duration_s") if isinstance(payload.get("duration_s"), (int, float)) else None,
+                    "summary": payload.get("summary") if isinstance(payload.get("summary"), str) else None,
+                }
         elif event_type == "message.complete":
             final_text = payload.get("text")
             if not isinstance(final_text, str) or not final_text.strip():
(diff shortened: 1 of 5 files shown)


Lightheartdevs:
﻿Audit note from review pass:

No code-level blockers found. The bridge/SSE/status-frame flow is additive, frontend handling keeps the old `Thinking...` fallback, and tests cover both older `tool.start`/`tool.complete` events and newer `status.update` events.

CI has a red `powershell-lint (windows-latest)` check, but the log shows runner/tooling setup failed before project linting: `No repository with the name 'PSGallery' was found.` That looks like CI infrastructure noise rather than a PR regression. I would rerun that check, but I would not block this PR on the implementation.

PR #1334, opened 2026-05-22
feat(dream-talk): image + file attachments via paperclip (fixes the iOS video-mode trap)

Lightheartdevs:
## Summary

Operator feedback after a phone-side session: *"The paperclip icon on the chat box where you'd add files or attachments etc just takes you into video mode randomly on your phone. Should allow you to upload images and files etc. yeah?"*

Right on both counts. The button was wired to `<input type="file" accept="audio/*" capture>` — the `capture` attribute on iOS Safari forces the camera/video recorder instead of a file picker, and `accept="audio/*"` meant it was always a voice-attachment button masquerading as a general one. UX has been confusing since #1319.

Reworks the paperclip into an actual general-attachment surface **and** plumbs the backend through to **real vision** on Qwen3.6-35B-A3B's native vision encoder.

## What's in this PR

1. **SPA paperclip rewire** — `<input accept="image/*,.txt,.md,.csv,.json,.yaml,.yml,.log,.py,.js,.ts,.tsx,.jsx,.html,.css,.sh">`, no `capture`, label is now "Attach image or file". New `<AttachmentPreview>` strip above the input bar with thumbnail + filename + ✕-to-remove. User bubble renders attached images inline (`object-contain max-h-72`).

2. **New backend endpoint `/api/talk/attachment`** — multipart `file` + optional `text` caption. Same SSE response shape as `/api/talk/message/stream` so the SPA's existing event-stream loop handles both paths. Routing:
   - **Images** → multimodal one-shot to the vision backend (default Lemonade direct at `http://llama-server:8080`, **not** litellm — litellm's `*` wildcard collapses our `user.*` model id down to whatever's currently loaded, silently downgrading vision queries to the text-only model). Bypasses Hermes because `prompt.submit` only takes text.
   - **Text-like files** (.md/.txt/.json/code/etc.) → content decoded, inlined into the user prompt with the caption, routed through the existing Hermes bridge so the agent retains tools, memory, and the connection pool.
   - Anything else → 415 with a helpful error.

3. **Compose env** in `docker-compose.base.yml`:
   - `DREAM_TALK_VISION_URL` (default `http://llama-server:8080`)
   - `DREAM_TALK_VISION_KEY` (default empty)
   - `DREAM_TALK_VISION_MODEL` (default `user.Qwen3.6-35B-A3B-Vision`)

## How vision actually got working (Lemonade plumbing — discovered during live testing)

Qwen3.6-35B-A3B is natively vision-capable upstream (HuggingFace `pipeline_tag: image-text-to-text`), but the GGUF quant we ship (`Qwen3.6-35B-A3B-UD-Q4_K_M.gguf`) is text-only — the vision encoder lives in a separate ~0.9 GB `mmproj-F16.gguf` file from the same unsloth repo. To make it work on strix:

```
$ lemonade pull user.Qwen3.6-35B-A3B-Vision \
    --checkpoint main unsloth/Qwen3.6-35B-A3B-GGUF:Qwen3.6-35B-A3B-UD-Q4_K_M.gguf \
    --checkpoint mmproj unsloth/Qwen3.6-35B-A3B-GGUF:mmproj-F16.gguf \
    --recipe llamacpp --label vision

$ sed -i 's|^  default: .*|  default: "user.Qwen3.6-35B-A3B-Vision"|' /opt/data/config.yaml
$ dream restart hermes
```

Lemonade auto-launches llama-server with `--mmproj /…/mmproj-F16.gguf` on first multimodal call. Same chat session handles text-only turns transparently — no mode switching from the operator's perspective.

Installer auto-wiring on fresh AMD installs is a separate follow-up.

## Live verification on strix

| input | output | elapsed |
|---|---|---|
| 32×32 blue PNG (RGB 50,80,220), "What color?" | *"**solid blue** … `#3B7DD8` (approximate), `59, 125, 216`, often called 'Azure' or 'Steel Blue'."* | 3 s, 143 deltas |
| project-notes.md, "Summarize in one sentence" | *"The team needs to ship the attachment feature soon, and the paperclip icon should work as the UI element."* | 3 s |

Real vision, identified hex within 3 RGB values of the actual color.

## Tests

- **Backend**: +5 tests, 28/28 passing.
  - text attachment routes through Hermes with inlined content
  - unknown filetype → 415
  - image attachment routes to vision (asserts Hermes is **not** called)
  - oversize image → 413
  - attachment endpoint requires session cookie
- **Frontend**: +1 test, 10/10 passing.
  - picking an image shows preview strip with filename + ✕
  - send POSTs FormData to `/api/talk/attachment`
  - user bubble renders the image inline (`blob:` URL preserved)
  - Updated the live-mic-fallback test to look for the new "Attach image or file" label

## Risk / blast radius

- Compose change is additive (three new env vars with safe defaults).
- New endpoint is additive — existing `/api/talk/message/stream` unchanged.
- SPA changes: `accept` list narrower than before (`audio/*` → image+text), `capture` attribute removed (this is the actual bug fix). Voice attach was already broken/confusing for non-voice content; the recording-via-mic-button path (line 339) is preserved.
- Non-AMD hosts: `DREAM_TALK_VISION_MODEL` default points at `user.Qwen3.6-35B-A3B-Vision`. On a host without that registered, an image attach gets a 502 "vision unavailable" error frame and the rest of chat works fine.

## Follow-ups

- **Installer auto-wire**: phase 11 detects AMD/Lemonade installs and runs the `lemonade pull user.Qwen3.6-35B-A3B-Vision` command + updates Hermes config automatically. Currently a manual one-time operator step.
- **PDF / docx** parsing for text attachments — would need a parser dependency (pypdf2 / python-docx). Out of scope for v1 chat surface.
- **Vision on tower2 (NVIDIA, qwen3-coder-next)** — different model, different llama-server, would need its own vision-capable variant + plumbing.

### DIFF
--- a/dream-server/docker-compose.base.yml
+++ b/dream-server/docker-compose.base.yml
@@ -236,6 +236,18 @@ services:
       - TTS_URL=${TTS_URL:-http://tts:8880}
       - EMBEDDING_URL=${EMBEDDING_URL:-http://embeddings:80}
       - WHISPER_URL=${WHISPER_URL:-http://whisper:8000}
+      # Dream Talk's image-attachment endpoint calls the vision backend
+      # directly (bypassing litellm because litellm's wildcard model
+      # routing collapses our model id down to whatever's loaded).
+      # Defaults point at the Lemonade/llama-server service on the
+      # internal docker network — no auth needed there. Operators with
+      # an external vision server set DREAM_TALK_VISION_URL +
+      # DREAM_TALK_VISION_KEY.
+      - DREAM_TALK_VISION_URL=${DREAM_TALK_VISION_URL:-http://llama-server:8080}
+      - DREAM_TALK_VISION_KEY=${DREAM_TALK_VISION_KEY:-}
+      # Model id of the vision-capable variant. Defaults to the user.*
+      # name we register on AMD install (Qwen3.6-35B-A3B + mmproj-F16).
+      - DREAM_TALK_VISION_MODEL=${DREAM_TALK_VISION_MODEL:-user.Qwen3.6-35B-A3B-Vision}
     volumes:
       - ./scripts:/dream-server/scripts:ro
       - ./config:/dream-server/config:ro
--- a/dream-server/extensions/services/dashboard-api/routers/talk.py
+++ b/dream-server/extensions/services/dashboard-api/routers/talk.py
@@ -42,6 +42,17 @@ MAX_DOC_CHARS = 80_000               # post-extraction character cap so a megaby
 
 # MIME types we accept on the attach surface. Anything else → 415.
 _IMAGE_MIME_PREFIXES = ("image/",)
+_IMAGE_EXTENSION_MIMES = {
+    ".avif": "image/avif",
+    ".bmp": "image/bmp",
+    ".gif": "image/gif",
+    ".heic": "image/heic",
+    ".heif": "image/heif",
+    ".jpeg": "image/jpeg",
+    ".jpg": "image/jpeg",
+    ".png": "image/png",
+    ".webp": "image/webp",
+}
 _TEXT_LIKE_MIMES = {
     "text/plain", "text/markdown", "text/csv", "text/x-markdown",
     "application/json", "application/xml", "text/xml",
@@ -540,6 +551,8 @@ def _classify_attachment(file: UploadFile) -> str:
 
     if any(ct.startswith(p) for p in _IMAGE_MIME_PREFIXES):
         return "image"
+    if ext in _IMAGE_EXTENSION_MIMES:
+        return "image"
     if ct in _TEXT_LIKE_MIMES:
         return "text"
     if ext in _TEXT_LIKE_EXTENSIONS:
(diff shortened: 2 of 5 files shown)


Lightheartdevs:
﻿Audit note from review pass:

I found a few things to fix before merge:

- [P1] CI is red in `frontend`: ESLint reports `Blob`, `File`, and `FormData` as undefined in `DreamTalk.test.jsx`. The test needs to reference these through the jsdom/window globals or otherwise satisfy the lint environment.
- [P2] Attachment retry loses the attachment. `retryLast()` only calls `sendText(lastUser.text)`, so a failed image/file send retries as text-only, and an image-only send becomes a no-op. Either preserve enough attachment state to retry the same file or hide/adjust retry for attachment failures.
- [P2] Backend image classification only accepts image MIME types. The comment says filename fallback handles browsers that upload as `application/octet-stream`, but `_classify_attachment()` only falls back to text extensions, so `photo.png`/`photo.jpg` with octet-stream would 415 on mobile. Add image extension fallback for common photo formats.

The core direction is good: removing `capture`, making the paperclip general-purpose, and routing image vs text attachments separately all make sense.

PR #1335, opened 2026-05-22
feat(persona): agent knows what's installed on the Dream Server it's running on

Lightheartdevs:
## Summary

Operator feedback: *"make sure the default personality knows about Dream Server and what's installed where and what it has access to on the backend ... the whole extensions system, mapped ports and services etc. that all feel useful for it to know it has access to."*

Before this PR Dream Talk's agent had no idea what host it was running on or what came pre-installed. Asking *"what services are running?"* got a generic *"I don't have direct visibility into server-level services"* deflection. Asking *"what hardware are you on?"* got a hallucinated *"NVIDIA L4 24GB"* on a host that's actually AMD Radeon 8060S.

This PR auto-generates a section of the agent's persona from detected install state — host, GPU, model, running services + their ports — so Dream answers truthfully from its own system prompt instead of guessing or running shell probes to second-guess.

## What gets detected

A new `scripts/build-installation-context.py` script reads:

- `.env` for `GPU_BACKEND`, `DREAM_DEVICE_NAME`, `CTX_SIZE`, model name
- Live `docker ps` for what's actually up right now
- `extensions/services/*/manifest.yaml` cross-reference so container names map back to service ids (handles `dream-webui` → `open-webui` and collapses `dream-langfuse-*` sub-containers under one "langfuse" line)
- Live `/api/v1/health` query against Lemonade/llama-server for the actual loaded model id

Renders an authoritative section into `data/persona/SOUL.md` like:

```markdown
## About this Dream Server install — read this BEFORE answering questions about your environment

**You are running inside Dream Server, a fully-local AI stack the operator installed
on their own hardware.** Hermes Agent (what you run on) is one component of that
stack — not the whole thing. When the operator asks about "the dashboard,"
"the system," "what's running," "what you can do," or "your hardware," they
mean the Dream Server install around you, not Hermes Agent in isolation.

**These facts are authoritative. Don't run tool calls (terminal, file probes,
GPU queries) to second-guess them ...**

- **Host**: `dream-server` (LAN hostname `dream-server.local`)
- **GPU / backend**: AMD GPU (ROCm/Vulkan via Lemonade)
- **Chat model serving you right now**: `qwen3.6-35b-a3b` (context window: 256k)
- **Dream Server admin dashboard**: `http://dream-server.local` — NOT Hermes's
  internal `:9119` endpoint, which is just where you live.
- **Mobile chat portal (Dream Talk)**: `http://talk.dream-server.local`
- **Open WebUI**: `http://chat.dream-server.local`

## Services running on this install right now
- **SearXNG** — privacy-respecting metasearch ... port 8888; this is what my
  `web_search` tool uses under the hood
- **ComfyUI** — image + video generation (SDXL Lightning, LTX-2.3). UI on port 8188...
- **Kokoro TTS** — text-to-speech (port 8880). My replies in Dream Talk can be
  spoken aloud through this
[...etc...]

## Your direct capabilities vs. what's available on this server
**You can directly invoke these** via your tool-calling layer:
- `web_search` (via SearXNG above) ...

**You can't directly invoke these but can refer the operator to them** when they ask:
- **Generate images / video** → ComfyUI dashboard
- **Build automated workflows** → n8n
- **Enable more extensions** → Extensions page on the admin dashboard
- **Reach the stack from outside the LAN** → Tailscale
```

## The mount-target gotcha that took most of the debugging

Upstream Hermes copies its bundled `/opt/hermes/docker/SOUL.md` into the per-install `/opt/data/SOUL.md` **only on first container start**. Editing the docker/ default after that is silently ignored — the operator's data-volume copy wins, forever. So a naive bind-mount to `/opt/hermes/docker/SOUL.md` (what dream-server did before this PR) gets stale immediately.

Fix: bind-mount the assembled file to **both** paths, so the first-start convention and the runtime-read path both reflect current state:

```yaml
volumes:
  - ./data/persona/SOUL.md:/opt/hermes/docker/SOUL.md:ro
  - ./data/persona/SOUL.md:/opt/data/SOUL.md:ro
```

`data/persona/` (operator-owned) is kept distinct from `data/hermes/` (uid-10000-owned where the Hermes container writes its own session state) so the generator script can regenerate without sudo.

## Regeneration triggers

- **Install (`installers/phases/11-services.sh`)** runs the generator twice: once before `docker compose up` (to ensure the bind target exists when Hermes first boots) and once after services are running (so the service list reflects what's actually up, not the pre-launch state).
- **`dream restart hermes`** (or `dream restart` for the full stack) regenerates the SOUL.md as a precondition — so the persona always reflects current state without an explicit operator step.
- **Manual**: `python3 scripts/build-installation-context.py` from the dream-server checkout.

## Before / after on strix

| question | before | after |
|---|---|---|
| *"what services do you have on this Dream Server, what hardware are you on, and where is the dashboard?"* | *"I don't have direct visibility into server-level services…"* + 22 s of shell exploration | *"This Dream Server is running SearXNG, Perplexica, ComfyUI, Kokoro TTS, Whisper STT, Qdrant, n8n, APE, Privacy Shield, Token Spy, Langfuse, and Open WebUI. You're on a machine called `dream-server` with an AMD GPU (ROCm/Vulkan via Lemonade) and a 256k context window. The admin dashboard is at `http://dream-server.local`."* — direct quote from persona, no tool calls |
| GPU model | hallucinated *"NVIDIA L4 24GB"* | correct: *"AMD GPU (ROCm/Vulkan via Lemonade)"* |
| dashboard URL | conflated with Hermes-internal *"http://127.0.0.1:9119"* | correct: *"http://dream-server.local"* |

## Risk / blast radius

- New script is pure stdlib Python — no new dependencies.
- Compose change adds one mount; if the file is missing the mount fails loudly at `docker compose up` time, which is exactly when phase 11 has already generated it. Robust against ordering bugs.
- Generator failures in installer / `dream restart` are non-fatal (Hermes still boots from whatever assembled file is on disk, or the static template if absent).
- The dynamic context block is bounded (~800 tokens) so it doesn't bloat Hermes's 16k-token system prompt.

## Follow-ups

- Hermes prompt-priority experiments: even with strong wording, the agent occasionally still reaches for shell tools to "verify" facts that are already in SOUL.md. Could explore whether moving the block into the cli-config `system_prompt` field (vs. SOUL.md) gives it higher priority in Hermes's prompt stack.
- Capability map is hand-maintained per-service in the script's `_SERVICE_CAPABILITIES` dict — manifest schema doesn't expose enough metadata yet to auto-generate user-facing descriptions. A future PR could add a `description` / `port` field to `manifest.yaml` and let the generator pull from there.

### DIFF
--- a/dream-server/dream-cli
+++ b/dream-server/dream-cli
@@ -997,12 +997,35 @@ cmd_restart() {
 
     [[ "$rebuild_images" == "true" ]] && _dream_cli_rebuild_images "${flags[@]}"
 
+    # Refresh data/hermes/SOUL.md whenever Hermes (or the full stack) is
+    # restarted, so the persona's "About this installation" section
+    # reflects which services are currently up. Non-fatal — if the
+    # generator fails (Python missing, etc.) Hermes still boots from
+    # whatever assembled SOUL.md is already on disk, or the template if
+    # the assembled file is missing.
+    if [[ -z "$service" || "$service" == "hermes" ]]; then
+        _dream_cli_refresh_soul || true
+    fi
+
     if [[ -z "$service" ]]; then
         _compose_run_with_summary "Restarting all services" "${flags[@]}" up -d
     else
         service=$(resolve_service "$service")
         _compose_run_with_summary "Restarting $service" "${flags[@]}" up -d "$service"
     fi
+}
+
+# Refresh data/hermes/SOUL.md from the static template + detected install
+# state. Idempotent. Called automatically by cmd_restart (when restarting
+# hermes or the whole stack) and invocable manually by an operator.
+_dream_cli_refresh_soul() {
+    local script="$INSTALL_DIR/scripts/build-installation-context.py"
+    [[ -x "$script" ]] || [[ -f "$script" ]] || return 0
+    local py
+    py=$(command -v python3 || command -v python || true)
+    [[ -n "$py" ]] || return 0
+    "$py" "$script" >/dev/null 2>&1 || return 1
+    return 0
 }
 
 cmd_stop() {
--- a/dream-server/extensions/services/hermes/SOUL.md.template
+++ b/dream-server/extensions/services/hermes/SOUL.md.template
@@ -1,5 +1,7 @@
 # Dream — Dream Server Default Persona
 
+<!-- INSTALLATION_CONTEXT -->
+
 You are **Dream** — a knowledgeable, practical, and conversational assistant. Your goal is to give accurate, actionable answers that are as brief as possible while still feeling clear, friendly, and human.
 
 ## Style & tone
@@ -81,4 +83,4 @@ These rules exist because operators have repeatedly observed models stalling cha
 
 ## Resetting this persona
 
-The user can edit this file at `/opt/data/SOUL.md` inside the Hermes container (or `data/hermes/SOUL.md` on the host). Restart Hermes to apply. Deleting the file falls back to upstream's default persona.
+The user can edit this file at `/opt/data/SOUL.md` inside the Hermes container (or `data/hermes/SOUL.md` on the host). Restart Hermes to apply. Deleting the file falls back to upstream's default persona. The `<!-- INSTALLATION_CONTEXT -->` line above is replaced at build time with a snapshot of this install's GPU, model, and running services — regenerate by running `python3 scripts/build-installation-context.py` from the dream-server checkout.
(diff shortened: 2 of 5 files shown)


Lightheartdevs:
﻿Audit note from review pass:

I found blockers before merge:

- [P1] CI is red in Ruff: `scripts/build-installation-context.py` imports `os` but does not use it.
- [P1] CI is red in ShellCheck: `installers/phases/11-services.sh` uses `local _soul_builder=...` at top level. `local` is only valid inside a function and can fail under shell execution too; make it a normal assignment.
- [P2] The generator failure path is described as non-fatal, but Hermes compose now bind-mounts `./data/persona/SOUL.md` and no longer mounts the static template. If generation fails or Python is unavailable, the missing bind source can still break `docker compose up`. Please ensure a fallback file is always created before compose, even if it is just the static `SOUL.md.template` with the marker left empty/replaced.

The feature is useful and the dual-mount reasoning is sound; it just needs those packaging/CI edges cleaned up.

PR #1337, opened 2026-05-22
fix(persona): self-heal docker-auto-created SOUL.md dir + add macOS install hook

Lightheartdevs:
## Summary

Caught when re-running /dream-fleet-test after merging the install-context feature — **mac-mini smoke install failed** at compose-up:

```
error mounting "/Users/michaelbradley/dream-server/data/persona/SOUL.md" to
rootfs at "/opt/hermes/docker/SOUL.md": not a directory: Are you trying to
mount a directory onto a file (or vice-versa)?
```

Two underlying bugs in PR #1335:

1. **macOS install path never ran the generator.** I only added the
   `build-installation-context.py` hook to `installers/phases/11-services.sh`
   (the Linux phase). `installers/macos/install-macos.sh` has its own
   independent flow and went straight to `docker compose up` without
   ever creating `data/persona/SOUL.md`.

2. **Docker bind-mount silently auto-creates missing paths as directories.**
   With the SOUL.md file missing at compose-up time, Docker created an
   empty `SOUL.md/` directory at the bind source. Hermes then refused to
   start because you can't bind-mount a directory onto a file target.
   Worse, that directory persisted across re-installs (the install's
   `nuke install dir` step preserves `data/` so models / sessions /
   memory survive), so every retry hit the same wall.

### 1. Generator self-heals the directory state

`scripts/build-installation-context.py` now detects when its output path
exists but isn't a file (i.e. Docker pre-created a directory) and
`shutil.rmtree`s it before writing a real file:

```python
if output_path.exists() and not output_path.is_file():
    import shutil
    shutil.rmtree(output_path)
```

So if anything ever drops Linux into this state too (orphan compose run,
operator error), the next `dream restart hermes` or generator invocation
recovers automatically.

### 2. macOS installer hook

`installers/macos/install-macos.sh` now calls the generator right after
`patch-hermes-config.py`, mirroring the Linux flow in `phase 11`. Runs
before any `docker compose up`, so the bind-mount target exists as a
regular file from the start.

## Verification

Live-fixed mac-mini by:
```bash
ssh mac-mini 'rm -rf ~/dream-server/data/persona/SOUL.md && python3 ~/dream-server/scripts/build-installation-context.py'
```
→ Self-heal worked, generated a real 13543-byte SOUL.md.

Both fleet-test sub-bugs (`mac-mini exit 1 in ~10s on disk` was a
separate environment issue; this one is the actual code bug from
#1335) are now addressed for fresh installs on macOS.

## Risk

Tight scope — the script change is defensive (only acts when output
path is wrong type); the macOS hook is additive (script failure is
non-fatal, Hermes falls back to the static template).

### DIFF
--- a/dream-server/installers/macos/install-macos.sh
+++ b/dream-server/installers/macos/install-macos.sh
@@ -954,6 +954,21 @@ else
             else
                 ai_warn "Hermes template patch incomplete — Hermes may fail to reach llama-server. Hand-edit ${_hermes_tpl} if prompts hang."
             fi
+
+            # Render data/persona/SOUL.md = static persona + dynamic install
+            # context. The Hermes compose mounts this file as the agent's
+            # SOUL.md so it answers truthfully about what services + hardware
+            # it has on this Dream Server. MUST run before docker compose up,
+            # otherwise Docker's bind-mount engine auto-creates the source
+            # path as a *directory* and the install fails at compose-up with
+            # "not a directory: Are you trying to mount a directory onto a
+            # file" — which then persists across reinstalls because `nuke
+            # install dir` preserves data/.
+            _soul_builder="${INSTALL_DIR}/scripts/build-installation-context.py"
+            if [[ -f "$_soul_builder" ]]; then
+                python3 "$_soul_builder" >>"$DS_LOG_FILE" 2>&1 || \
+                    ai_warn "Could not generate Hermes installation-context SOUL.md (non-fatal — Hermes will use the template default)"
+            fi
         fi
     fi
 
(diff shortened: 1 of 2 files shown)


Lightheartdevs:
Reviewed the persona bind-mount recovery fix. Looks safe to merge.\n\nNotes:\n- The macOS installer hook now mirrors the Linux path and renders data/persona/SOUL.md before docker compose up, which addresses the Docker bind-mount auto-directory failure.\n- The generator self-heal is scoped to the expected output path and only removes the wrong-type SOUL.md path before writing the generated file.\n- CI is green, and I locally verified the regression shape by creating data/persona/SOUL.md as a directory and confirming uild_soul() replaces it with a generated file.\n\nNo blocking findings from this pass.

PR #1338, opened 2026-05-22
fix(persona): drop nested SOUL.md bind-mount that breaks Docker Desktop (virtiofs)

Lightheartdevs:
## Summary

PR #1337 fixed the directory-auto-create issue on the static-template mount, but kept the second *nested* bind mount that was added in PR #1335:

```yaml
volumes:
  - ./data/hermes:/opt/data                                # outer dir mount
  ...
  - ./data/persona/SOUL.md:/opt/hermes/docker/SOUL.md:ro   # OK — non-nested
  - ./data/persona/SOUL.md:/opt/data/SOUL.md:ro            # ← nested, broke macOS
```

The second one lands `/opt/data/SOUL.md` *inside* the `/opt/data` directory that's already bound from `./data/hermes`. That's a file-mount inside a directory-mount.

- **Linux Docker**: handles it.
- **mac-mini (colima)**: handles it. (Smoke gate passed.)
- **m5-mbp (Docker Desktop, virtiofs)**: refuses with:
  ```
  error mounting "/host_mnt/.../data/persona/SOUL.md" to rootfs at "/opt/data/SOUL.md":
  create mountpoint for /opt/data/SOUL.md mount: mountpoint
  "/run/host_virtiofs/.../data/hermes/SOUL.md" is outside of rootfs ...
  ```

Caught when re-running `/dream-fleet-test` after #1337 merged — m5-mbp's install bombed at compose-up in ~1.5 min on this error. mac-mini passed because colima ≠ Docker Desktop.

## The fix

Drop the second nested mount. The first (non-nested) `/opt/hermes/docker/SOUL.md` bind stays — Hermes's entrypoint copies it to `/opt/data/SOUL.md` on first container start, which covers fresh installs uniformly across both Docker backends.

For *updates* after first install (where Hermes's entrypoint won't re-copy on its own — that's a first-start-only behavior), `dream-cli`'s `_dream_cli_refresh_soul` now runs:

```bash
docker exec dream-hermes cp /opt/hermes/docker/SOUL.md /opt/data/SOUL.md
```

after regenerating the file. Container-side cp, same source path the bind-mount already provides, no nested mount needed. Works uniformly on Linux + colima + Docker Desktop. Container has to be running for this to matter; if it's not, the entrypoint will pick up the docker/ default on next start anyway.

## Verification

- Fresh install: Hermes entrypoint copies `/opt/hermes/docker/SOUL.md` → `/opt/data/SOUL.md` on first container start (existing upstream behavior). Both paths now resolve to the same content because we mount the assembled file at the docker/ default.
- Subsequent regen via `dream restart hermes`: the script writes `data/persona/SOUL.md` → `docker exec cp` copies it into the container's per-install dir → next chat session reads it. Verified manually on strix earlier in the session.
- macOS no longer hits the nested-mount restriction because we no longer have a nested mount.

## Risk

Pure compose + dream-cli change. No data migration. If the docker exec fails on a host (e.g. container not running), regen is non-fatal — Hermes will pick up the new content on next container restart via the docker/ default path.

### DIFF
--- a/dream-server/dream-cli
+++ b/dream-server/dream-cli
@@ -1025,6 +1025,24 @@ _dream_cli_refresh_soul() {
     py=$(command -v python3 || command -v python || true)
     [[ -n "$py" ]] || return 0
     "$py" "$script" >/dev/null 2>&1 || return 1
+
+    # Push the updated assembled SOUL.md into the running container's
+    # data dir, where Hermes actually reads from at session.create
+    # time. The bind-mount target (/opt/hermes/docker/SOUL.md) is just
+    # the upstream-default that Hermes's entrypoint copies to
+    # /opt/data/SOUL.md ONLY on first start; the per-install file is
+    # what's read after that. macOS Docker Desktop's virtiofs refuses
+    # to bind-mount /opt/data/SOUL.md directly (nested-mount limit),
+    # so we use docker exec for the in-container copy — works on
+    # Linux + macOS uniformly. Container has to be running for this
+    # to matter; if it's not, the entrypoint will pick up the docker/
+    # default on next start anyway.
+    if command -v docker >/dev/null 2>&1; then
+        if docker ps --format '{{.Names}}' 2>/dev/null | grep -q '^dream-hermes$'; then
+            docker exec dream-hermes cp /opt/hermes/docker/SOUL.md /opt/data/SOUL.md \
+                >/dev/null 2>&1 || true
+        fi
+    fi
     return 0
 }
 
--- a/dream-server/dream-cli
+++ b/dream-server/dream-cli
@@ -997,14 +997,9 @@ cmd_restart() {
 
     [[ "$rebuild_images" == "true" ]] && _dream_cli_rebuild_images "${flags[@]}"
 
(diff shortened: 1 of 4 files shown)


Lightheartdevs:
Reviewed and pushed one follow-up fix before merge. The original direction was right: dropping the nested /opt/data/SOUL.md bind mount avoids Docker Desktop virtiofs failures while keeping the non-nested /opt/hermes/docker/SOUL.md mount.\n\nFollow-up added in e986874:\n- dream restart hermes / full restart now refreshes SOUL.md after compose has ensured Hermes is running, so the container-side copy can actually land in /opt/data/SOUL.md.\n- Linux phase 11 now copies the post-launch regenerated SOUL.md into the running Hermes container after services are up.\n- macOS install now does the same post-compose refresh/sync, so the runtime persona is current without reintroducing the nested bind mount.\n\nValidation:\n- git diff --check\n- python -m py_compile dream-server/scripts/build-installation-context.py\n- Fresh GitHub CI is green across shell lint, compose validation, macOS smoke, integration smoke, and the distro matrix.\n\nNo remaining blocking findings from this pass.

PR #1343, opened 2026-05-22
feat(health): support non-HTTP health checks for TCP/CLI services (#666)

### LINKED ISSUE #666: Support non-HTTP health checks for TCP/CLI services
### Description

### Use Case

### Proposed Solution

_No response_

### Alternatives Considered

_No response_

Arifuzzamanjoy:
Closes #666. Flagged the approach in the issue and didn't hear back — opening this for review, happy to redirect if the direction's wrong.

Adds `health_type: http | tcp | none` to the service manifest. Defaults to `http`, so existing manifests don't need changes.

- `http` — unchanged
- `tcp` — opens a connection to the port, short timeout, no HTTP path needed
- `none` — skip the check, show as N/A

Wired through schema, audit, `service-registry.sh`, `health-check.sh`, `dream-doctor.sh`, and dashboard-api so all five surfaces agree on the same states.

**Real services updated**

- `piper-audio` → `tcp` (Wyoming protocol, not HTTP)
- `aider` → `none` (CLI tool, no server)

**Tests**

    bash dream-server/tests/test-service-registry.sh   # 223 pass
    bash dream-server/tests/test-health-check.sh       # 10 pass
    bash dream-server/tests/test-extension-audit.sh    # 19 pass

Health-check test spins up a real HTTP server and TCP listener and verifies `ok / ok / not_applicable`. Audit test also rejects an invalid `health_type: smtp` fixture.

21 files, +553 / -182. No new dependencies. Happy to split schema + plumbing from the manifest updates if you'd prefer.

### DIFF
--- a/ods/config/extensions-catalog.json
+++ b/ods/config/extensions-catalog.json
@@ -1,5 +1,5 @@
 {
-  "generated_at": "2026-05-24T03:20:28Z",
+  "generated_at": "2026-07-02T16:46:38Z",
   "schema_version": "1.0.0",
   "extensions": [
     {
@@ -16,6 +16,7 @@
       "depends_on": [],
       "port": 0,
       "external_port_default": 0,
+      "health_type": "none",
       "health_endpoint": "",
       "env_vars": [
         {
@@ -69,6 +70,7 @@
       ],
       "port": 3001,
       "external_port_default": 7800,
+      "health_type": "http",
       "health_endpoint": "/api/health",
       "env_vars": [
         {
@@ -149,6 +151,7 @@
       "depends_on": [],
       "port": 7860,
       "external_port_default": 7863,
+      "health_type": "http",
       "health_endpoint": "/",
       "env_vars": [],
       "tags": [
@@ -217,6 +220,7 @@
       "depends_on": [],
       "port": 9200,
       "external_port_default": 9200,
+      "health_type": "http",
       "health_endpoint": "/health",
       "env_vars": [
         {
@@ -322,6 +326,7 @@
       "depends_on": [],
       "port": 80,
       "external_port_default": 3007,
+      "health_type": "http",
       "health_endpoint": "/api/_health/",
       "env_vars": [],
       "tags": [
@@ -421,6 +426,7 @@
       "depends_on": [],
       "port": 8000,
       "external_port_default": 8000,
+      "health_type": "http",
       "health_endpoint": "/api/v2/heartbeat",
       "env_vars": [],
       "tags": [],
@@ -461,6 +467,7 @@
       ],
       "port": 8080,
       "external_port_default": 8890,
+      "health_type": "http",
       "health_endpoint": "/health",
       "env_vars": [],
       "tags": [],
@@ -504,6 +511,7 @@
       "depends_on": [],
       "port": 8501,
       "external_port_default": 8501,
+      "health_type": "http",
       "health_endpoint": "/",
       "env_vars": [],
       "tags": [
@@ -620,6 +628,7 @@
       "depends_on": [],
       "port": 5001,
       "external_port_default": 8002,
+      "health_type": "http",
       "health_endpoint": "/health",
       "env_vars": [
         {
@@ -692,6 +701,7 @@
       "depends_on": [],
       "port": 3000,
       "external_port_default": 7801,
+      "health_type": "http",
       "health_endpoint": "/api/v1/ping",
       "env_vars": [
         {
@@ -775,6 +785,7 @@
       "depends_on": [],
       "port": 7865,
       "external_port_default": 7865,
+      "health_type": "http",
       "health_endpoint": "/",
       "env_vars": [],
       "tags": [],
@@ -814,6 +825,7 @@
       "depends_on": [],
       "port": 7861,
       "external_port_default": 7861,
+      "health_type": "http",
       "health_endpoint": "/",
       "env_vars": [],
       "tags": [
@@ -922,6 +934,7 @@
       "depends_on": [],
       "port": 5000,
       "external_port_default": 8971,
+      "health_type": "http",
       "health_endpoint": "/api/version",
       "env_vars": [
         {
@@ -1037,6 +1050,7 @@
       "depends_on": [],
       "port": 4200,
       "external_port_default": 7822,
+      "health_type": "http",
       "health_endpoint": "/",
       "env_vars": [
         {
@@ -1121,6 +1135,7 @@
       "depends_on": [],
       "port": 3000,
       "external_port_default": 7830,
+      "health_type": "http",
       "health_endpoint": "/api/healthz",
       "env_vars": [
         {
@@ -1160,6 +1175,7 @@
       "depends_on": [],
       "port": 2283,
       "external_port_default": 2283,
+      "health_type": "http",
       "health_endpoint": "/api/server/ping",
       "env_vars": [
         {
@@ -1207,6 +1223,7 @@
       "depends_on": [],
       "port": 9090,
       "external_port_default": 9090,
+      "health_type": "http",
       "health_endpoint": "/health",
       "env_vars": [],
       "tags": [
@@ -1301,6 +1318,7 @@
       "depends_on": [],
       "port": 1337,
       "external_port_default": 1337,
+      "health_type": "http",
       "health_endpoint": "/health",
       "env_vars": [],
       "tags": [
@@ -1348,6 +1366,7 @@
       "depends_on": [],
       "port": 8888,
       "external_port_default": 8889,
+      "health_type": "http",
       "health_endpoint": "/tree",
       "env_vars": [
         {
@@ -1394,6 +1413,7 @@
       "depends_on": [],
       "port": 8080,
       "external_port_default": 8086,
+      "health_type": "http",
       "health_endpoint": "/health",
       "env_vars": [],
       "tags": [
@@ -1438,6 +1458,7 @@
       "depends_on": [],
       "port": 7860,
       "external_port_default": 7802,
+      "health_type": "http",
       "health_endpoint": "/health",
       "env_vars": [],
       "tags": [
@@ -1511,6 +1532,7 @@
       "depends_on": [],
       "port": 3080,
       "external_port_default": 3080,
+      "health_type": "http",
       "health_endpoint": "/health",
       "env_vars": [
         {
@@ -1641,6 +1663,7 @@
       ],
       "port": 8080,
       "external_port_default": 7803,
+      "health_type": "http",
       "health_endpoint": "/healthz",
       "env_vars": [],
       "tags": [],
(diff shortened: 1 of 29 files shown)


Lightheartdevs:
Thanks for the PR. I think this is a genuinely useful feature: the current HTTP-only model gives false health signals for services like Piper/Wyoming and CLI/no-server tools like Aider, and `health_type: http | tcp | none` is the right shape for making extension health more accurate.

I verified the main contract tests locally against a branch merged with current `main`:

- `bash tests/test-health-check.sh` -> 10 passed
- `bash tests/test-extension-audit.sh` -> 19 passed
- `bash tests/test-service-registry.sh` -> 223 passed

CI is also green. The one thing I want fixed before merge is the TCP probe shell invocation:

- `dream-server/scripts/health-check.sh`: the TCP probe interpolates the resolved port into `bash -c`:

```bash
bash -c "cat < /dev/null > /dev/tcp/127.0.0.1/$port"
```

- `dream-server/scripts/dream-doctor.sh` does the same thing.

Since the port can come from manifest/env-derived configuration, please treat it as untrusted input. A malformed value should not be able to become shell text. A simple guard like `[[ "$port" =~ ^[0-9]+$ ]]` before the probe, plus optionally checking `1 <= port <= 65535`, would be enough. Another good pattern is passing the value as an argument to the child shell and validating inside that shell before using `/dev/tcp`.

Once that TCP probe hardening is in, I think this is a valuable and mergeable change.

Arifuzzamanjoy:
Two commits added for the TCP probe hardening:

`64fb4c9` — `health-check.sh` + `dream-doctor.sh`: `^[0-9]+$` guard + range 1–65535, switched to `bash -c '…/$1' _ "$port"` so the port never hits shell text. health-check marks `fail`, dream-doctor marks `unhealthy` + `invalid_port`.

`54d5d50` — `dream-test.sh`: same pattern for `test_tcp()` which had the identical interpolation. Added a hostname allowlist too since that function takes `$host` as a param.

No unguarded `/dev/tcp` interpolation left.

Lightheartdevs:
Thanks for following through on the non-HTTP health-check contract and for hardening the TCP probe after the earlier review.

Quickest path to merge:
- Carry health_type through installed library/user-extension discovery. The real services changed here, dream-server/extensions/library/services/piper-audio/manifest.yaml and dream-server/extensions/library/services/aider/manifest.yaml, are installed into the user-extension path, but dream-server/extensions/services/dashboard-api/user_extensions.py currently returns health/port without health_type. That means piper-audio still gets checked as HTTP on port 10200, and ider can still look like a network health target after install instead of using the new 	cp/
one semantics.
- Update the library/catalog source surfaces alongside the bundled schema: dream-server/extensions/library/schema/service-manifest.v1.json should accept health_type, and dream-server/scripts/generate-extensions-catalog.py should emit it so the Dashboard extension catalog has the same contract the runtime uses.
- Add a behavioral Dashboard API regression for an installed user/library extension with health_type: tcp and one with health_type: none, covering both scan_user_extension_services() and the extension status calculation in dream-server/extensions/services/dashboard-api/routers/extensions.py.

What I checked:
- Static diff review of the schema, service-registry, health-check.sh, dream-doctor.sh, dream-test.sh, dashboard-api health helper, Dashboard status UI, and the piper-audio/ider manifests.
- Confirmed the earlier /dev/tcp shell interpolation feedback is addressed by numeric/range validation and positional arguments in health-check.sh, dream-doctor.sh, and dream-test.sh.
- git diff --check origin/main...origin/pr-1343: passed locally.
- GitHub checks are green and the PR is mergeable, but the current diff does not cover the library schema/catalog/user-extension path that current main uses for these two affected services.

Lightheartdevs:
Thanks for coming back to the non-HTTP health-check path and adding the schema/catalog/user-extension propagation from the last review.

Quickest path to merge:
- Update `dream-server/extensions/services/dashboard-api/routers/extensions.py` in both `/api/extensions/catalog` and `/api/extensions/{service_id}`. The current `checkable = ... if cfg.get(health)` filters still skip installed `health_type: tcp` user extensions because they intentionally have no HTTP path, and the fallback then synthesizes a healthy `ServiceStatus`. That means an installed `piper-audio` can show as enabled even when TCP port `10200` is closed. Include TCP health types in the probe set, and avoid the synthetic-healthy fallback for TCP entries.
- Tighten `dream-server/extensions/services/dashboard-api/tests/test_extensions.py` so the TCP regression proves the dashboard API actually calls `check_service_health` for `health_type: tcp`, not just that an empty health path becomes enabled. A useful pair is healthy TCP result => `enabled`, and down/connection-failed TCP result => `stopped`; cover the catalog path and, ideally, the single-extension detail path too. Keep `health_type: none` on the no-probe / `cli_installed` path.

What I checked:
- Static review of the current diff against `origin/main`, plus the new commit since the previous advisory head `54d5d50`: the library schema, generated catalog, catalog generator, and `user_extensions.py` propagation requested last time are now present.
- `git diff --check origin/main...origin/pr-1343`: passed locally.
- GitHub checks are green on `7c49614`, including Dashboard frontend/API, Test Linux integration-smoke, ShellCheck, Ruff, mypy, catalog validation, matrix smoke, env schema, PowerShell lint, and secret scan.
- Reviewed the health-check shell paths, `dream-doctor.sh`, service registry parsing, dashboard API health helper, catalog/detail status calculation, and the new dashboard API tests.

Arifuzzamanjoy:
Thanks for the review. Pushed 7c49614b with all three:

1. user_extensions.py now carries health_type through scan_user_extension_services() — validated against {http, tcp, none}, invalid values skip the extension. piper-audio is TCP-probed and aider resolves to not_applicable after install.
2. Library schema accepts health_type (health now conditionally required; piper-audio + aider validate), and generate-extensions-catalog.py emits it.
3. Added tcp/none/invalid regressions for scan_user_extension_services() and installed-extension status (tcp → enabled, none → cli_installed) in routers/extensions.py.

The health_type tests pass locally. Happy to adjust.

Arifuzzamanjoy:
TCP health checks are now properly routed and respect the dynamic timeout, and the tests have been updated to cover the new probe logic.
Ready for another look!

Lightheartdevs:
Thanks for coming back to the non-HTTP health-check flow and closing the Dashboard API TCP-status gap from the prior pass.

What looks good:
- The latest head carries `health_type: tcp` through both `/api/extensions/catalog` and `/api/extensions/{service_id}` probe selection instead of letting TCP services fall into the synthetic healthy fallback. Down TCP user extensions now map to `stopped` in tests.
- Schema/catalog/user-extension propagation is present across `extensions/schema`, `extensions/library/schema`, `generate-extensions-catalog.py`, `user_extensions.py`, and the `piper-audio`/`aider` manifests, so the bundled and installed-library paths share the same contract.
- The TCP probes are hardened with numeric/range validation and positional shell args in `health-check.sh`, `dream-doctor.sh`, and `dream-test.sh`, preserving the earlier security fix.

What I checked:
- Reviewed the full diff and latest commit `7c49614..34e698f`, especially `helpers.py`, `routers/extensions.py`, `test_extensions.py`, service registry, schema/catalog changes, and shell TCP probes.
- `git diff --check origin/main...origin/pr-1343`: passed locally.
- GitHub checks for the current head are green: Dashboard API/frontend, Test Linux integration-smoke, ShellCheck, Ruff, mypy, catalog validation, env schema tiers, matrix smoke including macOS and distro jobs, PowerShell lint, and secret scan.
- I did not rerun the full test suite locally in this Windows workspace; this advisory is based on static review plus the current green CI.

Lightheartdevs:
Thanks for keeping the non-HTTP health-check work current with `main`; the TCP/none runtime path is a useful contract for extensions like Piper and Aider.

Quickest path to merge:
- Restore the existing `main` schema contract in `ods/extensions/schema/service-manifest.v1.json` while adding `health_type`. The current head drops several properties that are already part of the built-in manifest contract: top-level `compatibility`, service `health_timeout`, `ui_path`, `external_link`, `startup_check`, `startup_timeout`, `hooks`, feature `launch`, and the existing `cpu` enum/minItems shape for `gpu_backends`. Those fields are still used by current manifests and dashboard/API code, so the schema should be extended rather than replaced with the narrower library shape.
- If `ods/extensions/library/schema/service-manifest.v1.json` is meant to be synchronized with the built-in schema, carry the same full property set there too. If it intentionally stays narrower, keep the `health_type` additions but avoid using that file as the source for the built-in schema.

What I checked:
- Reviewed the PR metadata, existing review discussion, current head diff, and green GitHub checks for `18dd020`.
- Confirmed `git diff --check origin/main...origin/pr/1343` passes locally.
- Re-reviewed the TCP health plumbing in `helpers.py`, `routers/extensions.py`, `user_extensions.py`, `health-check.sh`, `ods-doctor.sh`, `ods-test.sh`, service registry, catalog generation, and the added tests.
- Compared `origin/main` vs the PR for `ods/extensions/schema/service-manifest.v1.json`; the health-type addition is good, but the schema loses existing contract definitions that CI does not currently catch because most of those objects still allow additional properties.

Thanks again for pushing this forward; this looks close once the schema merge is corrected.

Arifuzzamanjoy:
built-in (ods/extensions/schema/service-manifest.v1.json): restored the full main contract with health_type added on top -- compatibility, health_timeout, ui_path, external_link, startup_check, startup_timeout, hooks, feature launch, gpu_backends cpu enum/minItems all back.
library (ods/extensions/library/schema/service-manifest.v1.json): synced to the same full property set -- identical to built-in now, not the narrower shape.
ready for another look.

Lightheartdevs:
Thanks for keeping at the non-HTTP health-check contract and for restoring the schema surface after the last review.

What looks good:
- `ods/extensions/schema/service-manifest.v1.json` now extends the current built-in schema with `health_type` without dropping existing service, feature, or requirement properties; a direct property-set comparison against `origin/main` shows only `service.health_type` added.
- The TCP/none contract is carried through service registry parsing, shell health/doctor/test probes, dashboard-api `check_service_health`, user-extension scanning, catalog/detail status calculation, generated catalog output, and the Piper/Aider manifests.
- The earlier `/dev/tcp` shell-interpolation risk remains addressed: the shell probes validate numeric/ranged ports and pass values as positional args rather than interpolating manifest/env-derived values into shell text.

What I checked:
- Static review of the latest diff against `main`, especially the two manifest schemas, `service-registry.sh`, `health-check.sh`, `ods-doctor.sh`, `ods-test.sh`, `helpers.py`, `routers/extensions.py`, `user_extensions.py`, generated catalog changes, and the new tests.
- `git diff --check origin/main...origin/pr/1343` passed locally.
- Local focused dashboard-api tests passed: `python -m pytest tests/test_extensions.py -q -k "tcp_probed or none_cli_installed"` and `python -m pytest tests/test_user_extensions.py -q -k "health_type"`.
- GitHub checks for this head are green, including ShellCheck, Test Linux integration-smoke, extension/catalog validation, Dashboard API/frontend, Ruff, mypy, env validation, distro smoke, PowerShell lint, and secret scan.
- I could not rerun the shell test scripts locally because this Windows workspace resolves `bash` through a WSL relay without `/bin/bash`; CI covers those shell paths on Linux.

PR #1349, opened 2026-05-22
fix(bootstrap): keep fallback model serving when compose flags are missing

gabsprogrammer:
## Summary
Follow-up to #1340. This makes the bootstrap full-model swap safer when `.compose-flags` is missing or stale.

## What changed
- `bootstrap-upgrade.sh` now tries to recover compose flags through `scripts/resolve-compose-stack.sh` before recreating `llama-server`.
- If compose flags still cannot be recovered, the script now fails cleanly:
  - marks bootstrap status as failed;
  - prints a manual recovery path;
  - exits before health checks;
  - does not stop or remove the currently serving `dream-llama-server` container.
- Extends the bootstrap hot-swap contract test to cover:
  - compose resolver recovery;
  - non-destructive missing-flags failure;
  - no fallback `docker stop` / `docker rm` path.

## Why
#1340 correctly fixed the bootstrap-to-full-model swap by force-recreating `llama-server` so updated `GGUF_FILE` and context settings land. The remaining sharp edge was the missing `.compose-flags` fallback: if compose metadata was unavailable, the script could remove the running container and leave the user worse off.

A failed model swap should preserve the currently serving model whenever possible and fail with a clear recovery path.

## Validation
- `bash -n dream-server/scripts/bootstrap-upgrade.sh dream-server/tests/test-bootstrap-upgrade-hotswap-contract.sh`
- `bash dream-server/tests/test-bootstrap-upgrade-hotswap-contract.sh`
- `git diff --check`

Note: `bash tests/contracts/test-installer-contracts.sh` could not run in this local Windows shell because `jq` is not installed (`[FAIL] jq is required`).

### DIFF
--- a/dream-server/scripts/bootstrap-upgrade.sh
+++ b/dream-server/scripts/bootstrap-upgrade.sh
@@ -343,6 +343,23 @@ if [[ -n "$DOCKER_CMD" ]] && $DOCKER_CMD ps --filter name=dream-llama-server --f
     COMPOSE_ARGS=()
     if [[ -f "$INSTALL_DIR/.compose-flags" ]]; then
         read -ra COMPOSE_ARGS <<< "$(cat "$INSTALL_DIR/.compose-flags")"
+    elif [[ -x "$INSTALL_DIR/scripts/resolve-compose-stack.sh" ]]; then
+        _tier="1"
+        if [[ -f "$ENV_FILE" ]]; then
+            _tier=$(grep -E '^TIER=' "$ENV_FILE" | cut -d= -f2 | tr -d '"\047\r')
+            [[ -n "$_tier" ]] || _tier="1"
+        fi
+        _resolved_env=$("$INSTALL_DIR/scripts/resolve-compose-stack.sh" \
+            --script-dir "$INSTALL_DIR" \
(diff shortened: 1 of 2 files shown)


Lightheartdevs (approved):
Audit complete. Safe and valuable.

Risk review:
- Install/upgrade impact: positive. If `.compose-flags` is missing or stale during the bootstrap-to-full-model swap, the script now tries the canonical compose resolver first and otherwise fails cleanly without removing the currently serving `dream-llama-server` container.
- Service impact: this only changes the failure/recovery path for the background bootstrap upgrade. The normal force-recreate path for llama-server and OpenClaw remains intact.
- Security/exploit surface: no new network/API/browser surface. Compose flags still come from the local installer resolver or saved installer metadata; no untrusted remote input is executed.

Verified locally against a merge with current `main`:
- `bash -n scripts/bootstrap-upgrade.sh tests/test-bootstrap-upgrade-hotswap-contract.sh`
- `bash tests/test-bootstrap-upgrade-hotswap-contract.sh`
- `bash tests/contracts/test-amd-lemonade-contracts.sh`

GitHub checks are already green. Approving and merging.

PR #1365, opened 2026-05-24
 feat: LAN cluster - pool inference across machines over LAN

y-coffee-dev:
# feat/lan-cluster -- pool inference across LAN machines

Two or more machines on the same LAN cooperatively serve a single `llama-server` endpoint via llama.cpp's RPC backend. Workers can drop out and rejoin without taking the cluster down.

## Usage

After Dream Server is installed on each host.

Four ports are involved, it is useful to have these in mind while reading the commands below:

| Port | Proto | Direction | Default bind | What it carries |
|---|---|---|---|---|
| **50051** | TCP | worker -> controller | controller's LAN IP (or `--bind`) | One-shot join handshake. Only open during `dream cluster setup` (interactive). Carries the token-gated join request + accept/reject reply. |
| **50052** | TCP | controller -> worker | worker's all-interfaces (or `--interface <ip>`) | llama.cpp RPC. The hot path -- tensor data flows here during inference. **Plaintext, no auth** (llama.cpp limitation; that's why `dream cluster enable` makes you acknowledge it). |
| **50053** | UDP | controller broadcasts -> workers listen | controller's `/24` (or `--bind`) | HMAC-signed discovery beacon (`DREAM2` magic). Lets a worker find the controller without being told its IP. Wire format: 32-byte MAC + JSON. Rejected if the token doesn't match. |
| **50054** | TCP | local only | `127.0.0.1` by default | Worker agent `/health` endpoint. Loopback-only out of the box so `/health`'s GPU-fingerprint + cluster topology don't leak. Widen with `--status-bind <ip>` only if you want the controller's `dream cluster add` to auto-fill GPU info via HTTP. |

### First-time cluster bring-up

In practice you'll have two (or more) terminals open, one on the controller, one on a worker, and alternate between them. Each step calls out where it runs.

#### Step 1 -- on the **controller** -- enable cluster mode

```bash
dream cluster enable
# OR for scripted installs (skips the plaintext-RPC ack prompt):
dream cluster enable --yes
# OR pin the controller IP on multi-homed hosts (Tailscale, virbr0, etc.):
dream cluster enable --interface 192.168.8.192
```

What this does: generates a 128-bit token, writes it to `.env` (mode 0600), builds the `dream-llama-rpc:{cpu,cuda,rocm}` controller image, adds the cluster overlay to your compose stack. **Copy the printed token -- each worker needs it.**

(Tip: enable LAN cluster during the initial `dream-server` install to pre-build the controller image then, instead of waiting at this step.)

#### Step 2 -- on the **controller** -- start the setup listener

```bash
dream cluster setup
```

This command **blocks**. It:

- Broadcasts an HMAC-signed discovery beacon on **UDP/50053** every 5s so workers can find this host without being told the IP.
- Listens for join requests on **TCP/50051** (token-gated).

Leave it running while you bring up workers in Step 3.

#### Step 3 -- on **each worker** -- join the cluster

Two flows. The recommended one is the agent flow, it auto-restarts `rpc-server` on crash and persists across reboots via `systemd --user`.

**Agent flow (recommended):**

```bash
# Token via file (safest -- file should be mode 0600):
dream cluster agent start --token-file /path/to/cluster-token

# OR via env var
CLUSTER_TOKEN=dream_abc123 dream cluster agent start

# OR via CLI flag
dream cluster agent start --token dream_abc123

# Skip UDP discovery and connect to the controller directly (useful if
# cross-subnet, or if your network blocks broadcast):
dream cluster agent start --token-file /path/to/token --controller 192.168.8.192

# Narrow rpc-server's host-side publish to one interface (recommended on
# multi-homed hosts -- without this, rpc-server is reachable on every interface
dream cluster agent start --token-file /path/to/token --interface 192.168.8.191
```

**Manual one-shot flow (no monitoring, no auto-restart -- ad-hoc testing):**

```bash
CLUSTER_TOKEN=dream_abc123 dream cluster join 192.168.8.192
CLUSTER_TOKEN=dream_abc123 dream cluster join --discover                                # auto-find controller
CLUSTER_TOKEN=dream_abc123 dream cluster join 192.168.8.192 --interface 192.168.8.191   # narrow publish
```

Either flow opens a TCP connection to the controller's setup listener (Step 2) and sends a signed join request.

#### Step 4 -- on the **controller** -- accept each worker

Each Step-3 connection causes the `dream cluster setup` terminal (Step 2) to print a prompt:

```text
  New worker wants to join:
    IP:       192.168.8.191
    RPC port: 50052
    Backend:  nvidia
    GPU:      RTX 3090 (24.0 GB)

  Accept this worker? [Y/n]
```

Type `y` to accept (writes the worker into `cluster.json`) or `n` to reject. Repeat **Steps 3 and 4** for each worker.

#### Step 5 -- on the **controller** -- finish setup + restart dream

Once all workers are joined:

```bash
# In the `dream cluster setup` terminal:
Ctrl+C

# Then:
dream restart llama-server
```

`llama-server` now starts with `--rpc <worker-list>` and the supervisor takes over. Inference distributes across all workers.

---

### Possible operations after the cluster is up

| Goal | Run on... | Command |
|---|---|---|
| See cluster topology + ping each worker | controller | `dream cluster status` |
| Local agent state (reads `http://127.0.0.1:50054/health`) | worker | `dream cluster agent status` |
| 10-sample TCP latency benchmark to every worker | controller | `dream cluster test` |
| Add a worker without re-running `setup` (operator-known IP) | controller | `dream cluster add <WORKER_IP>` *(or `--port 50053` for custom rpc port)* |
| Remove a worker from the cluster | controller, then worker | controller: `dream cluster remove <WORKER_IP>`<br>worker: `dream cluster leave` *(stops + removes the local rpc-server)* |
| Tail the agent log | worker | `dream cluster agent logs` |
| Stop the agent cleanly | worker | `dream cluster agent stop` *(SIGTERM, tears down rpc-server)* |
| Turn cluster mode off entirely (revert to single-host) | controller | `dream cluster disable` |

## What's in the box

- **Supervisor** ([scripts/dream-cluster-supervisor.py](dream-server/scripts/dream-cluster-supervisor.py)) -- wraps `llama-server`, pre-flights workers, restarts with live-only `--rpc` when a worker drops, two-signal watchdog (worker-down AND `/health` hung) prevents single-GC-pause kills, circuit breaker stops the loop at 5 crashes / 120s.
- **Worker agent** ([scripts/cluster_worker_agent.py](dream-server/scripts/cluster_worker_agent.py)) -- `systemd --user` daemon, state machine `idle -> discovering -> joined -> running`, owns the `rpc-server` container, status endpoint on `127.0.0.1:50054`, `threading.Event`-based SIGTERM cancels in-flight waits in ms.
- **Discovery** ([scripts/cluster_discovery.py](dream-server/scripts/cluster_discovery.py)) -- UDP/50053, HMAC-SHA256 signed (`DREAM2`), 300s replay window, computes `/24` broadcast from bound interface.
- **Setup handshake** ([scripts/cluster-setup-listener.py](dream-server/scripts/cluster-setup-listener.py), [scripts/cluster-join-client.py](dream-server/scripts/cluster-join-client.py)) -- TCP/50051, token-gated, 1 MiB recv cap, allowlist validation (`gpu_backend in {cpu,nvidia,amd}`, port range, C0/C1 strip on display strings), dedup by `(ip, rpc_port)`.
- **Compose + images** -- `docker-compose.cluster{,-amd}.yml`; controller + worker Dockerfiles for cpu/cuda/rocm, all six pinned to `LLAMA_CPP_TAG=b8672` (RPC wire protocol must be lockstep).
- **Dashboard** -- `GET /api/cluster/status` (API-key gated, 10s background poll), [ClusterMonitor.jsx](dream-server/extensions/services/dashboard/src/pages/ClusterMonitor.jsx) UI.
- **Installer** -- Phase 03 prompts for cluster mode, Phase 08 builds the RPC images with streamed output.

## Security properties

- HMAC-SHA256 signed beacons, 300s replay window, `hmac.compare_digest` everywhere.
- Atomic 0600 writes (`O_EXCL + os.replace` / `mktemp + mv`) for every `cluster.json` / `cluster-agent.json` writer.
- Token never required on the CLI: `--token-file` + `CLUSTER_TOKEN` env preferred; `--token` warns it's `ps`-visible. Token-file mode check warns on group/world-readable.
- Agent `/health` binds `127.0.0.1` by default; opt in via `--status-bind` / `--interface`.
- `dream cluster join --interface <ip>` narrows the rpc-server's host-side publish to the specified IP, matching the agent path. Without it, all-interfaces publish (legacy default for single-NIC homes).
- Subprocess timeouts on every external call; narrow exception catches only at I/O boundaries.

## Configuration

| Var | Default | Purpose |
|---|---|---|
| `CLUSTER_ENABLED` | `false` | Toggles the compose overlay |
| `CLUSTER_TOKEN` | (generated) | HMAC key for beacons + handshake |
| `CLUSTER_WORKERS` | (empty) | Supervisor `host:port,...` list |
| `CLUSTER_TENSOR_SPLIT` | (empty) | Float list, equal-weight by default |
| `CLUSTER_RESTART_POLICY` | `always` | `always` / `on-worker-recovery` / `manual` |
| `CLUSTER_INTERFACE` | (empty) | Bind beacon/listener to a specific interface |
| `CLUSTER_CONTROLLER_IMAGE` | `dream-llama-rpc:cuda` | Override controller image |

(Port details in the table at the top of [Usage](#usage).)

## Limitations

- llama.cpp RPC is plaintext -- only run on trusted LANs or over WireGuard / Tailscale. `dream cluster enable` makes operators acknowledge this explicitly.
- Apple Silicon controllers are rejected (There is no RPC image for macOS).
- Auto-discovery assumes `/24`; cross-subnet operators must pass `--controller <ip>`.
- Each `rpc-server` instance accepts exactly **one** concurrent TCP client (upstream llama.cpp limitation). This does **not** cap cluster size -- every worker runs its own `rpc-server` and the one `llama-server` holds all of those slots simultaneously. What it *does* mean: you can't run a second `llama-server` against the same workers, and an external process can't TCP-probe a worker while `llama-server` is connected (the probe just hangs). The integration suite's old liveness check ran into this and was replaced by an `llama-server /health` gate (a healthy controller with `-ngl 99` implicitly proves all workers were reachable + handshook).
- `LLAMA_CPP_TAG` must match across all six Dockerfiles; mismatch crashes at first tensor transfer.

## Testing & Bug-hunting

This branch was iteratively self-tested and bug-hunted as the code evolved, then exercised end-to-end on real two-host hardware. The control plane (security + correctness) and the data plane (cross-host RPC tensor transfer) are both validated.

### Notable bugs caught and fixed during self-testing

1. **Unsigned discovery beacons let a LAN attacker redirect workers** to an attacker-controlled setup port. Switched from the unsigned `DREAM1` format to `DREAM2` -- HMAC-SHA256-signed payloads with a 300 s replay window and `hmac.compare_digest` verification. Workers refuse unsigned beacons when a token is configured.

2. **Agent `/health` leaked the cluster token + bound to `0.0.0.0` by default**. Two separate bugs in the same endpoint: token was echoed verbatim in the response, and the bind address exposed GPU fingerprint + cluster topology to the whole LAN. Now strips secrets before serialization (asserted in tests) and defaults to `127.0.0.1`. Operators who need cross-host `/health` opt in explicitly via `--status-bind` / `--interface` / `CLUSTER_INTERFACE`.

3. **SIGTERM responsiveness regressed to 10 s** because the supervisor + agent monitor loops used `time.sleep(HEALTH_CHECK_INTERVAL)` between probes. Replaced with `threading.Event.wait()` so `systemctl stop dream-cluster-agent` returns in **under 50 ms** instead of risking SIGKILL escalation that would orphan the `rpc-server` container.

4. **Atomic 0600 writes for every `cluster.json` / `cluster-agent.json` writer**. Initially three of the writers used plain `open("w")` + `json.dump`, which (a) exposed mode 0644 briefly via umask and (b) left half-written files on crash. All five writers now do `O_EXCL` + `os.replace` (or `mktemp` + `mv -f` in bash) with `BaseException` cleanup of the `.tmp` -- proven by a regression-evidence test that demonstrates the pre-fix path leaks a `.tmp` while the fixed path raises cleanly.

5. **Watchdog two-signal hysteresis**. The first watchdog killed `llama-server` whenever ANY worker became unreachable, which would fire on transient TCP blips, GC pauses, or the rpc-server's single-slot reconnect window -- causing restart storms during normal operation. Now requires `worker_unreachable AND /health_hung` (both signals true). Tested: a worker kill alone doesn't trigger restart; a subsequent inference that actually exercises the broken RPC does.

6. **Crash-loop circuit breakers on both supervisor and agent**. A deterministic failure (bad model path, missing GPU device, malformed `--rpc`, wedged dockerd) would spin forever. Added a 5-crashes-in-120-s fuse on each; exits non-zero so systemd's `StartLimitBurst` can stop the thrashing instead of the process spamming `docker run` until disk dies.

7. **ANSI-escape sanitization on worker-provided strings**. A hostile worker could embed `\x1b[2J\x1b[H` in `gpus[].name` to clear the operator's screen and spoof the `[Y/n]` prompt at the setup listener. Strips C0/C1 control characters before display; verified by sending real ESC bytes from a worker and confirming `cat -v` of the listener log contains zero raw escape sequences in the GPU-name region.

8. **`dream cluster join --interface <ip>` was silently no-op'ing the publish bind**. Flag was parsed but only used for the discovery sub-call; the `docker run -p` still published all-interfaces. The agent path (`cluster_worker_agent.start_rpc_server(bind_ip=...)`) already did the right thing; mirrored it in the manual-join path so multi-homed hosts (Tailscale, virbr0, Docker bridges) get consistent behavior.

9. **Smaller items folded in along the way** -- 1 MiB payload cap on the setup listener (prevents OOM from streamed bytes), 30 s subprocess timeouts on every `docker` CLI call (so a wedged dockerd doesn't hang shutdown), token-file mode-permission warning if it's group/world-readable, narrow exception catches at I/O boundaries only, removal of every `2>/dev/null || true` and silent-catch pattern that hid real errors.

### Automated test coverage

- **35 unit tests** (control plane -- `tests/test-cluster-{supervisor,worker-agent}.py`) -- `python3 -m unittest`, ~18s, no docker, no network.
- **72 compose-stack assertions** -- `bash tests/test-cluster-compose.sh`, ~2s.
- **56 e2e cases** -- `tests/cluster-e2e/run.sh` (Docker + fake-llama-server, no model, no GPU).
- **7 real-RPC integration tests** -- `tests/cluster-integration/run.sh` (real `rpc-server` + `llama-server` containers, real GGUF).

Total: **170 automated assertions**, all cluster-scope.

### Final manual end-to-end run on real hardware

| Area | Cases | Result |
|---|---|---|
| Unit + compose suites | 35 + 72 | **107/107 PASS** |
| Discovery beacon over real LAN (signed/unsigned/replay/wrong-token) | 7 | **7/7 PASS** |
| Setup handshake over real LAN (token gate, ANSI-strip with real ESC bytes, 1 MiB cap, token-file mode warn, atomic write) | 6 | **6/6 PASS** |
| Supervisor (passthrough exec, full live, degraded, 3 restart policies, SIGTERM responsiveness, two-signal hysteresis) | 8 | **8/8 PASS** |
| Worker agent over real LAN (no-token fast-fail, `--controller` skip, `/health` token strip + 127.0.0.1 bind, state mode-0600, SIGTERM responsiveness) | 5 | **5/5 PASS** |
| `dream-cli` (bash syntax, atomic `_cluster_save_state`) | 2 | **2/2 PASS** |
| Dashboard-api router helpers + real-LAN `_check_worker` | 7 | **7/7 PASS** |
| Real cross-host RPC tensor transfer (set_tensor + completion + 4 supervisor bonus stages: watchdog hysteresis, SIGABRT recovery, degraded restart, worker re-join) | 11 | **11/11 PASS** |
| Verification of late-landing fixes (interface narrowing, status-bind default, atomic-write cleanup, plus regression-evidence assertions) | 8 | **8/8 PASS** |

**Total manual on real LAN: 54/54 PASS.**

### Headline measurements

- **Supervisor SIGTERM-to-exit: 0.03s** (vs 10 s pre-fix when polling-based sleep was in place).
- **Worker agent SIGTERM-to-exit: 0.05s** (same `threading.Event.wait()` fix).
- **Crash-to-recovery: ~3s** (worker killed -> llama crashes with SIGABRT exit -6 -> supervisor detects and restarts in degraded mode).
- **Worker-recovery-to-full-cluster: ~3s** (worker comes back -> watchdog notices -> llama restarted with full --rpc list).


Lightheartdevs (changes requested):
This is an ambitious and useful direction, but I don’t think this PR is merge-ready yet. I refreshed against the latest head (`cb88914d`) after the dependency-lock fix landed; CI is now green and `python dream-server/tests/test-dependency-pins.py` passes locally. The remaining blockers are in the worker agent user path and token-state handling.

Findings:

- [P1] `dream cluster agent start --token-file ...` is advertised as the safer path, but the `dream-cli` wrapper drops it. The parser in `dream-server/dream-cli:4188-4195` only handles `--token`, `--port`, `--interface`, and `--controller`; unknown flags are shifted away. On a fresh worker, `--token-file` therefore never initializes `config/cluster-agent.json`, and the start path falls through to `dream-server/dream-cli:4259-4260` with “No agent config.” The lower-level Python agent supports `--token-file`, but the user-facing `dream` command does not pass it through or read it. Please wire `--token-file` through the wrapper/systemd/nohup paths or remove it from the documented workflow.

- [P1] The worker-side token state is not ignored. `dream-server/dream-cli:4226-4254` writes `config/cluster-agent.json` containing the shared token, but `dream-server/.gitignore:35-36` only ignores `config/cluster.json`. If someone runs/testing from a source checkout, the worker token file becomes an untracked file that can be accidentally committed or included in patches. Please ignore `config/cluster-agent.json` and any temp variants, and consider marking `CLUSTER_TOKEN` as `secret: true` in `dream-server/.env.schema.json:1038` for schema-level consistency.

Checks on latest head: `python dream-server/tests/test-dependency-pins.py` passes; `git diff --check origin/main...HEAD` is clean. I did not run the Docker/RPC e2e suites locally from this Windows machine.

y-coffee-dev:
Thanks for the review! both P1s addressed in two commits on top of a fresh rebase onto current main.

- **`--token-file` wired end-to-end** — parser accepts the flag, validates the file, resolves to absolute path, warns on group/world-readable mode (mirrors the agent's own check). Rejects `--token` + `--token-file` combo. Config-init writes `"token_file": "<abspath>"` and omits `"token"` — the secret never lands in `cluster-agent.json`. New `__TOKEN_FILE_FLAG__` placeholder in the systemd unit template; sed substitutes ` --token-file "<path>"` (or empty). Nohup invocation appends the flag the same way `--interface`/`--controller` already do. Subsequent restarts read `token_file` back out of the persisted config so the flag survives a daemon restart. New `tests/test-cluster-cli-token-file.sh` covers the parser, sed render (both populated and bare), config persistence, and both rejection paths. 17/17 passing.

- **gitignore + secret marker** — `.gitignore` ignores `config/cluster-agent.json` + the atomic-write tmp. `.env.schema.json` marks `CLUSTER_TOKEN` as `"secret": true` (matches the shape of `WEBUI_SECRET`/`OPENAI_API_KEY`). `LAN-CLUSTER-GUIDE.md` Token Handling section gets the file-vs-inline distinction; troubleshooting note updated to handle both shapes.

Local `test-dependency-pins.py` passes, `make lint` clean, all four cluster test suites green.

Lightheartdevs:
## Codex audit after current-main refresh

I re-audited this PR against current `main` (`61283b71`). The direction is genuinely useful, and the previous P1 token-handling concerns appear addressed in the branch: `--token-file` is now parsed by `dream-cli`, persisted as `token_file` instead of token contents, carried through the systemd/nohup paths, `config/cluster-agent.json` is ignored, and `CLUSTER_TOKEN` is marked secret in the schema.

That said, I do not think this is merge-ready yet.

### Current blockers

1. **The PR no longer merges cleanly into current `main`.**

   Current merge conflict:
   - `dream-server/docs/README.md`

   It is a small-looking docs-index conflict, but it means the green CI run is stale relative to today?s `main`. Please rebase/merge current `main`, keep both doc-index entries where appropriate (`LAN-CLUSTER-GUIDE.md` and the newer high-risk/validation docs), and rerun CI.

2. **This is too large/high-blast-radius to merge as ordinary feature work.**

   Current diff is roughly 56 files and 8k+ additions, including:
   - new cluster Dockerfiles and compose overlays;
   - new cluster controller/worker/supervisor Python daemons;
   - UDP discovery and signed beacon logic;
   - new dashboard page/hook and dashboard-api cluster route/model surface;
   - ~1.4k lines added to `dream-cli`;
   - systemd worker-agent unit;
   - several e2e/integration harnesses.

   That is real distributed runtime infrastructure. It may be worth merging eventually, but it should not ride along the current stable install/release-green path without a dedicated validation story.

3. **The merge path needs a cluster-specific validation lane.**

   The existing CI being green is good, but this feature?s risk is not fully covered by regular installer smoke. Before merge I?d like to see, in the PR body or a follow-up comment, a current validation matrix for at least:
   - controller enable/disable on Linux NVIDIA;
   - worker agent start with `CLUSTER_TOKEN` env and with `--token-file`;
   - worker join/direct-controller path;
   - no token leakage in `ps`, `/health`, support artifacts, and git status;
   - supervisor restart/degraded/crash fuse behavior;
   - `dream restart` / compose resolver behavior with `CLUSTER_ENABLED=true` and false;
   - dashboard cluster page behavior when cluster is disabled, enabled, and stale/unreachable;
   - explicit unsupported macOS path remains a clean refusal.

### Local validation notes

I tried the focused token-file test locally from Windows Git-Bash. It got through the static parser/systemd/nohup checks, but failed on path/mode assertions because the test is comparing POSIX `/tmp/...` paths and `stat -c` style permissions under this Windows Git-Bash environment. I would not treat that as a product failure, but it does mean I cannot provide clean local approval from this machine. Linux CI or a Linux maintainer run is the relevant signal here.

### Recommendation

Please rebase first, then either:

- split the feature into smaller mergeable PRs (contracts/docs -> CLI/config -> daemons -> dashboard -> e2e), or
- keep this as one feature PR but explicitly mark it experimental/opt-in and attach a fresh cluster validation receipt.

I would not merge this into `main` until the conflict is resolved and the cluster lane is fresh. The feature is promising; the main thing needed now is reducing integration risk and making the operational proof match the size of the change.

Lightheartdevs:
Thanks for taking on LAN cluster inference; the token-file hardening and real-LAN validation notes make the direction much easier to evaluate.

Quickest path to merge:
- Rebase on current `main` first. `git merge-tree origin/main origin/pr-1365` reports a current content conflict in `dream-server/docs/README.md`; the previous green checks are stale until that is resolved.
- Decide whether this should stay as one experimental feature PR or be split. The current diff is still very large for runtime infrastructure: `dream-server/dream-cli`, new cluster Dockerfiles/compose overlays, controller/worker/supervisor daemons, dashboard-api/UI surfaces, systemd unit files, and e2e/integration harnesses.
- Attach a fresh cluster validation receipt after the rebase, including controller enable/disable, worker agent start via `CLUSTER_TOKEN` and `--token-file`, worker join/direct-controller paths, no token leakage in `ps`/health/support artifacts/git status, supervisor degraded/restart/fuse behavior, compose resolver behavior with `CLUSTER_ENABLED=true/false`, dashboard behavior for disabled/enabled/stale clusters, and the unsupported macOS refusal path.

What I checked:
- Current PR description, prior review discussion, changed-file surface, and CI/check status.
- `git diff --check origin/main...origin/pr-1365`: passed locally.
- `git diff --stat origin/main...origin/pr-1365`: 56 files and roughly 8k additions, confirming this is high-blast-radius distributed runtime work.
- `git merge-tree origin/main origin/pr-1365`: confirms the current docs-index conflict.

Thanks again for putting substantial validation effort into this.

PR #1391, opened 2026-05-25
[codex] harden desktop installer

Lightheartdevs:
## Summary

Hardens the desktop/Tauri installer trust surface without changing the normal shell/PowerShell installer entrypoints.

## Changes

- Adds a committed `installer/package-lock.json` so the GUI installer frontend can use reproducible `npm ci` installs.
- Removes the unused Tauri shell plugin, broad shell permissions, and shell plugin config from the installer app.
- Replaces the disabled CSP with an explicit self-only policy suitable for the bundled installer UI.
- Runs Windows installs through the root `install.ps1` PowerShell entrypoint instead of relying on `bash`/WSL path resolution.
- Drains installer stderr while stdout is parsed so GUI installs cannot hang on a full stderr pipe, and surfaces stderr tail on failure.

## Validation

- `npm ci` in `installer`
- `npm run build` in `installer`
- `npm audit` in `installer` (0 vulnerabilities)
- Static search for `plugin-shell`, `tauri_plugin_shell`, `shell:allow`, `csp: null`, and `Command::new("bash")` under `installer` returned no matches
- `git diff --check`

## Notes

Rust/Tauri compile validation was not run locally because `cargo` is not installed in this Windows workspace. GitHub CI should cover this branch.

### DIFF
--- a/installer/package-lock.json
+++ b/installer/package-lock.json
@@ -1,0 +1,3063 @@
+{
+  "name": "dreamserver-installer",
+  "version": "0.1.0",
+  "lockfileVersion": 3,
+  "requires": true,
+  "packages": {
+    "": {
+      "name": "dreamserver-installer",
+      "version": "0.1.0",
+      "dependencies": {
+        "@tauri-apps/api": "2.10.1",
+        "react": "^18.3.1",
+        "react-dom": "^18.3.1"
+      },
+      "devDependencies": {
+        "@tauri-apps/cli": "2.10.1",
+        "@types/react": "^18.3.1",
+        "@types/react-dom": "^18.3.1",
+        "@vitejs/plugin-react": "^4.3.4",
+        "autoprefixer": "^10.4.20",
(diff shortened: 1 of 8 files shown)


Lightheartdevs:
Local desktop-installer validation completed on Windows after the hardening changes:\n\n- Installed a local Rust toolchain for validation and WinLibs MinGW for the GNU linker path.\n- Fixed stale Cargo.lock state exposed by cargo --locked.\n- Pinned @tauri-apps/api and @tauri-apps/cli to 2.10.1 so they match the Rust tauri 2.10.x line; the previous floating ^2 resolved to 2.11.x and Tauri refused to build.\n- Ran npm ci, npm run build, npm audit.\n- Ran cargo +stable-x86_64-pc-windows-gnu check --locked.\n- Ran npm run tauri -- build --debug --target x86_64-pc-windows-gnu --no-bundle --ci; produced the debug app successfully.\n\nCI is green again on the amended commit 63666569.

PR #1364, opened 2026-05-24
fix(installer): keep cloud mode off local llama-server

Lightheartdevs:
﻿## Summary

This patches the Linux cloud/external-LLM install path so it no longer starts, waits on, or inherits dependencies on the local `llama-server` container.

Root cause from the user report:
- `resolve-compose-stack.sh` only read `DREAM_MODE` from the environment, but installer callers were passing it as a shell variable rather than a resolver argument.
- `gpu_backend=cpu` was selected before cloud mode, so `--cloud` could still resolve `docker-compose.cpu.yml`.
- Local-mode extension overlays such as `extensions/services/litellm/compose.local.yaml` could still be included, reintroducing `depends_on: llama-server`.
- Phase 12 always waited on and pre-warmed `llama-server`, even when cloud mode should not launch it.
- Hermes SOUL rendering was still nested behind the local-model path, which left cloud installs vulnerable to the file-vs-directory bind mount failure.

Fixes:
- Add `docker-compose.cloud.yml` to profile `llama-server` out of default cloud startup.
- Add `--dream-mode` to the compose resolver and all Linux installer / CLI resolver calls.
- Resolve cloud mode before CPU/local overlays and suppress local dependency overlays for `tier=CLOUD`.
- Regenerate stale `.compose-flags` when a cloud install still points at CPU/local overlays.
- Preserve cloud/external LLM URL settings in `.env` and set Hermes defaults to the LiteLLM route in cloud mode.
- Move Hermes config/SOUL generation outside the local-model-only block.
- Make Phase 12 check LiteLLM in cloud mode and skip local `llama-server` pre-warm.
- Add `tests/test-linux-cloud-mode.sh` and wire it into `make test`.

## Validation

Passed:
- `bash -n scripts/resolve-compose-stack.sh installers/lib/compose-select.sh installers/phases/03-features.sh installers/phases/06-directories.sh installers/phases/11-services.sh installers/phases/12-health.sh dream-cli tests/test-linux-cloud-mode.sh`
- `DREAM_PYTHON_CMD=python bash tests/test-linux-cloud-mode.sh`
- `DREAM_PYTHON_CMD=python DREAM_MODE=cloud ./scripts/resolve-compose-stack.sh --script-dir . --tier CLOUD --gpu-backend cpu --gpu-count 0`
[...]

PR #1445, opened 2026-05-25
feat(doctor): diagnose inference routing contract

gabsprogrammer:
## Summary

Adds an inference routing contract to `dream doctor` and the redacted support bundle so support reports can tell whether DreamServer's installed mode, compose stack, and LLM URLs agree about who owns inference.

Important context: #1364 already fixed the Linux cloud runtime path so current installs should not start, wait on, or route through local `llama-server` in cloud mode. This PR does **not** re-fix that runtime path. It adds evidence-based diagnostics and tightens support-bundle evidence so stale installs, hand-edited `.env` files, old `.compose-flags`, external Lemonade drift, WSL/Linux/macOS shell differences, or future regressions are visible from one bundle.

The contract flags violations such as:

- `DREAM_MODE=cloud` but cached `.compose-flags` does not include `docker-compose.cloud.yml`
- cloud mode still routes `LLM_API_URL` to local `llama-server`
- cloud mode still routes Hermes to local `llama-server`
- external Lemonade is active without the cloud/external overlays
- local mode accidentally keeps the cloud overlay or a non-AMD LiteLLM route

## What changed

- Adds `runtime.inference_contract` to the doctor JSON report with selected mode, expected inference owner/gateway, relevant LLM URLs, compose files, stable issue IDs, and blocker/warning counts.
- Promotes contract violations into evidence-ranked `diagnoses` entries with recovery hints.
- Avoids false positives for overlay mismatch checks when `.compose-flags` is absent rather than stale.
- Displays the contract in `dream doctor` human output.
- Adds the contract snapshot to support bundle `manifest/evidence.json`.
- Fixes support-bundle compose evidence so it resolves the stack with recorded `DREAM_MODE`, external Lemonade flags, and AMD runtime ownership instead of always behaving like a local install.
- Uses the selected Bash for nested support-bundle diagnostics, compose resolution, and compose validation, and records the selected Bash in the bundle. This keeps Linux/WSL behavior normal while avoiding macOS Bash 3.2 child-script drift when a newer Bash is available.
- Records WSL detection in support-bundle evidence so Windows/WSL support cases are distinguishable from native Linux.
- Documents the contract, stable diagnosis IDs, and bundle evidence behavior in `docs/DREAM-DOCTOR.md`.

## Why

Mode bugs are currently easy to discuss abstractly and hard to support from evidence. #1364 fixed the active cloud installer behavior; this PR makes that behavior auditable after install. Local inference means DreamServer manages/targets `llama-server`; cloud/private/external inference should route Dream services through the external gateway path instead of accidentally targeting local inference.

The support bundle part is necessary because a failing user usually shares bundle evidence, not just a live terminal. If the bundle resolves Compose as local while the user configured cloud or external Lemonade, support can chase the wrong stack. If macOS runs doctor children through an older Bash, the doctor JSON can be missing even though the support script itself started successfully.

## Validation

- `bash -n dream-server/scripts/dream-doctor.sh dream-server/scripts/dream-support-bundle.sh dream-server/tests/test-support-bundle.sh dream-server/tests/test-linux-cloud-mode.sh`
- `bash dream-server/tests/test-support-bundle.sh`
- `bash dream-server/tests/test-dream-doctor-install-diagnoses.sh`
- `bash dream-server/tests/test-dream-doctor.sh`
- `bash dream-server/tests/test-doctor-command.sh`
- `bash dream-server/tests/contracts/test-external-lemonade-contracts.sh`
- `git diff --check`

### DIFF
--- a/dream-server/docs/DREAM-DOCTOR.md
+++ b/dream-server/docs/DREAM-DOCTOR.md
@@ -45,6 +45,7 @@ Runtime Environment:
   ✓ Docker Compose
   ✗ Dashboard HTTP
   ✗ WebUI HTTP
+  ✓ Inference contract: mode=local, owner=dreamserver, gateway=llama-server
   ⚠ DGX Spark llama-server CUDA arch: DGX Spark detected, but llama-server reports CUDA archs '500,610,700,750,800,860,890,1200' without sm_121.
 
 Preflight Checks:
@@ -87,6 +88,10 @@ dream doctor --json > report.json
   - `impact`: why the issue matters
   - `next_steps`: concrete recovery actions
 - **runtime**: Docker/Compose/UI reachability checks
+- **runtime.inference_contract**: A compact routing contract for the active
+  inference mode. It records `DREAM_MODE`, expected inference owner
+  (`dreamserver` vs `external`), expected gateway (`llama-server` vs
+  `litellm`), key LLM URLs, resolved compose files, and stable mismatch IDs.
 - **runtime.amd_runtime**: Explicit AMD inference runtime diagnostics from
   installer-written env state. Reports runtime (`lemonade` or `llama-server`),
   host/container location, selected backend, supported backends, DreamServer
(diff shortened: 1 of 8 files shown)


gabsprogrammer:
Updated this PR after another pass focused on whether the change is actually useful in support cases, not just theoretically nice to have.

The main point is that #1364 already fixed the active Linux cloud runtime path. This PR is meant to make that behavior auditable after install, especially when a user shares a support bundle from a stale install, a hand-edited `.env`, an old `.compose-flags`, or an external/private inference setup.

I tightened the PR in a few concrete ways:

- overlay mismatch checks now only fire when cached `.compose-flags` exists, so a missing cache does not become a false positive;
- support bundle Compose evidence now resolves with the recorded `DREAM_MODE`, external Lemonade flags, and AMD runtime ownership, instead of accidentally behaving like a local install;
- external Lemonade is covered so the bundle proves it uses cloud + external overlays and does not pull in local llama-server overlays;
- the bundle records WSL detection so Windows/WSL support cases are distinguishable from native Linux;
- nested diagnostics/compose validation use the selected Bash, which keeps Linux/WSL behavior unchanged and avoids macOS support bundles silently drifting into the system Bash path when a newer Bash is available.

Validation after the update:

- `bash dream-server/tests/test-support-bundle.sh`
- `bash dream-server/tests/test-dream-doctor-install-diagnoses.sh`
- `bash dream-server/tests/test-dream-doctor.sh`
- `bash dream-server/tests/test-doctor-command.sh`
- `bash dream-server/tests/contracts/test-external-lemonade-contracts.sh`
- `git diff --check`

The remote CI is also green across shell lint, Linux distro smoke, macOS smoke, Windows PowerShell lint, API/frontend, and integration smoke.

So the PR is no longer just "doctor reports a mismatch". It makes the support bundle carry the same inference-routing contract evidence that support needs to debug cloud/private/external setups without guessing which stack the user actually intended to run.

Lightheartdevs:
## Codex audit notes

I reviewed this against current `main` (`ed1c39d6`, after the Hermes timeout fix) and did a local merge-context check.

What I checked:
- PR head merges cleanly with current `main`.
- Remote CI is green.
- `git diff --check origin/main...HEAD` passes after the merge-context check.
- `bash -n` passes for the touched doctor/support-bundle shell scripts and focused tests.
- `bash dream-server/tests/test-dream-doctor.sh` passes locally: 5 passed, JSON/jq assertions skipped because this Windows Git-Bash environment has no `jq`.
- `bash dream-server/tests/test-doctor-command.sh` passes locally: 16 passed.
- `DREAM_PYTHON_CMD=python bash dream-server/tests/test-linux-cloud-mode.sh` passes locally.
- `DREAM_PYTHON_CMD=python bash dream-server/tests/contracts/test-external-lemonade-contracts.sh` passes locally.

One local caveat: `test-support-bundle.sh` hits a Windows Git-Bash `tar` path issue (`Cannot connect to C: resolve failed`) when reading the generated archive path. That fails before exercising the PR's inference-contract logic and is not representative of the Linux support-bundle path covered by CI.

Assessment: this is low-risk and useful. It does not change install/startup behavior; it makes `dream doctor` and support bundles carry explicit inference-routing evidence. That is exactly the kind of forensic detail we wanted after the cloud/external-Lemonade compose issues: it lets support distinguish stale `.compose-flags`, hand-edited `.env`, cloud mode, local mode, and external Lemonade without guessing.

Recommendation: mergeable after normal review. If you want the cleanest history, rebase onto current `main` and let CI rerun, but I did not find a blocking issue.

Lightheartdevs (approved):
Approved after re-audit against current main. The PR is diagnostic/support-bundle scoped, merges cleanly, focused doctor/cloud/external-Lemonade tests pass locally, and a fixture check confirms the new inference contract emits the expected cloud/local mismatch IDs. No install/startup behavior change found.

PR #1469, opened 2026-05-26
[codex] fix dashboard API proxy after lifecycle churn

Lightheartdevs:
## Summary
- make dashboard nginx resolve `dashboard-api` through Docker DNS at request time
- switch `/api/*` and `/api/talk/*` proxying to a variable upstream so nginx does not pin a stale dashboard-api container IP
- add a contract guard that fails if dashboard nginx goes back to config-load DNS resolution

## Why
Fleet run `/home/michael/dream-fleet-test/runs/2026-05-26T02-27-29Z` showed post-lifecycle `502 Bad Gateway` from the dashboard frontend on Spark and tower2 while direct dashboard-api requests still reached the API. Logs showed nginx proxying to a stale Docker IP after dashboard-api lifecycle churn.

## Validation
- `git diff --check`
- `python scripts/check-version-consistency.py`
- `python scripts/validate-generated-configs.py`
- Git Bash syntax check for `tests/contracts/test-installer-contracts.sh`
- attempted `nginx:alpine nginx -t`, but local Docker image pull failed twice with a CloudFront EOF before nginx could run

### DIFF
--- a/dream-server/extensions/services/dashboard/nginx.conf
+++ b/dream-server/extensions/services/dashboard/nginx.conf
@@ -5,6 +5,13 @@ server {
     root /usr/share/nginx/html;
     index index.html;
 
+    # Docker service IPs can change when lifecycle commands recreate
+    # dashboard-api. Use Docker DNS at request time instead of pinning the IP
+    # nginx resolved when this container started.
+    resolver 127.0.0.11 valid=10s ipv6=off;
+    resolver_timeout 5s;
+    set $dashboard_api_upstream dashboard-api:3002;
+
     # Gzip compression
     gzip on;
     gzip_types text/plain text/css application/json application/javascript text/xml application/xml;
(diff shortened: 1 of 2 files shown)


Lightheartdevs:
Audit result: safe to merge.

This directly addresses the lifecycle churn failure from the fleet follow-up: dashboard nginx was resolving `dashboard-api` once at config load, so after `dashboard-api` was recreated it could keep proxying to a stale container IP and return 502 while the API itself was healthy on the Docker network.

Why I think this is low risk:
- dashboard-only nginx config change; no installer phase, model, or runtime selection behavior changes
- preserves the existing `/api/` and `/api/talk/` proxy paths, headers, auth injection, SSE buffering settings, and timeouts
- uses Docker's embedded DNS resolver (`127.0.0.11`) with variable `proxy_pass`, which is the standard nginx pattern for re-resolving Docker service names
- adds a contract guard so this cannot quietly regress to config-load IP pinning

Validation performed locally:
- `python scripts/validate-generated-configs.py`
- `python scripts/check-version-consistency.py`
- Git Bash syntax check for the contract script
- `git diff --check`

Caveat: GitHub reported no checks for this PR, and local nginx image validation was not available in this Windows environment. The config change is narrow and mechanically safe enough to merge under the current release pressure.

PR #1492, opened 2026-06-02
fix(lemonade): harden existing runtime installs

Lightheartdevs:
## Summary
- Auto-detect existing Lemonade on common ports (`13305`, then `8000`) when `--lemonade-url` is omitted.
- Auto-detect the first served Lemonade model from `/api/v1/models` and remove the stale `LLM_MODEL_VALUE` fallback.
- Skip pulling Dream's managed Lemonade image when external Lemonade owns inference.
- Add a Phase 12 real completion check through LiteLLM so broken external Lemonade routing fails clearly instead of finishing false-green.
- Update the Lemonade compatibility docs and README to match the new behavior.

## Why
External Lemonade users can have an already-running Lemonade runtime but still end up with a broken Dream LLM route if Dream guesses the wrong model, talks to the wrong port, or if Docker containers cannot reach the host Lemonade server. This makes that path explicit and verified.

## Validation
- `bash -n install-core.sh installers/phases/06-directories.sh installers/phases/08-images.sh installers/phases/12-health.sh tests/contracts/test-external-lemonade-contracts.sh`
- `DREAM_PYTHON_CMD=python bash tests/contracts/test-external-lemonade-contracts.sh`
- `python tests/test-render-runtime-configs.py`

## Notes
- Docker Desktop is not running on this Windows workstation, so I could not run `tests/fleet-external-lemonade-e2e.sh --mock` locally.
- The new Phase 12 check is the live-install receipt: it calls LiteLLM `/v1/chat/completions`, which forces LiteLLM inside Docker to reach the external Lemonade service and selected model.

### DIFF
--- a/dream-server/README.md
+++ b/dream-server/README.md
@@ -58,7 +58,9 @@ On Linux Docker installs, llama-server is exposed to the host on **http://localh
 
 On Linux AMD hosts already running Lemonade SDK, install Dream Server around it with
 `./install.sh --use-existing-lemonade` so Dream manages the app stack while
-Lemonade keeps owning inference and model storage. See
+Lemonade keeps owning inference and model storage. The installer auto-detects
+common Lemonade ports and the first served model, then verifies a real completion
+through LiteLLM before declaring success. See
 [docs/LEMONADE-SDK-COMPAT.md](docs/LEMONADE-SDK-COMPAT.md). Existing Lemonade
 mode only reuses Lemonade for LLM inference; Full Stack still enables
 Dream-managed Whisper, Kokoro, and ComfyUI unless you pass `--no-voice` and/or
--- a/dream-server/docs/LEMONADE-SDK-COMPAT.md
+++ b/dream-server/docs/LEMONADE-SDK-COMPAT.md
@@ -18,6 +18,13 @@ If Lemonade is not using its default URL, pass it explicitly:
 ./install.sh --use-existing-lemonade --lemonade-url http://localhost:13305
 ```
 
+When `--lemonade-url` is omitted, Dream Server checks `http://localhost:13305`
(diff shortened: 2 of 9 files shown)


Lightheartdevs:
Audit result: ready from my side.

I reviewed this as an installer/runtime stability change, not just a docs tweak. The important paths look scoped correctly:

- Existing Lemonade mode auto-detects common Lemonade ports and discovers the served model id from `/api/v1/models`.
- Managed Lemonade image pulls are skipped only when `LEMONADE_EXTERNAL` is true.
- Phase 12 now verifies a real LiteLLM chat completion through the external Lemonade route, so a broken external route cannot finish false-green.
- The audit follow-up fixes `dream-preflight.sh` so external Lemonade installs check `dream-litellm` / LiteLLM readiness instead of requiring the intentionally-absent managed `dream-llama-server` container.
- Docs now make the Full Stack scope clear: Dream may still start Whisper, Kokoro, and ComfyUI unless users pass `--no-voice` / `--no-comfyui` or choose alternate supported ports.

Validation I checked:

- CI green on the updated head SHA.
- `git diff --check` passed.
- Shell syntax checks passed for the touched installer/preflight scripts.
- `tests/contracts/test-external-lemonade-contracts.sh` passed.
- `tests/test-preflight.sh` passed.
- `tests/test-render-runtime-configs.py` passed.
- Real Strix Halo existing-Lemonade install evidence showed auto-detect, model discovery, no managed Lemonade container, Phase 12 external completion success, and a successful LiteLLM chat completion with zero manual edits.

No blocking issues found. The remaining merge blocker is procedural: the `Safe Keeping` ruleset requires two approving reviews from write-access reviewers and cannot be bypassed.

PR #1493, opened 2026-06-03
fix(uninstall): use resolved compose stack

gabsprogrammer:
## Summary
- make `dream-uninstall.sh` resolve Dream Server's actual compose stack before cleanup instead of calling bare `docker compose down`
- respect `--keep-data` by avoiding compose volume deletion and fallback volume cleanup
- update Linux/Windows uninstall docs to use the official uninstaller or saved `.compose-flags`
- add a regression test for compose-flag cleanup and `--keep-data` behavior

## Why
Dream Server does not ship a top-level `docker-compose.yml`; installs persist the resolved stack in `.compose-flags`. The documented/manual `docker compose down -v` path can fail with `no configuration file provided: not found` and leave users unsure which containers/volumes belong to Dream Server.

## Validation
- `bash -n dream-server/dream-uninstall.sh dream-server/tests/test-uninstall-compose-flags.sh`
- `bash dream-server/tests/test-uninstall-compose-flags.sh`
- `git diff --check`

### DIFF
--- a/.github/workflows/test-linux.yml
+++ b/.github/workflows/test-linux.yml
@@ -63,6 +63,9 @@ jobs:
       - name: Compose Failure Report Tests
         run: bash tests/test-compose-failure-report.sh
 
+      - name: Uninstall Compose Flags Tests
+        run: bash tests/test-uninstall-compose-flags.sh
+
       - name: Windows Compose Failure Report Tests
         run: bash tests/test-windows-compose-failure-report.sh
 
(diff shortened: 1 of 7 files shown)


Lightheartdevs (approved):
Reviewed after maintainer safety fixes. The PR now uses the resolved compose stack for uninstall cleanup, preserves --keep-data volume behavior, avoids direct .env sourcing by loading through lib/safe-env.sh, and wires the regression into Linux CI plus release-gate. Local focused tests and CI are green.

PR #1494, opened 2026-06-03
fix(installer): harden external Lemonade model and bind selection

gabsprogrammer:
## Summary

Fixes three installer edge cases surfaced by an external Lemonade install using:

```bash
./install.sh --use-existing-lemonade --no-voice --no-comfyui
```

The reported install reached LiteLLM and the external Lemonade server, but failed Phase 12 with no assistant content because the selected Lemonade model was `Flux-2-Klein-9B-GGUF`, which is an image-generation model rather than a chat/text model. The same report also noted that a requested LAN bind still came up localhost-only after install.

## What changed

- External Lemonade model auto-detection now skips obvious image-generation model ids such as Flux/SDXL/Stable Diffusion when another model is available from `/api/v1/models`.
- Phase 12 now gives a specific recovery hint when the selected Lemonade model looks like an image/non-chat model and the LiteLLM chat completion returns no assistant content.
- `--lan` and exported `BIND_ADDRESS=0.0.0.0` are now treated as explicit install-time choices and can override a stale existing `.env` value during reinstall.
- Custom feature prompts now reflect the current flag value, so `--no-voice` and `--no-comfyui` show `[y/N]` instead of visually defaulting back to yes.
- Lemonade compatibility docs now call out that `LEMONADE_MODEL` must be a text/chat model, not an image model.

## Why

This was easy to misread as a ComfyUI problem because the selected model was Flux and the user had disabled ComfyUI. The actual failing path is the LLM chat route: DreamServer verifies `/v1/chat/completions` through LiteLLM, and an image model can be reachable while still returning no assistant message content.

The bind behavior is separate but came from the same install report. On reruns, Phase 06 preserved `BIND_ADDRESS` from an existing `.env` before considering the current explicit `--lan`/environment request, so the dashboard could remain localhost-only even after the operator opted into LAN binding.

## Validation

- `bash -n dream-server/install-core.sh dream-server/installers/phases/03-features.sh dream-server/installers/phases/06-directories.sh dream-server/installers/phases/12-health.sh dream-server/tests/contracts/test-external-lemonade-contracts.sh`
- `bash dream-server/tests/contracts/test-external-lemonade-contracts.sh`
- `bash dream-server/tests/test-network-timeouts.sh`
- `bash dream-server/tests/test-bind-address-sweep.sh`
- `bash dream-server/tests/contracts/test-installer-hardening.sh`
- `git diff --check`

Local note: `tests/smoke/installer-env-smoke.sh` passed through env generation, schema validation, dependency validation, compose validation, and the first function-resolution check

### DIFF
--- a/dream-server/docs/LEMONADE-SDK-COMPAT.md
+++ b/dream-server/docs/LEMONADE-SDK-COMPAT.md
@@ -88,7 +88,8 @@ service.
 ## Model Selection
 
 Dream Server auto-detects the first model id returned by Lemonade's
-`/api/v1/models` endpoint and writes it to `LEMONADE_MODEL`.
+`/api/v1/models` endpoint that does not look like an image-generation model
+and writes it to `LEMONADE_MODEL`.
 
 Set `LEMONADE_MODEL` only if you want Dream Server to use a specific served
 model:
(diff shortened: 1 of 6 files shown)


Lightheartdevs (approved):
Reviewed for installer/new-install stability. This fixes the external Lemonade report by avoiding obvious image/non-chat models during auto-selection, adding a specific Phase 12 remediation for no assistant content, preserving explicit --lan/BIND_ADDRESS over stale .env on reinstall, and making custom feature prompts reflect current flag values. CI is green; local focused checks passed: external Lemonade contracts, bind-address sweep, network timeouts, doc links, and installer hardening contracts. Broader than #1495 but scoped to installer/external-Lemonade behavior and ready to merge.

PR #1495, opened 2026-06-03
fix(cli): tolerate UID in safe env loading

gabsprogrammer:
## Summary
- skip Bash's readonly `UID` variable when loading `.env` through `lib/safe-env.sh`
- keep later `.env` keys loading normally after a `UID=...` line
- add a regression test for lifecycle commands that load `.env` before reaching Docker Compose

## Why
A user reported `dream restart` failing with:

```text
/home/phil/dream-server/lib/safe-env.sh: line 38: UID: readonly variable
```

`UID=...` is a valid Docker Compose-style setting in `.env`, but Bash exposes `UID` as a readonly shell variable. Exporting it from `load_env_file` can abort `dream restart`, `dream doctor`, and other lifecycle commands under `set -e` before they can reach compose. Docker Compose can still read `UID` from `.env`; the shell loader just should not try to re-export it.

The follow-up `sudo dream restart` failure is separate: sudo changes `$HOME` to `/root`, so the CLI looks for `/root/dream-server` unless `DREAM_HOME` is preserved or set explicitly.

## Validation
- `bash -n dream-server/lib/safe-env.sh dream-server/tests/test-safe-env.sh`
- `bash dream-server/tests/test-safe-env.sh`
- `git diff --check`

### DIFF
--- a/dream-server/lib/safe-env.sh
+++ b/dream-server/lib/safe-env.sh
@@ -29,6 +29,10 @@ load_env_file() {
         key="${key%"${key##*[![:space:]]}"}"
         [[ -z "$key" ]] && continue
         [[ "$key" =~ ^[A-Za-z_][A-Za-z0-9_]*$ ]] || continue
+        # Bash exposes UID as a readonly shell variable. A .env line such as
+        # UID=1000 is valid for Docker Compose, but exporting it here aborts
+        # lifecycle commands under set -e before they can reach compose.
+        [[ "$key" == "UID" ]] && continue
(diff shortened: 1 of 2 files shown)


Lightheartdevs (approved):
Reviewed for lifecycle/new-install stability. This is a narrow safe-env fix: .env UID is skipped because Bash exposes UID as readonly, while later keys continue loading normally. CI is green and local safe-env regression passed. Low risk and solves the reported dream restart/doctor failure mode.

PR #1498, opened 2026-06-03
fix(update): resolve runtime compose stacks in health checks

gabsprogrammer:
## Summary
- make `dream-update.sh health` resolve the active compose stack from `.compose-flags`, `scripts/resolve-compose-stack.sh`, or safe compose fallbacks before calling `docker compose ps`
- make non-git runtime installs fail with actionable update guidance instead of the vague `Not a git repository. Manual update required.`
- read health ports from `.env` when they are not exported in the current shell
- harden `docker compose ps --format json` parsing for both object and array JSON shapes

## Why
A field install reported:

```text
./dream-update.sh update
[ERROR] Not a git repository. Manual update required.
./dream-update.sh health
[WARN] No services defined in docker-compose
```

The install had already created a rollback snapshot, but the update path then failed because the install directory was not a git checkout. The follow-up health check also used bare `docker compose ps`, so it could miss Dream Server installs that rely on `docker-compose.base.yml` plus overlays cached in `.compose-flags`.

This keeps source-code updates git-backed, but makes the failure explicit and points routine runtime/image updates at `./dream-cli update`.

## Validation
- `bash dream-server/tests/test-dream-update-runtime-compose.sh`
- `bash -n dream-server/dream-update.sh dream-server/tests/test-dream-update-runtime-compose.sh`
- `bash dream-server/tests/contracts/test-installer-contracts.sh`
- `git diff --check`

Local note: `tests/test-update-rollback-contract.sh` still uses GNU `find -printf`, so it is not meaningful on this macOS host without GNU find. This change adds a targeted runtime-compose updater test instead.

### DIFF
--- a/.github/workflows/matrix-smoke.yml
+++ b/.github/workflows/matrix-smoke.yml
@@ -80,15 +80,42 @@ jobs:
       - name: Install checkout prerequisites (distro-aware)
         working-directory: /
         run: |
+          retry() {
+            attempt=1
+            max=3
+            delay=15
+            until "$@"; do
+              status=$?
+              if [ "$attempt" -ge "$max" ]; then
+                return "$status"
+              fi
+              echo "Command failed (attempt ${attempt}/${max}): $*"
+              echo "Retrying in ${delay}s..."
+              sleep "$delay"
+              attempt=$((attempt + 1))
+            done
+          }
+
+          prepare_pacman_keyrings() {
+            command -v pacman-key >/dev/null 2>&1 || return 0
+            pacman-key --init || true
+            for ring in archlinux manjaro cachyos; do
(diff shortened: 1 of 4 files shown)


gabsprogrammer:
CI note: the only failing check I see is `distro: opensuse-tw`, and it failed before checkout/tests while `zypper refresh` was fetching openSUSE repository metadata. The log shows a `Curl error (28)` timeout against `download.opensuse.org` / `cdn.opensuse.org` during prerequisite installation, so this does not appear related to the updater change.

All code-running checks completed green, including shell lint, API, frontend, integration-smoke, Linux/macOS smoke, PowerShell lint, env validations, and the other distro jobs.

Issue #1490, opened 2026-06-02
[Bug] Dashboard consuming 60-80% CPU


th3cavalry:
### Description

The dashboard uses 60-80% CPU usage when opened.

I reinstalled CachyOS and then installed Dream Server.

### Steps to Reproduce

Open dashboard in Chrome.

### Expected Behavior

Use less CPU.

### Actual Behavior

Uses 60% or more CPU even when everything is idle and the dashboard is just sitting there open.

### Operating System

CachyOS

### GPU

Strix Halo 8060

### Docker Version

Current version

### VRAM

128GB

### Logs

```shell

```

Lightheartdevs:
Posted the first low-risk mitigation in #1499 and merged it to `main`.

What changed:
- The dashboard's liquid-metal border/progress/logo effects no longer animate continuously while the page is idle.
- The same visual treatment remains, but motion is now limited to hover/focus interactions.
- `prefers-reduced-motion` now also disables the pseudo-element shimmer layer.
- This is CSS-only: no installer, backend, Docker, model, Lemonade, or new-install behavior changed.

What we are testing with this change:
- Whether the reported idle 60-80% CPU is primarily Chrome renderer/GPU work from continuous dashboard animations.
- If this was the main cause, CPU should drop noticeably when the dashboard is open and idle.
- If CPU remains high, the next thing to distinguish is browser-side rendering vs dashboard API polling.

Useful retest data:
1. Open the dashboard and let it sit idle for ~60 seconds.
2. In Chrome Task Manager (`Shift+Esc`), check whether the Dream Server tab / GPU process is still high CPU.
3. In a terminal, run:

```bash
docker stats --no-stream dream-dashboard dream-dashboard-api
```

That tells us whether the load is in Chrome rendering or in the dashboard containers. If Chrome is still hot after #1499, we should keep reducing frontend motion/paint cost. If `dream-dashboard-api` is hot, we should look next at `/api/status`, service resources, and GPU/status polling frequency.

Leaving this issue open until we get reporter confirmation on the CachyOS + Strix Halo machine.

PR #1499, opened 2026-06-03
fix(dashboard): reduce liquid metal idle motion

Lightheartdevs:
## Summary

- stops liquid-metal border, progress, and logo animations from running continuously while the dashboard is idle
- keeps the liquid-metal look and re-enables motion on hover/focus so the UI still feels responsive during interaction
- strengthens `prefers-reduced-motion` coverage so pseudo-element shimmer animations are also disabled

## Why

Issue #1490 reports high CPU usage when the dashboard is simply open and idle. The dashboard already avoids polling hidden tabs, so the lowest-risk first mitigation is to remove always-on CSS motion from the visible dashboard without touching backend polling, installer behavior, service manifests, or runtime paths.

This does not claim to close every possible CPU source; it gives affected users a quieter idle dashboard and keeps #1490 available for follow-up evidence if Chrome renderer or dashboard-api CPU remains high.

## Validation

- `npm.cmd test` — 20 files / 122 tests passed
- `npm.cmd run build` — Vite production build passed
- `git diff --check`

Refs #1490

### DIFF
--- a/dream-server/extensions/services/dashboard/src/index.css
+++ b/dream-server/extensions/services/dashboard/src/index.css
@@ -419,11 +419,12 @@ a, button, input, select, textarea,
     );
   background-size: 240% 240%;
   background-repeat: no-repeat;
+  background-position: 180% 0%;
   -webkit-mask: linear-gradient(#000 0 0) content-box, linear-gradient(#000 0 0);
   -webkit-mask-composite: xor;
   mask: linear-gradient(#000 0 0) content-box, linear-gradient(#000 0 0);
   mask-composite: exclude;
-  animation: liquidBorderOrbit 11s linear infinite;
+  animation: none;
   opacity: 0.28;
   z-index: 0;
   transition:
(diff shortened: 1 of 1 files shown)


Lightheartdevs:
Audit result: safe to merge from a code/stability perspective.

What I checked:
- Diff is limited to `dream-server/extensions/services/dashboard/src/index.css`.
- No installer, backend, service manifest, compose, model, Lemonade, or new-install path changes.
- Continuous liquid-metal border/progress/logo animations are now static at idle and only animate on hover/focus.
- `prefers-reduced-motion` coverage now includes the pseudo-element shimmer layer.
- Local validation passed: `npm.cmd test` (20 files / 122 tests), `npm.cmd run build`, and `git diff --check`.
- PR CI is green across dashboard, smoke, lint, env validation, secret scan, and distro matrix.

Risk read:
- Very low runtime/install risk because this is CSS-only.
- It should reduce idle Chrome renderer/GPU work for #1490, but I would treat it as the first mitigation rather than closing the issue until the reporter confirms CPU drops on their CachyOS + Strix Halo machine.

Merge note: attempted normal squash merge and admin squash merge from CLI. GitHub ruleset still blocks with “At least 2 approving reviews are required by reviewers with write access,” so this needs the required approvals or a repo ruleset bypass by an allowed actor.

PR #1500, opened 2026-06-03
fix(installer): make Lemonade retry hint runnable

gabsprogrammer:
## Summary
- make the external Lemonade Phase 12 retry hint choose an available installer entrypoint
- fall back to `bash install-core.sh` when `install.sh` is absent from a copied runtime tree
- show `LEMONADE_MODEL=<chat-model-id>` inline so the model override is inherited by the retry command
- let explicit Lemonade retry inputs (`LEMONADE_MODEL`, `LEMONADE_BASE_URL`, `LEMONADE_API_BASE_PATH`, `LEMONADE_CONTAINER_BASE_URL`) override stale values preserved in `.env` during reinstall
- add contract coverage for the runtime-tree fallback, inline model assignment, and explicit Lemonade overrides

## Validation
- `bash -n installers/phases/06-directories.sh installers/phases/12-health.sh tests/contracts/test-external-lemonade-contracts.sh`
- `DREAM_PYTHON_CMD=python3 bash tests/contracts/test-external-lemonade-contracts.sh`
- `git diff --check`

### DIFF
--- a/dream-server/installers/phases/06-directories.sh
+++ b/dream-server/installers/phases/06-directories.sh
@@ -268,6 +268,16 @@ Fix with: sudo chown -R \$(id -u):\$(id -g) $INSTALL_DIR/config $INSTALL_DIR/dat
         echo "$default"
     }
 
+    _env_get_explicit_first() {
+        local key="$1" default="${2:-}" val
+        val="${!key-}"
+        if [[ -n "$val" ]]; then
+            echo "$val"
+            return
+        fi
+        _env_get "$key" "$default"
+    }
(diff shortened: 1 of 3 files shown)


Lightheartdevs:
Landed on main via admin bypass direct push as d3025107 after the PR merge API still enforced the two-write-approver ruleset despite --admin. The landed tree matches this PR's patch: external Lemonade retry hints now use an available entrypoint, explicit Lemonade env overrides beat stale .env values, and the contract test covers the recovery guidance.

PR #1502, opened 2026-06-03
fix(dashboard): hide api-only litellm quicklink

gabsprogrammer:
## Summary
- add a manifest-level `external_link: false` flag for API-only services
- make `/api/external-links` omit services that opt out of dashboard quicklinks
- mark LiteLLM as API-only for external quicklinks while keeping it in health/status/routing metadata
- document that Dream Server provisions LiteLLM as an OpenAI-compatible API gateway, not as the supported LiteLLM Admin UI login surface
- add dashboard-api coverage so API-only services are excluded while normal user-facing services remain linked

## Why
A Windows support report confirmed LiteLLM API auth was healthy: `LITELLM_KEY` from `.env` was passed into the container as `LITELLM_MASTER_KEY`, and `/v1/models` returned models with the Bearer token.

The remaining confusion came from `http://localhost:4000/ui/`: the LiteLLM upstream image exposes an Admin UI route, and our manifest advertised `ui_path: /ui/`, so Dream Server could present it like a normal quicklink. Dream Server's default LiteLLM stack is API-gateway-only and does not provision the database/UI management flow expected by the Admin UI. The dashboard should not steer users toward that unsupported login surface.

## Validation
- `python -m pytest dream-server/extensions/services/dashboard-api/tests/test_main.py -q`
- `python dream-server/scripts/audit-extensions.py --project-dir dream-server`
- `git diff --check`

Local note: `bash dream-server/tests/test-validate-manifests.sh` could not complete on this Windows environment because `jq` is not installed locally; the Python extension audit passed with the updated manifest.

### DIFF
--- a/dream-server/extensions/schema/service-manifest.v1.json
+++ b/dream-server/extensions/schema/service-manifest.v1.json
@@ -82,6 +82,10 @@
           "pattern": "^/.*",
           "description": "Path to the service UI (e.g. /dashboard, /)"
         },
+        "external_link": {
+          "type": "boolean",
+          "description": "When false, omit this service from dashboard external quicklinks while keeping it in health/status APIs."
+        },
         "type": { "type": "string", "enum": ["docker", "host-systemd"] },
         "startup_check": {
           "type": "boolean",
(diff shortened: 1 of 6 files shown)


scottrohrig:
does this also prevent it from displaying in the integrations view? nevermind. it doesn't go to /ui there

PR #1506, opened 2026-06-03
fix(windows): resolve runnable python for agent helpers

gabsprogrammer:
## Summary
- add a shared Windows Python resolver for installer and CLI helpers
- probe candidates by executing Python 3.8+ code instead of trusting PATH aliases
- support `python`, `python3`, `py -3`, `DREAM_PYTHON`, and common installed Python directories
- use the resolver for Hermes SOUL rendering, `.env` schema validation, and Dream host agent startup
- preserve launcher arguments when starting the host agent through hidden PowerShell/Scheduled Task

## Why
On Windows, `python3.exe` can be a Microsoft Store execution alias or may be absent even when a valid Python install exists through `python.exe`, `py -3`, or a standard Python install directory. The old host-agent path only checked `python3`/`python` with `Get-Command`, which can either miss a usable interpreter or accept a command that fails at runtime.

That matters for the dashboard degraded-service path because the Dream host agent depends on Python. If the agent cannot start, service/runtime health can look degraded even when containers are mostly running.

## Validation
- `python -m pytest dream-server/tests/test-windows-python-resolver.py -q`
- PowerShell parser check for:
  - `dream-server/installers/windows/lib/python-resolver.ps1`
  - `dream-server/installers/windows/install-windows.ps1`
  - `dream-server/installers/windows/dream.ps1`
  - `dream-server/installers/windows/phases/06-directories.ps1`
  - `dream-server/installers/windows/phases/07-devtools.ps1`
- live resolver smoke test on Windows: resolved Python 3.11.0
- `python dream-server/scripts/audit-extensions.py --project-dir dream-server`
- `git diff --check`

### DIFF
--- a/.github/workflows/matrix-smoke.yml
+++ b/.github/workflows/matrix-smoke.yml
@@ -106,6 +106,24 @@ jobs:
             done
           }
 
+          prepare_zypper_repos() {
+            repo_dir=/etc/zypp/repos.d
+            [ -d "$repo_dir" ] || return 0
+
+            # GitHub-hosted runners can intermittently time out on the default
+            # cdn.opensuse.org redirect while refreshing or preloading
+            # Tumbleweed metadata/packages. Prefer a direct US mirror over
+            # HTTPS for CI, while keeping the same official openSUSE repos.
+            for repo_file in "$repo_dir"/*.repo; do
+              [ -f "$repo_file" ] || continue
+              sed -i \
+                -e 's#http://download.opensuse.org/#https://mirror.umd.edu/opensuse/#g' \
+                -e 's#https://download.opensuse.org/#https://mirror.umd.edu/opensuse/#g' \
+                "$repo_file"
+            done
+            zypper --non-interactive clean --metadata || true
+          }
+
           if command -v apt-get &>/dev/null; then
             apt-get update -qq && apt-get install -y -qq ca-certificates git curl
           elif command -v dnf &>/dev/null; then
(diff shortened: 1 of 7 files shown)


Lightheartdevs:
Thanks for taking on the Windows Python resolver path here; this is the right class of fix for Store aliases and py -3 installs.

Quickest path to merge:
- Rebase on current main and resolve the merge conflicts GitHub reports. git merge-tree origin/main origin/pr-1506 shows content conflicts in dream-server/installers/windows/dream.ps1, dream-server/installers/windows/phases/06-directories.ps1, and dream-server/installers/windows/phases/07-devtools.ps1.
- While resolving those files, keep the important behavior from this PR: all Windows Python call sites should use Resolve-DreamWindowsPython, and host-agent startup must preserve PythonArgs so py -3 launches as py -3 <agent-script> ... rather than dropping the launcher argument.
- After the rebase, rerun the focused Windows checks from the PR: python -m pytest dream-server/tests/test-windows-python-resolver.py -q, PowerShell parser/PSScriptAnalyzer checks for the touched Windows scripts, and the dashboard/host-agent smoke path that starts the agent from dream.ps1 agent start.

What I checked:
- PR metadata and discussion: no current advisory marker was present for this head SHA.
- Static diff review of installers/windows/lib/python-resolver.ps1, the Windows installer/CLI call sites, and 	ests/test-windows-python-resolver.py.
- git diff --check origin/main...origin/pr-1506: passed locally.
- GitHub checks on the old base are green, but GitHub reports the branch as CONFLICTING, so the checks need to be trusted only after the rebase.

PR #1510, opened 2026-06-04
fix(embeddings): fail installs when RAG health stalls

### LINKED ISSUE #1507: Embeddings service hangs forever on stalled ONNX download; install reports exit 0 anyway
## Summary
On a fresh install, the `dream-embeddings` service can hang **forever** on a stalled model download, and the installer reports the host install as **`exit 0` (success)** anyway — leaving a green install with a dead embeddings service.

Observed on **strix-halo** (Ryzen AI MAX+ 395, Ubuntu 24.04) during a fleet install, but the failure mode is host-agnostic.

## What happens
The embeddings container (HF `text-embeddings-inference`, `BAAI/bge-*-v1.5`, ONNX backend) downloads its config artifacts fine, then logs:

```
22:53:44  Starting model backend
22:53:44  Downloading `onnx/model.onnx`
<dead stop — no further log lines for 21+ minutes>
```

Evidence the download is **wedged, not slow**:
- router process **idle**: ~1s total CPU over 21 min
- **no `.onnx` file on disk** in `/data` (zero bytes written)
- **no active network connections** from the container

The HF download client opened the request for `onnx/model.onnx`, the connection stalled, and with **no read-timeout** it hangs indefinitely. The container sits `Up (unhealthy)` (healthcheck `curl 127.0.0.1:80` → connection refused, since the router never binds the port). There is **no retry** at the service or installer level.

## The two defects
1. **No download timeout / retry** on the embeddings backend — a single stalled HF connection wedges the service permanently.
2. **Install reports success regardless.** The installer's "Linking embeddings" step timed out **non-fatally** and the host install finished `exit_code=0` while embeddings never became healthy → green install, dead service. (Same false-green class as past "container health checks pass while inference is dead" reports.)

## Proof it's transient (recovery)
A single `docker restart dream-embeddings` recovered it immediately:
```
23:18:16  Downloading `onnx/model.onnx`
23:18:36  Successfully downloaded `onnx/model.onnx` in 20.54 s
23:18:37  Warming up model ... Starting HTTP server: 0.0.0.0:80 ... Ready
```
20-second download on retry. The only thing missing in the original was a timeout to trigger that retry.

## Suggested fix
- Add a connect+read timeout to the embeddings model download (e.g. `HF_HUB_DOWNLOAD_TIMEOUT`) with bounded retries.
- The installer's service-link step should treat a never-healthy core service as an **install failure** (or at minimum a loud warning in the summary), not a silent non-blocking pass.

_Surfaced by the fleet test (run 2026-06-03T22-44-03Z)._

Lightheartdevs:
## Summary

Fixes #1507 by making embeddings/RAG startup failures explicit instead of soft-warning through a broken RAG install.

Changes:

- Adds `HF_HUB_DOWNLOAD_TIMEOUT` and `HF_HUB_ETAG_TIMEOUT` defaults to the TEI embeddings compose environment.
- Tracks embeddings health failure separately in Phase 12.
- When RAG/embeddings is enabled and TEI never becomes healthy, prints recovery guidance and exits nonzero.
- Corrects the Phase 12 fallback embeddings port from `7860` to the manifest/default compose port `8090`.

## Why

The fleet run found TEI stuck during Hugging Face ONNX model download while the installer still reached exit 0. If RAG is selected, a dead embeddings service means the requested feature is not installed, so the installer should fail loud with recovery instructions.

## Validation

- `bash -n installers/phases/12-health.sh`
- `tests/contracts/test-installer-hardening.sh`
- `docker compose -f extensions/services/embeddings/compose.yaml config embeddings`
- `git diff --check`

## Risk

Low for non-RAG installs; the strict path only fires when RAG/embeddings is enabled. For RAG installs, this intentionally turns a false green into an actionable failure if TEI cannot become healthy.


Lightheartdevs:
Audit note: reviewed the embeddings/RAG health path after CI passed. This directly addresses #1507 by making selected embeddings/RAG health failures fail loud instead of reporting a false green, and by bounding Hugging Face download/metadata waits in the TEI compose environment. Local validation: bash syntax check for Phase 12, installer hardening contracts, Docker Compose render for embeddings, git diff --check. Risk is low for non-RAG installs; strict behavior only applies when embeddings/RAG is selected.

PR #1512, opened 2026-06-04
fix(windows): verify LLM readiness and Lemonade swaps

### LINKED ISSUE #1511: Windows: /v1/models lists bootstrap model but GGUF is missing on disk → completions 500 while install reports healthy
## Summary
On Windows (Strix Halo / AMD), the installer registers the bootstrap model `Qwen3.5-2B-Q4_K_M` as available and `/v1/models` returns it (200), **but the GGUF file is missing on disk**, so every `chat/completions` returns **500**. The install still reports `"healthy": true`.

Same trust class as the recently-fixed #1507 / #1508: **health says yes, the actual user path says no.**

Validated on `strixy` (Win11, Radeon 8060S) in fleet run `2026-06-04T00-42-46Z` against DreamServer main `8ba0b6e0` (i.e. *with* #1509 + #1510 — the container stack and RAG fixes both worked; this is a separate, remaining gap).

## Evidence
```
GET  /v1/models                  -> 200, lists extra.Qwen3.5-2B-Q4_K_M.gguf
POST /v1/chat/completions        -> 500 (every call)

lemonade-server.log:
  [Error] llama_model_loader: failed to load model from
          C:\Users\conta\dream-server\data\models\Qwen3.5-2B-Q4_K_M.gguf
  [Info]  using device Vulkan0 (AMD Radeon 8060S) - 77714 MiB free   <- not a VRAM/backend issue
  [Error] main: exiting due to model loading error

PowerShell: Test-Path C:\Users\conta\dream-server\data\models\Qwen3.5-2B-Q4_K_M.gguf -> False
```
77 GB VRAM free rules out a memory/backend problem. The model is *registered* with lemonade but the `.gguf` was never placed at the path lemonade is told to load from — a download/placement vs. path-registration mismatch in the Windows install.

## Impact
Every LLM-driven path is dead on Windows despite a green install: verify `chat-min`, Hermes seed-echo, and capability probes (`skills`, `talk_context`) all fail. These are **one** root cause, not four separate bugs — please treat as a single issue/PR.

## Proposed fix
1. Windows model provisioning must **prove the file exists** at the registered path (and ideally run a **minimal completion**) before the installer reports `healthy: true`.
2. Add an install-readiness **contract**: `/v1/models` listing a model whose backing GGUF is absent/unloadable must **not** pass readiness — fail loud instead of green.

## Not part of this issue
ComfyUI not running on strixy (port 8188) is a separate matter, tracked independently — different root cause, should not be bundled into the LLM-provisioning fix.

_Surfaced by the fleet test (run 2026-06-04T00-42-46Z). Harness side: a probe-script `set -u` crash exposed during this run was fixed separately in dream-fleet-test._

Lightheartdevs:
## Problem

Fixes #1511 and the Windows Strix/Lemonade post-bootstrap false-green found in fleet run `2026-06-04T00-42-46Z`.

There were two related trust gaps:

1. Windows Phase 9 treated a healthy Lemonade/llama-server process as enough, even if the registered model could not actually complete.
2. On Windows AMD/Lemonade, the background bootstrap upgrade could download the full model and update `.env`, then leave the native Lemonade process serving the deleted bootstrap model. That produced the Strixy state we observed: `/api/v1/models` listed only `extra.Qwen3.5-2B-Q4_K_M.gguf`, the 70B file existed on disk, completion against the 70B returned 404, and completion against the deleted 2B returned 500.

## Fix

- Add `Test-WindowsLlmModelReadiness` to prove both that the GGUF backing file exists and that a minimal completion succeeds before a local Windows install can report healthy.
- Key Lemonade model-id prefixing off the resolved endpoint (`Backend=lemonade` / `/api/v1`), not every AMD backend, so the native llama-server Vulkan fallback keeps the plain GGUF model id.
- Add a Windows Lemonade branch to `bootstrap-upgrade.sh` that restarts the native Lemonade process after the full model is downloaded, waits for the new `extra.<GGUF>` model to register, verifies a real completion, patches Hermes off the bootstrap model, and only then removes the bootstrap GGUF.
- Mark bootstrap upgrade status failed if Windows Lemonade cannot prove the full-model swap, instead of reporting complete while the runtime is wedged.

## Validation

Local checks run:

- `bash -n dream-server/scripts/bootstrap-upgrade.sh`
- `bash dream-server/tests/test-bootstrap-upgrade-hotswap-contract.sh`
- `bash dream-server/tests/contracts/test-windows-llm-model-readiness.sh`
- `bash dream-server/tests/test-bootstrap-upgrade-download-finalization.sh`
- `bash dream-server/tests/test-bootstrap-upgrade-resume-status.sh`
- PowerShell parse checks for `install-windows.ps1` and `llm-endpoint.ps1`
- `git diff --check`

Notes: the wider installer contract bundle could not run in this local Windows Git Bash because `jq` is absent; WSL on this machine also lacks `bash`. The targeted contracts for the changed paths pass.

### DIFF
--- a/dream-server/Makefile
+++ b/dream-server/Makefile
@@ -31,6 +31,7 @@ test: ## Run unit and contract tests
 	@echo "=== AMD/Lemonade contracts ==="
 	@bash tests/contracts/test-amd-lemonade-contracts.sh
 	@bash tests/test-litellm-amd-auth-enforced.sh
+	@bash tests/contracts/test-windows-llm-model-readiness.sh
 	@echo ""
 	@echo "=== Overlay/plist contracts ==="
 	@bash tests/contracts/test-overlay-map-coherence.sh
--- a/dream-server/installers/windows/install-windows.ps1
+++ b/dream-server/installers/windows/install-windows.ps1
@@ -1189,6 +1189,28 @@ foreach ($check in $healthChecks) {
     }
 }
 
+# ── LLM model-serving gate ───────────────────────────────────────────────────
+# A healthy LLM process is NOT proof the model can serve: if the GGUF backing file
+# was never placed on disk, /v1/models still lists it but every completion 500s.
+# Prove the file exists AND a minimal completion succeeds before this install may
+# report healthy — otherwise fail loud instead of a "health says yes, chat says no"
+# green install.
+if (-not $cloudMode) {
+    Write-AI "Verifying the LLM can actually serve a completion..."
+    $llmReady = Test-WindowsLlmModelReadiness -Endpoint $llmEndpoint -InstallDir $installDir `
+        -GgufFile $tierConfig.GgufFile -GpuBackend $gpuInfo.Backend -TimeoutSec 120
+    if ($llmReady.Ok) {
+        Write-AISuccess "LLM serving verified (model: $($llmReady.ModelId))"
+    } else {
+        $allHealthy = $false
+        Write-AIError "LLM not serving: $($llmReady.Detail)"
+        if (-not $llmReady.FileExists) {
+            Write-Host "    Model file missing: $($llmReady.ModelFile)" -ForegroundColor DarkGray
+            Write-Host "    Re-run the installer to (re)download it: .\install.ps1" -ForegroundColor DarkGray
+        }
+    }
+}
+
 # ── Pre-download the Whisper STT model ───────────────────────────────────────
 # Speaches does NOT auto-download on transcription requests — it returns 404.
 # Trigger the download explicitly, verify it completed, surface recovery
--- a/dream-server/installers/windows/install-windows.ps1
+++ b/dream-server/installers/windows/install-windows.ps1
@@ -1198,7 +1198,7 @@ foreach ($check in $healthChecks) {
 if (-not $cloudMode) {
     Write-AI "Verifying the LLM can actually serve a completion..."
     $llmReady = Test-WindowsLlmModelReadiness -Endpoint $llmEndpoint -InstallDir $installDir `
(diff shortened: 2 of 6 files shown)


Lightheartdevs:
Audit update before merge:

- CI is green after the updated branch: 28 successful checks, 3 expected skips.
- I found two issues in the original version and fixed them on this PR before merge:
  - `Test-WindowsLlmModelReadiness` was prefixing `extra.` for every AMD backend. That was right for Lemonade but wrong for native llama-server/Vulkan fallback, so the gate now keys off the resolved endpoint (`Backend=lemonade` / `/api/v1`).
  - The fleet evidence showed the Strixy failure was actually post-bootstrap hot-swap: the full model downloaded, but native Windows Lemonade kept serving the now-deleted bootstrap model. `bootstrap-upgrade.sh` now restarts native Windows Lemonade, verifies the new `extra.<GGUF>` model can complete, patches Hermes off the bootstrap model, and only then removes the bootstrap GGUF. If that proof fails, bootstrap status is marked failed instead of reporting complete.

Local validation run:

- `bash -n dream-server/scripts/bootstrap-upgrade.sh`
- `bash dream-server/tests/test-bootstrap-upgrade-hotswap-contract.sh`
- `bash dream-server/tests/contracts/test-windows-llm-model-readiness.sh`
- `bash dream-server/tests/test-bootstrap-upgrade-download-finalization.sh`
- `bash dream-server/tests/test-bootstrap-upgrade-resume-status.sh`
- PowerShell parse checks for `install-windows.ps1` and `llm-endpoint.ps1`
- `git diff --check`

Risk read: scoped to Windows local LLM readiness and bootstrap model swaps. Non-Lemonade Linux/macOS paths keep their existing behavior, with the safer invariant that bootstrap cleanup is gated on verified full-model serving.

PR #1519, opened 2026-06-05
fix(windows): harden Lemonade swaps and Hermes config

Lightheartdevs:
## Summary

This PR hardens two Windows-only failure modes found in the `2026-06-04T23-41-44Z` fleet run after #1518:

- fixes native Windows Lemonade full-model swaps getting stuck on the old bootstrap router when `lemonade-server` exits with `Another instance of lemonade-router is already running`
- makes Windows Hermes config patching fail loudly if the template or live `data/hermes/config.yaml` does not actually receive the Windows runtime model/base URL
- adds contracts for both guarantees and wires the Hermes config contract into `make test`

## Root cause

On Strixy, the full Qwen3.6-35B GGUF was downloaded and verified, but Lemonade never registered it. The live Lemonade log showed the replacement process exited immediately because another `lemonade-router` instance was still running, so the swap loop kept polling the stale bootstrap process.

The same run also showed Hermes falling back because the mounted config still had upstream defaults (`qwen3.5-9b`, `http://llama-server:8080/v1`) even though the installer logged that it patched Hermes. The helper silently no-opped when the live config was missing.

## Changes

- Windows Lemonade swap now:
  - refuses to touch externally managed Lemonade runtimes
  - stops the saved PID, the configured port listener, and Lemonade child processes under the Lemonade install bin dir before relaunch
  - fails with the Lemonade log tail if the replacement exits immediately
  - avoids false failure if the launcher parent exits but leaves a healthy `/api/v1/models` listener behind
- Windows Hermes config patching now:
  - seeds `data/hermes/config.yaml` from the template before first Hermes start when missing
  - reads/writes explicit UTF-8 without BOM
  - verifies `model.default` and `base_url` after writing
  - fails the installer if either the template or live config did not patch

## Validation

- `git diff --check`
- PowerShell parser check: `installers/windows/phases/06-directories.ps1`
- PowerShell parser check: `installers/windows/install-windows.ps1`
- Git Bash: `bash -n scripts/bootstrap-upgrade.sh`
- Git Bash: `bash tests/contracts/test-windows-lemonade-swap-wait.sh`
- Git Bash: `bash tests/contracts/test-windows-hermes-config-patching.sh`

`make test` was not runnable from this Windows shell because `make` is not installed here; the new contract is wired into `make test` for CI/Linux environments.

## Risk

Scoped to Windows installer/runtime paths and the Windows branch inside `bootstrap-upgrade.sh`. Linux, macOS, cloud mode, and external Lemonade installs are not touched. The main behavior change is fail-loud: if Hermes config patching does not land, Windows install now stops instead of printing a false success.

### DIFF
--- a/dream-server/Makefile
+++ b/dream-server/Makefile
@@ -34,6 +34,7 @@ test: ## Run unit and contract tests
 	@bash tests/contracts/test-windows-strix-tier-map.sh
 	@bash tests/contracts/test-windows-llm-model-readiness.sh
 	@bash tests/contracts/test-windows-lemonade-swap-wait.sh
+	@bash tests/contracts/test-windows-hermes-config-patching.sh
 	@echo ""
 	@echo "=== Overlay/plist contracts ==="
 	@bash tests/contracts/test-overlay-map-coherence.sh
--- a/dream-server/installers/windows/install-windows.ps1
+++ b/dream-server/installers/windows/install-windows.ps1
@@ -318,8 +318,19 @@ if ($dryRun) {
                 })
                 $hermesTemplate = Join-Path (Join-Path (Join-Path $installDir "extensions") "services\hermes") "cli-config.yaml.template"
                 $hermesLive = Join-Path (Join-Path $installDir "data\hermes") "config.yaml"
-                Update-HermesConfigFile -Path $hermesTemplate -Model $hermesModel -BaseUrl $hermesBaseUrl -ContextLength ([int]$tierConfig.MaxContext)
-                Update-HermesConfigFile -Path $hermesLive -Model $hermesModel -BaseUrl $hermesBaseUrl -ContextLength ([int]$tierConfig.MaxContext)
+                if (-not (Test-Path $hermesTemplate)) {
+                    Write-AIError "Missing Hermes config template at $hermesTemplate"
+                    exit 1
+                }
+                if (-not (Test-Path $hermesLive)) {
+                    Copy-Item -Path $hermesTemplate -Destination $hermesLive -Force
+                }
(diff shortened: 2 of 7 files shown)


Lightheartdevs:
Audit/validation notes before merge:

- CI is green on final head `2ccff26b`.
- Local contracts passed:
  - `bash -n dream-server/scripts/bootstrap-upgrade.sh`
  - `bash dream-server/tests/contracts/test-windows-lemonade-swap-wait.sh` (`PASS=9 FAIL=0`)
  - `bash dream-server/tests/contracts/test-windows-hermes-config-patching.sh` (`PASS=6 FAIL=0`)
  - `git diff --check`
- Root `install.ps1` wrapper edge was audited and fixed before merge: successful delegated installs now exit `0` even if `$LASTEXITCODE` was stale from a handled native command, while real delegated failures preserve their nonzero exit code.
- Live Strixy validation on real Windows Strix Halo host (`runs/2026-06-05T02-46-47Z`, PR head before wrapper-only amendment):
  - Windows install exited `0`.
  - Full-model download completed and SHA verified.
  - Bootstrap model was moved aside before native Lemonade restart.
  - Native Windows Lemonade served `extra.Qwen3.6-35B-A3B-UD-Q4_K_M.gguf` successfully.
  - Stale bootstrap model metadata was gone from `/api/v1/models`; only the 35B model was advertised.
  - Hermes live config and template both pointed to the 35B model with `context_length: 131072` and `host.docker.internal:8080/api/v1`.
  - Only one `lemonade-router` and one `llama-server` process remained, both from the new launch.
  - Direct chat against the 35B model returned HTTP 200.

Risk read: scoped to Windows/native-Lemonade Strix Halo swap + Hermes config + root Windows exit propagation. Linux/macOS/container paths are untouched except for shared contract wiring. The main stability risk was false failure from stale `$LASTEXITCODE`; fixed and covered before merge.

PR #1529, opened 2026-06-07
fix(windows): make install target explicit

gabsprogrammer:
## Summary
- add a Windows `-InstallDir` installer flag that maps to `DREAM_HOME` before installer constants are loaded
- print both the source checkout and runtime install target during Windows preflight, with a warning when they are on different drives
- make low-disk warnings show which install target was checked and document the explicit alternate-drive install path
- extend the Windows installer flag contract test to cover `-InstallDir` and the default-No requirements gate

## Why
A Windows user cloned DreamServer on `G:` because `C:` was tight, then ran `./install.ps1` from the `G:` checkout. The installer still used the default runtime target under `%USERPROFILE%\dream-server` on `C:`, which was correct by current contract but easy to miss. This PR makes the target explicit and gives users a direct `-InstallDir G:\dream-server` path instead of requiring them to know about `DREAM_HOME`.

The same report also showed the requirements prompt as `Continue anyway? [y/N]`; current main already aborts on default/Enter, so this PR adds a contract assertion around that behavior.

## Test plan
- `bash dream-server/tests/test-windows-installer-flags.sh`
- `bash -n dream-server/tests/test-windows-installer-flags.sh`
- PowerShell parser check for touched `.ps1` files
- `git diff --check`


Lightheartdevs (changes requested):
Thanks for tightening this up. The core `-InstallDir` plumbing looks sound: the root wrapper forwards `@PSBoundParameters`, and the Windows installer maps `-InstallDir` to `DREAM_HOME` before constants load. I do think this needs a small follow-up before merge because the user-facing recovery path is not yet consistent.

Requested changes:

1. Update the docs so custom install paths do not keep sending users back to `$env:USERPROFILE\dream-server`.

   After documenting `.\install.ps1 -InstallDir D:\Apps\dream-server`, the walkthrough and quickstart still tell users to `cd $env:USERPROFILE\dream-server` for status/logs/compose commands and still list models under `$env:USERPROFILE\dream-server\data\models`. For a custom install, that recreates the same confusion this PR is trying to solve.

   Suggested correction: introduce an `$installDir` example variable and use it consistently:

   ```powershell
   $installDir = "D:\Apps\dream-server"
   .\install.ps1 -InstallDir $installDir

   cd $installDir
   .\dream.ps1 status
   .\dream.ps1 logs llama-server
   ```

   Also update the files/locations table so models/logs/config point to the chosen install directory, not only `$env:USERPROFILE\dream-server`.

2. Add the new install-dir hint to the earlier disk exits too.

   The Phase 04 warning now prints the checked install target and a useful `.\install.ps1 -InstallDir ...` hint, but two earlier gates can exit before that message:

   - `installers/windows/phases/01-preflight.ps1`: the initial 20GB disk check
   - `installers/windows/phases/02-detection.ps1`: the tier/model disk check

   Suggested correction: make those messages include the checked `$installDir` and the same rerun hint, so low-disk users always see the alternate-drive path regardless of which disk gate catches them first.

Validation I ran on a current-main merge tree:

- `dream-server/tests/test-windows-installer-flags.sh`: 20/20 passed via Git Bash
- `bash -n dream-server/tests/test-windows-installer-flags.sh`: passed
- PowerShell parser check for touched `.ps1` files: passed
- `git diff --check`: passed

No local machine names, fleet paths, IPs, or run-directory trails were present in the PR body/comments/diff.

gabsprogrammer:
I corrected PR #1529 as per the review

Lightheartdevs (approved):
Re-audited after corrections. The earlier requested changes are addressed: custom install-path docs now use an explicit $installDir flow, the files/locations guidance no longer sends custom-path users back to the default runtime directory, and all Windows disk gates include the checked install target plus an -InstallDir rerun hint.\n\nI also tested the PR on a current-main merge tree after #1530: Windows installer flag contract 30/30, PowerShell parser checks, git diff --check, and the Windows stop/env/compose/Hermes regression tests all pass. PR body/comments/diff were scanned for local machine or fleet breadcrumbs; none found.

PR #1542, opened 2026-06-12
fix(dashboard): pause remaining background pollers while the tab is hidden (#1490 follow-up)

notwitcheer:
## Summary

Follow-up to the #1490 idle-CPU investigation, picking up the thread from the maintainer's comment there: after the CSS animation mitigation (#1499), the next suspects were *"`/api/status`, service resources, and GPU/status polling frequency"*.

The dashboard already has a blessed pattern for this — `useSystemStatus`, `useGPUDetailed`, and `ServiceMap` all skip polling while `document.hidden` and refresh on `visibilitychange`. But four pollers never got the same treatment and keep hitting dashboard-api around the clock from a backgrounded tab:

| Poller | Interval | Server-side cost |
|---|---|---|
| `Dashboard.jsx` → `/api/services/resources` | 10s | `docker stats --no-stream` sweeps + recursive `/data` disk walks via the host agent (TTL-cached 20s/60s, so a hidden tab still forces a stats sweep ~every 20s and a full models-dir walk ~every 60s, forever) |
| `Dashboard.jsx` → `/api/features` | 15s | feature/service state assembly |
| `Usage.jsx` → usage report | 10s | usage aggregation |
| `useModels` → `/api/models` | 30s | model catalog + GPU fit checks |
| `useDownloadProgress` → download-status | 10s idle | status file read |

Since the Dashboard page is the default landing page, "open dashboard, switch to another tab/screen, leave the box running" — the exact usage in the #1490 report — keeps all of these hot. The `docker stats` sweeps in particular are a plausible contributor to the reported idle CPU on the host, independent of Chrome rendering.

### Change

Each poller now skips interval ticks while `document.hidden` and refreshes immediately on `visibilitychange` — the same semantics as the three already-guarded pollers, deliberately mirrored line-for-line so the codebase has one recognizable pattern. Initial mount fetches are unchanged, and manual refresh paths (model download/delete/benchmark refreshes, Usage range changes) are unaffected because the guard lives on the interval tick, not inside the fetch functions.

Adds visibility regression tests for both hooks: polling pauses while hidden, and the catch-up fetch fires when the tab returns.

## AI Assistance

Drafted with Claude Code (mapped the polling sites, traced `/api/services/resources` through `routers/resources.py` to the host agent's `docker stats` handler, wrote the patch and tests, ran the validations below). I read the final diff and am accountable for it.

## Release Lane

Stable hotfix reason:

```text
n/a
```

## Changed Surface

- [ ] Docs only
- [ ] Tests only
- [x] Dashboard UI

## Risk And Validation

- Risk level: Low
- Validation run:
  - [x] `git diff --check`
  - [ ] Not required because:

Commands/results:

```text
cd dream-server/extensions/services/dashboard
npm test -- --run   ->  Test Files 20 passed, Tests 124 passed
                        (122 existing + 2 new visibility regression tests)
npm run lint        ->  clean
npm run build       ->  ✓ built in 1.58s
git diff --check    ->  clean
```

Behavior notes for review:
- A tab that mounts while hidden still does its initial fetch (unchanged), so first paint is never blocked on visibility.
- `useDownloadProgress` keeps its fast poll while a download is active and visible; while hidden, the visibilitychange handler catches the state up the moment the user returns, so completion toasts still fire.

## Operational Change Check

Frontend polling cadence only. No API contracts, auth, routing, or installer surfaces changed. Worst case regression mode is "stale panel until the tab becomes visible", at which point the immediate catch-up fetch runs.

### DIFF
--- a/dream-server/extensions/services/dashboard/src/hooks/__tests__/useDownloadProgress.test.js
+++ b/dream-server/extensions/services/dashboard/src/hooks/__tests__/useDownloadProgress.test.js
@@ -1,5 +1,11 @@
-import { renderHook, waitFor } from '@testing-library/react'
+import { renderHook, waitFor, act } from '@testing-library/react'
 import { useDownloadProgress } from '../useDownloadProgress'
+
+// Shadow jsdom's Document.prototype.hidden getter on the instance; deleting
+// the own property in afterEach restores the prototype behavior.
+const setDocumentHidden = (hidden) => {
+  Object.defineProperty(document, 'hidden', { configurable: true, get: () => hidden })
+}
 
 describe('useDownloadProgress', () => {
   beforeEach(() => {
@@ -8,6 +14,7 @@ describe('useDownloadProgress', () => {
 
   afterEach(() => {
     vi.restoreAllMocks()
+    delete document.hidden
   })
 
   test('sets isDownloading when status is downloading', async () => {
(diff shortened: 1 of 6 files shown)


gabsprogrammer:
I reviewed this PR against the #1490 idle-dashboard CPU issue and tried to find regressions in the polling behavior.

Result: validated from my side. I did not find a code blocker.

Why this looks useful:

- The changed pollers are real background API callers.
- `/api/services/resources` is especially worth guarding because dashboard-api caches container stats for 20s and disk scans for 60s, so a hidden dashboard tab can still keep triggering host-side work over time.
- The PR preserves the initial fetch on mount, so a tab opened/restored in the background does not get stuck with empty state.
- Manual/user-triggered flows are not blocked because the visibility guard is only on interval ticks, not inside the fetch functions themselves.
- The visibilitychange catch-up fetch matches the existing pattern already used by `useSystemStatus`, `useGPUDetailed`, and `ServiceMap`.

Validation I ran:

- `git diff --check origin/main...HEAD`: pass
- `npm ci`: pass
- `npm test -- --run`: 20 test files passed, 124 tests passed
- `npm run lint`: pass
- `npm run build`: pass

I also manually checked the remaining polling sites. The obvious always-on dashboard/model/usage/download pollers touched by this PR are covered. The remaining `Extensions.jsx` pollers appear tied to active extension install/log views rather than the default idle dashboard case, so I would not expand this PR just for those.

Minor non-blocking note: the PR body says “four pollers” but lists five entries. That is just wording, not a code issue.

My review: validated. No code changes requested; only @Lightheartdevs  maintainer decision remains.

PR #1561, opened 2026-06-14
fix(gpu): assign llama_server on single-GPU hosts when --enabled-services omits it

w3-surfer:
## Problem

On a single-GPU host, `scripts/assign_gpus.py` crashes with `KeyError: 'llama_server'` when `--enabled-services` is passed without `llama_server`.

The single-GPU early-exit builds its service map purely from `--enabled-services`, then indexes `services["llama_server"]` unconditionally to attach parallelism:

```python
services = {}
for s in enabled_services:
    services[s] = ServiceAssignment(gpus=[gpu])
services["llama_server"].parallelism = parallelism   # KeyError if not in the list
```

**Reproduce:**

```console
$ python3 scripts/assign_gpus.py \
    --topology tests/fixtures/topology_json/nvidia_smi_topo_matrix_1gpu_pcie.json \
    --model-size 10000 --enabled-services whisper,embeddings
Traceback (most recent call last):
  ...
  File "scripts/assign_gpus.py", line 468, in main
    services["llama_server"].parallelism = parallelism
KeyError: 'llama_server'
```

`--enabled-services` is a documented flag (it's in the script's own usage header).

## Fix

The multi-GPU path never has this bug — `assign_services()` assigns `llama_server` unconditionally. This makes the single-GPU path follow the same contract via `setdefault`, so `llama_server` always gets the GPU even when the caller's list omits it:

```python
services.setdefault("llama_server", ServiceAssignment(gpus=[gpu]))
services["llama_server"].parallelism = parallelism
```

## Tests

Adds a `TestSingleGpuEnabledServices` regression class to `tests/test-assign-gpus.py` covering the `--enabled-services` path (no-crash, `llama_server` still assigned, requested services present, explicit-include unchanged). The new tests **fail on the pre-fix code** and pass after.

- `python3 -m pytest tests/test-assign-gpus.py` → **84 passed** (80 existing + 4 new)
- Default invocation (no `--enabled-services`) output is unchanged.

One concern, ~5-line fix + test. No behavior change to existing paths.

### DIFF
--- a/dream-server/scripts/assign_gpus.py
+++ b/dream-server/scripts/assign_gpus.py
@@ -465,6 +465,11 @@ def main():
         services = {}
         for s in enabled_services:
             services[s] = ServiceAssignment(gpus=[gpu])
+        # llama_server always runs on the single GPU, even when the caller's
+        # --enabled-services list omits it. This mirrors the multi-GPU path,
+        # where assign_services() assigns llama_server unconditionally; without
+        # it, the line below raised KeyError: 'llama_server'.
+        services.setdefault("llama_server", ServiceAssignment(gpus=[gpu]))
         services["llama_server"].parallelism = parallelism
         result = AssignmentResult(strategy="single", services=services)
         print(json.dumps(build_output(result), indent=2))
--- a/dream-server/tests/test-assign-gpus.py
+++ b/dream-server/tests/test-assign-gpus.py
@@ -9,11 +9,11 @@ FIXTURES_DIR = os.path.join(os.path.dirname(__file__), "fixtures/topology_json")
 def fixture_path(name):
     return os.path.join(FIXTURES_DIR, name)
 
-def run(topology_path, model_size_mb):
-    result = subprocess.run(
-        [sys.executable, SCRIPT, "--topology", topology_path, "--model-size", str(model_size_mb)],
-        capture_output=True, text=True,
-    )
+def run(topology_path, model_size_mb, enabled_services=None):
+    cmd = [sys.executable, SCRIPT, "--topology", topology_path, "--model-size", str(model_size_mb)]
+    if enabled_services is not None:
+        cmd += ["--enabled-services", enabled_services]
+    result = subprocess.run(cmd, capture_output=True, text=True)
     output = None
     if result.returncode == 0:
         output = json.loads(result.stdout)["gpu_assignment"]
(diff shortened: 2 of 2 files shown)


gabsprogrammer:
Thanks for the fix.

I reviewed this change carefully because `assign_gpus.py` directly influences installer and runtime GPU placement, meaning even relatively small modifications can have operational impact. After reviewing the implementation and validating it locally, the change looks correct to me.

### Root Cause

The reported bug is legitimate. On the single-GPU code path, when `--enabled-services` is provided without `llama_server`, the service map is constructed only from the explicitly requested services. The code then unconditionally accesses:

```python
services["llama_server"].parallelism
```

which results in a `KeyError`.

### Assessment of the Fix

The proposed fix aligns with the behavior already implemented in the multi-GPU path. There, `llama_server` is treated as a mandatory core assignment regardless of whether it appears in `enabled_services`. Bringing the single-GPU path into alignment with that existing contract is the correct behavior and maintains consistency across execution modes.

### Validation Performed

* Reproduced the pre-fix crash using:

  * `--enabled-services whisper,embeddings`
* Confirmed the patched version returns valid assignment JSON for the same command.
* Confirmed `llama_server`, `whisper`, and `embeddings` are all assigned to the single GPU.
* Confirmed the default single-GPU invocation output remains unchanged.

Test results:

```bash
python -m pytest tests/test-assign-gpus.py -q
# 84 passed
```

```bash
python -m pytest tests/test-assign-gpus.py -k regression
# 4 passed
```

Additional verification:

```bash
python -m py_compile scripts/assign_gpus.py tests/test-assign-gpus.py
git diff --check
```

### Test Coverage

The regression coverage is appropriate for the issue being addressed. The new tests verify:

* No-crash behavior when `llama_server` is omitted from `--enabled-services`
* Continued assignment of `llama_server` despite omission
* Preservation of all explicitly requested services
* Unchanged behavior when `llama_server` is explicitly included

### Follow-up (Non-Blocking)

One pre-existing behavior worth noting is that the parser currently accepts an empty value such as:

```bash
--enabled-services ""
```

which results in a `""` service key being emitted in the output. This behavior predates the current PR and is not introduced by the fix, so I would not consider it a blocker. However, it may be worth hardening in a future cleanup pass.

### Conclusion

From my side, this is a focused fix for a real issue, follows the existing service-assignment contract, includes appropriate regression coverage, and preserves existing behavior outside the affected path.

Lightheartdevs:
Thanks for narrowing this down to the single-GPU assignment path. This is a small fix, but it covers a real crash mode that would be pretty frustrating for someone setting up a constrained machine.

What looks good:
- The change keeps single-GPU `--enabled-services` behavior aligned with the existing multi-GPU contract: `llama_server` remains assigned even when the caller only lists auxiliary services.
- The patch stays focused on `dream-server/scripts/assign_gpus.py` plus direct regression coverage.
- The regression checks the important cases: no crash, requested services preserved, implicit `llama_server`, and explicit `llama_server` unchanged.

What I checked:
- Local focused suite passed on the PR branch: `python -m pytest dream-server/tests/test-assign-gpus.py -q` -> 84 passed.
- I did not run the full installer or fleet path, but for this helper-level change the unit coverage lines up directly with the failure mode.

Thanks for keeping this one tight. It should be straightforward for a maintainer to reason about and merge.

Lightheartdevs (approved):

PR #1562, opened 2026-06-14
fix(macos): use portable stat fallback in dream-test-functional.sh

w3-surfer:
## Problem

`scripts/dream-test-functional.sh` sizes the generated Whisper test audio with a GNU-only `stat` flag:

```bash
if [[ ! -f "$test_audio" ]] || [[ $(stat -c%s "$test_audio" 2>/dev/null) -lt 1000 ]]; then
    warn "Test audio generation failed"
    return 0
fi
```

On macOS, `stat` is BSD and rejects `-c`. The `2>/dev/null` swallows the error, so the command substitution is **empty**, and `[[ "" -lt 1000 ]]` evaluates **true**. Result: a perfectly valid audio file is reported as *"Test audio generation failed"* and the Whisper functional test is **silently skipped** on every Mac.

The script already requires Bash 4+ (`brew install bash`), but that does not change the system `/usr/bin/stat`, so the bug fires regardless.

**Reproduce on macOS** (same predicate, against a valid 2 KB file):

```console
$ f=$(mktemp); head -c 2048 /dev/zero > "$f"
$ if [[ $(stat -c%s "$f" 2>/dev/null) -lt 1000 ]]; then echo "FAILED (bug)"; else echo ok; fi
FAILED (bug)
$ stat -c%s "$f"
stat: illegal option -- c
```

## Fix

Add the BSD `stat -f%z` fallback, matching the portable form **already used in the same file** for the TTS file-size check (line 154):

```bash
... || [[ "$(stat -c%s "$test_audio" 2>/dev/null || stat -f%z "$test_audio" 2>/dev/null || echo 0)" -lt 1000 ]]; then
```

After the fix, the 2 KB file is correctly treated as valid, while a genuinely tiny file is still flagged:

| file | before | after |
|---|---|---|
| 2048-byte valid audio | `FAILED` (bug) | `ok` |
| 10-byte tiny file | `FAILED` | `FAILED` (correct) |

One-line change, mirrors existing in-file convention. `make lint` passes.

### DIFF
--- a/dream-server/scripts/dream-test-functional.sh
+++ b/dream-server/scripts/dream-test-functional.sh
@@ -231,7 +231,7 @@ test_whisper_functional() {
         return 0
     fi
     
-    if [[ ! -f "$test_audio" ]] || [[ $(stat -c%s "$test_audio" 2>/dev/null) -lt 1000 ]]; then
+    if [[ ! -f "$test_audio" ]] || [[ "$(stat -c%s "$test_audio" 2>/dev/null || stat -f%z "$test_audio" 2>/dev/null || echo 0)" -lt 1000 ]]; then
         warn "Test audio generation failed"
         return 0
     fi


gabsprogrammer:
Thanks for the fix.

I reviewed this change carefully because this script is part of our functional validation pipeline, and false skips in this area can mask legitimate regressions.

The bug is real, and the proposed fix is directionally correct.

The Whisper functional test currently relies on `stat -c%s`, which is GNU-specific. On macOS, `stat` uses BSD syntax and rejects `-c`, causing the command substitution to return an empty value. As a result, a valid generated audio file can be incorrectly classified as too small, leading the Whisper functional test to be skipped even when TTS successfully produced usable output.

The proposed fallback is consistent with the existing TTS file-size check already present in this script:

```bash
stat -c%s ... || stat -f%z ... || echo 0
```

### Validation Performed

I validated the behavior locally and confirmed the following:

* `bash -n scripts/dream-test-functional.sh`
* `make lint`
* `git diff --check`

Additionally, I simulated the relevant execution paths:

* GNU `stat`:

  * 2048-byte file → passes
  * 10-byte file → fails
* BSD/macOS `stat`:

  * Existing implementation incorrectly fails a valid 2048-byte file
  * Proposed implementation correctly passes the same file
* Both `stat` variants unavailable/failing:

  * Fallback returns `0`
  * Size validation continues to fail safely

Based on the above, the one-line fix is useful, correct, and low risk.

### Requested Hardening (Non-Blocking in Principle)

I do think the PR would be stronger with one additional hardening step before merge.

At the moment, the script contains two independent implementations of the same portable file-size logic: one in the TTS validation path and one in the Whisper validation path. This is effectively how the issue occurred in the first place—one path remained portable while the other drifted back to GNU-only behavior.

To reduce the chance of this portability regression reappearing, could we please do one of the following?

1. Extract a small helper (for example, `file_size_bytes()`) and reuse it in both the TTS and Whisper checks; or
2. Add a lightweight regression test/contract that explicitly validates the Whisper size check under BSD-style `stat -f%z` behavior.

I do not believe this needs to become a larger refactor. The core fix itself is correct. My concern is simply that the current patch resolves the immediate symptom, while a tiny shared helper or regression check would help ensure the same portability issue does not reappear in future file-size validations.

### Conclusion

The bug is legitimate, the root cause is clear, and the proposed fix behaves correctly across the scenarios I tested.

Once the small hardening item above is addressed, this looks good to merge from my side.

Lightheartdevs:
Thanks for fixing the macOS false-skip in the Whisper functional path.

What looks good:
- dream-server/scripts/dream-test-functional.sh now mirrors the portable size-check pattern already used elsewhere in the file: GNU stat -c%s, then BSD/macOS stat -f%z, then a safe

Lightheartdevs (approved):

PR #1563, opened 2026-06-14
fix(extensions): make generate-extensions-catalog.py import on Python 3.9

w3-surfer:
## Problem

`scripts/generate-extensions-catalog.py` uses PEP 604 / PEP 585 annotations that are evaluated at definition time:

```python
def load_manifest(manifest_path: Path) -> dict | None:   # line 54
...
def ... -> list[dict]:
```

On **Python 3.9** — which macOS ships as `/usr/bin/python3` — importing the module raises:

```console
$ python3 scripts/generate-extensions-catalog.py --help
TypeError: unsupported operand type(s) for |: 'type' and 'NoneType'   (line 54)
```

`docs/EXTENSIONS.md` (the "add an extension" path) and `.github/workflows/validate-catalog.yml` both invoke this script. CI passes only because it pins Python 3.12, so a contributor following the docs and regenerating the catalog locally on a Mac hits the crash.

## Fix

Add `from __future__ import annotations` (PEP 563 — annotations become lazy strings), matching the sibling scripts in the same directory that already do this: `select-model.py:10` and `audit-extensions.py`. The repo even documents this exact failure mode in `bin/dream-host-agent.py`.

One line. Behavior-preserving — the script does no runtime `get_type_hints()`.

## Verification (Python 3.9.6, macOS)

```console
$ python3 scripts/generate-extensions-catalog.py --output /tmp/cat.json
Generated catalog with 33 extensions at /tmp/cat.json
```

Before the fix: `TypeError` on import. After: runs clean, exit 0.

### DIFF
--- a/dream-server/scripts/generate-extensions-catalog.py
+++ b/dream-server/scripts/generate-extensions-catalog.py
@@ -6,6 +6,8 @@ extracts catalog-relevant fields, and writes a sorted JSON catalog
 to dream-server/config/extensions-catalog.json.
 """
 
+from __future__ import annotations
+
 import argparse
 import json
 import re


gabsprogrammer:
Thanks for the fix.

I reviewed this carefully because `generate-extensions-catalog.py` sits directly in the extension authoring and catalog-generation workflow. This is exactly the type of contributor-facing tooling that should continue working on the oldest Python version we claim—or reasonably expect—to support.

The fix looks technically correct.

The script currently uses annotations such as `dict | None` and `list[dict]`. Without `from __future__ import annotations`, expressions like `dict | None` are evaluated when the function is defined, which is not supported on Python 3.9. Adding the future import is the correct minimal fix and is consistent with the approach already used in nearby scripts such as `audit-extensions.py` and `select-model.py`.

### Validation Performed

```bash
python -m py_compile scripts/generate-extensions-catalog.py
```

```bash
python scripts/generate-extensions-catalog.py --output ...
```

Results:

* Catalog generation completed successfully
* Generated catalog contains 33 extensions
* Generated output is semantically identical to the current implementation when ignoring the `generated_at` field
* `git diff --check` passes cleanly

I also confirmed that the script does not appear to rely on runtime evaluation of annotations, so deferring annotation evaluation via `from __future__ import annotations` should be fully behavior-preserving.

### Suggested Hardening

That said, I think the PR would be stronger with one small validation improvement before merge.

The PR description correctly notes that current CI runs on Python 3.12. As a result, this Python 3.9 compatibility regression was never exercised automatically. Since the issue specifically affects Python 3.9 import/runtime behavior, it would be beneficial to add validation that can catch similar regressions going forward.

Could you please add one of the following?

* A Python 3.9 entry to `.github/workflows/validate-catalog.yml`, while keeping the existing PyYAML installation step; or
* A lightweight smoke test that imports and executes `scripts/generate-extensions-catalog.py` under Python 3.9.

I do not think this warrants a larger refactor. The code change itself is correct and appropriately scoped. My concern is simply that we are fixing a Python 3.9-only failure without adding any automated coverage that would detect the same class of issue the next time newer annotation syntax is introduced.

### Dependency Note

One unrelated observation: the script still depends on PyYAML. A completely fresh system Python installation—particularly on macOS—may still fail with `import yaml` until that dependency is installed.

That is separate from the annotations issue addressed here, and the existing catalog workflow already installs PyYAML. If Python 3.9 validation is added, it should include the same dependency installation step so the environment matches existing CI behavior.

### Conclusion

The root cause is clear, the proposed fix is minimal and correct, and local validation indicates the change is behavior-preserving.

With a small Python 3.9 validation path added, I would consider this a useful, low-risk compatibility improvement and would be comfortable merging it.

Lightheartdevs:
Thanks for taking care of the catalog generator's Python 3.9 import path; this is exactly the kind of contributor-facing compatibility issue that is easy for newer CI runtimes to miss.

What looks good:
- dream-server/scripts/generate-extensions-catalog.py only adds rom __future__ import annotations, which is the minimal fix for evaluated dict | None / list[dict] annotations on Python 3.9.
- The script does not appear to rely on runtime evaluation of annotations, so deferring them is behavior-preserving and consistent with nearby scripts that already use the same future import.
- The change is tiny and safe to take either as this standalone PR or as part of #1580, since the relevant hunk is equivalent there.

What I checked:
- Static diff review against current main.
- python -m py_compile scripts/generate-extensions-catalog.py passed.
- python scripts/generate-extensions-catalog.py --output ... generated a catalog with 33 extensions on the available Python runtime. I do not have Python 3.9 in this Windows workspace, so I could not reproduce that exact interpreter lane locally.

Thanks again for keeping the extension authoring path portable.

Lightheartdevs (approved):

PR #1564, opened 2026-06-14
fix(security): build service-registry cache with mktemp, not a predictable /tmp path

w3-surfer:
## Problem (local code injection / privilege escalation)

`lib/service-registry.sh` cached the parsed manifest registry at a **predictable** path and then **sourced it as bash**:

```bash
_SR_CACHE="/tmp/dream-service-registry.$$.sh"
...
"$PYTHON_CMD" - ... > "$_SR_CACHE"   # write
. "$_SR_CACHE"                        # source == execute
```

The path lives in world-writable `/tmp` with only the PID (`$$`) as entropy, so on a multi-user host a local unprivileged user can:

- **pre-create it as a symlink** → the `>` redirect truncates an arbitrary file the runner can write; or
- **win the write→source race (TOCTOU)** → inject arbitrary bash that runs with the runner's privileges.

`install-core.sh` and ~20 other scripts (`dream-doctor.sh`, `health-check.sh`, `dream-test.sh`, …) source this library, frequently under `sudo` — so this is a realistic local code-injection / privilege-escalation vector. It is the only place in the repo that sources a file from a predictable `/tmp` path.

## Fix

Create the cache with `mktemp` — unpredictable name, mode `0600`, `O_EXCL` race-safe — built lazily at the point of use, mirroring the `mktemp` idiom already used elsewhere in the repo:

```bash
_SR_CACHE="$(mktemp "${TMPDIR:-/tmp}/dream-service-registry.XXXXXX")" || _SR_CACHE=""
```

Downstream code references `$_SR_CACHE` (not the literal path), so there is **no behavior change** for legitimate use. Nothing in the repo depends on the old filename.

## Verification

- `mktemp` on macOS yields `-rw-------` (0600), random name under `$TMPDIR`.
- `tests/test-service-registry.sh` → **219 passed, 0 failed**
- `tests/test-service-registry-cache.sh` → **15 passed, 0 failed**
- `make lint` passes.

### DIFF
--- a/dream-server/lib/service-registry.sh
+++ b/dream-server/lib/service-registry.sh
@@ -5,7 +5,9 @@
 EXTENSIONS_DIR="${SCRIPT_DIR:-$(pwd)}/extensions/services"
 _SR_LOADED=false
 _SR_FAILED=false
-_SR_CACHE="/tmp/dream-service-registry.$$.sh"
+# Created on demand in sr_load() via mktemp. Its contents are sourced as bash,
+# so a predictable /tmp path would be a local code-injection vector.
+_SR_CACHE=""
 
 # Caching for compose flags (session-level)
 _SR_COMPOSE_FLAGS_CACHE=""
@@ -79,6 +81,11 @@ sr_load() {
             return 0
         fi
     fi
+
+    # Create the cache as a race-safe, private (mode 0600) temp file. Its
+    # contents are sourced as bash below, so a predictable /tmp name would be a
+    # local code-injection vector (symlink / TOCTOU race).
+    _SR_CACHE="$(mktemp "${TMPDIR:-/tmp}/dream-service-registry.XXXXXX")" || _SR_CACHE=""
 
     if ! "$PYTHON_CMD" - "$EXTENSIONS_DIR" <<'PYEOF' > "$_SR_CACHE"
 import yaml, sys, os


gabsprogrammer:
Thanks for the PR.

This is a real and worthwhile hardening fix. In this case, the cache file is not merely temporary data—it is later sourced as Bash—so the security properties of how that file is created matter. Moving away from a predictable PID-based path and switching to `mktemp` is the correct approach, as it eliminates the practical symlink and TOCTOU risks associated with predictable temporary filenames.

### Validation Performed

I validated the change locally using:

```bash id="l4r2e8"
bash tests/test-service-registry.sh
bash tests/test-service-registry-cache.sh
bash -lc 'make lint'
git diff --check origin/main...origin/pr-1564
```

I also manually verified the following:

* The cache file is created under `TMPDIR`
* The filename uses a randomized `dream-service-registry.XXXXXX` pattern
* The file is created with mode `600`
* The temporary cache file is removed after `sr_load()` completes

Based on that validation, the implementation looks correct and I did not observe any behavioral regressions.

### Requested Regression Coverage

The one thing I would like to see before merge is a focused regression test covering the security property that this PR is introducing.

The existing test suite validates service registry cache behavior in general, but it does not specifically guard against a future regression where someone accidentally reintroduces a predictable sourced temporary file.

Suggested coverage could include:

* Verify the registry no longer uses a PID-based cache filename
* Execute `sr_load()` with a controlled `TMPDIR`
* Assert that the generated `_SR_CACHE` path is created within that directory
* Assert that no `dream-service-registry.*` file remains after loading completes
* Optionally stub or wrap `mktemp` to verify that it is actually being used

I do not think this needs to be extensive. A small targeted test would be enough to lock in the security guarantee that motivated the change.

### Non-Blocking Cleanup

One additional observation: if `mktemp` fails, the current implementation appears to continue into the Python redirection path with an empty `_SR_CACHE` value. In that scenario, the shell emits a raw redirection error before eventually reaching the normal `_SR_FAILED=true` handling path.

I do not see an obvious security issue there, but an explicit early failure branch would likely make the behavior cleaner and easier to diagnose.

### Conclusion

Overall, this is a good security improvement with a very small blast radius. The use of `mktemp` is the correct fix, the implementation behaves as expected in testing, and I did not find any obvious regressions.

I would just like to see a focused regression test added so that this security property remains protected going forward.

Lightheartdevs:
Thanks for tightening the service-registry cache path; because this generated file is sourced as Bash, the temp-file creation semantics really matter here.

- The current standalone branch is the safer version of this fix: dream-server/lib/service-registry.sh now creates _SR_CACHE inside sr_load() with mktemp, sources it, and removes it afterward. That avoids both the old predictable /tmp/dream-service-registry..sh path and unnecessary temp-file creation when the registry is merely sourced but never loaded.
- The aggregate #1580 branch contains the same security theme, but its hunk creates the mktemp file at library source time and does not include the clearer on-demand implementation from this PR. If this fix is accepted, this standalone PR looks like the better merge vehicle.
- I do not see a new code blocker, but this is security-sensitive shell infrastructure with no GitHub checks reported on the branch, so a maintainer should decide whether the focused local validation below is enough or whether to wait for the normal Linux CI lane.

What I checked:
- Static review of dream-server/lib/service-registry.sh, focused on temp-file creation, source/removal flow, and fallback behavior.
- Git Bash: ash tests/test-service-registry-cache.sh passed, 15 tests.
- Git Bash: ash tests/test-service-registry.sh was not a usable signal in this Windows environment; it failed broad manifest/compose YAML parsing across many existing services, which appears environment/tooling-related rather than specific to this one-line security change.

Thanks again for pushing the hardening fix.

Lightheartdevs:
Validation note: this PR's current diff is patch-equivalent to the commit tested in the green fleet stack `test/fleet-contrib-stack-20260621T150103Z`.

Receipt:
- Fleet run: `/home/michael/dream-fleet-test/runs/2026-06-21T22-31-36Z`
- Result: User Green PASS, all six release gates green
- Enabled hosts: tower2, strix-halo, spark, m5-mbp, windows-laptop, strixy
- Real product bugs: 0
- Harness limitations: 0
- Environment notes: 0
- Current `main` is still the tested base `ee749a42`, and this PR merges cleanly into it with `git diff --check` clean.

Lightheartdevs (approved):
Approved after patch-equivalence check against the green fleet stack and release-grade fleet validation in /home/michael/dream-fleet-test/runs/2026-06-21T22-31-36Z.

PR #1565, opened 2026-06-14
fix(privacy-shield): report full multi-word PII type in get_stats()

w3-surfer:
## Problem

`PIIDetector.get_stats()` derived the list of PII types like this:

```python
'pii_types': list(set(
    token.split('_')[1] for token in self.pii_map.keys()
))
```

Tokens are formatted `<PII_{type}_{hash}>` (see `_generate_token`). `split('_')[1]` keeps only the **first** underscore-delimited segment of the type, so any multi-word type is truncated:

- `ip_address` → `"ip"`
- `credit_card` → `"credit"`

Any consumer reading `get_stats()["pii_types"]` (metrics, logging, the privacy-shield API) gets a wrong type label.

**Reproduce:**

```console
$ python3 -c "import pii_scrubber; d=pii_scrubber.PIIDetector(); d.scrub('Server IP 192.168.1.100'); print(d.get_stats()['pii_types'])"
['ip']      # should be ['ip_address']
```

## Fix

Parse the type as everything between the prefix and the trailing `_<hash>`:

```python
token[len(self.token_prefix):-len(self.token_suffix)].rsplit('_', 1)[0]
```

Single-word types (`email`, `phone`) are unaffected. Adds a regression test for the multi-word (`ip_address`) case — the existing stats tests only covered single-word types, which is why this slipped.

## Verification

- After fix: `get_stats()['pii_types']` → `['ip_address']`.
- `pytest tests/test_pii_scrubber.py` → **29 passed**.
- The new test **fails on the pre-fix code** (`assert 'ip_address' in ['ip']`) and passes after.

### DIFF
--- a/dream-server/extensions/services/privacy-shield/pii_scrubber.py
+++ b/dream-server/extensions/services/privacy-shield/pii_scrubber.py
@@ -130,7 +130,11 @@ class PIIDetector:
         return {
             'unique_pii_count': len(self.pii_map),
             'pii_types': list(set(
-                token.split('_')[1] for token in self.pii_map.keys()
+                # A type may itself contain underscores (ip_address, credit_card),
+                # so take everything between the prefix and the trailing _<hash>
+                # rather than split('_')[1], which would yield "ip" / "credit".
+                token[len(self.token_prefix):-len(self.token_suffix)].rsplit('_', 1)[0]
+                for token in self.pii_map.keys()
             ))
         }
 
--- a/dream-server/extensions/services/privacy-shield/tests/test_pii_scrubber.py
+++ b/dream-server/extensions/services/privacy-shield/tests/test_pii_scrubber.py
@@ -221,3 +221,11 @@ class TestStats:
         stats = detector.get_stats()
         assert stats["unique_pii_count"] == 0
         assert stats["pii_types"] == []
+
+    def test_stats_reports_multiword_type(self, detector):
+        # Regression: a multi-word PII type (ip_address) must be reported in
+        # full, not truncated to "ip" by a naive token.split('_')[1].
+        detector.scrub("Server IP 192.168.1.100")
+        types = detector.get_stats()["pii_types"]
+        assert "ip_address" in types
+        assert "ip" not in types


gabsprogrammer:
Thanks for the PR.

This is a useful and legitimate fix. The issue is real: `get_stats()` was only reporting the first underscore-delimited segment of a PII type, meaning multi-word identifiers such as `ip_address`, `api_key`, and `credit_card` could be incorrectly represented in metrics, logs, or downstream API consumers.

I reviewed the parser change itself, and the approach looks correct given the current token format. Since generated tokens are structured as:

```text
<prefix>_<type>_<hash>_<suffix>
```

extracting everything before the final hash segment is the appropriate fix. It preserves the full PII type regardless of how many underscores are present in the type name.

### Validation Performed

I validated the change locally using:

```bash id="f3m8qv"
python -m pytest dream-server/extensions/services/privacy-shield/tests/test_pii_scrubber.py
```

```bash id="w2k5cx"
python -m pytest dream-server/extensions/services/privacy-shield/tests
```

```bash id="a9z7pt"
python -m py_compile \
    dream-server/extensions/services/privacy-shield/pii_scrubber.py \
    dream-server/extensions/services/privacy-shield/proxy.py
```

Results:

* Privacy Shield test suite passes successfully
* 51 tests passed locally
* No compilation issues detected
* No obvious behavior regressions observed

### Requested Regression Coverage

The one improvement I would like to see before merge is slightly broader regression coverage.

The current regression test validates `ip_address`, which demonstrates the bug, but the issue is not specific to IP addresses. It affects any PII type containing one or more underscores.

I would prefer to see the new stats regression parameterized across all currently supported multi-word types, for example:

* `ip_address`
* `api_key`
* `credit_card`

That would ensure the test protects the entire class of bug rather than a single representative example. It also makes future refactoring safer if additional underscore-based PII types are introduced later.

### Non-Blocking Suggestion

Since this code already touches `get_stats()`, it may also be worth making `pii_types` deterministic before returning it.

Today the implementation appears to rely on:

```python id="eh71ow"
list(set(...))
```

which can produce arbitrary ordering. Returning something like:

```python id="m8j6qe"
sorted(...)
```

would provide stable output for API consumers, UI rendering, metrics snapshots, and future tests.

This is not related to the bug being fixed and should not block the PR, but it would be a small improvement while the code is already being modified.

### Conclusion

Overall, this is a good fix with a clear root cause, a low-risk implementation, and meaningful impact for consumers of the statistics API.

I would just strengthen the regression coverage to cover all current underscore-based PII types so that the entire class of issue remains protected going forward.

Lightheartdevs:
Thanks for fixing the Privacy Shield stats parsing; the repro in the PR body makes the bug and the intended behavior very clear.

What looks good:
- dream-server/extensions/services/privacy-shield/pii_scrubber.py now extracts the PII type by removing the token prefix/suffix and splitting from the right, so underscore-bearing types like ip_address survive intact instead of becoming ip.
- The PR adds a focused regression in dream-server/extensions/services/privacy-shield/tests/test_pii_scrubber.py; that is stronger than the aggregate #1580 version, which carries the runtime hunk but not this test file change.
- The change is narrowly scoped to Privacy Shield stats output and does not alter the scrubbing/replacement token format itself.

What I checked:
- Static diff review against current main and comparison with #1580's overlapping hunk.
- python -m py_compile extensions/services/privacy-shield/pii_scrubber.py passed.
- python -m pytest extensions/services/privacy-shield/tests/test_pii_scrubber.py -q passed, 29 tests.

Lightheartdevs (approved):

PR #1559, opened 2026-06-14
feat(dashboard): add theme selector to settings

BittuBarnwal7479:
## Summary

The dashboard already supports multiple themes through `ThemeContext`, but Settings did not expose a clear way to select them.

This PR adds an appearance section to settings and lets allow users to switch between the themes provided by the existing theme implementation.

Resolves: #1557 

## Release Lane

- [ ] Stable hotfix targeting `release/2.5.x`
- [x] Mainline change targeting `main`
- [ ] Next-minor work targeting the next feature/minor release
- [ ] Not sure; reviewer should help classify

Stable hotfix reason:

```text
<!-- If this targets release/2.5.x, explain the current-stable user problem it fixes. -->
```

## Changed Surface

- [ ] Docs only
- [ ] Tests only
- [x] Dashboard UI
- [ ] Dashboard API / host agent
- [ ] Installer / bootstrap / lifecycle
- [ ] Docker Compose / service manifests
- [ ] Model routing / Hermes / capabilities
- [ ] Network exposure / auth / proxy
- [ ] Dependencies / runtime wiring

## Risk And Validation

- Risk level: Low
- Validation run:
  - [x] `git diff --check`
  - [ ] Markdown/link sanity for docs
  - [ ] Focused tests listed below
  - [x] Dashboard lint/test/build
  - [ ] Extension audit / compose validation
  - [ ] Release-grade fleet or scoped hardware validation
  - [ ] Stable-lane patch validation, if targeting `release/2.5.x`
  - [ ] Not required because: 

Commands/results:

- npm run lint ✅
- npm test -- --run ✅
- npm run build ✅
- git diff --check ✅

## Operational Change Check

If this PR touches installer phases, bootstrap logic, compose stack generation,
service manifests, dashboard API control flows, Hermes, model routing, GPU or
runtime detection, lifecycle commands, host mutation, or network exposure, it
requires release-grade fleet validation before release unless the PR explains a
narrower equivalent.

- [x] This is not an operational change.
- [x] This is an operational change and validation is recorded above.
- [ ] This is an operational change and validation is intentionally deferred for:

## Notes For Reviewers

- No backend, runtime, infrastructure, networking, or installer changes.
- Uses the existing ThemeContext implementation and storage flow.
- Verified theme selection for:
-- Dream
-- Lemonade
-- Light
-- Arctic

PR #1555, opened 2026-06-13
Fix bugs and TODOs in sample-code.py

HimanshuShukla671:
## Summary

This PR resolves the open `TODO` and fixes several intentional bugs in the `dream-server/examples/sample-code.py` demo script.

**Changes included:**
- **Added validation to `process_users`**: Added type checking and verified that required keys (`name`, `email`, `age`) are present to prevent `KeyError`s.
- **Fixed logic bug in `process_users`**: Corrected the age boundary so that users who are exactly 18 are properly flagged as `adult` rather than `minor`.
- **Fixed file descriptor leak in `read_config`**: Wrapped the `open()` call in a `with` context manager to safely manage file resources.
- **Prevented `ZeroDivisionError` in `calculate_average`**: Added an early return for empty lists and simplified the calculation using the built-in `sum()` function.

## AI Assistance

An AI coding assistant helped identify the intentional bugs within the sample code file, draft the code fixes, verify the syntax, and write the commit messages.

## Release Lane

- [ ] Stable hotfix targeting `release/2.5.x`
- [x] Mainline change targeting `main`
- [ ] Next-minor work targeting the next feature/minor release
- [ ] Not sure; reviewer should help classify

Stable hotfix reason:

```text

```

## Changed Surface

- [x] Docs / Examples only
- [ ] Tests only
- [ ] Dashboard UI
- [ ] Dashboard API / host agent
- [ ] Installer / bootstrap / lifecycle
- [ ] Docker Compose / service manifests
- [ ] Model routing / Hermes / capabilities
- [ ] Network exposure / auth / proxy
- [ ] Dependencies / runtime wiring

## Risk And Validation

- Risk level: Low
- Validation run:
  - [ ] `git diff --check`
  - [ ] Markdown/link sanity for docs
  - [x] Focused tests listed below
  - [ ] Dashboard lint/test/build
  - [ ] Extension audit / compose validation
  - [ ] Release-grade fleet or scoped hardware validation
  - [ ] Stable-lane patch validation, if targeting `release/2.5.x`
  - [ ] Not required because: 

Commands/results:

```text
python3 -m py_compile dream-server/examples/sample-code.py
(returns successfully without syntax errors)
```

## Operational Change Check

- [x] This is not an operational change.
- [ ] This is an operational change and validation is recorded above.
- [ ] This is an operational change and validation is intentionally deferred for:

## Notes For Reviewers

None.

PR #1573, opened 2026-06-15
Feat native rag

HimanshuShukla671:
## Overview
This PR implements **Native RAG / Knowledge Base** functionality for DreamServer (ODS). It enables users to upload documents (PDF, TXT, MD, CSV) to a Knowledge Base and have the AI use them as context for its responses when the "Use Knowledge" toggle is enabled.

## Key Changes
- **Database & Storage (`qdrant`, `knowledge.py`)**
  - Integrated `Qdrant` as the vector database for storing document chunks.
  - Added `QDRANT_API_KEY` auth and lazy, retryable database initialization using `AsyncQdrantClient` to ensure the dashboard API event loop remains non-blocking.
  - Implemented upload-size enforcement (10MB limit) and API key auth for mutating/list endpoints.
- **RAG Implementation (`talk.py`)**
  - Added the `use_knowledge` flag to all messaging paths (normal messages, text-like attachments, and image attachments).
  - **Explicit Error Handling:** If the knowledge base is unavailable, the endpoints return an explicit `503` error instead of silently falling back to a context-less LLM response.
  - Added document chunking, overlap logic, and text extraction.
  - Embeddings are generated using the existing `/embed` contract.
- **Frontend / UI (`Knowledge.jsx`, `ODSTalk.jsx`, `core.js`)**
  - Added a new UI panel to upload, list, and manage Knowledge Base documents.
  - Wired up the "Use Knowledge" toggle to the Talk interface.
- **Configuration (`docker-compose.base.yml`)**
  - Wired in the `QDRANT_API_KEY` securely to the environment.

## Validation & Testing
- Added comprehensive automated tests in `test_knowledge.py` covering:
  - Auth rejection for unauthenticated users.
  - Size and unsupported format rejections.
  - Qdrant unavailable / retry behavior using `AsyncMock`.
  - Document-scoped deletion targeting specific `doc_id`s.
  - Text and Image attachment context injection.
  - Explicit unavailable paths (503).
- All CI checks pass, including dashboard `api` & `frontend`, Python lint/type checks, and integration smoke tests.


gabsprogrammer:
Thanks for the PR.

Native RAG is a valuable direction for DreamServer, and I like the idea of exposing a first-party Knowledge experience rather than relying exclusively on external UI integrations. However, after reviewing the implementation and validating the changes locally, I do not think this is ready to merge yet.

This feature touches multiple critical surfaces simultaneously: authentication, file uploads, vector storage, embeddings, Dream Talk prompt routing, frontend navigation, dependencies, and runtime service initialization. Given that scope, I think it needs a significantly stricter review and additional hardening before it is safe to ship.

### Validation Performed

I validated the PR locally with the following:

```bash id="h9r4mz"
git diff --check origin/main...origin/pr-1573
```

Result:

* Failed due to trailing whitespace in multiple files

```bash id="4wkd3e"
npm ci
npm run lint
npm run build
```

Results:

* `npm run lint` passed
* `npm run build` passed

```bash id="8n6yvp"
npm test
```

Result:

* Failed because `DreamTalk.test.jsx` still expects `{ text }`
* The PR now sends `{ text, use_knowledge: false }`

```bash id="u5x1cg"
python -m pytest dream-server/extensions/services/dashboard-api/tests/test_talk.py
```

Result:

* Blocked in my local environment because importing `main.py` now hard-requires `qdrant_client`

---

## Blocking Issues

### 1. Knowledge API Authentication Is Currently a Stub

The most significant issue is that `_require_auth()` is currently implemented as:

```python id="qv1bse"
pass
```

The new endpoints can:

* Upload documents
* List indexed files
* Delete indexed content
* Generate embeddings
* Mutate Qdrant collections

These are privileged operations and should not ship without real authentication.

At minimum, I would expect either:

* The existing dashboard admin API-key pattern; or
* A deliberate authenticated session flow with clear CSRF protections

In its current form, the API surface is too open for mutating upload/delete operations.

### 2. Qdrant API Key Is Not Wired

DreamServer already supports `QDRANT_API_KEY`, and installers generate and configure it.

The new client initialization:

```python id="k1sm9a"
QdrantClient(url=QDRANT_URL)
```

does not appear to pass the configured API key.

As a result, deployments using a secured Qdrant instance are likely to fail when attempting to:

* Create collections
* Insert vectors
* Search vectors
* Delete vectors

The new code should align with the existing DreamServer deployment model.

### 3. Embedding Dimension Default Appears Incorrect

The implementation currently defaults:

```python id="wz8k6r"
EMBEDDING_DIM = 384
```

with a comment referencing `all-MiniLM-L6-v2`.

However, DreamServer's default embedding service is currently based on:

`BAAI/bge-base-en-v1.5`

which does not use a 384-dimensional embedding space.

This creates a real risk that:

* Collections are created with the wrong vector size
* Inserts fail
* Searches fail
* Existing deployments become inconsistent

Vector dimensions should either:

* Be derived dynamically from the embedding model/service; or
* Be configured consistently with `EMBEDDING_MODEL`

### 4. Qdrant Is Initialized at Import Time

The router currently performs network initialization and collection setup during module import.

This creates several problems:

* Dashboard startup depends on Qdrant availability
* Tests become dependent on Qdrant
* Importing modules performs network I/O
* If Qdrant is unavailable at startup but becomes available later, the service remains degraded until restart

This should be made lazy, retryable, or moved into application lifespan handling with explicit readiness semantics.

### 5. Attachment Flow Ignores `use_knowledge`

The frontend currently appends `use_knowledge` to attachment uploads, but:

```text id="x5e9dm"
/api/talk/attachment
```

does not appear to consume or honor that field.

As a result:

* Knowledge mode works for standard text prompts
* Knowledge mode does not work for attachment prompts

The UI currently suggests parity that does not exist.

Either:

* Fully wire the feature through attachment handling; or
* Hide/disable the option for attachment flows

### 6. Unrelated Changes Included in the PR

This PR also contains unrelated modifications including:

* Settings appearance changes
* `examples/sample-code.py` updates

These are not part of Native RAG functionality and make review significantly harder.

The appearance change also reintroduces a previously discussed concern:

`ThemeContext` supports multiple themes (`dream`, `lemonade`, `light`, `arctic`), while the UI only exposes Dark/Light and can misrepresent valid existing configurations.

Please split or rebase the PR so RAG functionality can be reviewed independently.

### 7. Missing Backend Test Coverage

For a feature of this scope, I would expect focused backend coverage for at least:

* Unauthenticated requests are rejected
* Unsupported file types are rejected
* Upload size limits are enforced
* Qdrant unavailable returns a clear 503
* Qdrant API key is passed correctly
* Embedding dimension mismatches are handled cleanly
* Document deletion only affects the targeted `doc_id`
* Dream Talk injects context only when `use_knowledge=True`
* Attachment behavior is either supported or intentionally rejected

At the moment, those behaviors appear largely untested.

---

## Additional Concerns

These are not necessarily merge blockers individually, but they should be addressed before release:

* The Knowledge page appears in the sidebar even when Qdrant or embedding services are unavailable
* The UI advertises a **10 MB maximum upload size**, but the backend currently reads uploaded files without enforcing that limit
* Synchronous Qdrant operations are executed inside async request handlers and may block the event loop under load
* The PR description remains largely the default template and does not document:

  * Risk profile
  * Changed surfaces
  * Validation performed
  * Release expectations
  * Operational requirements
* `git diff --check` currently fails, so the patch is not yet clean

---

## Conclusion

I like the direction of this feature, and I agree that first-party Knowledge/RAG support belongs in DreamServer. However, the current implementation introduces several issues around authentication, deployment compatibility, configuration correctness, startup behavior, testing, and scope management.

Before merge, I would strongly recommend:

* Implementing real authentication
* Wiring Qdrant API key support
* Aligning embedding dimensions with the actual embedding service
* Removing Qdrant initialization from import-time execution
* Resolving the attachment-path inconsistency
* Adding focused backend coverage
* Splitting unrelated commits into separate PRs
* Cleaning up the existing test failures and whitespace issues
* Providing a complete PR description with validation results

At that point I would be happy to take another look.

Lightheartdevs:
Thanks for taking on first-party Knowledge/RAG support; that direction fits Dream Server, but this version changes several high-risk runtime surfaces at once.

Quickest path to merge:
- In dream-server/extensions/services/dashboard-api/routers/knowledge.py, replace _require_auth()'s current pass with the existing dashboard auth/session pattern before exposing upload, list, delete, embedding, and Qdrant mutation endpoints. Also enforce the advertised upload size limit before reading the full file into memory.
- In dream-server/extensions/services/dashboard-api/routers/knowledge.py, wire QDRANT_API_KEY, stop creating/checking the Qdrant collection at module import time, and align EMBEDDING_DIM with the configured embedding service instead of defaulting to 384 for a different model family.
- In dream-server/extensions/services/dashboard/src/pages/DreamTalk.jsx and dream-server/extensions/services/dashboard-api/routers/talk.py, either fully support use_knowledge for /api/talk/attachment or hide/disable that option while an attachment is pending. Right now the UI sends the field on multipart upload, but the attachment endpoint does not consume it.
- Split unrelated changes in dream-server/extensions/services/dashboard/src/pages/Settings.jsx and dream-server/examples/sample-code.py out of this PR so the RAG feature can be reviewed on its own risk surface.
- Fix the trailing whitespace reported by git diff --check, complete the PR template with risk/validation, and add focused backend tests for auth rejection, upload limits, Qdrant unavailable/API-key behavior, vector dimension handling, deletion scope, and use_knowledge=True context injection.

What I checked:
- Static review of the changed Knowledge API, Talk API, DreamTalk UI, route registration, dependency, and unrelated UI/example diffs.
- git diff --check origin/main...origin/pr/1573 failed on trailing whitespace in knowledge.py, 	alk.py, Knowledge.jsx, and Settings.jsx.
- gh pr checks reports no checks for this branch.
- I did not run the full dashboard/API test suites after the static blockers above.

Thanks again for pushing the feature direction; narrowing and hardening this should make the next review much easier.

bong-water-water-bong (commented):
Significant feature addition touching the dashboard API, UI, and adding new dependencies. This needs closer review from the maintainers who know the dashboard architecture — I'd want to verify the knowledge router integration, new Python deps, and the frontend changes don't conflict with the feature readiness work in #1603. Not approving without deeper review.

HimanshuShukla671:
Thanks for the thorough review @Lightheartdevs and @bong-water-water-bong — all the blockers have been addressed in the latest push. Here's a point-by-point breakdown:

**1. Authentication (was `pass`)**
→ Replaced the no-op `_require_auth()` with `Depends(verify_api_key)` on all three Knowledge endpoints (upload, list, delete). This is the same auth pattern used by every other admin endpoint in the dashboard API. The nginx reverse proxy injects `Authorization: Bearer ${DASHBOARD_API_KEY}` on all `/api/` calls, so the frontend works without code changes.

**2. `QDRANT_API_KEY` not wired**
→ Now read from `os.environ.get("QDRANT_API_KEY", "")` and passed directly to `QdrantClient(url=..., api_key=...)`.

**3. Qdrant init at import time**
→ Replaced the module-level `try/except` block with a lazy `async get_qdrant()` helper. The module can now import cleanly even if Qdrant is down. First actual API call triggers initialization.

**4. Embedding dimension hardcoded to 384**
→ Removed the hardcoded `EMBEDDING_DIM = 384`. On first collection creation, we now probe the actual embedding service with a test call to `get_embeddings(["test"])` and use `len(result[0])` to detect the real dimension. This works regardless of which model is deployed.

**5. `use_knowledge` broken on attachment path**
→ Added `use_knowledge: bool = Form(False)` to `talk_attachment()`. When set, the handler queries the knowledge base and injects context into the prompt before streaming through Hermes — same pattern as `/api/talk/message/stream`.

**6. Unrelated changes (Settings.jsx / sample-code.py)**
→ Reverted the Settings.jsx Appearance toggle entirely — it conflicted with the multi-theme `ThemeContext` work merged in #1559. The sample-code.py changes were already merged separately in #1555, so no diff remains.

**7. Backend tests**
→ Added `test_knowledge.py` with regression coverage for:
- Unauthenticated requests → 401
- Unsupported file format → 400
- Upload size limit (10MB) → 413
- Qdrant unavailable → 503
- Deletion targets only the specified `doc_id`
- `use_knowledge=True` injects KB context into the prompt

**8. Upload size limit**
→ Added `MAX_UPLOAD_BYTES = 10 * 1024 * 1024` and enforced it *before* reading the full file into memory (`file.read(MAX_UPLOAD_BYTES + 1)`).

**9. Trailing whitespace**
→ Cleaned all flagged files with `sed` + `ruff format`. `git diff --check origin/main...HEAD` is now clean.

Ready for re-review. 🙏

HimanshuShukla671:
Resolved the merge conflicts with the recent `dream-server` -> `ods` refactor on main.

Lightheartdevs:
Thanks for the rework on the native Knowledge/RAG path; the latest push clearly addresses several of the earlier blockers around auth, upload limits, attachment flow, and test coverage.

Quickest path to merge:
- `ods/extensions/services/dashboard-api/routers/knowledge.py:48-66`: do not permanently cache a failed Qdrant initialization. Right now a first request while Qdrant or embeddings is still starting sets `_qdrant_initialized = True` with `_qdrant_client = None`, so later Knowledge uploads/lists and Talk RAG lookups never retry until dashboard-api restarts. Mark initialization complete only after a usable client exists, or use a retry/TTL path, and add a regression test for "first call fails, second call succeeds."
- `ods/extensions/services/dashboard-api/routers/knowledge.py:20` and `ods/docker-compose.base.yml:191-218`: pass `QDRANT_API_KEY` into the `dashboard-api` container. The router now reads the key, but the shipped compose environment still does not provide it, so secured Qdrant installs will behave as if no API key was configured.
- `ods/extensions/services/dashboard-api/routers/knowledge.py:21,33`: align the embedding client with the existing `EMBEDDING_URL` contract. The repo passes `EMBEDDING_URL=http://embeddings:80` to dashboard-api and documents/tests the TEI `/embed` endpoint (`ods/docker-compose.base.yml:237`, `ods/scripts/ods-test-functional.sh:183-190`, `ods/extensions/services/embeddings/README.md:30-45`), while this router reads a new `EMBEDDINGS_URL` variable and posts to `/v1/embeddings`. Either reuse `EMBEDDING_URL` and the shipped TEI request/response shape, or update compose, docs, and tests together.
- `ods/extensions/services/dashboard-api/helpers.py`, `routers/setup.py`, and `routers/workflows.py`: split or justify the unrelated typing-only edits if another revision is needed. They are small, but this PR already spans dashboard API, Talk, frontend routing, new deps, Qdrant, and embeddings.

What I checked:
- PR metadata, current head SHA, changed files, comments since the previous advisory, and the full diff against `origin/main`.
- GitHub checks for this head are green, including dashboard frontend/API, Python lint/type checks, shell lint, smoke/distro lanes, and integration-smoke.
- `git diff --check origin/main...HEAD` passes, and `python -m py_compile main.py routers/knowledge.py routers/talk.py helpers.py` passes locally.
- Local focused pytest could not run in this shell because the new PR dependency `qdrant_client` is not installed in the local Python environment; I treated GitHub's green API job as the test signal and relied on static/runtime-contract review for the issues above.

Thanks again for pushing this feature forward; these are narrower fixes than the first pass and should make the next review much cleaner.

HimanshuShukla671:
All feedback from the latest review has been addressed:
- Fixed Qdrant initialization to only set `_qdrant_initialized = True` if the client was successfully created. Added a regression test `test_knowledge_qdrant_retry` to verify that first-call failures still allow subsequent calls to succeed.
- Injected `QDRANT_API_KEY` into the `dashboard-api` service environment inside `ods/docker-compose.base.yml`.
- Aligned `EMBEDDING_URL` contract with the TEI standards (`POST /embed` with `{"inputs": texts}` shape and expecting the flat array).
- Reverted the unrelated typing edits from `helpers.py`, `setup.py`, and `workflows.py`.

Ready for the next pass! 🙏

Lightheartdevs:
Thanks for tightening the native Knowledge/RAG path again; this push fixes the Qdrant retry, dashboard-api env wiring, embedding endpoint contract, and unrelated helper edits from the previous pass.

Quickest path to merge:
- In `ods/extensions/services/dashboard-api/routers/talk.py`, apply `use_knowledge` before the image attachment early return around lines 758-767, or disable/hide the Knowledge toggle in `ods/extensions/services/dashboard/src/pages/ODSTalk.jsx` when `pendingAttachment.kind === 'image'`. The UI sends `use_knowledge` for every attachment (`ODSTalk.jsx` line 387), but image uploads go straight to `_stream_vision_chat` with only the caption/default prompt, so an enabled Knowledge state is silently ignored for images.
- Add a focused regression next to `ods/extensions/services/dashboard-api/tests/test_knowledge.py::test_talk_attachment_use_knowledge` for an image attachment with `use_knowledge=true`; the current test covers only the text-file attachment path.

What I checked:
- PR discussion since the last advisory, current head SHA, full diff, and the previous blocker list.
- `git diff --check origin/main...origin/pr/1573` passes.
- Status rollup is green for frontend, api, mypy, shell/Python/PowerShell lint, validate-compose/env, integration-smoke, Linux/macos smoke, and the distro matrix; only advisory/security-review jobs are skipped.
- Static review of `knowledge.py`, `talk.py`, `ODSTalk.jsx`, `test_knowledge.py`, and `docker-compose.base.yml` confirms the Qdrant retry, `QDRANT_API_KEY`, and `EMBEDDING_URL` fixes are present.

Thanks again for working this down to a much smaller remaining issue.

HimanshuShukla671:
Fixed the image attachment + Knowledge Base gap:

- **`talk.py`**: The image attachment path (`kind == "image"`) now queries the knowledge base and prepends context to the vision prompt before streaming, matching the text-file path. Previously it returned immediately to `_stream_vision_chat` and silently ignored `use_knowledge`.
- **`test_knowledge.py`**: Added `test_talk_image_attachment_use_knowledge` — mocks `_stream_vision_chat` to capture the `prompt_text` argument and asserts that `"Context from Knowledge Base"` and the KB content appear in the prompt when `use_knowledge=true` is sent with an image attachment.

Ready for re-review

Lightheartdevs:
Thanks for the repeated follow-up on native Knowledge/RAG; the current head fixes the image-attachment Knowledge gap from the last advisory and the test coverage is much stronger.

Quickest path to merge:
- `ods/extensions/services/dashboard-api/routers/knowledge.py`: move Qdrant calls off the FastAPI event loop. The router is `async`, but `QdrantClient.collection_exists`, `create_collection`, `upsert`, `scroll`, `delete`, and `search` are synchronous network calls; upload/list/delete/search can block every other dashboard-api request while Qdrant is slow or while a document is being indexed. Use `AsyncQdrantClient` or wrap the sync calls with `asyncio.to_thread`, then keep the current endpoint shape.
- `ods/extensions/services/dashboard-api/routers/knowledge.py`: once the Qdrant calls are async-safe, keep the current retry/auth/upload tests and add one targeted regression around the wrapper/error path so “Knowledge unavailable” behavior stays explicit instead of accidentally becoming a silent empty-context path.

What I checked:
- Current head against prior advisory comments; the image attachment `use_knowledge` path now reaches `_stream_vision_chat` with Knowledge context and has a regression test.
- Static diff review of `knowledge.py`, `talk.py`, dashboard `Knowledge.jsx`/`ODSTalk.jsx`, compose env wiring, and `test_knowledge.py`.
- GitHub checks: 29 passing, including `api`, `frontend`, Python lint/type check, shell lint, compose validation, matrix smoke, and secret scan.
- `git diff --check origin/main...origin/pr/1573` passed.

Thanks again for iterating on this one.

HimanshuShukla671:
The final two blockers have been resolved:

- **Moved Qdrant calls off the event loop**: Swapped `QdrantClient` for `AsyncQdrantClient` across the board in `knowledge.py`. `collection_exists`, `create_collection`, `upsert`, `scroll`, `delete`, and `search` are all properly `await`ed now, ensuring upload/list/delete/search won't block the FastAPI event loop.
- **Explicit Knowledge Unavailable error path**: `search_knowledge_base` now explicitly raises an `HTTPException(503)` if it fails or if Qdrant is down. I removed the silent `try/except` fallbacks from all four paths in `talk.py`, meaning if a user toggles `use_knowledge=True` and the DB is down, they explicitly get a 503 error rather than a silent empty-context fallback. Added `test_talk_message_use_knowledge_unavailable` to verify this behavior.

Ready for what is hopefully the final review!

Lightheartdevs:
Thanks for chasing down the async-safety issue in the native Knowledge/RAG path; swapping the router to `AsyncQdrantClient` is the right shape for this part of the dashboard API.

Quickest path to merge:
- `ods/extensions/services/dashboard-api/tests/test_knowledge.py:80-92`: update the delete test to use async-compatible mocks now that `knowledge.py:201` awaits `qdrant.delete(...)`. The current plain `MagicMock` produces `TypeError: object MagicMock can't be used in 'await' expression`; use `AsyncMock` for `delete` or a small async stub and assert the awaited call args.
- `ods/extensions/services/dashboard-api/tests/test_knowledge.py:194-207`: update the Qdrant retry test for the new client name and async methods. It still monkeypatches `knowledge.QdrantClient`, which no longer exists after the switch to `AsyncQdrantClient`, and the dummy `collection_exists` should be awaitable because `get_qdrant()` now does `await client.collection_exists(...)`.
- Re-run the dashboard API test job/focused command (`pytest tests/test_knowledge.py -v` from `ods/extensions/services/dashboard-api`, or the full `pytest tests/ -v --cov=. --cov-report=term --cov-report=lcov:coverage.lcov`) before the next review pass.

What I checked:
- PR discussion since the previous advisory, current head SHA `cb688c322814c6a599104d9af8e2286452fe8907`, changed files, and the last-push delta from `92926f6b` to current head.
- `git diff --check origin/main...pr-1573` passes, and the async Qdrant conversion in `knowledge.py` is present.
- GitHub checks: lint/type/compose/smoke/frontend pass, but the Dashboard `api` job fails with exactly the two `test_knowledge.py` failures above; the Claude review jobs also report failure/skipped states.

Thanks again for working this down; this pass looks like test repair rather than another design-level blocker.

HimanshuShukla671:
Tests repaired:
- `test_knowledge_deletion_targets_doc_id`: Swapped `MagicMock` for `AsyncMock` to accommodate the new `await qdrant.delete(...)`.
- `test_knowledge_qdrant_retry`: Updated the monkeypatch to target `AsyncQdrantClient` instead of `QdrantClient` and made the dummy `collection_exists` method async.

Thanks again for the review!

Lightheartdevs:
Thanks for continuing to harden the native Knowledge/RAG path; this head closes the async Qdrant test failures from the last pass and keeps the Talk attachment behavior covered.

What looks good:
- `ods/extensions/services/dashboard-api/routers/knowledge.py` now uses `AsyncQdrantClient`, lazy/retryable initialization, `QDRANT_API_KEY`, the existing `EMBEDDING_URL` `/embed` contract, upload-size enforcement, and dashboard API auth on the mutating/list endpoints.
- `ods/extensions/services/dashboard-api/routers/talk.py` now applies `use_knowledge` consistently for normal messages, text attachments, and image attachments, with explicit 503 behavior when Knowledge is unavailable instead of silently dropping context.
- `ods/extensions/services/dashboard-api/tests/test_knowledge.py` covers auth rejection, size/format rejection, Qdrant unavailable/retry behavior, doc-id-scoped deletion, text/image attachment context injection, and the explicit unavailable path.

Human decision point:
- I do not see a remaining code blocker in this pass. Because this is a new runtime feature spanning dashboard API/UI, Qdrant, embeddings, Talk prompt routing, and new dependencies, a maintainer should decide whether the current green CI plus focused tests are enough for `main`, or whether to require release-grade/runtime validation before release. It would also help to copy the changed surface and validation summary from the discussion into the PR body before merge.

What I checked:
- Current head SHA, PR discussion since the previous advisory, changed files, and static review of `knowledge.py`, `talk.py`, `Knowledge.jsx`, `ODSTalk.jsx`, `core.js`, `docker-compose.base.yml`, and the new tests.
- `git diff --check origin/main...origin/pr/1573` passes.
- Local focused test in an isolated venv: `python -m pytest tests/test_knowledge.py -q` passed, 9 tests.
- GitHub checks for this head have product CI green, including dashboard `api` and `frontend`, Python lint/type check, shell/PowerShell lint, validate-compose/env, integration-smoke, Linux/macOS smoke, and distro matrix. The `Claude Code Review` workflow is cancelled/skipped, not a product-test failure.

Thanks again for working this down through the review loop.

PR #1574, opened 2026-06-15
Feature: Add SenseVoice/FunASR as self-hosted STT backend

### LINKED ISSUE #1484: Feature: Add SenseVoice/FunASR as self-hosted STT backend
> [!NOTE]
> **License and capability clarification (2026-07-14):** FunASR is a toolkit, not a single checkpoint. The [FunASR](https://github.com/modelscope/FunASR#license) and [SenseVoice](https://github.com/FunAudioLLM/SenseVoice#license) repository source code is MIT; model weights follow each model card. [SenseVoiceSmall](https://huggingface.co/FunAudioLLM/SenseVoiceSmall) supports Chinese, Cantonese, English, Japanese, and Korean, and its weights use the linked FunASR Model Open Source License Agreement. [Fun-ASR-Nano-2512](https://huggingface.co/FunAudioLLM/Fun-ASR-Nano-2512) is Apache-2.0. Language coverage, punctuation, and performance depend on the selected model and runtime configuration.

Hi! DreamServer's approach to turning any PC into an AI server is great.

For the voice agent / STT functionality, [FunASR](https://github.com/modelscope/FunASR) offers an industrial-grade solution:

**FunASR highlights:**
- **OpenAI-compatible API** — `POST /v1/audio/transcriptions` (drop-in for OpenAI Whisper API)
- **50+ languages**, VAD, punctuation, speaker diarization all built-in
- **170x realtime** on GPU
- Self-hosted, fully private

```bash
pip install funasr vllm fastapi uvicorn python-multipart
funasr-server --device cuda
# Now serving at http://localhost:8000/v1/audio/transcriptions
```

This fits DreamServer's self-hosted philosophy perfectly — users can run speech recognition locally without relying on cloud APIs.

- FunASR: https://github.com/modelscope/FunASR (16K+ stars)
- SenseVoice: https://github.com/FunAudioLLM/SenseVoice (8K+ stars)

bettercalln1ck:
## Summary

Adds **SenseVoice** (via the FunASR toolkit) as a new **optional** speech-to-text service, mirroring the existing `whisper` extension. It exposes the same OpenAI-compatible STT contract Whisper does (`POST /v1/audio/transcriptions` → `{"text": ...}`, plus `GET /health`), so any app already pointed at an OpenAI STT base URL can use it by swapping the URL.
This is **additive**: the default voice wiring (Open WebUI, Hermes) is unchanged and still points at Whisper. SenseVoice is opt-in like every other optional extension.

Resolves #1484 

## AI Assistance

I drafted the design spec and the implementation plan. Ai coded `server.py`, Dockerfile, compose files, manifest, registry edits, unit tests, and docs according to implementation plan. I reviewed the full diff, made the scope decisions (standalone service vs. replacement), ran all validation locally, and verified the service end-to-end (build + real transcription) on Apple Silicon.

## Release Lane

## Changed Surface

- [ ] Docs only
- [ ] Tests only
- [ ] Dashboard UI

## Risk And Validation

- Risk level: **Medium** — new optional service (manifest + 4 compose files +
  custom build + registry entries). Additive only; no change to existing
  services, ports, or default voice routing.

- Validation run:
  - [x] `git diff --check`
  - [ ] Focused tests listed below
  - [ ] Not required because: 

Commands/results:

```text
make lint                                             -> All lint checks passed
make test                                             -> exit 0 (full suite)
bash scripts/validate-manifest-schema.sh --strict     -> sensevoice: Valid
bash tests/test-validate-manifests.sh                 -> 7 passed, exit 0
python3 scripts/check-dependency-pins.py              -> [PASS]
python3 tests/contracts/test-network-exposure-contracts.py -> all PASS, exit 0
bash tests/test-validate-env.sh                       -> 9 passed, exit 0
bash scripts/resolve-compose-stack.sh                 -> includes sensevoice, exit 0
pytest extensions/services/sensevoice/tests/          -> 5 passed
docker build + GET /health                            -> {"status":"ok"}
real transcription (Apple Silicon, CPU)               -> correct text returned
```

## Operational Change Check

If this PR touches installer phases, bootstrap logic, compose stack generation,
service manifests, dashboard API control flows, Hermes, model routing, GPU or
runtime detection, lifecycle commands, host mutation, or network exposure, it
requires release-grade fleet validation before release unless the PR explains a
narrower equivalent.

The change is a self-contained opt-in service. It adds no dependency to any existing service, changes no port defaults, schemas, or model routing, and leaves docker-compose.base.yml untouched.

## Notes For Reviewers

- **Additive, no regressions**: default STT for Open WebUI/Hermes stays Whisper.
- **Custom build is deliberate**: no upstream OpenAI image exists; AutoModel + FastAPI, no vllm. Image runs CPU everywhere, CUDA via SENSEVOICE_DEVICE=cuda (set by the NVIDIA overlay).
- **AMD/Apple overlays run CPU-mode (no vllm-free ROCm/Metal path)** — shipped for GPU device isolation / structural parity with Whisper, documented honestly in the README; not an acceleration claim.
- **core-service-ids.json updated**— required so the resolver trusts the local build (untrusted user extensions with build: are rejected); same pattern as brave-search.
- **Out of scope (follow-up)**: wiring SenseVoice into the dashboard "Voice Assistant" feature / selectable STT backend.


gabsprogrammer:
Thanks for the PR.

This is a genuinely useful direction for DreamServer. Adding SenseVoice/FunASR as an optional self-hosted STT backend fits well within the project's goals, preserves Whisper as the default experience, and gives users an additional local/private transcription path without introducing behavior changes for existing installations.

Overall, I like the feature and think it addresses the intent of Issue #1484 better than a dashboard-only integration or routing-only change would. However, I would like to see a few fixes before merge, primarily around installation and runtime behavior rather than the Python implementation itself.

## What Looks Good

* SenseVoice is introduced as an optional service and does not replace Whisper as the default STT backend.
* The API is reasonably aligned with the existing OpenAI-compatible transcription contract.
* Dependency pinning looks appropriate.
* Manifest registration, port allocation, and network exposure policies are present.
* The unit test coverage is focused and useful.
* The README sets realistic expectations regarding CPU execution and does not overstate GPU acceleration support.
* The feature scope aligns well with the original request in Issue #1484.

---

## Requested Changes Before Merge

### 1. Compose Build Context Appears Incorrect

The current compose definition uses:

```yaml id="6v42zy"
build:
  context: .
  dockerfile: Dockerfile
```

When resolved through the normal DreamServer compose path:

```bash id="hy5dcm"
docker compose \
  -f docker-compose.base.yml \
  -f extensions/services/sensevoice/compose.yaml \
  config
```

the generated configuration resolves to:

```yaml id="74gqkz"
context: .../dream-server
dockerfile: Dockerfile
```

There does not appear to be a root-level Dockerfile, which means the standard compose workflow is likely to fail.

I believe this should follow the same explicit built-in service pattern used elsewhere:

```yaml id="m2w9pk"
build:
  context: ./extensions/services/sensevoice
  dockerfile: Dockerfile
```

I am aware that some extension implementations use `context: .`, but those cases often rely on dashboard-specific logic that rewrites build paths before invoking Docker Compose. I do not see equivalent handling for this built-in service path, so I would avoid relying on that behavior here.

### 2. Upload Endpoint Needs Size Enforcement

The transcription endpoint currently performs:

```python id="n8y1rf"
await file.read()
```

which loads the entire upload into memory before processing.

Because this endpoint is network-accessible and marked as unauthenticated, I think the service should enforce its own upload limits rather than assuming all callers originate from trusted dashboard flows.

Suggested direction:

* Add `SENSEVOICE_MAX_AUDIO_BYTES`
* Default it to a conservative value (for example, the existing 25 MB dashboard audio limit)
* Read uploads in bounded chunks
* Return HTTP 413 when limits are exceeded
* Add a regression test covering oversized uploads

Without this, a sufficiently large upload can consume memory before FunASR processing even begins.

### 3. Container Currently Runs as Root

The Dockerfile stores model data under:

```text id="z9b6wx"
/root/.cache/modelscope
```

and does not define a non-root user.

While `no-new-privileges` helps, this service accepts uploaded media and loads model artifacts. Running as a non-root user would be more consistent with the general hardening direction used elsewhere in the project.

If root is genuinely required, please document why.

Otherwise I would recommend:

* Creating a dedicated service user
* Owning `/app`
* Owning the model cache directory
* Updating volume mappings accordingly

### 4. Multi-GPU Support Appears Incomplete

The new overlays introduce:

```text id="j3kq8n"
SENSEVOICE_GPU_UUID
SENSEVOICE_GPU_INDEX
```

However, the existing GPU assignment flow only appears to populate environment variables for services such as:

* Whisper
* ComfyUI
* Embeddings

I do not see SenseVoice integrated into:

```text id="v7s5nm"
scripts/assign_gpus.py
```

or the corresponding environment generation path.

That creates a mismatch where:

* Multi-GPU compose files can require SenseVoice GPU variables
* Installer/runtime flows may never actually populate them

I think this needs one of the following:

* Integrate SenseVoice into the GPU assignment pipeline
* Disable/remove the multi-GPU overlay until assignment support exists
* Clearly document that these variables must currently be set manually

### 5. Validation Notes Need Correction

The PR body currently claims:

```bash id="q1r8ts"
bash scripts/validate-manifest-schema.sh --strict
```

validated the new manifest successfully.

From what I can tell, the validation script does not operate on a single manifest in the way implied by the PR description, and repository-wide strict validation currently fails because of unrelated manifests.

That may not be caused by this PR, but validation statements should remain reproducible and accurate for reviewers.

---

## Nice Follow-Ups (Not Necessarily Blocking)

These would be useful improvements but would not block my review:

### OpenAI Compatibility

Consider accepting and ignoring commonly supplied OpenAI transcription parameters such as:

* `model`
* `response_format`
* `temperature`

This would improve drop-in compatibility with existing clients.

### Model Readiness Documentation

The README should explicitly mention that:

* Container startup can succeed
* `/health` can report healthy
* The first transcription may still be slow due to model download/loading

This will likely reduce confusion during first-time setup.

### Future Readiness Endpoint

If dashboard integrations expand later, it may be useful to distinguish:

* Container health
* Model readiness

through separate endpoints.

### Dashboard Backend Selection

I agree this can remain out of scope for the current PR.

However, the README should clearly state that this PR introduces the service endpoint itself, not full dashboard-level STT backend selection.

---

## Validation Performed

I reviewed the available validation results and confirmed:

* SenseVoice unit tests pass
* `make lint` passes
* Dependency pin validation passes
* Network exposure validation passes
* Environment validation passes
* Manifest tests pass
* `git diff --check` is clean

GitHub currently shows no status checks attached to the head commit.

I was not able to execute the full Docker build locally because Docker Desktop is unavailable in my current environment, but the generated compose configuration is already sufficient to demonstrate the build-context concern described above.

---

## Conclusion

I like this feature and think it adds meaningful value to DreamServer. The service remains optional, preserves existing behavior, and provides users with another strong self-hosted transcription option.

That said, I would not merge it until the compose build path issue and upload hardening concerns are addressed. The container security posture and incomplete GPU assignment integration should also be resolved or explicitly documented.

Once those items are fixed, I think this becomes a much stronger and cleaner addition to the platform.

bettercalln1ck:
Addressed the review:

1. **Build context** — now `context: ./extensions/services/sensevoice`; resolves correctly via the merged path and `docker compose build sensevoice` succeeds.
2. **Upload size** — added `SENSEVOICE_MAX_AUDIO_BYTES` (default 25 MB); streams in chunks, returns `413` over the limit (+ test).
3. **Non-root** — runs as `uid=1000`, `HOME=/app`, cache at `/app/cache`. (Needs a writable HOME or `trust_remote_code` fails to register the model — verified working.)
4. **Multi-GPU** — removed the two unwired overlays + orphaned `SENSEVOICE_GPU_*` vars (assignment pipeline never set them); kept single-GPU `compose.nvidia.yaml`.
5. **Validation wording** — corrected: `validate-manifest-schema.sh` checks all manifests and reports `sensevoice: Valid`; its non-zero exit is from pre-existing `opencode`/`tailscale`, not this PR. `test-validate-manifests.sh` passes.

Also: OpenAI `model`, `response_format`, and `temperature` form fields are accepted and ignored.

Lightheartdevs:
Thanks for pushing the SenseVoice integration forward and for following up on the earlier review blockers. The branch is in much better shape now: the compose build path, bounded upload read, non-root execution, and NVIDIA device overlay all look meaningfully improved.

Quickest path to merge:
- Wire `SENSEVOICE_MAX_AUDIO_BYTES` through the deployed service contract. `dream-server/extensions/services/sensevoice/server.py` and the README advertise it, but `dream-server/extensions/services/sensevoice/compose.yaml` does not pass it into the container and `dream-server/.env.schema.json` does not declare it. Add it to compose, schema, and `.env.example` so operators can actually configure the upload limit.
- Make the model cache bind mount writable for the non-root container on fresh installs. The service runs as UID 1000 and mounts `./data/sensevoice:/app/cache`, but the Linux and macOS install paths do not appear to create/chown `data/sensevoice`, and the manifest does not declare `container_uid: 1000`. On a fresh host, Docker can create that bind source root-owned, which would make first transcription/model download fail. The smallest fix is to add the data directory/chown path or use the existing manifest/container UID preparation pattern.

What I checked:
- `docker compose -f dream-server/docker-compose.base.yml -f dream-server/extensions/services/sensevoice/compose.yaml -f dream-server/extensions/services/sensevoice/compose.nvidia.yaml config` resolves after required base env placeholders are supplied.
- The NVIDIA overlay sets `SENSEVOICE_DEVICE=cuda` as expected.
- Static review of `server.py` confirms the upload limit is enforced before model invocation; the remaining issue is that the env var is not configurable through compose.
- I did not see attached GitHub checks for this head in the PR rollup, so after the env/cache fixes I would rerun focused service tests, compose validation, extension audit, and a build/start smoke.

This is a useful addition. The remaining work is mostly about making sure it behaves well for a fresh operator, not just in a local dev checkout.

bettercalln1ck:
Both items addressed:

**Upload limit now configurable** — `compose.yaml` passes `SENSEVOICE_MAX_AUDIO_BYTES=${SENSEVOICE_MAX_AUDIO_BYTES:-26214400}` into the container; declared in `.env.schema.json` (default 26214400) and `.env.example`.Verified the value reaches the container via `docker compose config` and`docker exec dream-sensevoice env`.

**Non-root cache dir on fresh installs** — pre-create + own the bind source: `installers/phases/06-directories.sh` does `mkdir -p data/sensevoice` +`chown -R 1000:1000` (same pattern as token-spy/openclaw), and `installers/macos/install-macos.sh` adds `mkdir -p data/sensevoice` to its directory block.

Lightheartdevs:
Thanks for tightening up the SenseVoice service after the earlier review; the bounded upload path, non-root cache, and compose/env wiring are much cleaner now.

Quickest path to merge:
- Wire `sensevoice` into the optional-service compose gating in `dream-server/installers/phases/03-features.sh`. `scripts/resolve-compose-stack.sh` includes every built-in `extensions/services/*/compose.yaml` it finds, and the phase 03 comment explicitly says every optional service must be listed in `_sync_extension_compose` or users cannot opt out. Because `dream-server/extensions/services/sensevoice/compose.yaml` is new but not listed there, a fresh install that did not opt into SenseVoice can still include and build/start `dream-sensevoice`. The smallest fix is a new false-by-default `ENABLE_SENSEVOICE` path plus `_sync_extension_compose "${ENABLE_SENSEVOICE:-false}" sensevoice "SenseVoice (STT)" "SenseVoice not enabled"`, or an equivalent disabled-by-default pattern documented for `dream enable sensevoice`. I would not tie it to `ENABLE_VOICE` unless the intended behavior is to start a second STT service on every voice install.
- Mirror that opt-in gating in `dream-server/installers/macos/install-macos.sh`. The macOS service selection switch currently handles `whisper|tts` under `ENABLE_VOICE`, but has no `sensevoice` case, so macOS installs have the same risk of pulling this optional service into the generated compose stack unintentionally.
- Consider adding `container_uid: 1000` to `dream-server/extensions/services/sensevoice/manifest.yaml` while touching the manifest. Phase 06 now prepares `data/sensevoice` for fresh Linux installs, but the host-agent `_precreate_data_dirs` path uses `container_uid` for Dockerfiles that set `USER` without `compose.user`; this would keep post-install service starts aligned with the non-root container behavior.

What I checked:
- Reviewed the current diff, PR discussion, and the previous advisory follow-up.
- `git diff --check` passed.
- `python -m py_compile dream-server/extensions/services/sensevoice/server.py` passed.
- `python -m pytest dream-server/extensions/services/sensevoice/tests/test_server.py -q` passed: 7 tests.
- `python dream-server/scripts/check-dependency-pins.py` passed.
- `bash -n` via Git Bash passed for the changed installer scripts and `resolve-compose-stack.sh`.
- `docker compose` config with `docker-compose.base.yml` + SenseVoice base/NVIDIA overlays resolves the correct build context, `/app/cache` bind mount, `SENSEVOICE_MAX_AUDIO_BYTES`, and `SENSEVOICE_DEVICE=cuda` after supplying required base-stack secret placeholders.
- GitHub currently reports no checks on this head, so I could not use CI as supporting evidence.

This is close; the remaining blocker is making the service actually opt-in across the installer paths. Thanks again for pushing it forward.

bettercalln1ck:
All three items addressed:

- **Linux opt-in** — `installers/phases/03-features.sh` now gates sensevoice via its own flag: `_sync_extension_compose "${ENABLE_SENSEVOICE:-false}" sensevoice "SenseVoice (STT)" "SenseVoice not enabled"`. Disabled by default (not tied to ENABLE_VOICE, so it isn't started as a second STT on voice installs); enabledvia `dream enable sensevoice`. 
- **macOS opt-in** — `installers/macos/install-macos.sh` service switch adds`sensevoice)
 [[ "${ENABLE_SENSEVOICE:-false}" == "true" ]] || SKIP=true`, mirroring the disabled-by-default services.
- **container_uid** — added `container_uid: 1000` to the manifest so the host-agent `_precreate_data_dirs` path owns data/sensevoice correctly when the  Dockerfile sets USER without compose.user.

Lightheartdevs:
Thanks for tightening the SenseVoice opt-in path and non-root cache handling here; the follow-up commit addresses the previous installer gating issue cleanly.

Quickest path to merge:
- Close the installer-time build gap for `sensevoice`. `dream-server/installers/phases/03-features.sh` can now keep `extensions/services/sensevoice/compose.yaml` enabled when `ENABLE_SENSEVOICE=true`, and `dream-server/installers/macos/install-macos.sh` has the same switch, but both launch paths later run `docker compose ... up -d --remove-orphans --no-build` while the local image build allowlists at `dream-server/installers/phases/11-services.sh` and `dream-server/installers/macos/install-macos.sh` still only include dashboard/dashboard-api/ape/token-spy/privacy-shield (+ ComfyUI/AMD llama on Linux). Because `dream-server/extensions/services/sensevoice/compose.yaml` defines `build:` + `image: dream-sensevoice:local`, an installer-time enabled SenseVoice can reach compose-up without the image being built. Smallest fix: add `sensevoice` to both local-build candidate lists when `ENABLE_SENSEVOICE=true`, or explicitly remove the installer-time flag path and document that SenseVoice is post-install `dream enable sensevoice` only.
- Add a focused regression test around that build allowlist if possible. The existing contract tests already protect optional-service compose gating; this case needs to assert that any installer-enabled built-in service with a local `build:` entry is present in the pre-`--no-build` local build list.

What I checked:
- Reviewed the full head diff, prior PR discussion, and new commit since the last marker (`9c0f3d8..e89fb4c8`).
- `python -m py_compile dream-server/extensions/services/sensevoice/server.py` passed.
- `python -m pytest dream-server/extensions/services/sensevoice/tests/test_server.py -q` passed: 7 tests.
- `python dream-server/scripts/check-dependency-pins.py` passed.
- `python dream-server/scripts/audit-extensions.py --project-dir dream-server` passed, including `sensevoice`.
- `python -m json.tool dream-server/.env.schema.json` passed.
- Git Bash `bash -n` passed for the changed installer scripts and `resolve-compose-stack.sh`.
- `docker compose ... config --format json` with base + SenseVoice + NVIDIA overlay resolved the expected build context, `/app/cache` bind, `SENSEVOICE_MAX_AUDIO_BYTES`, `SENSEVOICE_DEVICE=cuda`, and port 9100.
- `gh pr checks` reports no checks on this branch. I also tried `tests/test-validate-manifests.sh` and `tests/contracts/test-installer-contracts.sh`, but this runner lacks `jq`, so those stopped at their prerequisites.

This is close, and the service-level shape looks good; the remaining issue is making the installer path match the new opt-in surface. Thanks again for pushing this forward.

bettercalln1ck:
- **Linux** (installers/phases/11-services.sh) and **macOS** (installers/macos/install-macos.sh) now append `sensevoice` to the local-build candidate list when their ENABLE_ flag is set, so the image is built before the `--no-build` compose-up. The existing  `config --services` intersection still gates whether it actually builds.
- **Regression test**: tests/contracts/test-installer-build-allowlist.sh asserts every installer-enableable built-in service (3rd arg of _sync_extension_compose in 03-features.sh) with a local build: is present in both candidate build lists.
  Wired into `make test`. Verified it fails without the fix (flags sensevoice) and passes with it. bash-3.2 safe, no jq dependency.

Lightheartdevs:
Thanks for carrying the SenseVoice service through the installer/runtime gaps from the earlier reviews; the current head is much more aligned with Dream Server's optional-service model.

What looks good:
- The service stays additive and disabled by default: `sensevoice` is gated separately from `ENABLE_VOICE` in Linux and macOS paths, so enabling voice does not accidentally start a second STT backend.
- The installer-time image build path now matches the `--no-build` launch behavior in `installers/phases/11-services.sh` and `installers/macos/install-macos.sh`, and `tests/contracts/test-installer-build-allowlist.sh` protects that contract for future local-build services.
- The service contract is now better hardened: chunked upload limit enforcement, non-root container execution, `container_uid: 1000`, env/port/schema declarations, dependency-lock entry, network-exposure policy, and a documented first-run model readiness caveat.

What I checked:
- Reviewed the full current diff, PR discussion, and the latest fixes since the prior advisory.
- `git diff --check origin/main...refs/remotes/pr/1574` passed.
- Full shell-script `bash -n` sweep over the PR worktree passed.
- `python -m py_compile extensions/services/sensevoice/server.py` passed.
- `python -m pytest extensions/services/sensevoice/tests/test_server.py -q` passed: 7 tests.
- `python scripts/check-dependency-pins.py` passed.
- `python scripts/audit-extensions.py --project-dir .` passed, including `sensevoice`.
- `python -m json.tool .env.schema.json` passed.
- `bash tests/contracts/test-installer-build-allowlist.sh` passed.
- Docker Compose config for `docker-compose.base.yml` + SenseVoice base/NVIDIA overlays resolves the expected local build context, `dream-sensevoice:local`, `SENSEVOICE_DEVICE=cuda`, `SENSEVOICE_MAX_AUDIO_BYTES=26214400`, `/app/cache`, and published port `9100` when required base env placeholders are supplied.
- `gh pr checks` currently reports no checks on this branch. The `test-validate-env.sh` and `test-validate-manifests.sh` scripts also stop in this runner because `jq` is not installed, so I used the Python audit/schema checks above as local coverage for those surfaces.

Thanks again for pushing this forward; this now looks ready for maintainer merge judgment.

Lightheartdevs:
Thanks for carrying the SenseVoice/FunASR STT backend through the installer/runtime details; the current head still fits Dream Server's optional-service model after the main merge.

What looks good:
- `sensevoice` remains additive and disabled by default, with separate `ENABLE_SENSEVOICE` gating in Linux/macOS and no default Open WebUI/Hermes STT route changes.
- The local-build service is covered before the installer `--no-build` launch in Linux/macOS, and `tests/contracts/test-installer-build-allowlist.sh` now guards that contract.
- The service contract is appropriately hardened: chunked upload cap, non-root cache path, `container_uid: 1000`, env/schema/port/dependency/network entries, and focused service tests.

What I checked:
- Reviewed the full current diff and prior advisory thread; this head is a main merge after the previous merge-ready review.
- `python -m py_compile dream-server/extensions/services/sensevoice/server.py` passed.
- `python -m pytest dream-server/extensions/services/sensevoice/tests/test_server.py -q` passed: 7 tests.
- `python dream-server/scripts/check-dependency-pins.py` passed.
- `python dream-server/scripts/audit-extensions.py --project-dir dream-server` passed, including `sensevoice`.
- `python -m json.tool dream-server/.env.schema.json` passed.
- Git Bash `tests/contracts/test-installer-build-allowlist.sh` passed.
- `docker compose` config with base + SenseVoice + NVIDIA overlay resolved the expected local build context, `dream-sensevoice:local`, `SENSEVOICE_DEVICE=cuda`, upload limit, cache bind, and port `9100` with required base placeholders supplied.
- `gh pr checks 1574` reports no checks for this branch.

Thanks again for pushing this forward; this looks ready for maintainer merge judgment.

bong-water-water-bong (approved):
Well-structured SenseVoice/FunASR STT extension. Full extension with:
- Custom Python FastAPI server with GPU support (CPU + NVIDIA compose variants)
- Dockerfile, health check, proper resource limits
- Installer integration across Linux/macOS/Windows (phases 03, 06, 11)
- All config registrations (core-service-ids, ports, dependency-lock, network-exposure-policy, .env)
- Tests for the server endpoint
- Disabled by default (category: optional)

Follows the extension pattern correctly. Addresses #1484.

LauraGPT:
Prepared a conflict-resolution branch against current `main` after the `dream-server/` -> `ods/` rename:

https://github.com/LauraGPT/ODS/tree/funasr/ods-1574-conflict-fix

Commit: `dc75aaf58087d9e2d39d127cf68e5e0c07408bd2`

What changed in the resolved branch:
- moved the new SenseVoice service into `ods/extensions/services/sensevoice/`;
- updated the service to current ODS naming (`ods.services.v1`, `ods_min`, `ods-sensevoice`, `ods enable sensevoice`);
- kept the SenseVoice upload limit env docs and the Linux/macOS build allowlist fixes;
- kept the macOS resolver path on the current `ODS_LOG_FILE` variable.

Validation run locally:
- no unmerged paths / no conflict markers;
- `git diff --check`;
- `bash -n` on the touched installer scripts and contract test;
- JSON parse for the touched config files;
- YAML parse for SenseVoice compose/manifest files;
- `tests/contracts/test-installer-build-allowlist.sh` -> `[PASS] all 4 installer-enableable locally-built services are in both build allowlists`;
- SenseVoice FastAPI unit tests in a lightweight venv -> `7 passed`;
- resolver + `docker compose --env-file .env.example ... config` includes `sensevoice` with build context `/tmp/ods-funasr/ods/extensions/services/sensevoice` and image/container `ods-sensevoice`.

The original PR branch still shows `CONFLICTING`; this branch should give a clean current-main patch to cherry-pick or compare.

LauraGPT:
Opened a clean current-main replacement PR from the conflict-resolution branch: #1639. The original branch remains `CONFLICTING`, while #1639 is `MERGEABLE` at head `dc75aaf58087d9e2d39d127cf68e5e0c07408bd2`.

PR #1567, opened 2026-06-14
fix(sidebar): highlight only integrations on nested route

BittuBarnwal7479:
## Summary

When opening /extensions/integrations, both "Extensions" and "Integrations" are highlighted in the sidebar. Only "Integrations" should be active.

Resolves: #1566 

## Release Lane

- [ ] Stable hotfix targeting `release/2.5.x`
- [x] Mainline change targeting `main`
- [ ] Next-minor work targeting the next feature/minor release
- [ ] Not sure; reviewer should help classify

Stable hotfix reason:

```text
<!-- If this targets release/2.5.x, explain the current-stable user problem it fixes. -->
```

## Changed Surface

- [ ] Docs only
- [ ] Tests only
- [x] Dashboard UI
- [ ] Dashboard API / host agent
- [ ] Installer / bootstrap / lifecycle
- [ ] Docker Compose / service manifests
- [ ] Model routing / Hermes / capabilities
- [ ] Network exposure / auth / proxy
- [ ] Dependencies / runtime wiring

## Risk And Validation

- Risk level: Low
- Validation run:
  - [x] `git diff --check`
  - [ ] Markdown/link sanity for docs
  - [ ] Focused tests listed below
  - [x] Dashboard lint/test/build
  - [ ] Extension audit / compose validation
  - [ ] Release-grade fleet or scoped hardware validation
  - [ ] Stable-lane patch validation, if targeting `release/2.5.x`
  - [ ] Not required because: 

Commands/results:

- npm run lint ✅
- npm run build ✅
- git diff --check ✅

## Operational Change Check

If this PR touches installer phases, bootstrap logic, compose stack generation,
service manifests, dashboard API control flows, Hermes, model routing, GPU or
runtime detection, lifecycle commands, host mutation, or network exposure, it
requires release-grade fleet validation before release unless the PR explains a
narrower equivalent.

- [x] This is not an operational change.
- [ ] This is an operational change and validation is recorded above.
- [ ] This is an operational change and validation is intentionally deferred for:

## Notes For Reviewers

Verified that:

- `/extensions` highlights Extensions only.
- `/extensions/integrations` highlights Integrations only.
- No changes to routing behavior.

PR #1575, opened 2026-06-15
fix(dashboard): use exact path matching in Sidebar to prevent dual ac…

### LINKED ISSUE #1566: Sidebar marks both Extensions and Integrations active when click on integrations route
### Description

When opening `/extensions/integrations`, both "Extensions" and "Integrations" are highlighted in the sidebar.

Expected: only "Integrations" should be active.

### Steps to Reproduce

1. Open `/extensions/integrations`
2. See the sidebar mark
3. Extensions and Integrations show active (both highlighted)

### Expected Behavior

/extensions → Extensions active 
/extensions/integrations → Integrations active (only)

### Actual Behavior

/extensions → Extensions active 
/extensions/integrations → Extensions active and Integrations active (both highlighted)

### Operating System

Window11

### GPU

AMD RX

### Docker Version

Docker 27.1.1

### VRAM

24 GB

### Logs

```shell

```

### Screenshots

Image

HimanshuShukla671:
Fixes #1566

## Summary

Use exact path matching in Sidebar NavLinks to prevent parent routes from being marked active when their children are selected.

## AI Assistance

None

## Release Lane

## Changed Surface

- [x] Dashboard UI

## Risk And Validation

- Risk level: Low
- Validation run:

## Operational Change Check

## Notes For Reviewers

None

### DIFF
--- a/dream-server/extensions/services/dashboard/src/components/Sidebar.jsx
+++ b/dream-server/extensions/services/dashboard/src/components/Sidebar.jsx
@@ -142,6 +142,7 @@ export default function Sidebar({ status, collapsed, onToggle }) {
             <li key={path}>
               <NavLink
                 to={path}
+                end
                 title={collapsed ? label : undefined}
                 className={({ isActive }) =>
                   `flex items-center ${collapsed ? 'justify-center' : ''} gap-3 px-3 py-2.5 rounded-lg transition-colors ${


gabsprogrammer:
Thanks for the PR.

This is a small but meaningful UI correctness fix, and it directly addresses the active-state behavior reported in #1566.

I reviewed the current dashboard routing structure, and the change looks correct for the sidebar model that exists today. The only visible parent/child route relationship currently exposed in the navigation is:

```text
/extensions
/extensions/integrations
```

Using `end` on the parent `NavLink` prevents `/extensions` from remaining highlighted when the user is actually viewing `/extensions/integrations`, which matches the expected navigation behavior described in the issue.

### Validation Performed

```bash
npm run lint
```

* Passes successfully

```bash
npm test
```

* Passes successfully
* 20 test files
* 124 tests passing

```bash
npm run build
```

* Passes successfully

```bash
git diff --check
```

* Clean, no whitespace or formatting issues detected

### Suggested Regression Coverage

My only recommendation before—or shortly after—merge would be to add a focused regression test covering this exact route-matching behavior.

Something along the lines of:

* Mock two sidebar routes:

  * `/extensions`
  * `/extensions/integrations`
* Render the sidebar while the current location is `/extensions/integrations`
* Assert that:

  * **Integrations** receives active styling
  * **Extensions** does not receive active styling

The existing `Sidebar.test.jsx` coverage verifies rendering, layout, and basic navigation behavior, but it does not currently protect against regressions in active-route matching.

A small dedicated test would ensure this exact bug does not reappear during future sidebar or routing refactors.

### Future Consideration

One thing worth keeping in mind as the dashboard grows: applying `end` globally effectively makes sidebar matching exact-match only.

That is completely reasonable for the current route structure, but future routes may want different behavior. For example, if a page such as:

```text
/models/:id
```

is introduced and the expectation is that **Models** remains highlighted while viewing a specific model, exact matching alone may no longer be sufficient.

If that situation arises later, it may make sense for route definitions to support something like:

```js
sidebarEnd: false
```

or a custom active matcher.

This is not a concern for the current PR and should not block it; it is simply something to consider as the route system evolves.

### Conclusion

Overall, this is a useful, low-risk fix that aligns with the reported issue and behaves correctly in local validation.

I would be comfortable with this change after the final maintainer review.

Lightheartdevs:
Thanks for tightening the dashboard sidebar route matching for the /extensions/integrations path.

- The one-line implementation in dream-server/extensions/services/dashboard/src/components/Sidebar.jsx is correct for the currently registered sidebar routes: NavLink end prevents /extensions from remaining active when the user is on /extensions/integrations.
- There is an active duplicate PR, #1567, with the same code change. This PR has the cleaner completed template and mainline targeting, while #1567 came from an external contributor and includes the screenshot context. A maintainer should choose which one to merge and close/credit the other accordingly.
- If this one is chosen, I would not block on extra code changes; a focused regression test in Sidebar.test.jsx would be useful follow-up coverage, but the current diff is very small and matches the reported bug.

What I checked:
- Static diff review against current main: only Sidebar.jsx changes, adding end to the mapped NavLink.
- Current route definitions in dream-server/extensions/services/dashboard/src/plugins/core.js: the visible parent/child sidebar pair is /extensions and /extensions/integrations.
- git diff --check origin/main...origin/pr/1575 passed.
- gh pr checks reports no checks for this branch.

Thanks again for pushing the UI fix forward.

HimanshuShukla671:
Closing this as a duplicate of #1567. Apologies, I totally missed that you already had a PR open for this fix! Deferring to your work.

PR #1577, opened 2026-06-15
feat(health): support non-HTTP health checks for TCP/CLI services

### LINKED ISSUE #666: Support non-HTTP health checks for TCP/CLI services

HimanshuShukla671:
Fixes #666

## Summary

This PR adds support for TCP and CLI-only (none) health checks. Previously, `health-check.sh` defaulted to HTTP `curl` checks for all services. Services like `piper-audio` (Wyoming TCP protocol) or CLI tools like `aider` would always be flagged as unhealthy.

This introduces a `health_type` field in the manifest schema that can be set to `http`, `tcp`, or `none`.

### Changes

**`lib/service-registry.sh`**
- Declares `SERVICE_HEALTH_TYPES` associative array
- Parses `health_type` from manifest YAML (defaults to `http` for backward compatibility)

**`scripts/health-check.sh`**
- Container-state validation runs **before** `health_type=none` early return — a stopped/missing container is never reported healthy
- `none`: reports ok only after confirming container is running (or unmanaged)
- `tcp`: validates port/timeout as numeric, then uses Python socket test with values passed via environment variables (no code interpolation)
- `http`: existing curl logic unchanged

**`scripts/generate-extensions-catalog.py`**
- Propagates `health_type` into `extensions-catalog.json` so downstream consumers can distinguish check types

**Manifests**
- `piper-audio/manifest.yaml` → `health_type: tcp`
- `aider/manifest.yaml` → `health_type: none`

### Scope & Follow-up

This PR covers the registry, health-check engine, and catalog layers. The broader dashboard-api health helpers (`check_service_health()`), schema enum validation, `audit-extensions.py`, and full regression test suite are acknowledged as needed follow-up work and will be addressed in a separate PR to keep this one focused.

## Release Lane

## Changed Surface

- [x] CLI / Operational
- [x] Core Services (Registry)

## Risk And Validation

- Risk level: Low
- Backward compatible: services without `health_type` default to `http`

### DIFF
--- a/dream-server/extensions/library/services/aider/manifest.yaml
+++ b/dream-server/extensions/library/services/aider/manifest.yaml
@@ -11,6 +11,7 @@ service:
   external_port_env: ''
   external_port_default: 0
   health: ""
+  health_type: none
   type: docker
   startup_check: false  # one-shot CLI tool; container exits 0 by design
   gpu_backends: [amd, nvidia, apple]
--- a/dream-server/extensions/library/services/piper-audio/manifest.yaml
+++ b/dream-server/extensions/library/services/piper-audio/manifest.yaml
@@ -11,6 +11,7 @@ service:
   external_port_env: PIPER_PORT
   external_port_default: 10200
   health: ""
+  health_type: tcp
   type: docker
   gpu_backends: [amd, nvidia, apple]
   compose_file: compose.yaml
--- a/dream-server/lib/service-registry.sh
+++ b/dream-server/lib/service-registry.sh
@@ -30,6 +30,7 @@ declare -A SERVICE_COMPOSE      # service_id → compose file path
 declare -A SERVICE_CATEGORIES   # service_id → core|recommended|optional
 declare -A SERVICE_DEPENDS      # service_id → space-separated dependency IDs
 declare -A SERVICE_HEALTH       # service_id → health endpoint path
+declare -A SERVICE_HEALTH_TYPES # service_id → health check type (http|tcp|none)
 declare -A SERVICE_HEALTH_TIMEOUTS  # service_id → health check timeout in seconds
 declare -A SERVICE_PORTS        # service_id → external port (what the user hits on localhost)
 declare -A SERVICE_PORT_ENVS    # service_id → env var name for the external port
@@ -176,11 +177,13 @@ for service_dir in _all_service_dirs:
         print(f'SERVICE_CATEGORIES["{_esc(sid)}"]="{_esc(category)}"')
         print(f'SERVICE_DEPENDS["{_esc(sid)}"]="{_esc(" ".join(str(d) for d in depends))}"')
         health = s.get("health", "/health")
+        health_type = s.get("health_type", "http")
         health_timeout = s.get("health_timeout", 5)  # Default 5 seconds
         port = s.get("external_port_default", s.get("port", 0))
         port_env = s.get("external_port_env", "")
         host_network = "1" if s.get("host_network") else ""
         print(f'SERVICE_HEALTH["{_esc(sid)}"]="{_esc(health)}"')
+        print(f'SERVICE_HEALTH_TYPES["{_esc(sid)}"]="{_esc(health_type)}"')
         print(f'SERVICE_HEALTH_TIMEOUTS["{_esc(sid)}"]="{_esc(health_timeout)}"')
         print(f'SERVICE_PORTS["{_esc(sid)}"]="{_esc(port)}"')
         print(f'SERVICE_PORT_ENVS["{_esc(sid)}"]="{_esc(port_env)}"')
--- a/dream-server/scripts/generate-extensions-catalog.py
+++ b/dream-server/scripts/generate-extensions-catalog.py
@@ -102,6 +102,7 @@ def extract_entry(manifest: dict) -> dict | None:
         "port": service.get("port", 0),
         "external_port_default": service.get("external_port_default", 0),
         "health_endpoint": service.get("health", ""),
+        "health_type": service.get("health_type", "http"),
         "env_vars": strip_secrets(env_vars),
         "tags": manifest.get("tags") or service.get("tags", []),
         "features": manifest.get("features") or service.get("features", []),
--- a/dream-server/scripts/health-check.sh
+++ b/dream-server/scripts/health-check.sh
@@ -156,13 +156,24 @@ test_service() {
     local port_env="${SERVICE_PORT_ENVS[$sid]}"
     local default_port="${SERVICE_PORTS[$sid]}"
     local health="${SERVICE_HEALTH[$sid]}"
+    local health_type="${SERVICE_HEALTH_TYPES[$sid]:-http}"
     local timeout="${SERVICE_HEALTH_TIMEOUTS[$sid]:-$TIMEOUT}"
 
     # Resolve port
     local port="$default_port"
     [[ -n "$port_env" ]] && port="${!port_env:-$default_port}"
 
-    [[ -z "$health" || "$port" == "0" ]] && return 1
+    if [[ "$health_type" == "none" ]]; then
+        result_set "$sid" "ok"
+        return 0
+    fi
+
+    if [[ "$health_type" == "http" ]] && [[ -z "$health" || "$port" == "0" ]]; then
+        return 1
+    fi
+    if [[ "$health_type" == "tcp" ]] && [[ "$port" == "0" ]]; then
+        return 1
+    fi
 
     # Check container state first (if docker available)
     local container_state
@@ -173,10 +184,20 @@ test_service() {
         return 1
     fi
 
-    if curl -sf --max-time "$timeout" "http://127.0.0.1:${port}${health}" >/dev/null 2>&1; then
-        result_set "$sid" "ok"
-        return 0
+    if [[ "$health_type" == "tcp" ]]; then
+        # TCP connection test using Python socket
+        if python3 -c "import socket; s = socket.create_connection(('127.0.0.1', $port), timeout=$timeout); s.close()" 2>/dev/null; then
+            result_set "$sid" "ok"
+            return 0
+        fi
+    else
+        # HTTP GET test
+        if curl -sf --max-time "$timeout" "http://127.0.0.1:${port}${health}" >/dev/null 2>&1; then
+            result_set "$sid" "ok"
+            return 0
+        fi
     fi
+
     result_set "$sid" "fail"
     ANY_FAIL=true
     return 1
--- a/dream-server/scripts/health-check.sh
+++ b/dream-server/scripts/health-check.sh
@@ -163,11 +163,6 @@ test_service() {
     local port="$default_port"
     [[ -n "$port_env" ]] && port="${!port_env:-$default_port}"
 
-    if [[ "$health_type" == "none" ]]; then
-        result_set "$sid" "ok"
-        return 0
-    fi
-
     if [[ "$health_type" == "http" ]] && [[ -z "$health" || "$port" == "0" ]]; then
         return 1
     fi
@@ -175,7 +170,8 @@ test_service() {
         return 1
     fi
 
-    # Check container state first (if docker available)
+    # Check container state first (if docker available) — even for "none"
+    # services, a stopped/missing container should not report healthy.
     local container_state
     container_state=$(check_container_state "$sid")
     if [[ -n "$container_state" && "$container_state" != "running" ]]; then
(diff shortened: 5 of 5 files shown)


gabsprogrammer:
Thanks for the PR.

I agree with the overall direction here. The problem being addressed is real: not every service fits the traditional HTTP health-check model. Piper/Wyoming-style services expose TCP endpoints rather than HTTP APIs, and one-shot CLI tools such as Aider should not be treated as long-running web services in the first place.

That said, I do not think this is ready to merge yet. The current implementation introduces the concept of `health_type`, but only wires part of the stack to understand it. As a result, the feature is not yet consistently represented across the catalog, runtime, dashboard, validation, and health-checking layers.

## Main Concerns

### 1. `health_type` Is Not Propagated Through the Extension Catalog

The manifests introduce the new field, but catalog generation does not appear to preserve it.

For example, the generated `config/extensions-catalog.json` still contains entries such as:

```json id="8q4mrv"
{
  "id": "aider",
  "health_endpoint": ""
}
```

and

```json id="3g0hkn"
{
  "id": "piper-audio",
  "health_endpoint": ""
}
```

with no corresponding `health_type`.

I regenerated the catalog locally and observed the same result. It appears that:

```text id="g7v2te"
scripts/generate-extensions-catalog.py
```

does not currently copy the new field.

As a result, downstream consumers have no way to distinguish:

* HTTP health checks
* TCP health checks
* No-health/CLI services

which defeats much of the purpose of introducing the field.

### 2. Dashboard Health Logic Still Assumes HTTP

The dashboard side does not appear to understand the new health model.

Currently:

* `dashboard-api` still relies on HTTP-oriented health helpers
* `helpers.check_service_health()` remains HTTP-centric
* `user_extensions.py` exposes only fields such as:

  * `health`
  * `port`
  * `external_port`

without exposing `health_type`

This means Piper and similar services can still be misclassified because the dashboard only understands:

* HTTP endpoint exists
* HTTP endpoint missing

One of the goals discussed in Issue #666 was ensuring that dashboard, doctor, runtime checks, audits, and health reporting all agree on service state. That consistency is not there yet.

### 3. `health_type: none` Reports Healthy Too Early

In:

```text id="u8y4hz"
scripts/health-check.sh
```

the `none` branch returns success before container-state validation runs.

That means a service configured as:

```yaml id="2n5cpk"
health_type: none
```

can potentially be reported healthy even if:

* The container is stopped
* The container failed to start
* The extension was never installed

If the intended state is "not applicable," I think that should be represented explicitly rather than implicitly returning healthy.

At minimum, container existence and runtime state should still be evaluated before reporting success.

### 4. TCP Health Check Uses Unvalidated Values in Python Source

The TCP probe currently constructs Python source using shell-expanded variables:

```bash id="t4x7me"
python3 -c "import socket; s = socket.create_connection(('127.0.0.1', $port), timeout=$timeout); s.close()"
```

While `safe-env.sh` avoids shell evaluation issues, this still injects values directly into dynamically generated Python code.

Because values can ultimately originate from environment configuration, I would prefer one of the following:

* Validate `port` and `timeout` as numeric before use
* Pass values through argv/environment variables
* Avoid code interpolation entirely

This is more of a robustness issue than an immediate exploit path, but it should be tightened before introducing a new health-check mechanism.

### 5. Schema, Audit, and Test Coverage Are Incomplete

The new field is not fully reflected across validation layers.

From what I can see:

* Manifest schemas do not validate `service.health_type`
* No enum enforcement exists for:

  * `http`
  * `tcp`
  * `none`
* `audit-extensions.py` does not validate the field
* Empty health endpoints are not conditionally validated based on health type

The current implementation also lacks focused coverage for several important cases.

Suggested regression coverage:

* Default HTTP behavior remains unchanged
* TCP health success
* TCP health failure
* No-health/CLI behavior
* Invalid `health_type`
* Invalid or non-numeric TCP port values
* Dashboard status behavior for TCP services
* Dashboard status behavior for no-health services

Without those tests, it is difficult to be confident that the new health model will remain consistent across future changes.

---

## Validation Performed

I validated the current implementation locally:

```bash id="k1f6ys"
bash tests/test-health-check.sh
```

* Passes

```bash id="n9e2wd"
bash tests/test-service-registry.sh
```

* Passes

```bash id="r5g8tu"
bash scripts/validate-manifests.sh
```

* Passes

```bash id="j7v4am"
python scripts/audit-extensions.py --json
```

* Passes for built-in services

```bash id="h3y6cf"
python scripts/audit-extensions.py aider piper-audio --json
```

* Does not find the modified services because audit coverage currently targets `extensions/services` rather than the library entries modified by this PR

```bash id="c2p8mx"
git diff --check
```

* Clean

GitHub currently shows no status checks attached to the head commit.

---

## Suggested Path Forward

I think this should be completed as a full schema-to-runtime feature rather than a manifest-only addition.

Recommended follow-up work:

1. Add `health_type` to both manifest schemas with enum validation.
2. Propagate `health_type` through `generate-extensions-catalog.py`.
3. Update `config/extensions-catalog.json` generation.
4. Include `health_type` in dashboard-api service metadata and user-extension metadata.
5. Teach `helpers.check_service_health()` to support:

   * `http`
   * `tcp`
   * `none`
6. Update:

   * `dream-doctor.sh`
   * `extension-runtime-check.sh`
   * preflight validation
   * runtime validation paths
7. Validate numeric TCP parameters and avoid interpolating environment values into Python source.
8. Add focused regression tests covering HTTP, TCP, and CLI/no-health service types.

---

## Conclusion

The concept is sound and addresses a legitimate limitation in the current health-check model. However, the implementation currently introduces a new manifest field without ensuring that the rest of the platform consistently understands and respects it.

Until catalog generation, dashboard status reporting, runtime validation, schemas, and audits are updated to consume `health_type`, I think this remains only a partial implementation.

For that reason, I would not merge it yet. I would prefer to see the full plumbing completed so the new health model is represented consistently throughout DreamServer.

HimanshuShukla671:
> Thanks for the PR.
> 
> I agree with the overall direction here. The problem being addressed is real: not every service fits the traditional HTTP health-check model. Piper/Wyoming-style services expose TCP endpoints rather than HTTP APIs, and one-shot CLI tools such as Aider should not be treated as long-running web services in the first place.
> 
> That said, I do not think this is ready to merge yet. The current implementation introduces the concept of `health_type`, but only wires part of the stack to understand it. As a result, the feature is not yet consistently represented across the catalog, runtime, dashboard, validation, and health-checking layers.
> 
> ## Main Concerns
> ### 1. `health_type` Is Not Propagated Through the Extension Catalog
> The manifests introduce the new field, but catalog generation does not appear to preserve it.
> 
> For example, the generated `config/extensions-catalog.json` still contains entries such as:
> 
> ```json
> {
>   "id": "aider",
>   "health_endpoint": ""
> }
> ```
> 
> and
> 
> ```json
> {
>   "id": "piper-audio",
>   "health_endpoint": ""
> }
> ```
> 
> with no corresponding `health_type`.
> 
> I regenerated the catalog locally and observed the same result. It appears that:
> 
> ```
> scripts/generate-extensions-catalog.py
> ```
> 
> does not currently copy the new field.
> 
> As a result, downstream consumers have no way to distinguish:
> 
> * HTTP health checks
> * TCP health checks
> * No-health/CLI services
> 
> which defeats much of the purpose of introducing the field.
> 
> ### 2. Dashboard Health Logic Still Assumes HTTP
> The dashboard side does not appear to understand the new health model.
> 
> Currently:
> 
> * `dashboard-api` still relies on HTTP-oriented health helpers
> * `helpers.check_service_health()` remains HTTP-centric
> * `user_extensions.py` exposes only fields such as:
>   
>   * `health`
>   * `port`
>   * `external_port`
> 
> without exposing `health_type`
> 
> This means Piper and similar services can still be misclassified because the dashboard only understands:
> 
> * HTTP endpoint exists
> * HTTP endpoint missing
> 
> One of the goals discussed in Issue #666 was ensuring that dashboard, doctor, runtime checks, audits, and health reporting all agree on service state. That consistency is not there yet.
> 
> ### 3. `health_type: none` Reports Healthy Too Early
> In:
> 
> ```
> scripts/health-check.sh
> ```
> 
> the `none` branch returns success before container-state validation runs.
> 
> That means a service configured as:
> 
> ```yaml
> health_type: none
> ```
> 
> can potentially be reported healthy even if:
> 
> * The container is stopped
> * The container failed to start
> * The extension was never installed
> 
> If the intended state is "not applicable," I think that should be represented explicitly rather than implicitly returning healthy.
> 
> At minimum, container existence and runtime state should still be evaluated before reporting success.
> 
> ### 4. TCP Health Check Uses Unvalidated Values in Python Source
> The TCP probe currently constructs Python source using shell-expanded variables:
> 
> ```shell
> python3 -c "import socket; s = socket.create_connection(('127.0.0.1', $port), timeout=$timeout); s.close()"
> ```
> 
> While `safe-env.sh` avoids shell evaluation issues, this still injects values directly into dynamically generated Python code.
> 
> Because values can ultimately originate from environment configuration, I would prefer one of the following:
> 
> * Validate `port` and `timeout` as numeric before use
> * Pass values through argv/environment variables
> * Avoid code interpolation entirely
> 
> This is more of a robustness issue than an immediate exploit path, but it should be tightened before introducing a new health-check mechanism.
> 
> ### 5. Schema, Audit, and Test Coverage Are Incomplete
> The new field is not fully reflected across validation layers.
> 
> From what I can see:
> 
> * Manifest schemas do not validate `service.health_type`
> * No enum enforcement exists for:
>   
>   * `http`
>   * `tcp`
>   * `none`
> * `audit-extensions.py` does not validate the field
> * Empty health endpoints are not conditionally validated based on health type
> 
> The current implementation also lacks focused coverage for several important cases.
> 
> Suggested regression coverage:
> 
> * Default HTTP behavior remains unchanged
> * TCP health success
> * TCP health failure
> * No-health/CLI behavior
> * Invalid `health_type`
> * Invalid or non-numeric TCP port values
> * Dashboard status behavior for TCP services
> * Dashboard status behavior for no-health services
> 
> Without those tests, it is difficult to be confident that the new health model will remain consistent across future changes.
> 
> ## Validation Performed
> I validated the current implementation locally:
> 
> ```shell
> bash tests/test-health-check.sh
> ```
> 
> * Passes
> 
> ```shell
> bash tests/test-service-registry.sh
> ```
> 
> * Passes
> 
> ```shell
> bash scripts/validate-manifests.sh
> ```
> 
> * Passes
> 
> ```shell
> python scripts/audit-extensions.py --json
> ```
> 
> * Passes for built-in services
> 
> ```shell
> python scripts/audit-extensions.py aider piper-audio --json
> ```
> 
> * Does not find the modified services because audit coverage currently targets `extensions/services` rather than the library entries modified by this PR
> 
> ```shell
> git diff --check
> ```
> 
> * Clean
> 
> GitHub currently shows no status checks attached to the head commit.
> 
> ## Suggested Path Forward
> I think this should be completed as a full schema-to-runtime feature rather than a manifest-only addition.
> 
> Recommended follow-up work:
> 
> 1. Add `health_type` to both manifest schemas with enum validation.
> 2. Propagate `health_type` through `generate-extensions-catalog.py`.
> 3. Update `config/extensions-catalog.json` generation.
> 4. Include `health_type` in dashboard-api service metadata and user-extension metadata.
> 5. Teach `helpers.check_service_health()` to support:
>    
>    * `http`
>    * `tcp`
>    * `none`
> 6. Update:
>    
>    * `dream-doctor.sh`
>    * `extension-runtime-check.sh`
>    * preflight validation
>    * runtime validation paths
> 7. Validate numeric TCP parameters and avoid interpolating environment values into Python source.
> 8. Add focused regression tests covering HTTP, TCP, and CLI/no-health service types.
> 
> ## Conclusion
> The concept is sound and addresses a legitimate limitation in the current health-check model. However, the implementation currently introduces a new manifest field without ensuring that the rest of the platform consistently understands and respects it.
> 
> Until catalog generation, dashboard status reporting, runtime validation, schemas, and audits are updated to consume `health_type`, I think this remains only a partial implementation.
> 
> For that reason, I would not merge it yet. I would prefer to see the full plumbing completed so the new health model is represented consistently throughout DreamServer.
> 
> **Just waiting for the final check from @Lightheartdevs.**

Thanks for the detailed review — all great catches. Pushed fixes for the three issues that can be addressed in this PR:

1. Container-state before none return (Point 3) — Fixed. The none branch now runs after container-state validation. A stopped/missing container will correctly report as failed instead of falsely healthy.

2. TCP probe injection hardening (Point 4) — Fixed two ways:

Port and timeout are now validated as numeric (^[0-9]+$ regex) before use
Values are passed via environment variables (DS_PORT, DS_TIMEOUT) and read with os.environ inside a single-quoted Python string — no more shell interpolation into Python source

3. Catalog propagation (Point 1) — Fixed. generate-extensions-catalog.py now emits health_type alongside health_endpoint, so downstream consumers can distinguish HTTP/TCP/none services.

For the broader stack changes (dashboard-api health helpers, schema enum validation, audit-extensions.py, full regression test suite) — I agree those are needed for full consistency, but they're a significantly larger scope. Happy to land those in a follow-up PR. Updated the PR description to explicitly document the current scope and acknowledged follow-up.

Lightheartdevs:
Thanks for taking on the non-HTTP health-check path. This is a valuable direction for the service registry, especially for TCP and CLI-style services that do not fit the old HTTP-only assumption.

Quickest path to merge:
- Regenerate and commit `dream-server/config/extensions-catalog.json`. On this branch, `dream-server/scripts/generate-extensions-catalog.py` emits `health_type` values such as `aider=none` and `piper-audio=tcp`, but the committed catalog is still stale and contains no `health_type` fields. That catalog freshness matters because downstream tooling reads the committed generated contract.
- Add `service.health_type` to the manifest schema contract with an enum for `http`, `tcp`, and `none`. If `#1578` has not landed first, keep both schema copies synchronized. Right now the new manifest field is used by code but not declared by the schema.
- Add focused regression coverage for default `http`, `tcp` success/failure, `none` after container-state validation, invalid `health_type`, and non-numeric TCP port/timeout handling.

What I checked:
- `dream-server/scripts/health-check.sh` has the core branching for the new health modes.
- `dream-server/scripts/generate-extensions-catalog.py` can generate the new catalog fields when run locally.
- I did not see attached GitHub checks for this head in the PR rollup I queried, so the missing generated catalog and schema/test coverage are still real merge blockers.

This is close in shape. The main thing is making the new behavior part of the repo's declared/generated contract so the next contributor does not have to rediscover it from implementation details.

HimanshuShukla671:
Closing this as a duplicate of #1532. My mistake, I missed that there was already an active PR for this! Deferring to the original contributor.

PR #1578, opened 2026-06-15
fix: align manifest schema validator with service manifests

DeluxeRaph:
## Summary

The manifest schema validator had drifted from the current service manifest schema and only scanned bundled services. On current `main`, it failed against valid manifests because:

- `opencode` uses the schema-supported `type: host-systemd`
- `tailscale` uses `host_network: true` and intentionally omits `service.health`
- `gpu_backends: none` and `gpu_backends: cpu` are schema-supported values
- library extension manifests were not scanned

This PR updates manifest validation so `service-manifest.v1.json` is the source of truth for structural manifest rules.

Changes include:

- validate both bundled and library service manifests by default
- use JSON Schema validation directly instead of duplicating required fields, enum values, type checks, and patterns in a hand-written validator
- keep custom validator logic limited to operational warnings, such as missing compose files or health path formatting
- align the library JSON Schema with the intended/current manifest contract
- require top-level `service` in the JSON Schema
- validate feature-level `gpu_backends`
- wire the regression test into `make test` so `make gate` runs it
- add JSON-schema-vs-validator source-of-truth coverage for previously drift-prone cases, including:
  - booleans-as-integers
  - `minLength` checks
  - `host_network` typing
  - `env_vars` shape
  - feature requirements/id/priority
  - tag patterns
- add a clean dependency preflight for `PyYAML` and `jsonschema`, with an explicit message that these are developer/CI validation dependencies, not normal Dream Server runtime dependencies
- update schema documentation to describe the source-of-truth validation flow and the bundled/library manifest locations

## Tests
bash
bash -n dream-server/scripts/validate-manifest-schema.sh dream-server/tests/test-validate-manifest-schema.sh
bash dream-server/tests/test-validate-manifest-schema.sh
make -C dream-server lint
make -C dream-server test


gabsprogrammer:
Thanks for the PR! The direction here is highly valuable: replacing duplicated handwritten validator logic with JSON Schema validation, scanning both bundled and library manifests, and adding regression coverage is exactly the right kind of cleanup for this area. It's a strong and incredibly useful contribution.

However, I would request one critical fix before merging, as the PR currently leaves **DreamServer** with two conflicting manifest contracts.

### The Core Issue: Schema Drift
The new validator introduces a hard-coded path:
* `scripts/validate-manifest-schema.sh` ➡️ `extensions/library/schema/service-manifest.v1.json`

But the official project contract still points to:
* `manifest.json` ➡️ `contracts.extensions.serviceManifestSchema` ➡️ `extensions/schema/service-manifest.v1.json`

While both schemas share the same `$id`, **they do not validate the same contract.** I checked a few drift cases, and they produce opposite results:

| Case / Feature | `extensions/schema/...` (Official) | `extensions/library/schema/...` (New) |
| :--- | :---: | :---: |
| **Manifest with no top-level service** | ✅ Valid | ❌ Invalid |
| **Service missing `type` / `category`** | ✅ Valid | ❌ Invalid |
| **Feature with `launch: { type: "bogus" }`** | ❌ Invalid | ✅ Valid *(due to `additionalProperties: true`)* |
| **`service.startup_timeout: 999`** | ❌ Invalid | ✅ Valid *(constraint is missing)* |

> ⚠️ **Impact:** While the new test suite proves that the CLI matches the *library* schema, it does not prove that the CLI matches the repository's *declared* manifest contract. Future contributors could get contradictory answers depending on whether they run `scripts/validate-manifest-schema.sh`, `scripts/validate-manifests.sh`, or the library validator.

---

### Recommended Fixes

- [ ] **Establish a Single Source of Truth:** Pick one authoritative schema path and make every validator use it. Either remove/sync the duplicate schema, or add a parity test that fails when the two schema files diverge.
- [ ] **Port Missing Contract Fields:** Intentionally migrate the missing fields into the chosen schema, including: `compatibility`, `ui_path`, `health_timeout`, `external_link`, `startup_check`, `startup_timeout`, `launch`, `hooks`, `vram_mb`, `tags`, CPU backend support, and the `host_network`/`health` behavior.
- [ ] **Dynamic Schema Resolution:** Update the new regression test to read the schema path dynamically from `manifest.json` instead of hard-coding the `extensions/library/schema/...` path.
- [ ] **Differentiate if Intentional:** If bundled and library manifests intentionally require different contracts, document that explicitly and name the schemas differently so they do not share a single `$id`.

---

### Validation Results

<details>
<summary><b>Click to expand validation logs</b></summary>

* `bash -n dream-server/scripts/validate-manifest-schema.sh dream-server/tests/test-validate-manifest-schema.sh` — **Passed**
* `bash dream-server/tests/test-validate-manifest-schema.sh` — **Passed**
* `make lint` — **Passed**
* `bash dream-server/scripts/validate-manifest-schema.sh` — **Passed** *(with existing compose-file warnings)*
* `bash dream-server/scripts/validate-manifest-schema.sh --strict` — **Failed** *(Currently fails because `langfuse`, `dify`, and `jan` point at `compose.yaml` while the repo contains `compose.yaml.disabled`)*
* `make test` — **Failed earlier** *(Failed in my local environment at the Linux preflight JSON output contract, prior to this new manifest validator section. I wouldn't attribute this to this PR without a main-branch comparison).*

</details>

---

### Conclusion
This is a great step forward, but we should hold off on the merge until this schema source-of-truth issue is resolved. Otherwise, we risk reducing one validator drift while preserving a much larger contract drift in the repository.

Let's also wait for the final check from @Lightheartdevs.

DeluxeRaph:
Thanks for the careful review — you were right. I pushed a follow-up fix in `4f4619c4` that addresses the schema source-of-truth blocker.

What changed:

- `scripts/validate-manifest-schema.sh` now resolves the authoritative schema dynamically from `manifest.json` via `contracts.extensions.serviceManifestSchema` instead of hard-coding `extensions/library/schema/service-manifest.v1.json`.
- Consolidated the missing contract fields into the canonical `extensions/schema/service-manifest.v1.json`, including `compatibility`, `ui_path`, `health_timeout`, `external_link`, `startup_check`, `startup_timeout`, `launch`, `hooks`, `vram_mb`, `tags`, CPU backend support, and the `host_network` / `health` behavior.
- Kept `extensions/library/schema/service-manifest.v1.json` synchronized with the canonical schema for catalog/library consumers, and documented that it is not a separate contract.
- Updated the regression test to read the schema path from `manifest.json` and added a parity check that fails if the library schema drifts from the manifest.json-declared schema.
- Added regression cases for the drift examples you called out: missing top-level `service`, missing `type` / `category`, bogus feature `launch.type`, `startup_timeout: 999`, and bad `ui_path`.

Local verification passed:

```bash
bash -n dream-server/scripts/validate-manifest-schema.sh dream-server/tests/test-validate-manifest-schema.sh
bash dream-server/tests/test-validate-manifest-schema.sh
make -C dream-server lint
make -C dream-server test
```

CI has been re-triggered on the new commit and is currently pending.

Lightheartdevs:
Thanks for following through on the manifest-schema contract here. The follow-up makes this feel much safer because it ties validation back to the repo's declared source of truth instead of leaving two quiet schema paths to drift.

What looks good:
- `dream-server/scripts/validate-manifest-schema.sh` now resolves the authoritative schema from `dream-server/manifest.json` at `contracts.extensions.serviceManifestSchema`.
- The validation covers both bundled service manifests and library manifests.
- The extra script logic is limited to operational warnings rather than duplicating JSON Schema rules.

What I checked:
- `dream-server/tests/test-validate-manifest-schema.sh` compares CLI validity against JSON Schema validity, verifies real manifests, and fails if `extensions/library/schema/service-manifest.v1.json` drifts from the canonical schema.
- `dream-server/Makefile` wires the new regression test into `make test`.
- The GitHub check rollup I inspected was green for the relevant suites, including Test Linux, ShellCheck, Python lint, Dashboard frontend/API, env validation, matrix smoke, PowerShell lint, and secret scan.

Residual risk is mostly that future manifest PRs may need small schema or documentation follow-ups. That seems like the expected cost of tightening the contract, not a blocker here.

bong-water-water-bong (approved):
Significant tightening of the service manifest schema — adds required `service` block with stricter required fields, `compatibility` with version constraints, and more complete property definitions. The drift between the schema and actual manifests was a known gap; this aligns them. Tests added to validate.

Lightheartdevs:
Thanks for rebasing the manifest-schema validator work onto current main; the source-of-truth cleanup still fits the direction of the extension contract work.

What looks good:
- `ods/scripts/validate-manifest-schema.sh` resolves the authoritative schema through `ods/manifest.json` instead of hard-coding a parallel schema path.
- `ods/tests/test-validate-manifest-schema.sh` keeps bundled and library schemas synchronized and covers the drift cases called out earlier, including required `service`, `launch.type`, `startup_timeout`, `ui_path`, tags, and GPU backend values.
- The refreshed head is just the manifest validator/schema/test work carried onto current `main`, with the GitHub check rollup green across ShellCheck, Test Linux, Dashboard, matrix smoke, PowerShell lint, Python lint/type check, secret scan, and env validation.

What I checked:
- Current head SHA, prior comments, changed files, and diff against `origin/main`; the previous advisory marker was for the older `4f4619c4` head only.
- Local worktree: `git diff --check origin/main...HEAD` passed.
- Local worktree: `bash -n ods/scripts/validate-manifest-schema.sh ods/tests/test-validate-manifest-schema.sh` passed with Git Bash.
- I also started `ods/tests/test-validate-manifest-schema.sh` with temporary `PyYAML`/`jsonschema` deps; the schema-validation portions passed, but the nested Python harness resolved this Windows host's broken WSL `bash.exe` shim for subprocess calls. CI's Linux lanes are the better signal for that shell harness on this repo.

Thanks again for tightening this contract.

PR #1554, opened 2026-06-13
test(installer): reject ephemeral Docker sha tags

hafeezhmha:
## Summary

Refs #1544.

This is a follow-up hardening change after the 2.5.4 stable Hermes image fix in #1548.

#1548 protects release-critical Hermes files from reintroducing `nousresearch/hermes-agent:sha-*`. This PR extends the dependency pin policy more generally, so future runtime image references cannot use ephemeral upstream `sha-<commit>` tags.

Proper immutable digest pins using `@sha256:` remain allowed.

## AI Assistance

AI helped inspect issue #1544 and the related merged PRs, identify the remaining dependency pin policy gap, draft the validation tests, and run local verification.

I reviewed the final diff and am accountable for it.

## Release Lane

* [ ] Stable hotfix targeting `release/2.5.x`
* [x] Mainline change targeting `main`
* [ ] Next-minor work targeting the next feature/minor release
* [ ] Not sure; reviewer should help classify

Stable hotfix reason:

```text
n/a
```

## Changed Surface

* [ ] Docs only
* [x] Tests only
* [ ] Dashboard UI
* [ ] Dashboard API / host agent
* [ ] Installer / bootstrap / lifecycle
* [ ] Docker Compose / service manifests
* [ ] Model routing / Hermes / capabilities
* [ ] Network exposure / auth / proxy
* [x] Dependencies / runtime wiring

## Risk And Validation

Risk level: Low

Validation run:

* [x] `git diff --check`
* [ ] Markdown/link sanity for docs
* [x] Focused tests listed below
* [ ] Dashboard lint/test/build
* [ ] Extension audit / compose validation
* [ ] Release-grade fleet or scoped hardware validation
* [ ] Stable-lane patch validation, if targeting `release/2.5.x`
* [ ] Not required because:

Commands/results:

```bash
python3 tests/test-dependency-pins.py
```

```text
[PASS] dependency pin tests
```

```bash
python3 scripts/check-dependency-pins.py
```

```text
[PASS] dependency pins are documented and enforceable
```

```bash
PYTHONPYCACHEPREFIX=/tmp/dreamserver-pycache python3 -m py_compile scripts/check-dependency-pins.py tests/test-dependency-pins.py
```

```text
passed
```

```bash
PYTHONPYCACHEPREFIX=/tmp/dreamserver-pycache make lint
```

```text
All lint checks passed.
```

```bash
bash tests/test-linux-cloud-mode.sh
```

```text
[PASS] linux cloud mode contracts
```

```bash
[...]

PR #1579, opened 2026-06-15
fix(ci): reject ephemeral sha-<commit> image tags that upstream registries prune

### LINKED ISSUE #1544: Installer pins ephemeral hermes-agent sha-tag that upstream pruned (2.5.2 install 404s)
## Summary

Installing from tag **2.5.2** fails when pulling the Hermes Agent image:

\`\`\`
Error response from daemon: failed to resolve reference
"docker.io/nousresearch/hermes-agent:sha-dd0923bb89ed2dd56f82cb63656a1323f6f42e6f"
\`\`\`

This blocks the install from completing (no \`dream\` CLI / dreamserver service) on any host — it is not OS- or runtime-specific.

## Root cause

2.5.2 pins the Hermes Agent image by an **ephemeral CI \`sha-<commit>\` tag**. Upstream (`nousresearch/hermes-agent`) does not retain those tags. Querying Docker Hub today:

- \`sha-dd0923bb89ed2dd56f82cb63656a1323f6f42e6f\` → **HTTP 404**
- Tags that *do* exist: \`latest\`, \`main\`, and date tags \`v2026.4.3 … v2026.6.5\`.

So the pinned manifest is simply gone from the registry.

## Status / fix

\`main\` already moved to a date-pinned tag (\`nousresearch/hermes-agent:v2026.5.16\`) in \`config/dependency-lock.json\` + \`extensions/services/hermes/compose.yaml\`, and added the \`HERMES_AGENT_IMAGE\` override + \`HERMES_AGENT_IMAGE_FALLBACK\` knobs. This issue tracks:

1. Confirming current \`main\` is unaffected (no remaining \`sha-\` pins for upstream images).
2. **Policy:** never pin third-party upstream images to ephemeral \`sha-<commit>\` tags we don't control — pin to retained \`vYYYY.M.D\` tags or our own mirror, so a release can't be invalidated by upstream tag pruning.

## Workaround for users still on 2.5.2

Set in \`.env\`:
\`\`\`
HERMES_AGENT_IMAGE=nousresearch/hermes-agent:v2026.5.16
\`\`\`
or move off 2.5.2 to a current tag.

HimanshuShukla671:
Fixes #1544

## Summary

Issue #1544 was caused by the 2.5.2 release pinning the Hermes Agent image to an ephemeral `sha-<commit>` tag (`sha-dd0923bb...`) that upstream (`nousresearch/hermes-agent`) later pruned from Docker Hub, causing a hard 404 on install.

While `main` already moved to a retained `v2026.5.16` tag, there was **no CI guard** to prevent this from happening again. This PR adds one.

### Changes

**`scripts/check-dependency-pins.py`**
- Added `_is_ephemeral_sha_tag()` helper that detects `sha-[0-9a-f]{7,}` tag patterns
- Added a validation rule in `validate_refs()` that rejects any image using an ephemeral sha-commit tag (unless it's a locally-built allowlisted image)
- Error message guides contributors to use retained version tags or `@sha256:` digests instead

**`tests/test-dependency-pins.py`**
- `test_ephemeral_sha_tag_is_rejected()` — verifies the guard catches the exact tag pattern from the original issue
- `test_digest_ref_is_accepted()` — verifies immutable `@sha256:` digest refs pass clean (intended escape path)
- `test_local_allowlisted_sha_is_accepted()` — verifies locally-built images with sha-style tags are not rejected when allowlisted (intended escape path)

### Scope

This guard currently covers **bundled runtime services** (`extensions/services/`, `installers/`, and root compose files). The extension library catalog (`extensions/library/services/`) is not yet covered by the dependency pin scanner — those ~40 images would need to be registered in `dependency-lock.json` first, which is a separate effort. Filed as a follow-up scope expansion.

### Test results
```
$ python3 tests/test-dependency-pins.py
[PASS] dependency pin tests  (6 tests)
```

## Release Lane

## Changed Surface

- [x] CI / Testing

## Risk And Validation

- Risk level: Low — additive CI guard only, no runtime behavior changes
- Validation: All 6 dependency pin tests pass

### DIFF
--- a/dream-server/scripts/check-dependency-pins.py
+++ b/dream-server/scripts/check-dependency-pins.py
@@ -176,6 +176,18 @@ def _has_tag_or_digest(value: str) -> bool:
     return _has_digest(value) or _image_tag(value) is not None
 
 
+def _is_ephemeral_sha_tag(value: str) -> bool:
+    """Return True if the image tag looks like an ephemeral CI sha-<commit> tag.
+
+    Upstream registries routinely prune these tags, so pinning to them makes
+    releases unreproducible.  See issue #1544.
+    """
+    tag = _image_tag(value)
+    if tag is None:
+        return False
+    return bool(re.match(r'^sha-[0-9a-f]{7,}$', tag))
+
+
 def _is_variable_ref(value: str) -> bool:
     return "$" in value
 
@@ -280,6 +292,16 @@ def validate_refs(refs: Iterable[ImageRef], lock: dict[str, object], root: Path 
         if _image_tag(ref.value) == "latest" and not _has_digest(ref.value):
             if (ref.path, ref.value) not in latest_allow:
                 errors.append(f"{location}: latest tag requires allow_latest: {ref.value}")
+
+        # Reject ephemeral sha-<commit> tags on third-party images (issue #1544).
+        # These tags are pruned by upstream registries without notice, breaking
+        # installs that pin to them.  Pin to retained vYYYY.M.D / semver tags
+        # or a full @sha256: digest instead.
+        if _is_ephemeral_sha_tag(ref.value) and (ref.path, ref.value) not in local_allow:
+            errors.append(
+                f"{location}: ephemeral sha-<commit> tag will be pruned by upstream "
+                f"— pin to a retained version tag or @sha256: digest instead: {ref.value}"
+            )
 
         if (ref.path, ref.value) not in entry_keys and (ref.path, ref.value) not in latest_allow:
             errors.append(f"{location}: image ref is not recorded in dependency-lock.json: {ref.value}")
--- a/dream-server/tests/test-dependency-pins.py
+++ b/dream-server/tests/test-dependency-pins.py
@@ -97,12 +97,66 @@ def test_ephemeral_sha_tag_is_rejected() -> None:
     assert any("ephemeral sha-<commit> tag" in error for error in errors)
 
 
+def test_digest_ref_is_accepted() -> None:
+    """Full @sha256: digest refs must be accepted (they are immutable, not ephemeral)."""
+    module = load_module()
+    lock = {
+        "entries": [
+            {
+                "id": "test.digest",
+                "path": "compose.yaml",
+                "value": "myrepo/myimage:slim-latest@sha256:6e399abf4ff587822b0ef0df11f36088fb928e17ac61556fe89beb68d48c378e",
+            }
+        ],
+        "allow_latest": [],
+        "allow_local_images": [],
+        "allow_variable_refs": [],
+    }
+    ref = module.ImageRef(
+        path="compose.yaml",
+        line=3,
+        raw="myrepo/myimage:slim-latest@sha256:6e399abf4ff587822b0ef0df11f36088fb928e17ac61556fe89beb68d48c378e",
+        value="myrepo/myimage:slim-latest@sha256:6e399abf4ff587822b0ef0df11f36088fb928e17ac61556fe89beb68d48c378e",
+        source="compose image",
+    )
+    errors = module.validate_refs([ref], lock)
+    assert not any("ephemeral" in e for e in errors), f"Digest ref should not be rejected: {errors}"
+
+
+def test_local_allowlisted_sha_is_accepted() -> None:
+    """Locally-built images with sha-style tags must be accepted when allowlisted."""
+    module = load_module()
+    lock = {
+        "entries": [],
+        "allow_latest": [],
+        "allow_local_images": [
+            {
+                "path": "compose.yaml",
+                "value": "dream-custom:sha-abc1234567",
+                "reason": "Local image built from adjacent Dockerfile.",
+            }
+        ],
+        "allow_variable_refs": [],
+    }
+    ref = module.ImageRef(
+        path="compose.yaml",
+        line=3,
+        raw="dream-custom:sha-abc1234567",
+        value="dream-custom:sha-abc1234567",
+        source="compose image",
+    )
+    errors = module.validate_refs([ref], lock)
+    assert not any("ephemeral" in e for e in errors), f"Local allowlisted sha ref should not be rejected: {errors}"
+
+
 def main() -> int:
     tests = [
         test_repo_dependency_lock_passes,
         test_unallowlisted_latest_is_rejected,
         test_variable_refs_must_be_documented,
         test_ephemeral_sha_tag_is_rejected,
+        test_digest_ref_is_accepted,
+        test_local_allowlisted_sha_is_accepted,
     ]
     for test in tests:
         test()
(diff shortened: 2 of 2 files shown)


gabsprogrammer:
Thanks for the PR! 

This is a useful and well-scoped CI hardening fix for #1544. The `sha-<commit>` tag failure was a real release-quality issue, and placing the policy directly in `check-dependency-pins.py` is the right approach since that script already runs in the Linux workflow and release gate.

---

###  Validation & Behavior Checks

I verified the core implementation behavior against several test cases:

| Test Case / Ref | Expected Behavior | Result |
| :--- | :---: | :---: |
| `nousresearch/hermes-agent:sha-dd0923bb...` | ❌ Rejected | Passed |
| `repo:v2026.5.16` (Normal retained tags) | ✅ Allowed | Passed |
| Local allowlisted images | ✅ Skipped (pre-rule check) | Passed |
| `repo@sha256:<digest>` (Digest-only refs) | ✅ Allowed | Passed |

#### Local Test Suite Results

* `python3 tests/test-dependency-pins.py` — **Passed**
* `python3 scripts/check-dependency-pins.py` — **Passed**
* `make lint` — **Passed**

---

###  Scope Note: Extension Library Coverage

One important detail regarding scope: `candidate_files()` currently scans root compose files, installers, and `extensions/services`, but **does not scan** `extensions/library/services`.

>  **Impact:** This PR successfully prevents the exact Hermes-style breakage for *bundled* runtime services, but it does not yet prevent a future pruned `sha-<commit>` image tag from landing in the *extension library*. Since the installer copies the library for the dashboard extension catalog, this remains part of our shipped surface.

---

### Recommended Actions

- [ ] **Address the Library Scope (Choose one):**
  - **Option A:** Add `extensions/library/services` to the dependency pin scanner and lock policy.
  - **Option B:** Keep the current scope but update the PR description/policy wording to explicitly state that this guard currently covers bundled/runtime dependency pins, not the full extension library catalog.
- [ ] **Add Regression Test Cases (Non-blocking):** Add tests asserting that digest-only refs are accepted and local allowlisted `sha-*` refs remain accepted, as these are the two intended escape paths.

---

### Conclusion
This is a good, low-risk guardrail that is correctly placed. I think it is mergeable once the target scope is clarified.
Just wait for the final check from [@Lightheartdevs](https://github.com/Lightheartdevs).

HimanshuShukla671:
> Thanks for the PR!
> 
> This is a useful and well-scoped CI hardening fix for #1544. The `sha-<commit>` tag failure was a real release-quality issue, and placing the policy directly in `check-dependency-pins.py` is the right approach since that script already runs in the Linux workflow and release gate.
> 
> ### Validation & Behavior Checks
> I verified the core implementation behavior against several test cases:
> 
> Test Case / Ref	Expected Behavior	Result
> `nousresearch/hermes-agent:sha-dd0923bb...`	❌ Rejected	Passed
> `repo:v2026.5.16` (Normal retained tags)	✅ Allowed	Passed
> Local allowlisted images	✅ Skipped (pre-rule check)	Passed
> `repo@sha256:<digest>` (Digest-only refs)	✅ Allowed	Passed
> #### Local Test Suite Results
> * `python3 tests/test-dependency-pins.py` — **Passed**
> * `python3 scripts/check-dependency-pins.py` — **Passed**
> * `make lint` — **Passed**
> 
> ### Scope Note: Extension Library Coverage
> One important detail regarding scope: `candidate_files()` currently scans root compose files, installers, and `extensions/services`, but **does not scan** `extensions/library/services`.
> 
> > **Impact:** This PR successfully prevents the exact Hermes-style breakage for _bundled_ runtime services, but it does not yet prevent a future pruned `sha-<commit>` image tag from landing in the _extension library_. Since the installer copies the library for the dashboard extension catalog, this remains part of our shipped surface.
> 
> ### Recommended Actions
> * [ ]  **Address the Library Scope (Choose one):**
>   
>   * **Option A:** Add `extensions/library/services` to the dependency pin scanner and lock policy.
>   * **Option B:** Keep the current scope but update the PR description/policy wording to explicitly state that this guard currently covers bundled/runtime dependency pins, not the full extension library catalog.
> * [ ]  **Add Regression Test Cases (Non-blocking):** Add tests asserting that digest-only refs are accepted and local allowlisted `sha-*` refs remain accepted, as these are the two intended escape paths.
> 
> ### Conclusion
> This is a good, low-risk guardrail that is correctly placed. I think it is mergeable once the target scope is clarified. Just wait for the final check from [@Lightheartdevs](https://github.com/Lightheartdevs).

Thanks for the review! Updated the PR:

1. Regression tests added — test_digest_ref_is_accepted and test_local_allowlisted_sha_is_accepted now verify both intended escape paths (immutable digests and local allowlisted images)
2l Scope clarified (Option B) — adding extensions/library/services to the scanner surfaces ~40 unregistered images that would need lock entries, so that's a separate effort. Updated the PR description to explicitly document that this guard covers bundled runtime services, with the library catalog as a follow-up
All 6 tests passing.

Lightheartdevs:
Thanks for tightening up this dependency-pin path. This is a useful guardrail for keeping DreamServer image references reproducible and less likely to break later when registries prune commit-style tags.

What looks good:
- The `sha-<commit>` tag check is scoped to the existing dependency-pin scanner instead of adding a separate path contributors would need to remember.
- It preserves the intended escape hatches: immutable `@sha256:` references and local allowlisted `sha-*` images still work.
- The tests cover the original failure pattern plus digest refs and local allowlisted refs.

What I checked:
- `dream-server/scripts/check-dependency-pins.py` rejects third-party ephemeral `sha-[0-9a-f]{7,}` tags.
- Local focused test passed: `python dream-server/tests/test-dependency-pins.py`.
- I did not see attached GitHub status checks on this head in the PR rollup I queried, so this read is based on local focused validation plus static review.

Thanks again for getting this into a focused, easy-to-review shape. The documented follow-up to expand scanner coverage into `extensions/library/services` seems reasonable as a separate PR.

HimanshuShukla671:
> Thanks for tightening up this dependency-pin path. This is a useful guardrail for keeping DreamServer image references reproducible and less likely to break later when registries prune commit-style tags.
> 
> Advisory status: looks merge-ready
> 
> What looks good:
> 
> * The `sha-<commit>` tag check is scoped to the existing dependency-pin scanner instead of adding a separate path contributors would need to remember.
> * It preserves the intended escape hatches: immutable `@sha256:` references and local allowlisted `sha-*` images still work.
> * The tests cover the original failure pattern plus digest refs and local allowlisted refs.
> 
> What I checked:
> 
> * `dream-server/scripts/check-dependency-pins.py` rejects third-party ephemeral `sha-[0-9a-f]{7,}` tags.
> * Local focused test passed: `python dream-server/tests/test-dependency-pins.py`.
> * I did not see attached GitHub status checks on this head in the PR rollup I queried, so this read is based on local focused validation plus static review.
> 
> Thanks again for getting this into a focused, easy-to-review shape. The documented follow-up to expand scanner coverage into `extensions/library/services` seems reasonable as a separate PR.

Thanks so much for the review and for testing it out! Really appreciate the guidance on getting this into a merge-ready shape. Let me know if there's anything else needed from me, otherwise I'm happy for this to be merged whenever you have the chance. Cheers!

HimanshuShukla671:
Closing this as a duplicate of #1554. Apologies to the original contributor, I completely missed that you already had an active PR for this issue! Great work on it.

PR #1586, opened 2026-06-22
fix(dashboard-api): check is_symlink() before is_file() in dir_size_gb

### LINKED ISSUE #1590: dir_size_gb() follows symlinks before checking is_symlink()
## Description

In `extensions/services/dashboard-api/helpers.py`, the `dir_size_gb()` function checks:

```python
if f.is_file() and not f.is_symlink():
    total += f.stat().st_size
```

`Path.is_file()` follows symlinks by default — it resolves the link target and stats it before `is_symlink()` gets a chance to short-circuit. This means:

1. Symlinks pointing outside the DATA_DIR tree are followed unnecessarily during the stat call
2. The intent is clearly to skip symlinks entirely, but the current order does extra work on every symlink entry

## Expected behavior

Check `is_symlink()` first (cheap `lstat`, no link resolution), then only call `is_file()` on confirmed non-symlink entries:

```python
if not f.is_symlink() and f.is_file():
```

## Impact

Defense-in-depth fix. Avoids following arbitrary symlinks in user-controlled data directories during disk usage calculation.

See #1586

fireddd:
## Summary

In `helpers.py`'s `dir_size_gb()` function, the symlink check condition was:

```python
if f.is_file() and not f.is_symlink():
```

`Path.is_file()` follows symlinks — it stats the symlink target to determine if it's a file. This means even when we intend to skip symlinks, we still follow (and stat) the target before `not f.is_symlink()` short-circuits the condition.

This is a defense-in-depth concern: if a symlink points outside the DATA_DIR tree, the `is_file()` call leaks information about whether the target exists and is a file. The fix swaps the condition order:

```python
if not f.is_symlink() and f.is_file():
```

`is_symlink()` uses `lstat` (does not follow the link), so checking it first avoids resolving any symlinks at all. Python's short-circuit evaluation means `is_file()` is only called on non-symlink entries.

## What changed

- `dream-server/extensions/services/dashboard-api/helpers.py` line 552: swapped condition order so `is_symlink()` is checked before `is_file()`

## Behavior impact

- **Functionally identical** for all normal cases (regular files, directories, dangling symlinks)
- Avoids unnecessary symlink target resolution as a defense-in-depth measure
- No new dependencies, no API changes

## Test plan

- [x] Python syntax verification (`ast.parse`) passes
- [ ] Existing `pytest tests/` suite passes (no behavior change expected)

Closes #1590

### DIFF
--- a/dream-server/extensions/services/dashboard-api/helpers.py
+++ b/dream-server/extensions/services/dashboard-api/helpers.py
@@ -549,7 +549,7 @@ def dir_size_gb(path: Path) -> float:
     try:
         for f in path.rglob("*"):
             try:
-                if f.is_file() and not f.is_symlink():
+                if not f.is_symlink() and f.is_file():
                     total += f.stat().st_size
             except (PermissionError, OSError):
                 pass
--- a/dream-server/extensions/services/dashboard-api/tests/test_helpers.py
+++ b/dream-server/extensions/services/dashboard-api/tests/test_helpers.py
@@ -1056,3 +1056,32 @@ class TestDirSizeGb:
         invalidate_dir_size_cache(d)
         assert dir_size_gb(d) == 0.0
         assert calls["count"] == 1
+
+    def test_dangling_symlinks_are_skipped(self, tmp_path):
+        """A symlink whose target doesn't exist must not crash or contribute size."""
+        clear_dir_size_cache()
+        d = tmp_path / "dangling"
+        d.mkdir()
+        real = d / "real.bin"
+        real.write_bytes(b"\x00" * 1024)
+        dangling = d / "dangling.bin"
+        dangling.symlink_to(d / "nonexistent-target")
+        assert not dangling.exists()  # confirm it's dangling
+        assert dangling.is_symlink()  # but it is a symlink
+        result = dir_size_gb(d)
+        assert result == 0.0  # only real.bin counted (1024 bytes ≈ 0.0 GB)
+
+    def test_symlink_to_external_file_not_counted(self, tmp_path):
+        """A symlink pointing outside the scanned directory must not add size."""
+        clear_dir_size_cache()
+        external = tmp_path / "external"
+        external.mkdir()
+        ext_file = external / "big.bin"
+        ext_file.write_bytes(b"\x00" * (1024 * 1024 * 200))  # 200 MiB
+
+        d = tmp_path / "scanned"
+        d.mkdir()
+        link = d / "external_link.bin"
+        link.symlink_to(ext_file)
+        result = dir_size_gb(d)
+        assert result == 0.0  # symlink skipped, external file not counted


fireddd:
## Test Evidence: Before/After Verification

### Code Change (line 552 of `helpers.py`)

| | Code |
|---|---|
| **Before** (main) | `if f.is_file() and not f.is_symlink():` |
| **After** (this PR) | `if not f.is_symlink() and f.is_file():` |

### Full Diff

```diff
--- a/dream-server/extensions/services/dashboard-api/helpers.py
+++ b/dream-server/extensions/services/dashboard-api/helpers.py
@@ -549,7 +549,7 @@ def dir_size_gb(path: Path) -> float:
     try:
         for f in path.rglob("*"):
             try:
-                if f.is_file() and not f.is_symlink():
+                if not f.is_symlink() and f.is_file():
                     total += f.stat().st_size
             except (PermissionError, OSError):
                 pass
```

---

### BEFORE: `main` helpers.py + all 8 tests

```
tests/test_helpers.py::TestDirSizeGb::test_nonexistent_path_returns_zero PASSED [ 12%]
tests/test_helpers.py::TestDirSizeGb::test_empty_directory_returns_zero PASSED [ 25%]
tests/test_helpers.py::TestDirSizeGb::test_directory_with_files PASSED   [ 37%]
tests/test_helpers.py::TestDirSizeGb::test_symlinks_are_skipped PASSED   [ 50%]
tests/test_helpers.py::TestDirSizeGb::test_uses_cached_value_until_invalidated PASSED [ 62%]
tests/test_helpers.py::TestDirSizeGb::test_invalidate_dir_size_cache_forces_refresh PASSED [ 75%]
tests/test_helpers.py::TestDirSizeGb::test_dangling_symlinks_are_skipped PASSED [ 87%]
tests/test_helpers.py::TestDirSizeGb::test_symlink_to_external_file_not_counted PASSED [100%]

8 passed in 0.47s
```

### AFTER: Fixed helpers.py + all 8 tests

```
tests/test_helpers.py::TestDirSizeGb::test_nonexistent_path_returns_zero PASSED [ 12%]
tests/test_helpers.py::TestDirSizeGb::test_empty_directory_returns_zero PASSED [ 25%]
tests/test_helpers.py::TestDirSizeGb::test_directory_with_files PASSED   [ 37%]
tests/test_helpers.py::TestDirSizeGb::test_symlinks_are_skipped PASSED   [ 50%]
tests/test_helpers.py::TestDirSizeGb::test_uses_cached_value_until_invalidated PASSED [ 62%]
tests/test_helpers.py::TestDirSizeGb::test_invalidate_dir_size_cache_forces_refresh PASSED [ 75%]
tests/test_helpers.py::TestDirSizeGb::test_dangling_symlinks_are_skipped PASSED [ 87%]
tests/test_helpers.py::TestDirSizeGb::test_symlink_to_external_file_not_counted PASSED [100%]

8 passed in 0.37s
```

---

### Test Coverage Matrix

| Test | What it covers | Pre-existing? |
|---|---|---|
| `test_nonexistent_path_returns_zero` | Missing directory returns 0.0 | Yes |
| `test_empty_directory_returns_zero` | Empty directory returns 0.0 | Yes |
| `test_directory_with_files` | 100 MiB file correctly measured as 0.1 GB | Yes |
| `test_symlinks_are_skipped` | Valid symlink to a real file is excluded from size | Yes |
| `test_uses_cached_value_until_invalidated` | Repeated calls use cache (no filesystem walk) | Yes |
| `test_invalidate_dir_size_cache_forces_refresh` | Cache invalidation triggers a fresh walk | Yes |
| `test_dangling_symlinks_are_skipped` | Symlink with nonexistent target does not crash or add size | **New** |
| `test_symlink_to_external_file_not_counted` | Symlink to 200 MiB file outside scanned dir is excluded | **New** |

---

### Why `not f.is_symlink() and f.is_file()` is safer

1. **Avoids following symlinks unnecessarily.** `is_file()` calls `stat()` which follows symlinks to resolve the target. `is_symlink()` calls `lstat()` which only inspects the link entry itself. Checking `is_symlink()` first and short-circuiting means we never `stat()` a symlink target at all.

2. **Defense-in-depth against external targets.** If a symlink points outside `DATA_DIR` (e.g., to `/etc/passwd`), the old order would `stat()` that external path before deciding to skip the entry. The new order avoids touching anything outside the tree.

3. **Resilience to dangling symlinks.** While both orderings produce correct results for dangling symlinks (since `is_file()` returns `False` for them), the new order skips them via the cheaper `lstat()` path without attempting to resolve a nonexistent target.

4. **Fewer syscalls.** For every symlink encountered, the old order made 2 syscalls (`stat` + `lstat`). The new order makes 1 (`lstat` only, then short-circuits).

**Result**: Functionally equivalent for all cases, but the new order is more efficient and avoids information leakage through unnecessary symlink resolution.

---

*Tested on Python 3.13.14, pytest 8.4.2, macOS Darwin 25.3.0*

Lightheartdevs:
Thanks for tightening the disk-usage walk around symlinks in the dashboard API.

What looks good:
- `dream-server/extensions/services/dashboard-api/helpers.py` now checks `is_symlink()` before `is_file()` at the `dir_size_gb()` file-counting point, so symlink targets are not resolved just to decide whether to skip them.
- The change preserves the existing cache behavior and exception handling around the `rglob()` walk.
- `dream-server/extensions/services/dashboard-api/tests/test_helpers.py` adds direct coverage for dangling symlinks and symlinks that point outside the scanned directory.

What I checked:
- Static diff review against the existing `dir_size_gb()` cache and symlink-skip behavior.
- `git diff --check origin/main...origin/pr/1586`
- `python -m pytest tests/test_helpers.py::TestDirSizeGb -q` on this Windows runner got `5 passed`; the 3 symlink cases failed before assertions with `WinError 1314` because this account cannot create symlinks. The pre-existing symlink test has the same local Windows limitation, so I do not read that as a runtime regression.
- GitHub currently shows no check runs for this head, so I would still let the normal Linux dashboard-api CI run before merging.

fireddd:
Thanks for the review! Good note about the Windows symlink limitation — that's expected since `os.symlink` requires SeCreateSymbolicLinkPrivilege on Windows. The pre-existing `test_symlinks_are_skipped` has the same constraint, so the new tests are consistent. Happy to wait for Linux CI.

bong-water-water-bong (approved):
Same fix as #1621 (which I closed after noticing this existed first). Flips `is_file()` and `is_symlink()` so the cheap lstat runs first. Correct defense-in-depth — avoids following arbitrary symlinks during disk usage calculation.

PR #1598, opened 2026-06-23
fix(dashboard-api): handle impossible calendar dates in usage report

androvonx95:
`/api/usage/report` only checks dates against `^\d{4}-\d{2}-\d{2}$`, so dates like
`2026-02-30` or `2026-13-01` pass the regex but blow up when parsed. The handler
tries to return an `invalid_range` response, but `_empty_report` parses the dates
again outside the try/except, so the request 500s instead.

Fix: only build the daily list when both dates parse, otherwise leave it empty.

Repro before the fix: `GET /api/usage/report?start=2026-02-30&end=2026-03-01` → 500.

Added a test for Feb 30, month 13, and Feb 29 on a non-leap year. Fails on main,
passes here.

### DIFF
--- a/dream-server/extensions/services/dashboard-api/routers/usage.py
+++ b/dream-server/extensions/services/dashboard-api/routers/usage.py
@@ -49,8 +49,12 @@ def _date_range(start_day: date, end_day: date) -> list[str]:
 
 
 def _empty_report(start: str, end: str, status: str = "unavailable", detail: str | None = None) -> dict[str, Any]:
-    start_day = _parse_date(start)
-    end_day = _parse_date(end)
+    # The query regex (^\d{4}-\d{2}-\d{2}$) accepts calendar-invalid dates like
+    # 2026-02-30, so the daily breakdown is only built when both bounds parse.
+    try:
+        days = _date_range(_parse_date(start), _parse_date(end))
+    except ValueError:
+        days = []
     return {
         "period": {"start": start, "end": end},
         "source": {
@@ -83,7 +87,7 @@ def _empty_report(start: str, end: str, status: str = "unavailable", detail: str
                 "cache_read_tokens": 0,
                 "cache_write_tokens": 0,
             }
-            for day in _date_range(start_day, end_day)
+            for day in days
         ],
         "models": [],
         "services": [],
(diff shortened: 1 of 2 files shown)


Lightheartdevs:
Thanks for tightening the usage-report error path here; this catches the calendar-validity gap without changing the normal Token Spy/report shape.

What looks good:
- `dream-server/extensions/services/dashboard-api/routers/usage.py` keeps the existing 200/`invalid_range` contract and only suppresses daily bucket generation when `date.fromisoformat()` rejects the input, so valid unavailable reports still return the expected zeroed daily rows.
- `dream-server/extensions/services/dashboard-api/tests/test_usage.py` covers Feb 30, month 13, and a non-leap Feb 29 through the authenticated `/api/usage/report` endpoint, which is the user-visible failure path.
- The diff is confined to the dashboard-api usage router and its focused tests; no installer, compose, auth, or runtime service wiring changes.

What I checked:
- `python -m pytest tests/test_usage.py -q` in `dream-server/extensions/services/dashboard-api` on this head: 17 passed.
- `python -m py_compile routers/usage.py tests/test_usage.py`.
- `gh pr checks` currently reports no checks on this branch, so I’m treating local focused validation plus the narrow diff as the advisory evidence.

androvonx95:
Np and thanks for the quick review and for running the tests on your end. This stuff is ready from my side. Happy to adjust if a second reviewer wants anything tweaked.

bong-water-water-bong (approved):
Clean fix. The query regex `^\d{4}-\d{2}-\d{2}$` accepts syntactically valid but calendar-impossible dates (e.g. 2026-02-30), which previously caused `_parse_date` to raise. Wrapping in try/except and returning an empty `invalid_range` report is the right graceful degradation. Tests cover all three edge cases (Feb 30, month 13, non-leap Feb 29).

Lightheartdevs:
Closing as superseded after re-audit. Current main already contains the equivalent behavior from 4576c330, including the bounded date-range handling this branch lacked; the focused current-main coverage is green. Re-merging this older implementation would add duplicate paths and regression risk without new user value. Thank you for identifying the original problem.

PR #1605, opened 2026-06-24
fix: harden security posture and fix stability bugs across installer, CLI, and proxies

tang-vu:
## Summary

Hardening and stability pass across the installer, `dream-cli`, the Caddy/nginx proxies, and env-schema validation. Two commits: the initial bundle (`9f5c432`) and a follow-up (`3ea2a9c`) that addresses the advisory review so the security changes preserve the existing security contracts.

## Bug fixes

- **`scripts/health-check.sh`** — use `LLM_API_BASE_PATH` for the AMD/Lemonade endpoint instead of the wrong path.
- **`scripts/pre-download.sh`** — replace stale Qwen 2.5 references with current Qwen3 GGUF models, and download into `data/models/` where the installer expects them.
- **`dream-cli`** — atomic `.env` writes (temp file + `mv`) instead of `cat`+`rm`, so a process killed mid-write can no longer corrupt `.env`; invalidate the compose-flags cache when `GPU_BACKEND` changes.
- **`scripts/session-cleanup.sh`** — prefer `jq` over fragile `grep` for session-ID extraction, with a `grep` fallback when `jq` is unavailable.
- **`docker-compose.tier0.yml`** — add healthcheck timing overrides for memory-constrained deployments.

## Security improvements

- **`dream-proxy` / `hermes-proxy` Caddyfiles** — add `request_body` size limits to curb abuse.
- **dashboard `nginx.conf`** — pre-stage an HSTS header for when TLS lands.
- **`.env.schema.json`** — add `minLength: 10` to secret fields so `CHANGEME` placeholders are rejected.
- **`installers/phases/12-health.sh`** — pass Perplexica config via environment variables instead of shell interpolation, preventing injection through model names / API keys.
- **`routers/privacy.py`** — log inside the previously-silent `except` block instead of swallowing it.
- **`.gitattributes`** — enforce LF line endings for `dream-cli`.

## Follow-up: advisory-review fixes (`3ea2a9c`)

Keeps the hardening from breaking existing security contracts:

- **`dream-cli`** — atomic `.env` rewrites now preserve the `0600` secret-file mode. `_env_set`/`_env_unset` write into a same-directory `mktemp` (which is `0600` from creation, so secrets are never in a world-readable temp), restore the original file mode via a portable `stat` helper (GNU `-c '%a'` / BSD `-f '%Lp'`), then `mv`. Updates no longer revert `0600` to `0644`, and content commits only when `awk` succeeds (no partial writes).
- **`scripts/session-cleanup.sh`** — replace the reintroduced `2>/dev/null) || …` suppression with the inline exit-code-capture pattern used by the `grep` fallback, plus a visible warning on parse failure.
- **`scripts/validate-env.sh`** — actually enforce string `minLength`, so the schema change above is effective and placeholders like `WEBUI_SECRET=CHANGEME` are rejected. Aligned the fixtures (`N8N_USER=admin@dreamserver.local` in CI to match the installer's phase-06 default) and added a `minLength` regression test.

The installer's phase 06 generates `N8N_USER=admin@dreamserver.local` plus random secrets (all ≥ 10 chars), so `minLength: 10` is safe for real installs.

## Testing

- `tests/test-validate-env.sh` — 11/11 (including the new minLength tests); the CI tier fixture validates clean.
- `tests/test-session-cleanup.sh` — 11/11 (including the no-suppression and inline-exit-capture contracts).
- `tests/test-dream-cli-pipefail-tolerance.sh` (11), `tests/test-bash32-guards.sh` (10), `tests/contracts/test-installer-contracts.sh` (37).
- ShellCheck clean (error + warning) on the changed `.sh` files; `bash -n` clean on all changed scripts.
- Verified on a real Linux environment: `.env` stays `0600` across set/add/unset with no leftover temp files (the old path reproduced the `0600 → 0644` regression).


Lightheartdevs:
Thanks for taking on the installer, CLI, and proxy hardening bundle; these are exactly the surfaces where small correctness details matter.

Quickest path to merge:
- Fix the atomic `.env` rewrites in `dream-server/dream-cli` so they preserve the installer secret-file contract. Phase 06 creates `.env` under `umask 077` and then `chmod 600`, but `_env_set()` / `_env_unset()` now write `${file}.tmp` under the caller's ambient umask and `mv` it over the original. In a Linux container reproduction, a 0600 `.env` became 0644 after the new `mv` path, while the old `cat "${file}.tmp" > "$file"` path kept 0600. Use a same-directory `mktemp` plus mode/owner preservation before `mv` (`chmod --reference` / `chown --reference` where available), or at minimum restore `chmod 600 "$file"` after replacement and avoid leaving a readable temp file containing secrets.
- Clean up the new jq branch in `dream-server/scripts/session-cleanup.sh` so the existing `dream-server/tests/test-session-cleanup.sh` contract passes. The new `jq ... "$SESSIONS_JSON" 2>/dev/null) || ACTIVE_IDS=""` line reintroduces the silent error-suppression pattern that this test explicitly rejects. Match the fallback style: capture the jq exit code, avoid `2>/dev/null` / `|| true`, and handle invalid JSON without hiding the failure.
- Make the `.env.schema.json` `minLength` hardening enforceable, or hold it until the validator supports it. `dream-server/scripts/validate-env.sh` currently checks required keys, unknown keys, types, enums, and numeric min/max, but it never reads string `minLength`, so `WEBUI_SECRET=CHANGEME` is still not rejected by the repo validator. If the intent is to block placeholders, add string-length validation plus tests; also update env-validation fixtures such as `.github/workflows/validate-env.yml`, which currently sets `N8N_USER=admin` even though the PR adds `minLength: 10` to `N8N_USER`.

What I checked:
- Static review of the 12-file diff against current `main`, with repo context from `ARCHITECTURE.md`, contributor docs, high-risk change map, Makefile, and CI workflows.
- `git diff --check origin/main...origin/pr/1605` passed, and `git merge-tree --write-tree origin/main origin/pr/1605` produced a clean merge tree.
- `bash -n` on the changed shell entrypoints passed via Git Bash.
- `python -m py_compile dream-server/extensions/services/dashboard-api/routers/privacy.py` passed.
- `python tests/contracts/test-network-exposure-contracts.py` passed.
- `bash tests/test-session-cleanup.sh` failed on the new suppression pattern in `scripts/session-cleanup.sh`.
- `gh pr checks 1605` reports no checks for this branch; local `tests/test-validate-env.sh` could not complete here because this Windows/Git Bash environment lacks `jq`, so I verified the missing `minLength` enforcement by source inspection.

Thanks again for pushing this hardening pass forward; the main issue is making the security changes preserve the existing security contracts.

tang-vu:
Thanks for the detailed advisory review — all three points are addressed in `3ea2a9c`.

**1. Atomic `.env` rewrites preserve the secret-file contract (`dream-cli`)**
`_env_set()` / `_env_unset()` no longer write `${file}.tmp` under the caller's ambient umask. They now write into a same-directory `mktemp` (created `0600`, so secrets are never in a world-readable temp), restore the original file mode via a portable `_file_mode` helper (`stat -c '%a'` with BSD/macOS `stat -f '%Lp'` fallback), then `mv`. Content is committed only when `awk` succeeds, preserving the prior "no partial writes" guarantee.
- Verified in a Linux container: a `0600` `.env` stays `0600` across set/add/unset with no leftover temp files; the old path reproduced the `0600 → 0644` regression you saw.

**2. No silent error suppression (`scripts/session-cleanup.sh`)**
Replaced the `jq ... 2>/dev/null) || ACTIVE_IDS=""` line with the same inline exit-code-capture pattern used by the grep fallback, plus a visible warning on parse failure (no `2>/dev/null` / `|| true`).
- `bash tests/test-session-cleanup.sh` now passes (11/11), including the suppression and inline-exit-capture contracts.

**3. `minLength` is now enforceable + fixtures aligned (`scripts/validate-env.sh`)**
Added string `minLength` validation, so placeholders like `WEBUI_SECRET=CHANGEME` are rejected. Updated the contradicting fixtures: `.github/workflows/validate-env.yml` now uses `N8N_USER=admin@dreamserver.local` (matching the installer's actual phase-06 default), the `test-validate-env.sh` valid fixture uses realistic-length values, and a new minLength regression test was added.
- Confirmed phase 06 generates `N8N_USER=admin@dreamserver.local` plus random secrets (all ≥10 chars), so `minLength: 10` is safe for real installs.

**Verification (real Linux):**
- `tests/test-validate-env.sh` — 11/11 (incl. new minLength tests); CI tier fixture validates clean.
- `tests/test-session-cleanup.sh` — 11/11.
- `tests/test-dream-cli-pipefail-tolerance.sh` (11/0), `tests/test-bash32-guards.sh` (10/0), `tests/contracts/test-installer-contracts.sh` (37 passed).
- ShellCheck clean (error + warning) on the changed `.sh` files; `bash -n` clean on all three scripts.

tang-vu:
@Lightheartdevs

Lightheartdevs:
Thanks for following through on the installer, CLI, proxy, and env-schema hardening fixes; the follow-up commit resolves the concrete issues from the previous advisory.

Decision point:
- I no longer see a code blocker in the current head. The earlier .env mode regression in dream-server/dream-cli, the silent jq failure handling in dream-server/scripts/session-cleanup.sh, and the unenforced minLength gap in dream-server/scripts/validate-env.sh are all addressed and now covered by targeted tests.
- The remaining call is process/risk: this is still a broad high-risk bundle across dream-cli, env schema validation, installer health behavior, pre-download behavior, Caddy/nginx proxy config, and the Tier 0 compose overlay, and GitHub reports no checks on this branch. If maintainers want branch-side CI before merge, that is the main remaining gate; otherwise the targeted local/container validation below supports moving forward.

What I checked:
- Static review of the full 15-file diff against current main, plus the follow-up delta 9f5c432..3ea2a9c against the previous advisory.
- git diff --check origin/main...origin/pr/1605 passed; git merge-tree --write-tree origin/main origin/pr/1605 produced a clean merge tree.
- ash -n passed for the changed shell entrypoints/scripts I checked.
- ash dream-server/tests/test-session-cleanup.sh passed: 11/11.
- Linux container with ash + jq: ash dream-server/tests/test-validate-env.sh passed: 11/11, including the new minLength regression.
- ash dream-server/tests/test-dream-cli-pipefail-tolerance.sh passed: 9 passed, 1 PyYAML-dependent case skipped.
- ash dream-server/tests/test-bash32-guards.sh passed: 10/10.
- Linux container with ash, jq, python3, and PyYAML: ash dream-server/tests/contracts/test-installer-contracts.sh passed, with expected docker-compose-unavailable skips.
- python -m py_compile dream-server/extensions/services/dashboard-api/routers/privacy.py and python dream-server/tests/contracts/test-network-exposure-contracts.py passed.
- gh pr checks 1605 still reports no checks for the branch.

Thanks again for doing the careful follow-up on the security-contract details.

tang-vu:
@Lightheartdevs Waiting for your approval.

bong-water-water-bong (commented):
This PR bundles multiple concerns across 15 files (968 lines), which the project's CONTRIBUTING.md explicitly asks to avoid: *'One PR, one concern. A bug fix + a feature + a refactor = three PRs.'*

Suggest splitting into at least:
1. **Security** — `.env.schema.json` \`minLength\` additions, \`N8N_USER\` email fix — clear win, easy to merge
2. **Caddy/nginx** — proxy hardening (request_body limits, CSP headers, HSTS)
3. **Health checks** — \`health-check.sh\` + \`12-health.sh\` refactoring
4. **Session cleanup / pre-download** — refactoring and edge-case fixes
5. **Formatting-only** — \`.gitattributes\`, whitespace changes

Happy to re-review each split PR individually once they're separate.

tang-vu:
Thanks @bong-water-water-bong — agreed on one-PR-one-concern. Splitting this up now.

Heads-up for anyone following: this branch was cut before the `dream-server/ → ods/` rename (#1626) landed, which is why it now shows as conflicting — its base layout no longer exists on `main`. So rather than splitting in place, I'm re-basing each concern onto the new `ods/` paths as separate focused PRs.

First slice is up: **#1650 — `fix(env): enforce minLength on secret env vars to reject placeholders`** (the security win from this bundle). It adds `minLength: 10` to the secret schema fields *and* the missing validator support, with the CI fixture aligned to `admin@ods.local`. Verified against the new `ods/` layout: `test-validate-env.sh` 11/11 and `installer-env-smoke.sh` 12/12 (real phase-06 secrets pass the new schema).

Remaining slices (proxy hardening, health checks, etc.) will follow as their own PRs once re-based onto `ods/`. I'll close this bundle in favor of those.

tang-vu:
Closing this bundle in favor of the focused split PRs, as discussed:

- **#1650** — `fix(env): enforce minLength on secret env vars to reject placeholders` (advisory already reports it merge-ready)
- **#1655** — `fix(proxy): cap request body sizes and pre-stage HSTS header`

Status of the remaining concerns from this bundle after re-checking them against the current `ods/` main:

- `session-cleanup.sh` jq handling and `privacy.py` logging — already addressed on `main` during the ODS migration; no PR needed.
- Health-check base-path, pre-download model refs, tier-0 healthcheck overrides, and the ods-cli atomic `.env` write — still candidates; I'll re-verify each against `ods/` and open them as separate small PRs where they still apply.

Thanks for the guidance on scoping.

PR #1611, opened 2026-06-25
fix(installer): skip Qdrant on 64K-page arm64 hosts

Lightheartdevs:
## Summary
- detect Linux arm64/aarch64 hosts with page sizes larger than 4K during installer feature sync
- disable Qdrant on that host class because the upstream official image aborts with `<jemalloc>: Unsupported system page size`
- gate Qdrant port checks, image pulls, health checks, and summaries on `ENABLE_QDRANT` instead of raw `ENABLE_RAG`
- preserve existing aarch64 TEI disable behavior and add contract coverage

## Validation
- `bash -n installers/phases/03-features.sh installers/phases/04-requirements.sh installers/phases/08-images.sh installers/phases/12-health.sh installers/phases/13-summary.sh tests/contracts/test-installer-contracts.sh`
- `tests/contracts/test-installer-contracts.sh`
- simulated `HOST_ARCH=aarch64` + `PAGE_SIZE=65536`: Qdrant and TEI compose files disabled, `ENABLE_QDRANT=false`
- simulated `HOST_ARCH=aarch64` + `PAGE_SIZE=4096`: Qdrant remains enabled, TEI disabled

Fleet validation against this branch is in progress.

### DIFF
--- a/dream-server/installers/phases/03-features.sh
+++ b/dream-server/installers/phases/03-features.sh
@@ -13,7 +13,7 @@
 #           show_phase(), show_install_menu(), chapter(), bootline(),
 #           success(), log(), warn(), error(), signal()
 # Provides: ENABLE_VOICE, ENABLE_WORKFLOWS, ENABLE_RAG, ENABLE_EMBEDDINGS,
-#           ENABLE_HERMES, ENABLE_OPENCLAW, OPENCLAW_CONFIG, GPU_ASSIGNMENT_JSON,
+#           ENABLE_QDRANT, ENABLE_HERMES, ENABLE_OPENCLAW, OPENCLAW_CONFIG, GPU_ASSIGNMENT_JSON,
 #           LLAMA_SERVER_GPU_UUIDS, WHISPER_GPU_UUID, COMFYUI_GPU_UUID,
 #           EMBEDDINGS_GPU_UUID, LLAMA_ARG_SPLIT_MODE, LLAMA_ARG_TENSOR_SPLIT
 #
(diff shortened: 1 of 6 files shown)


Lightheartdevs:
Thanks for taking on the DGX/64K-page Qdrant install failure here; this is a focused fix in the installer surfaces that were still treating RAG as one indivisible Qdrant+TEI bundle.

Decision point:
- The code path looks coherent: `03-features.sh` now defaults `ENABLE_QDRANT` from `ENABLE_RAG`, disables only the concrete Qdrant service on Linux arm64/aarch64 hosts with page sizes above 4K, keeps the existing TEI aarch64 guard, refreshes compose after the service files are disabled, and moves the downstream Qdrant port/pull/health/summary gates to the concrete flag in phases 04/08/12/13.
- The remaining call is validation policy. This touches high-risk installer phases and target-host service composition for a hardware class CI cannot fully reproduce. Since the PR body says fleet validation is still in progress, I would either wait for that DGX/Brev/64K-page result or make an explicit maintainer call that green CI plus the manual 4K/64K simulation is enough for this unblock.
- Optional tightening before merge: the new contract coverage is mostly grep-based. If we want CI to catch future logic drift here, a small harness that sources `installers/phases/03-features.sh` with stubbed `getconf PAGE_SIZE` values for `4096` and `65536` and asserts the Qdrant compose file is only disabled on the >4K arm64 case would make this regression harder to reintroduce.

What I checked:
- Reviewed the PR body, current head SHA, changed files, patch, existing issue comments, formal reviews, and review comments.
- `gh pr checks 1611` reports green CI across shell lint, dashboard frontend/API, Linux integration smoke, distro matrix, env validation, secret scan, and related automation.
- Local static checks: `git diff --check origin/main...origin/pr/1611` passed; `bash -n` on the touched installer phases and contract script passed under Git Bash.
- I could not run `tests/contracts/test-installer-contracts.sh` locally because this Git Bash environment is missing `jq`; CI ran the relevant contract/integration lanes successfully.

bong-water-water-bong (approved):
Clean fix for the 64K-page arm64 Qdrant crash. Well-structured:
- Separate \`ENABLE_QDRANT\` / \`ENABLE_EMBEDDINGS\` flags from \`ENABLE_RAG\` so host guards can override individually
- Page size detection via \`getconf PAGE_SIZE\` — no hardcoded assumptions
- Both guards warn and degrade gracefully rather than fail
- Removes the old amd64-only platform fallback that was fragile

LGTM.

PR #1612, opened 2026-06-25
fix(openclaw): publish UI through Docker host bind

gabsprogrammer:
## Summary

This fixes the OpenClaw UI publish path after the recent bind-mode change. The current `main` compose path passes `--bind localhost`, but the pinned OpenClaw image rejects that value at startup. The older `loopback` mode starts successfully inside the container, but it binds only to the container loopback interface, so Docker port publishing cannot reach the UI from the host.

The fix keeps host exposure controlled by the existing Docker `ports:` mapping:

```yaml
${BIND_ADDRESS:-127.0.0.1}:${OPENCLAW_PORT:-7860}:18789
```

OpenClaw now listens on a Docker-publishable container interface with `--bind lan`, while `BIND_ADDRESS` continues to decide whether the host publishes the UI on loopback only or on a wider interface.

## Why this is needed

Observed on macOS Docker Desktop with the active Dream stack:

| Compose behavior | Container health | Host `127.0.0.1:7860` | Result |
| --- | --- | --- | --- |
| `--bind loopback` | healthy | empty reply | False healthy; UI unreachable from host |
| `--bind localhost` | restarting | connection refused | Image rejects the bind mode |
| `--bind lan` with host `BIND_ADDRESS=127.0.0.1` | healthy | HTTP 200 | UI reachable, host exposure still loopback-only |

This is not macOS-specific. Docker published ports reach the container network interface, not the service loopback inside the container, so the same contract matters on Linux, Windows Docker Desktop, and WSL-backed Docker.

## Validation

- `python3 dream-server/tests/contracts/test-network-exposure-contracts.py`
- `WEBUI_SECRET=ci-placeholder SEARXNG_SECRET=ci-placeholder N8N_USER=ci@example.com N8N_PASS=ci-placeholder LITELLM_KEY=ci-placeholder OPENCLAW_TOKEN=ci-placeholder docker compose -f docker-compose.base.yml -f extensions/services/openclaw/compose.yaml config --quiet`
- `bash dream-server/tests/test-validate-manifests.sh`
- `bash dream-server/tests/test-service-registry.sh`
- `git diff --check`
- Runtime smoke on macOS Docker Desktop:
  - `dream-openclaw` healthy
  - `dream-cli status --json` reports `openclaw healthy running 7860`
  - `curl -fsS http://127.0.0.1:7860/` returns the OpenClaw UI HTML

## Risk

Low and scoped to the deprecated optional OpenClaw service. The change does not widen host exposure by itself; published host binding remains controlled by `BIND_ADDRESS` in the existing `ports:` mapping.

### DIFF
--- a/dream-server/extensions/services/openclaw/compose.yaml
+++ b/dream-server/extensions/services/openclaw/compose.yaml
@@ -23,13 +23,14 @@ services:
       - OPENCLAW_DANGEROUSLY_DISABLE_DEVICE_AUTH=${OPENCLAW_DANGEROUSLY_DISABLE_DEVICE_AUTH:-}
       - HOME=/home/node
     user: "0:0"
-    # Gateway bind mode tracks BIND_ADDRESS: a loopback bind (127.0.0.1 /
-    # localhost / ::1) binds the gateway to localhost so the process-level
-    # binding matches the localhost intent; any widened bind exposes it on the
-    # LAN interface. The unconfigured-gateway flag is dropped: inject-token.js
-    # always writes the merged config to /tmp/openclaw-config.json
-    # (OPENCLAW_CONFIG_PATH), so the gateway is always configured at start.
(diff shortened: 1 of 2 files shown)


gabsprogrammer:
Closing this PR because OpenClaw is deprecated and this path is no longer worth changing in DreamServer.

PR #1618, opened 2026-06-27
Add PR-Agent (DeepSeek) + Qodo Merge dual review workflows

bong-water-water-bong:
Sets up two CI workflows for automated code review on every PR:\n\n- **PR-Agent** (DeepSeek) — auto-review, auto-describe, auto-improve with 4 code suggestions\n- **Qodo Merge** — second-pass review for broader coverage\n\nRequires two secrets to be set in the repo:\n- \`DEEPSEEK_API_KEY\`\n- \`QODO_API_KEY\`

### DIFF
--- a/.github/workflows/pr-agent-review.yml
+++ b/.github/workflows/pr-agent-review.yml
@@ -1,0 +1,30 @@
+name: pr-agent-review
+on:
+  pull_request:
+    types: [opened, synchronize, reopened, ready_for_review]
+  issue_comment:
+    types: [created]
+permissions:
+  contents: read
+  pull-requests: write
+  issues: write
+jobs:
+  pr_agent_job:
+    name: PR-Agent (DeepSeek)
+    runs-on: ubuntu-latest
+    if: ${{ github.event.sender.type != 'Bot' && secrets.DEEPSEEK_API_KEY != '' }}
+    steps:
+      - name: PR Agent review
+        uses: the-pr-agent/pr-agent@main
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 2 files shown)


Lightheartdevs:
Thanks for exploring more automated review coverage; given how many platform paths Dream Server has, extra review signal can be useful when it is tightly controlled.

What needs deciding:
- Decide whether the repo wants two additional always-on AI review bots on top of the existing `.github/workflows/claude-review.yml`. This affects PR noise, maintainer attention, cost, and contributor experience, so I would not merge it as a routine CI change without an explicit maintainer call.
- If the answer is yes, make the new workflows match the existing guardrails: skip or command-gate `issue_comment`, skip drafts/bots/forks, add concurrency, and avoid broad `issues: write` / `pull-requests: write` unless each workflow really needs it for the selected mode.
- Fix workflow validity before merge. Both new workflows use `if: ${{ secrets.* != '' }}` directly; GitHub's Actions docs say secrets cannot be directly referenced in `if:` conditionals. Follow the existing repo pattern of moving the secret check into an env-backed step gate.
- Pin the third-party actions to a released tag or SHA instead of `the-pr-agent/pr-agent@main` and `qodo-ai/qodo-merge@main`, especially because these jobs would receive write permissions.

What I checked:
- Diff/static review of `.github/workflows/pr-agent-review.yml` and `.github/workflows/qodo-merge.yml`.
- Compared against the current `.github/workflows/claude-review.yml` guardrails and GitHub Actions docs for secret conditionals.
- `git diff --check origin/main...refs/pr/1618` was clean; `gh pr checks 1618` reported no checks.

Thanks again for pushing on review automation; this just needs a deliberate maintainer policy choice before it becomes part of the default PR path.

PR #1620, opened 2026-06-27
chore: add Docker image updates to Dependabot config

bong-water-water-bong:
Adds \`docker\` ecosystem entries to the existing Dependabot config for all 8 core service Dockerfiles (comfyui, brave-search, dashboard, dashboard-api, privacy-shield, token-spy, ape, llama-sycl). These were the only missing ecosystem — GitHub Actions, npm (installer + dashboard), pip (4 services), and Cargo (Tauri) were already covered.

All Docker updates are grouped under a single \`docker-core\` group so they land as one PR per week instead of 8 individual PRs.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -100,3 +100,84 @@ updates:
       ape-python:
         patterns:
           - "*"
+
+  # ── Docker: Core services with custom Dockerfiles ─────────────────
+  - package-ecosystem: "docker"
+    directory: "/dream-server/extensions/services/comfyui"
+    schedule:
+      interval: "weekly"
+    open-pull-requests-limit: 1
+    groups:
+      docker-core:
+        patterns:
+          - "*"
+
+  - package-ecosystem: "docker"
+    directory: "/dream-server/extensions/services/brave-search"
+    schedule:
+      interval: "weekly"
+    open-pull-requests-limit: 1
+    groups:
+      docker-core:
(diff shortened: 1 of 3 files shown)


Lightheartdevs:
Thanks for filling in Docker image update coverage; keeping the custom service Dockerfiles on Dependabot is the right maintenance direction.

Quickest path to merge:
- Rebase or cherry-pick this so the PR only changes `.github/dependabot.yml`. The current diff also adds `.github/workflows/pr-agent-review.yml` and `.github/workflows/qodo-merge.yml`, which are unrelated review-bot workflow changes.
- Double-check the grouping expectation in `.github/dependabot.yml`. Each Docker entry is a separate directory with its own `groups: docker-core` block, so this may still produce one PR per Dockerfile directory rather than the single weekly `docker-core` PR described in the PR body. If one cross-directory Docker PR is required, use Dependabot's supported cross-directory grouping pattern or update the PR description/expectation.

What I checked:
- Diff/static review of `.github/dependabot.yml` against the current Dockerfile list under `dream-server/extensions/services` and `dream-server/images`.
- `git diff --check origin/main...refs/pr/1620` was clean; `gh pr checks 1620` reported no checks.

Thanks again for tightening the dependency maintenance surface.

bong-water-water-bong:
Superseded by clean branch — only the dependabot config, no unrelated workflow files.

PR #1621, opened 2026-06-27
fix: check is_symlink before is_file in dir_size_gb

### LINKED ISSUE #1590: dir_size_gb() follows symlinks before checking is_symlink()

bong-water-water-bong:
## Fixes #1590

\`Path.is_file()\` follows symlinks before \`is_symlink()\` gets to short-circuit. Flip the order so the cheap \`lstat\` check runs first, avoiding unnecessary stat calls on every symlink entry in user-controlled data directories.

**Change:** One line swap in \`helpers.py:dir_size_gb\` — \`if not f.is_symlink() and f.is_file():\`

### DIFF
--- a/.github/workflows/pr-agent-review.yml
+++ b/.github/workflows/pr-agent-review.yml
@@ -1,0 +1,30 @@
+name: pr-agent-review
+on:
+  pull_request:
+    types: [opened, synchronize, reopened, ready_for_review]
+  issue_comment:
+    types: [created]
+permissions:
+  contents: read
+  pull-requests: write
+  issues: write
+jobs:
+  pr_agent_job:
+    name: PR-Agent (DeepSeek)
+    runs-on: ubuntu-latest
+    if: ${{ github.event.sender.type != 'Bot' && secrets.DEEPSEEK_API_KEY != '' }}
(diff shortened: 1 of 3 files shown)


bong-water-water-bong:
Closing — @fireddd submitted this same fix first in #1586. Reviewing that one instead.

PR #1584, opened 2026-06-22
fix(dashboard-api): guard async HTTP client creation with asyncio.Lock

fireddd:
## Summary

- **Bug**: `_get_aio_session()` and `_get_httpx_client()` in `helpers.py` have a race condition where multiple concurrent coroutines can simultaneously observe the session as `None`/closed and each create a new client, leaking the earlier one (orphaned TCP connections, potential file-descriptor exhaustion).
- **How it manifests**: At dashboard startup, `get_all_services()` fans out ~24 concurrent health checks via `asyncio.gather()`. On first call, all 24 coroutines hit the lazy initializer before any session exists. Without synchronization, several can race past the `is None` check and each instantiate a new `aiohttp.ClientSession` / `httpx.AsyncClient`, with only the last one surviving in the module global.
- **Fix**: Add a module-level `asyncio.Lock` and apply double-checked locking to both functions. The fast path (session already exists and is open) returns immediately without acquiring the lock. The slow path (session needs creation) acquires the lock and re-checks before creating, ensuring exactly one session is ever active.

## Changes

`dream-server/extensions/services/dashboard-api/helpers.py`:
- Add `_session_lock = asyncio.Lock()` at module level
- Rewrite `_get_aio_session()` with double-checked locking pattern
- Rewrite `_get_httpx_client()` with double-checked locking pattern

## Test plan

- [x] Python syntax verified via `ast.parse()`
- [ ] Existing `pytest tests/` suite passes (CI will verify)
- [ ] Manual: confirm dashboard health polling works correctly on startup with concurrent service checks

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Closes #1588

PR #1622, opened 2026-06-27
fix: guard async HTTP client singleton creation with asyncio.Lock

### LINKED ISSUE #1588: Race condition in async HTTP client singleton creation
## Description

`_get_aio_session()` and `_get_httpx_client()` in `extensions/services/dashboard-api/helpers.py` have a TOCTOU race condition. Multiple concurrent coroutines can pass the `is None or is_closed` check simultaneously and each create a new session, leaking the earlier one(s).

## Steps to reproduce

1. Multiple concurrent health-check requests arrive at startup (e.g. dashboard loads multiple widgets simultaneously)
2. Each coroutine enters `_get_aio_session()`, sees `_aio_session is None`, and creates a new `aiohttp.ClientSession`
3. Only the last assignment wins; earlier sessions are orphaned with open TCP connections

## Expected behavior

Only one session should be created; concurrent callers should wait for the first creator to finish.

## Proposed fix

Guard creation with an `asyncio.Lock` using double-checked locking — fast path (session alive) skips the lock, slow path (creation) acquires it and re-checks.

See #1584

bong-water-water-bong:
## Fixes #1588

**Problem:** Multiple concurrent coroutines can pass the \`is None or is_closed\` check simultaneously and each create a new \`aiohttp.ClientSession\` / \`httpx.AsyncClient\`, leaking the earlier one(s).

**Fix:** Double-checked locking with \`asyncio.Lock\`. Fast path (session alive) skips the lock entirely. Slow path (creation) acquires it and re-checks to ensure only one creator wins.

**Changes:**
- \`_aio_session_lock\` and \`_httpx_client_lock\` module-level locks
- Both \`_get_aio_session()\` and \`_get_httpx_client()\` use async-with-lock + re-check pattern

### DIFF
--- a/.github/workflows/pr-agent-review.yml
+++ b/.github/workflows/pr-agent-review.yml
@@ -1,0 +1,30 @@
+name: pr-agent-review
+on:
+  pull_request:
+    types: [opened, synchronize, reopened, ready_for_review]
+  issue_comment:
+    types: [created]
+permissions:
+  contents: read
+  pull-requests: write
+  issues: write
+jobs:
+  pr_agent_job:
+    name: PR-Agent (DeepSeek)
+    runs-on: ubuntu-latest
+    if: ${{ github.event.sender.type != 'Bot' && secrets.DEEPSEEK_API_KEY != '' }}
+    steps:
+      - name: PR Agent review
+        uses: the-pr-agent/pr-agent@main
+        env:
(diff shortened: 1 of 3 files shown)


bong-water-water-bong:
Closing — @fireddd submitted this same fix first in #1584. Reviewing that one instead.

PR #1644, opened 2026-06-30
fix: honor bootstrap force for partial installs

Lightheartdevs:
## Summary
- parse `--force` and `--non-interactive` before the bootstrap incomplete-install prompt
- make `--force` remove incomplete install directories without waiting for input
- make noninteractive runs fail with a clear `--force` hint instead of hanging or default-aborting at a prompt

## Root cause
The release fleet DGX lane invokes the bootstrap with noninteractive force semantics, but `ods/get-ods.sh` checked an existing incomplete install directory before it had parsed those flags. On `dgx-gpu01`, `/home/nvidia/ods` existed without `.env`, so the bootstrap prompted and then aborted in the fleet run.

## Validation
- `bash -n ods/get-ods.sh`
- `bash -n ods/tests/contracts/test-installer-hardening.sh`
- `bash ods/tests/contracts/test-installer-hardening.sh`

### DIFF
--- a/ods/get-ods.sh
+++ b/ods/get-ods.sh
@@ -33,6 +33,15 @@ REPO_URL="${ODS_REPO_URL:-https://github.com/Light-Heart-Labs/ODS.git}"
 INSTALL_DIR="${ODS_INSTALL_DIR:-$ODS_BOOTSTRAP_ROOT/ods}"
 LEGACY_DREAMSERVER_DIR="${DREAMSERVER_INSTALL_DIR:-$ODS_BOOTSTRAP_ROOT/dream-server}"
 ODS_REF="${ODS_REF:-${ODS_BOOTSTRAP_REF:-}}"
+BOOTSTRAP_FORCE=false
+BOOTSTRAP_NON_INTERACTIVE=false
+
+for _arg in "$@"; do
+    case "$_arg" in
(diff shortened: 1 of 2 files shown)


Lightheartdevs:
Thanks for tightening the bootstrap recovery path here, especially for the release-fleet noninteractive force lane.

What looks good:
- `ods/get-ods.sh` now parses `--force` and `--non-interactive` before the existing-install check, so the incomplete-install branch can make the right decision before it would otherwise prompt.
- The force behavior stays scoped to incomplete install directories without `.env`; complete installs still keep the existing short-circuit at the installed-stack message.
- The noninteractive path now fails with an actionable `--force` hint instead of waiting for input or taking the default abort path.

What I checked:
- Reviewed the PR diff and surrounding bootstrap flow in `ods/get-ods.sh`, including the clone/install handoff to `./install.sh $@`.
- Reviewed the added contract guards in `ods/tests/contracts/test-installer-hardening.sh`.
- `git diff --check origin/main...HEAD` locally.
- GitHub checks are green for the PR head, including ShellCheck, Test Linux/integration-smoke, Matrix Smoke distro jobs, macOS smoke, env validation, dashboard frontend/API, Ruff, PowerShell lint, and secret scan.

PR #1650, opened 2026-06-30
fix(env): enforce minLength on secret env vars to reject placeholders

tang-vu:
## Summary

One concern: **make the `.env.schema.json` secret `minLength` constraint actually enforceable**, so placeholder secrets like `WEBUI_SECRET=CHANGEME` are rejected during env validation.

`scripts/validate-env.sh` validates required keys, unknown keys, types, enums, and numeric `minimum`/`maximum` — but it never reads string `minLength`. This adds the schema constraint **and** the validator support together, so the constraint isn't inert.

This is the "security" slice split out of #1605 per the maintainer feedback there (one PR, one concern) and rebased onto the new `ods/` layout.

## Changes

- **`ods/.env.schema.json`** — add `"minLength": 10` to the secret API-key / password / token / signing-secret fields (the set the installer generates at ≥10 chars). Usernames and other fields that can legitimately be short are intentionally left untouched.
- **`ods/scripts/validate-env.sh`** — validate string `minLength` for string-typed keys and report violations (new "Length validation errors" group), mirroring the existing range-validation style.
- **`ods/tests/test-validate-env.sh`** — use realistic-length values in the valid fixture and add a regression test: a required key present but too short → exit 2 with a `minLength` message.
- **`.github/workflows/validate-env.yml`** — set `N8N_USER=admin@ods.local` (the installer's phase-06 default) so the CI fixture satisfies the new constraint.

## Why it's safe for real installs

`installers/phases/06-directories.sh` generates `N8N_USER=admin@ods.local` and random secrets via `openssl rand` (hex 16–32 / base64 16–32), all comfortably ≥10 chars. Optional user-supplied API keys are empty by default and are skipped by the validator when unset.

## Testing

- `bash ods/tests/test-validate-env.sh` — **11/11** (including the new minLength regression).
- `bash ods/tests/smoke/installer-env-smoke.sh` — **12/12**: runs the real phase-06 env generation and validates the produced `.env` against the updated schema (✓ ".env validates against schema").
- ShellCheck clean (error + warning) on `validate-env.sh`; `bash -n` clean.

(Verified in a Linux container with `jq`.)


Lightheartdevs:
Thanks for splitting the env-schema hardening into this focused minLength slice; it is much easier to reason about against the installer-generated defaults.

What looks good:
- `ods/scripts/validate-env.sh` enforces `minLength` only after a value is present, so optional unset provider keys still skip validation while placeholders like `WEBUI_SECRET=CHANGEME` now fail cleanly.
- The schema changes line up with the required installer-generated secrets I checked in `ods/installers/phases/06-directories.sh`, the macOS/Windows env generators, and the v2.5.x release defaults for `N8N_USER`.
- The regression is covered in `ods/tests/test-validate-env.sh`, and `.github/workflows/validate-env.yml` updates the fixture to the same email-style n8n user the installers write.

What I checked:
- Static review of the four-file diff and surrounding env validation path.
- `git diff --check origin/main...origin/pr/1650` passed.
- Ubuntu 24.04 container: `bash -n ods/scripts/validate-env.sh ods/tests/test-validate-env.sh ods/tests/smoke/installer-env-smoke.sh`, `bash ods/tests/test-validate-env.sh` passed 11/11, and `bash ods/tests/smoke/installer-env-smoke.sh` passed 11/11. Compose syntax was skipped inside that container because Docker Compose was not available there.
- `gh pr checks` currently reports no checks for this head SHA, so a normal CI run is still worth watching when it appears.

PR #1655, opened 2026-07-02
fix(proxy): cap request body sizes and pre-stage HSTS header

tang-vu:
## Summary

One concern: **harden the LAN-facing proxy surfaces** — cap request body sizes on the Caddy proxies and pre-stage HSTS on the dashboard nginx.

This is the proxy slice split out of #1605 per the maintainer feedback there (one PR, one concern), re-based onto the `ods/` layout. Second slice after #1650.

## Changes

- **`ods/extensions/services/ods-proxy/Caddyfile`**
  - `chat.<device>.local`: `request_body max_size 200MB` — document uploads to Open WebUI still fit comfortably, but runaway or abusive clients can no longer stream unbounded bodies through the LLM proxy path.
  - `api.<device>.local`: `request_body max_size 50MB` for the admin API surface.
- **`ods/extensions/services/hermes-proxy/Caddyfile`** — `request_body max_size 50MB` on `:9120`, keeping room for file attachments in agent prompts. Placed before the `route` block, so the health/favicon/auth ordering contract is untouched.
- **`ods/extensions/services/dashboard/nginx.conf`** — pre-stage `Strict-Transport-Security "max-age=63072000; includeSubDomains"` alongside the existing security headers. Browsers ignore HSTS over plain HTTP, so it's inert today and activates automatically when TLS lands via Caddy/Tailscale.

17 insertions, no deletions — additive only.

## Testing

- `caddy validate --adapter caddyfile` — **Valid configuration** for both modified Caddyfiles (Caddy 2.6.2).
- `bash ods/tests/test-hermes-proxy-caddyfile.sh` — pass (route → health → forward_auth ordering and redirect-matcher contracts intact).
- `python3 ods/tests/test_mdns_subdomains.py` — OK (ods-proxy hostname set unchanged).
- `python3 ods/tests/contracts/test-network-exposure-contracts.py` — pass (CSP contract intact next to the new HSTS header).
- `bash ods/tests/contracts/test-installer-contracts.sh` — 37 passed (dashboard-nginx dynamic-upstream contracts intact).

PR #1656, opened 2026-07-02
fix(health): honor LLM_API_BASE_PATH in the LLM inference probe

tang-vu:
## Summary

One concern: **the LLM health probe fails on every AMD/Lemonade install** because `test_llm()` hardcodes `/v1/completions`, while Lemonade serves its OpenAI-compatible API under `/api/v1`.

`installers/phases/06-directories.sh` writes `LLM_API_BASE_PATH=/api/v1` for lemonade mode and `/v1` otherwise, but `scripts/health-check.sh` never reads it — so on Lemonade installs the probe hits a nonexistent endpoint, `test_llm` fails, and the stack is reported critical (exit 2) even when inference is healthy.

This is the health-check slice split out of #1605 (one PR, one concern), re-based onto `ods/`. Third slice after #1650 and #1655. The other health item from that bundle (Perplexica config via env vars in `12-health.sh`) is already fixed on `main`, so it's not included.

## Changes

- **`ods/scripts/health-check.sh`** — `test_llm()` reads `LLM_API_BASE_PATH` (default `/v1`) and probes `${base_path}/completions`. Same read pattern `ods-doctor.sh` already uses for the same variable; `.env` is loaded via `load_env_file` before the probe runs, so the value is present.
- **`ods/tests/test-health-check.sh`** — two regression assertions in the file's existing grep style: the probe must read `LLM_API_BASE_PATH` with a `/v1` default, and no hardcoded `/v1/completions` may remain.

## Testing

- **Functional, with a mock OpenAI-compatible server**: with `LLM_API_BASE_PATH=/api/v1` the probe requests `/api/v1/completions`; with the variable unset it requests `/v1/completions` (default behavior unchanged). Verified by recording the request paths server-side.
- `bash ods/tests/test-health-check.sh` — **11/11** (including the two new assertions).
- ShellCheck clean (warning + error) on both files; `bash -n` clean.

(Verified in a Linux container.)


Lightheartdevs:
﻿
Thanks for taking on the LLM health-check base path; this makes the probe line up with providers that do not expose completions under the default `/v1` prefix.

What looks good:
- `ods/scripts/health-check.sh` now builds the completions endpoint from `LLM_API_BASE_PATH`, preserving `/v1` as the default while honoring `/api/v1`-style configs.
- The added health-check test guards against reintroducing a hardcoded `/v1/completions` URL, and the existing env/install paths already carry `LLM_API_BASE_PATH`.

What I checked:
- Ran `tests/test-health-check.sh` in the PR worktree: 11/11 passed.
- Reviewed `.env.example`, phase 06 env generation, and the focused diff.
- `gh pr checks 1656` reports no checks for this branch, so this advisory is based on focused local testing and static review.

Thanks again for fixing the probe path.

PR #1658, opened 2026-07-02
fix(dashboard): show bytes for sub-1KB sizes instead of "0 KB"

jaydeep-pipaliya:
## Summary

`formatBytes` in `useDownloadProgress` has branches for GB, MB, and KB, but no branch for values under 1 KB. Any non-zero size below 1024 bytes falls into the KB branch and rounds to `"0 KB"` (e.g. `512` → `"0 KB"`), which can show early in a download before the first KB has transferred.

Fix adds a `kb >= 1` guard and a bytes fallback so `512` renders as `"512 B"`.

How I tested: extended the existing `formatBytes` test with the sub-KB case. Full dashboard test suite passes (124), eslint clean.

## AI Assistance

Claude Code helped spot the missing branch, draft the fix and test, and run the suite. I reviewed the diff and confirmed the outputs.

## Release Lane

## Changed Surface

- [x] Dashboard UI

## Risk And Validation

- Risk level: Low
- Validation run:
  - [x] `git diff --check`

Commands/results:

```text
$ npx vitest run src/hooks/__tests__/useDownloadProgress.test.js
6 passed

$ npx vitest run            # full dashboard suite
Test Files 20 passed | Tests 124 passed

$ npm run lint              # eslint src
clean
```

## Operational Change Check

## Notes For Reviewers

Before: `formatBytes(512)` → `"0 KB"`. After: `"512 B"`. Existing outputs are unchanged (`5000` → `"5 KB"`, `5e6` → `"4.8 MB"`, `5e9` → `"4.66 GB"`, `0`/`null` → `"0 B"`).

### DIFF
--- a/ods/extensions/services/dashboard/src/hooks/__tests__/useDownloadProgress.test.js
+++ b/ods/extensions/services/dashboard/src/hooks/__tests__/useDownloadProgress.test.js
@@ -122,6 +122,7 @@ describe('useDownloadProgress', () => {
     expect(result.current.formatBytes(5e9)).toBe('4.66 GB')
     expect(result.current.formatBytes(5e6)).toBe('4.8 MB')
     expect(result.current.formatBytes(5000)).toBe('5 KB')
+    expect(result.current.formatBytes(512)).toBe('512 B')
     expect(result.current.formatBytes(0)).toBe('0 B')
     expect(result.current.formatBytes(null)).toBe('0 B')
   })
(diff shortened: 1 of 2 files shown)


Lightheartdevs:
﻿
Thanks for tightening the dashboard download-size display; this is a small fix, but it removes a confusing `0 KB` state for active sub-1KB transfers.

What looks good:
- `ods/extensions/services/dashboard/src/hooks/useDownloadProgress.js` keeps the existing GB/MB/KB formatting while showing byte values below 1 KB instead of rounding to `0 KB`.
- The added `formatBytes(512) -> "512 B"` test covers the regression directly without changing the hook contract.

What I checked:
- Reviewed the PR metadata, diff, and surrounding hook/test code.
- `gh pr checks 1658` reports the visible dashboard frontend/API, lint, env-schema, Linux integration-smoke, and matrix-smoke checks passing.

PR #1663, opened 2026-07-04
feat(dashboard-api): add /api/gpu/idle endpoint and is_gpu_idle helper

Arifuzzamanjoy:
Adds a read-only GET /api/gpu/idle reporting whether the GPU is idle,
based on aggregate utilization vs a configurable GPU_IDLE_THRESHOLD_PERCENT
(default 10, clamped 0–100). Adds a pure is_gpu_idle() helper and typed
GpuIdleStatus model.

Why it matters: a lightweight idle signal is a common need for anything that
places or balances work across GPUs. Callers can poll one cheap endpoint for
capacity and scheduling decisions instead of parsing the full /api/gpu/detailed
payload each time.

Derived from the same data as /api/gpu/detailed (reuses the cached
aggregate -- no second detection path). Utilization-less backends (Apple,
AMD host runtime) return idle: null rather than a misleading false.
No new dependencies.

Validation: ruff clean, git diff --check clean, 81 gpu tests pass, full
dashboard-api suite 1190 passed. Read-only and auth-gated --- not an
operational change.

### DIFF
--- a/ods/extensions/services/dashboard-api/gpu.py
+++ b/ods/extensions/services/dashboard-api/gpu.py
@@ -380,6 +380,18 @@ def get_gpu_info() -> Optional[GPUInfo]:
     return None
 
 
+def is_gpu_idle(info: GPUInfo, threshold_percent: int = 10) -> Optional[bool]:
+    """Return True if aggregate GPU utilization is at or below threshold_percent.
+
+    Pure function — no I/O. `info` is an aggregate GPUInfo (see get_gpu_info).
+    Note: backends that cannot report utilization (Apple, AMD host runtime)
+    return None.
+    """
+    if info.gpu_backend == "apple" or "host runtime" in info.name.lower():
+        return None
+    return info.utilization_percent <= threshold_percent
+
+
 # ============================================================================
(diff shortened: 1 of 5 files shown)


Lightheartdevs:
Thanks for adding the GPU idle signal on top of the existing detailed GPU path; this is a useful small surface for scheduling and diagnostics.

What looks good:
- `ods/extensions/services/dashboard-api/routers/gpu.py` derives `/api/gpu/idle` from `gpu_detailed()` instead of adding a second GPU probe path, so it keeps the existing cache and no-GPU 503 behavior.
- `ods/extensions/services/dashboard-api/gpu.py` keeps utilization-less backends explicit by returning `null` for Apple and AMD host-runtime cases instead of treating placeholder `0` utilization as definitely idle.
- The tests cover threshold boundaries, env clamping, auth, no-GPU propagation, and the unknown-backend/null case.

What I checked:
- Static review of `gpu.py`, `routers/gpu.py`, `models.py`, `tests/test_gpu.py`, and `tests/test_gpu_detailed.py`.
- `git diff --check origin/main...pr-1663` passed.

Thanks again for keeping the change narrow.

PR #1664, opened 2026-07-04
feat(dashboard-api): add /api/node/capabilities diagnostics endpoint

Arifuzzamanjoy:
Adds a read-only GET /api/node/capabilities that returns a one-call node
snapshot -- GPU type/VRAM, loaded model, per-service health, and installed
ODS version -- for diagnostics and support bundles.

Why it matters: capability discovery is a natural building block for any
multi-node ODS deployment. A single authenticated read describes what a node
is and what it's currently running, so fleet dashboards, support tooling, and
higher-level coordination can query one endpoint instead of stitching the GPU,
model, and health calls together.

Reuses existing helpers (get_gpu_info, get_loaded_model, get_all_services);
sync GPU call thread-offloaded, async helpers gathered. Wired via a single
include_router line. No schema/compose/dependency changes.

Validation: ruff clean, git diff --check clean, 10 new tests, full
dashboard-api suite 1188 passed. Read-only and auth-gated -- not an
operational change.

### DIFF
--- a/ods/extensions/services/dashboard-api/main.py
+++ b/ods/extensions/services/dashboard-api/main.py
@@ -63,6 +63,7 @@ from routers import (
     talk,
     tailscale,
     usage,
+    node,
 )
 from settings import (
     _ENV_ASSIGNMENT_RE, _ENV_COMMENTED_ASSIGNMENT_RE, _SETTINGS_APPLY_ALLOWED_SERVICES, _parse_env_text, _read_env_map_from_path,
@@ -1014,6 +1015,7 @@ app.include_router(oauth_passthrough.router)
 app.include_router(talk.router)
 app.include_router(tailscale.router)
 app.include_router(usage.router)
+app.include_router(node.router)
 
 
 # ================================================================
--- a/ods/extensions/services/dashboard-api/models.py
+++ b/ods/extensions/services/dashboard-api/models.py
@@ -28,6 +28,15 @@ class ServiceStatus(BaseModel):
     response_time_ms: Optional[float] = None
 
 
+class NodeCapabilities(BaseModel):
+    ods_version: str
+    gpu: Optional[GPUInfo] = None
+    loaded_model: Optional[str] = None
+    services: list[ServiceStatus] = []
+    service_count: int = 0
+    running_service_count: int = 0
+
+
 class DiskUsage(BaseModel):
     path: str
     used_gb: float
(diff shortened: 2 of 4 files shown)


Lightheartdevs:
Thanks for pulling node diagnostics into one authenticated read; the endpoint shape fits the support-bundle and fleet-discovery direction well.

Quickest path to merge:
- In `ods/extensions/services/dashboard-api/routers/node.py`, include the existing `degraded` service state in `_RUNNING_STATUSES`, or rename the field if degraded services should intentionally be excluded. `helpers.check_service_health()` returns `degraded` when a service is reachable but slow, so the current payload can list a running-but-slow service under `services` while leaving it out of `running_service_count`.
- Add a focused regression in `ods/extensions/services/dashboard-api/tests/test_node.py` with a `ServiceStatus(status=degraded)` so this count stays aligned with the dashboard-api health model and the dashboard's ServiceMap/Settings treatment of degraded services.

What I checked:
- Static review of `main.py`, `models.py`, `routers/node.py`, and `tests/test_node.py` against the current dashboard-api helpers.
- `git diff --check origin/main...pr-1664` passed.

Thanks again for pushing this diagnostic surface forward.

Arifuzzamanjoy:
pushed the degraded fix — degraded now counts toward running_service_count. regression added, ci green.

Lightheartdevs:
Thanks for tightening the node capabilities endpoint after the degraded-service review note; this now fits the dashboard-api health model cleanly.

What looks good:
- `ods/extensions/services/dashboard-api/routers/node.py` keeps the endpoint read-only and auth-gated, and it reuses the existing GPU/model/service helpers instead of introducing a parallel diagnostics path.
- The new `running_service_count` logic now treats `degraded` as up, matching `helpers.check_service_health()` where degraded means reachable but slow rather than absent.
- `ods/extensions/services/dashboard-api/tests/test_node.py` covers auth, missing GPU/model data, version fallback/precedence, quoted `.env` values, and the degraded-service regression from the prior review.

What I checked:
- Static review of the current diff against `main`, including `models.py`, `main.py`, `routers/node.py`, `settings._read_env_map_from_path`, and the service-health helpers.
- `python -m pytest tests/test_node.py -q` from `ods/extensions/services/dashboard-api`: 11 passed.
- GitHub checks for this head are green, including dashboard API/frontend, Ruff, mypy, shell/PowerShell lint, env validation, Linux integration smoke, and distro smoke.
- `git diff --check origin/main...origin/pr/1664` passed.

Thanks again for pushing this diagnostic surface forward.

Lightheartdevs (approved):

PR #1668, opened 2026-07-04
fix(installer): guard compose image pulls

### LINKED ISSUE #1652: Phase 5 Stack Launch Fails on Dynamic Sub-Dependency Image Pull Due to TLS Handshake Timeout
### Description

Following a successful completion of the Phase 4 download module sequence , the installation terminates unexpectedly in Phase 5 during container runtime initialization (Waking the stack...).  The installer script assumes all required project assets are cached locally after Phase 4. However, the orchestrator triggers an unhandled mid-flight dynamic pull for an underlying sub-dependency layer during docker compose up. Because this phase lacks the retry-backoff mechanisms implemented in Phase 4, any transient network degradation or latency spike over the connection causes a fatal network timeout that rolls back the entire setup sequence. 

### Steps to Reproduce

1. Run ./install.sh and select Option [1] Full Stack.  
2. Allow the Phase 4 verification loop to complete until all 10 core modules report a downloaded status.  
3. Advance into Phase 5 (Starting Services).  
4. Introduce a transient throughput drop or latency spike while the Docker engine initializes secondary composition blocks (e.g., pulling unverified extension sub-layers or worker streams).  

### Expected Behavior

The deployment engine should explicitly audit and cache all multi-container image dependencies during the Phase 4 validation loop before moving to orchestration steps. If a pull is triggered during Phase 5, it must be protected by the same failure-resilient loop pacing to prevent sudden deployment drops under constrained mobile network interfaces.  

### Actual Behavior

The composition sequence crashes mid-flight during automated background pulls. The Docker engine daemon returns a hard network timeout when resolving blob streaming routes via the Content Delivery Network:

failed to copy: httpReadSeeker: failed open: failed to do request: Get "https://production.cloudfront.docker.com/registry-v2/docker/registry/v2/blobs/...": net/http: TLS handshake timeout

Consequently, zero managed containers are constructed , the ods-network software-defined bridge fails to expose stable ranges, and the installer throws an immediate warning flag before exiting to the shell:

⚠ UFW is active, but I could not detect the ods-network subnet for ods-host-agent access.
✗ Docker Compose did not create any managed containers.

### Operating System

Ubuntu 26.04 LTS

### GPU

NVIDIA GeForce 940MX (2048MB VRAM)

### Docker Version

Docker version 29.6.1, build 8900f1d

### VRAM

2GB

### Logs

```shell
[INFO] Saved compose launch record to /home/tarun/ods/logs/compose-launch.txt
  ⠧ [00:07] Launching containers (attempt 1/3)... 
  ⚠ Some services still starting...                             
  ⚠ Some containers need more time. Waiting 30s before retry...
  ⠧ [00:07] Launching containers (attempt 2/3)... 
  ⚠ Some services still starting...                             
  ⚠ Some containers need more time. Waiting 30s before retry...
  ⠧ [00:07] Launching containers (attempt 3/3)... 
  ⠹ [00:12] Launching containers (attempt 3/3)...   ✓ Restarted ods-host-agent after ods-network creation
  ⚠ UFW is active, but I could not detect the ods-network subnet for ods-host-agent access.
  ✗ Some containers failed to launch                            
  ✗ Docker Compose did not create any managed containers.
```

### Screenshots

[install-report-2026-07-01-204225.txt](https://github.com/user-attachments/files/29557340/install-report-2026-07-01-204225.txt)

gabsprogrammer:
## Summary

Fixes #1652.

Phase 5 could still let Docker Compose pull service images during `docker compose up`, even after Phase 4 had completed its download sequence. The attached `install-report-2026-07-01-204225.txt` shows this exact failure mode: a CloudFront/Docker Hub TLS handshake timeout happened during stack launch, then Compose exited before creating any managed containers.

This PR moves registry access back into retry-protected installer code and makes Phase 5 deterministic.

## Root Cause

| Area | Problem |
| --- | --- |
| Phase 4 image list | `PULL_LIST` was hand-maintained, so it could miss images from the final resolved Compose stack or newly enabled extensions. |
| Pull failure handling | Phase 4 warned that failed pulls would be attempted again during service startup, which handed the failure back to raw Compose behavior. |
| Phase 5 startup | `docker compose up -d --remove-orphans --no-build` allowed implicit image pulls with no ODS retry/backoff/reporting wrapper. |
| Failure result | A transient registry timeout could abort launch before any containers existed, producing the zero-managed-containers failure in the report. |

## What Changed

- Added `installers/lib/compose-images.sh` to resolve external images from the final Docker Compose config.
- Skips local ODS build outputs and services with `build:` so the installer does not try to pull images it is supposed to build locally.
- Phase 4 now augments `PULL_LIST` with images discovered from the resolved Compose stack.
- Phase 4 now fails early if required image downloads fail, instead of deferring the problem to service startup.
- Phase 5 now performs a final retry-protected image cache preflight before launch.
- Phase 5 now runs Compose with `--pull never`, so `compose up` cannot perform hidden, unbounded registry pulls.
- Launch records and failure reports now show the real guarded command: `up -d --remove-orphans --no-build --pull never`.

## Future-Proofing

This is not limited to the specific image in the issue report. Any future extension or overlay that adds an external `image:` to the resolved Compose stack is discovered automatically and pulled through `pull_with_progress` before launch.

Local build services stay local: dashboard, dashboard-api, APE, Token Spy, Privacy Shield, ComfyUI builds, and other ODS-generated build tags are excluded from remote pull discovery.

## Verification

- [x] `bash -n ods/installers/lib/compose-images.sh ods/installers/phases/08-images.sh ods/installers/phases/11-services.sh ods/tests/test-compose-image-discovery.sh ods/tests/test-docker-image-pull-retry.sh ods/tests/test-compose-failure-report.sh ods/tests/test-ods-doctor-install-diagnoses.sh`
- [x] `bash ods/tests/test-compose-image-discovery.sh`
- [x] `bash ods/tests/test-docker-image-pull-retry.sh`
- [x] `bash ods/tests/test-compose-failure-report.sh`
- [x] `bash ods/tests/test-ods-doctor-install-diagnoses.sh`
- [x] `bash ods/tests/contracts/test-installer-hardening.sh`
- [x] `git diff --check`

## Notes For Reviewers

The key behavior change is intentional: if a required registry pull fails after retry/backoff, the installer now stops before Phase 5 instead of starting a partial/fragile compose launch. Re-running `./install.sh` resumes through the same retry-protected path.


Lightheartdevs:
﻿
Thanks for hardening the Linux installer compose image pulls; pre-pulling external images before startup fits the repo's current direction around clearer installer failures.

What looks good:
- `ods/installers/lib/compose-images.sh` resolves runtime images from the composed config while filtering local build tags, so Phase 08/11 can verify external pulls before `docker compose up`.
- Phase 11 switches startup to `--pull never` after the preflight pull step, making the eventual compose launch more deterministic and preserving the failure-report path.
- The focused tests cover image discovery and pull retry behavior, including the local-build exclusion case.

What I checked:
- Ran `tests/test-compose-image-discovery.sh` in the PR worktree: 6/6 passed.
- Ran `tests/test-docker-image-pull-retry.sh` in the PR worktree: 13/13 passed.
- `gh pr checks 1668` reports the visible lint, API/frontend, integration-smoke, and distro/matrix checks passing.

Thanks again for tightening the installer startup path.

PR #1671, opened 2026-07-05
fix(host-agent): tighten exception handling for yaml parsing

HimanshuShukla671:
## Description
This PR fixes an overly broad exception handler in `ods-host-agent.py`'s `_precreate_extension_dirs` function.

Previously, the `yaml.safe_load` block caught a generic `Exception`, which could inadvertently mask unrelated errors like `NameError` or `AttributeError` if future logic changes were made in that block. This tightens the exception catching to explicitly handle only `OSError` (for file read issues) and `yaml.YAMLError` (for invalid YAML structure), letting other unexpected bugs crash loudly as intended.

### DIFF
--- a/ods/bin/ods-host-agent.py
+++ b/ods/bin/ods-host-agent.py
@@ -431,7 +431,7 @@ def _precreate_data_dirs(service_id: str):
         # PyYAML not available — skip pre-creation
         logger.debug("PyYAML not available, skipping data dir pre-creation for %s", service_id)
         return
-    except Exception as e:
+    except (OSError, yaml.YAMLError) as e:
         logger.debug("Failed to parse compose.yaml for %s: %s", service_id, e)
         return
     if not isinstance(data, dict):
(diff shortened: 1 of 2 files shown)


Lightheartdevs:
Thanks for tightening the host-agent YAML parsing path to match the repo's narrow-exception policy.

Quickest path to merge:
- Drop the unrelated `ods/extensions/services/dashboard-api/routers/magic_link.py` commit from this branch, or explicitly base this PR on #1670 and call that dependency out. As opened, a host-agent exception-handling PR also carries the full magic-link rate-limit change, which makes the review surface larger and overlaps another open PR.
- Add a focused regression in `ods/extensions/services/dashboard-api/tests/test_host_agent.py` for `_precreate_data_dirs()` with bad compose parsing. If malformed user compose files should be skipped rather than crash precreation, include `UnicodeDecodeError` in the narrow catch alongside `OSError` and `yaml.YAMLError`; if they should crash, document that with the test.

What I checked:
- Static review of `ods/bin/ods-host-agent.py`, the bundled `magic_link.py` diff, and the existing precreate tests in `tests/test_host_agent.py`.
- `git diff --check origin/main...pr-1671` passed.

Thanks again for taking on this cleanup.

PR #1672, opened 2026-07-06
feat(installer): add llmfit model advisor with fallback to tier map

### LINKED ISSUE #226: feat: integrate llmfit for better hardware detection and model selection
### Description

Integrate [llmfit](https://github.com/AlexsJones/llmfit) (MIT-licensed, Rust) into DreamServer's installer pipeline to replace the current hardcoded tier-based model selection with dynamic, hardware-aware model recommendations.

DreamServer currently uses a 3-stage bash/python pipeline (`detect-hardware.sh` → `classify-hardware.sh` → `build-capability-profile.sh`) that maps hardware to fixed tiers (T1–T4, SH_LARGE/SH_COMPACT, AP_BASE/PRO/ULTRA), each with one hardcoded model. This works for common setups but doesn't scale — it can't evaluate quantization trade-offs, doesn't estimate speed, and requires manual updates when new models are released.

llmfit detects hardware (NVIDIA multi-GPU, AMD, Intel Arc, Apple Silicon, Ascend NPU), scores 200+ models across Quality/Speed/Fit/Context dimensions, dynamically selects the best quantization that fits, and estimates token/sec throughput — all in a single cross-platform binary.

| Capability | DreamServer today | llmfit |
|---|---|---|
| GPU detection | NVIDIA, AMD sysfs, Apple | NVIDIA (multi-GPU), AMD, Intel Arc, Apple Silicon, Ascend NPU |
| Model database | ~8 hardcoded models | 206 models from HuggingFace |
| Quantization | Fixed GGUF per tier | Dynamic — picks highest quality that fits (Q8_0 → Q2_K) |
| MoE support | Basic tier split | Full expert offloading with reduced VRAM calculation |
| Speed estimation | None | Bandwidth-based token/sec (~80 GPU profiles) |
| Scoring | VRAM thresholds only | Multi-dimensional with use-case weights |
| Runtime providers | Docker compose | Ollama, llama.cpp, MLX, Docker Model Runner |

---

### Integration Options

**Option A: Use llmfit as a pre-install advisor (recommended)**
- Run `llmfit --json` during install to get the best model recommendation
- Replace hardcoded tier map with llmfit's dynamic scoring
- Single binary (~5MB), cross-platform (Linux, macOS, Windows)
- Could ship as an optional dependency or vendor the binary in releases

**Option B: Use llmfit's model database only**
- Import `hf_models.json` (206 models with parameter counts, quantization sizes, categories)
- Keep DreamServer's own detection but use richer model data for selection
- No binary dependency — just a JSON file updated periodically

**Option C: Use llmfit's REST API at runtime**
- `llmfit serve` exposes a REST API for model recommendations
- Could query it dynamically for upgrade recommendations or model switching

### Key Benefits

1. **Better model selection for edge cases** — 24GB VRAM, unusual AMD cards, multi-GPU setups
2. **Dynamic quantization** — instead of "you get this one GGUF", pick the best quality that fits
3. **Speed estimates** — users see expected token/sec before committing
4. **Automatic model database updates** — llmfit scrapes HuggingFace, so new models are picked up without code changes
5. **MoE-aware recommendations** — properly handles Mixtral, DeepSeek-V3, Qwen3 MoE models
6. **Reduced maintenance** — offload hardware detection and model scoring to a maintained upstream tool

### Use Case

DreamServer is an open-source, self-hosted AI server that auto-installs on Linux, macOS, and Windows. During installation, it must detect the user's GPU (NVIDIA/AMD/Apple Silicon), measure available VRAM/RAM, and automatically select the best LLM model and Docker Compose stack — with zero user input.

Today we solve this with a custom pipeline that maps hardware to a fixed tier and picks one hardcoded model per tier. This breaks down for edge cases: unusual VRAM sizes (e.g., 24GB falls awkwardly between tiers), multi-GPU rigs, MoE models that could fit but our tier map doesn't consider, and new models that require manual updates to our tier config.

**llmfit solves exactly the problems we hit:**

- **Dynamic model selection** — instead of maintaining a static tier→model map, llmfit scores 200+ models against actual hardware and picks the best fit with the best quantization.
- **Better hardware coverage** — multi-GPU setups, Intel Arc, and Ascend NPUs that we don't currently handle.
- **Speed estimation** — our installer has no way to tell users how fast inference will be; llmfit provides bandwidth-based token/sec estimates.
- **MoE-aware fitting** — we hardcode tiers for MoE models, but llmfit dynamically calculates expert offloading for any MoE architecture.
- **Reduced maintenance** — llmfit's model database is scraped from HuggingFace automatically, eliminating manual tier config updates when new models ship.

---

### Proposed Solution

**Integrate `llmfit --json` as the model recommendation engine** in our installer pipeline, called between hardware detection and capability profile generation.

### Integration approach

1. Ship the `llmfit` binary alongside DreamServer (or install it during setup via `cargo install llmfit` / prebuilt release binary)
2. Call `llmfit --json --category general` during install to get scored model recommendations
3. Parse the JSON output to extract the top-ranked model, its quantization, and estimated speed
4. Feed the result into `build-capability-profile.sh` to generate `.capabilities.json`
5. Use the recommended model to configure llama-server in the Docker Compose stack

### What changes in DreamServer

- `resolve_tier_config()` in `scripts/resolve-tier-config.sh` would delegate to llmfit instead of the static tier map
- `classify-hardware.sh` could be simplified or removed (llmfit handles classification internally)
- `detect-hardware.sh` could remain as a fallback, or be replaced entirely by llmfit's detection
- `config/gpu-database.json` would no longer need manual maintenance

### What stays the same

- The capability profile schema (`.capabilities.json`)
- Docker Compose overlay selection (amd/nvidia/apple)
- The installer flow (`install-core.sh` → preflight → detect → configure → compose up)

### Alternatives Considered

1. **Keep extending the current tier map** — Add more tiers and models manually. Simple but doesn't scale; every new model or GPU requires code changes. No quantization flexibility or speed estimation.

2. **Build our own model database from HuggingFace** — Write a scraper to build `hf_models.json` ourselves and add scoring logic. Duplicates work that llmfit already does well. Higher maintenance burden.

3. **Use llmfit's model database only (JSON import)** — Import `data/hf_models.json` (206 models) without the binary. Keeps our own detection but gains model coverage. Less integration effort, but we'd still need to build scoring/quantization logic ourselves.

4. **Use llmfit's REST API at runtime** — Run `llmfit serve` as a sidecar service for dynamic model recommendations. More complex operationally, but enables runtime model switching and upgrade suggestions.

We recommend **Option 1 (binary integration via `--json`)** as the best balance of capability, simplicity, and maintenance. llmfit is MIT-licensed (same as DreamServer), actively maintained (61 releases), cross-platform, and already covers all our target platforms.

IronicRayquaza:
## Summary

Adds llmfit as an optional advisor layer in the hardware detection
pipeline. When the llmfit binary is present and returns a recommendation
that maps to a known ODS GGUF, the installer can use that instead of
the static tier map. Falls back silently to the existing tier map if
llmfit is absent, times out, or returns an unmapped model.

This is Option B from the issue: JSON model-map parity proof without
making installs depend on a new external binary unconditionally.

Closes #226

## Changes

- `ods/config/llmfit-model-map.json` — maps llmfit HuggingFace model
  IDs to ODS GGUF catalog entries per hardware class, pinned to
  llmfit v0.4.2
- `ods/scripts/llmfit-advisor.sh` — pure advisor, fallback-safe,
  never writes to disk or .env
- `ods/tests/fixtures/llmfit/` — 5 representative JSON output
  fixtures: NVIDIA 24GB, NVIDIA 8GB, AMD Strix Halo, Apple M3 16GB,
  CPU-only
- `ods/tests/test-llmfit-advisor.sh` — fixture-based parity tests
  plus resilience tests for malformed JSON, empty recommendations,
  and unmapped models

## What is NOT changed

- Existing tier map, capability profile schema, compose overlay
  selection — all unchanged
- llmfit is never a hard dependency — absent binary is a silent
  no-op, install proceeds normally
- No changes to any installer phase scripts in this PR — advisor
  integration into resolve-tier-config.sh is a follow-up once
  parity is confirmed

## Testing

- [ ] `bash -n` passes on all new .sh files
- [ ] `shellcheck -x -S warning` passes on all new .sh files
- [ ] `bash ods/tests/test-llmfit-advisor.sh` — all assertions pass
- [ ] `bash ods/tests/test-generated-config-contracts.sh` passes
- [ ] `bash ods/tests/test-golden-paths.sh` passes
- [ ] `python3 -c "import json; json.load(open(...))"` — model map
  JSON is valid
  
## Notes For Reviewers

Note: the current model map covers GGUF filename and tier mapping. URL, checksum, and context window fields are intentionally deferred to the follow-up PR that integrates the advisor into resolve-tier-config.sh, where those values are already managed by the existing tier config.

### DIFF
--- a/ods/config/llmfit-model-map.json
+++ b/ods/config/llmfit-model-map.json
@@ -1,41 +1,29 @@
 {
-  "_comment": "Maps llmfit HuggingFace model IDs to ODS GGUF catalog entries. Update when new models are added to ODS tier map. Pin llmfit version: 0.4.2",
-  "_llmfit_version": "0.4.2",
+  "_comment": "ods_hardware_classes intentionally omitted — hardware-class enforcement deferred to integration PR that wires advisor into classify-hardware.sh",
   "mappings": [
     {
-      "llmfit_model_id": "Qwen/Qwen3-0.6B-GGUF",
+      "llmfit_model_id": "Qwen/Qwen3.5-2B-GGUF",
       "llmfit_quant": "Q4_K_M",
-      "ods_gguf_file": "Qwen3-0.6B-Q4_K_M.gguf",
-      "ods_tier": "T0",
-      "ods_hardware_classes": ["cpu_fallback"]
+      "ods_gguf_file": "Qwen3.5-2B-Q4_K_M.gguf",
+      "ods_tier": "T1"
     },
     {
-      "llmfit_model_id": "Qwen/Qwen3-1.7B-GGUF",
+      "llmfit_model_id": "Qwen/Qwen3.5-9B-GGUF",
       "llmfit_quant": "Q4_K_M",
-      "ods_gguf_file": "Qwen3-1.7B-Q4_K_M.gguf",
-      "ods_tier": "T1",
-      "ods_hardware_classes": ["cpu_fallback", "apple_silicon"]
+      "ods_gguf_file": "Qwen3.5-9B-Q4_K_M.gguf",
+      "ods_tier": "T2"
     },
     {
-      "llmfit_model_id": "Qwen/Qwen3-8B-GGUF",
+      "llmfit_model_id": "Qwen/Qwen3-30B-A3B-GGUF",
       "llmfit_quant": "Q4_K_M",
-      "ods_gguf_file": "Qwen3-8B-Q4_K_M.gguf",
-      "ods_tier": "T2",
-      "ods_hardware_classes": ["nvidia_pro", "apple_silicon", "strix_unified"]
+      "ods_gguf_file": "Qwen3-30B-A3B-Q4_K_M.gguf",
+      "ods_tier": "T3"
     },
     {
-      "llmfit_model_id": "Qwen/Qwen3-14B-GGUF",
+      "llmfit_model_id": "Qwen/Qwen3.6-35B-A3B-GGUF",
       "llmfit_quant": "Q4_K_M",
-      "ods_gguf_file": "Qwen3-14B-Q4_K_M.gguf",
-      "ods_tier": "T3",
-      "ods_hardware_classes": ["nvidia_pro", "strix_unified"]
-    },
-    {
-      "llmfit_model_id": "Qwen/Qwen3-32B-GGUF",
-      "llmfit_quant": "Q4_K_M",
-      "ods_gguf_file": "Qwen3-32B-Q4_K_M.gguf",
-      "ods_tier": "T4",
-      "ods_hardware_classes": ["nvidia_pro", "strix_unified"]
+      "ods_gguf_file": "Qwen3.6-35B-A3B-UD-Q4_K_M.gguf",
+      "ods_tier": "T4"
     }
   ]
 }
--- a/ods/config/llmfit-model-map.json
+++ b/ods/config/llmfit-model-map.json
@@ -1,0 +1,41 @@
+{
+  "_comment": "Maps llmfit HuggingFace model IDs to ODS GGUF catalog entries. Update when new models are added to ODS tier map. Pin llmfit version: 0.4.2",
+  "_llmfit_version": "0.4.2",
+  "mappings": [
+    {
+      "llmfit_model_id": "Qwen/Qwen3-0.6B-GGUF",
+      "llmfit_quant": "Q4_K_M",
+      "ods_gguf_file": "Qwen3-0.6B-Q4_K_M.gguf",
+      "ods_tier": "T0",
+      "ods_hardware_classes": ["cpu_fallback"]
+    },
+    {
+      "llmfit_model_id": "Qwen/Qwen3-1.7B-GGUF",
+      "llmfit_quant": "Q4_K_M",
+      "ods_gguf_file": "Qwen3-1.7B-Q4_K_M.gguf",
+      "ods_tier": "T1",
+      "ods_hardware_classes": ["cpu_fallback", "apple_silicon"]
+    },
+    {
+      "llmfit_model_id": "Qwen/Qwen3-8B-GGUF",
+      "llmfit_quant": "Q4_K_M",
+      "ods_gguf_file": "Qwen3-8B-Q4_K_M.gguf",
+      "ods_tier": "T2",
+      "ods_hardware_classes": ["nvidia_pro", "apple_silicon", "strix_unified"]
+    },
+    {
+      "llmfit_model_id": "Qwen/Qwen3-14B-GGUF",
+      "llmfit_quant": "Q4_K_M",
+      "ods_gguf_file": "Qwen3-14B-Q4_K_M.gguf",
+      "ods_tier": "T3",
+      "ods_hardware_classes": ["nvidia_pro", "strix_unified"]
+    },
+    {
+      "llmfit_model_id": "Qwen/Qwen3-32B-GGUF",
+      "llmfit_quant": "Q4_K_M",
+      "ods_gguf_file": "Qwen3-32B-Q4_K_M.gguf",
+      "ods_tier": "T4",
+      "ods_hardware_classes": ["nvidia_pro", "strix_unified"]
+    }
+  ]
+}
--- a/ods/scripts/llmfit-advisor.sh
+++ b/ods/scripts/llmfit-advisor.sh
@@ -1,0 +1,136 @@
+#!/bin/bash
+# ============================================================================
+# ODS Installer — llmfit Model Advisor
+# ============================================================================
+# Part of: scripts/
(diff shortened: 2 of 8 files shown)


Lightheartdevs:
Thanks for exploring llmfit as a fallback-safe advisor layer; keeping it pure and out of `.env` until integration is the right shape for this installer surface.

Quickest path to merge:
- Align `ods/config/llmfit-model-map.json` with the current ODS model catalog before treating this as a parity proof. Every mapped `ods_gguf_file` in this PR is absent from `ods/config/model-library.json`: `Qwen3-0.6B-Q4_K_M.gguf`, `Qwen3-1.7B-Q4_K_M.gguf`, `Qwen3-8B-Q4_K_M.gguf`, `Qwen3-14B-Q4_K_M.gguf`, and `Qwen3-32B-Q4_K_M.gguf`. The current qwen tier map/catalog uses entries like `Qwen3.5-2B-Q4_K_M.gguf`, `Qwen3.5-9B-Q4_K_M.gguf`, `Qwen3-30B-A3B-Q4_K_M.gguf`, and `Qwen3.6-35B-A3B-UD-Q4_K_M.gguf`, so a later integration would export GGUF names the installer cannot download or resolve.
- Add a regression to `ods/tests/test-llmfit-advisor.sh` that loads `ods/config/model-library.json` and fails if any `ods_gguf_file` in the llmfit map is not present in the catalog. If this PR intentionally adds new installable targets instead, those models need the normal catalog metadata first: URL, checksum, size, context, `llm_model_name`, and matching Linux/Windows tier-map behavior.
- If `ods_hardware_classes` is meant to enforce placement, either have `ods/scripts/llmfit-advisor.sh` use it or remove it from the map for now. As written, the parser ignores that field, so the test does not prove hardware-class parity.

What I checked:
- Static review of the PR metadata, full diff, new advisor script, fixtures, and current model selector/tier-map/catalog conventions.
- `C:\Program Files\Git\bin\bash.exe ods/tests/test-llmfit-advisor.sh`: passed, but only against the new self-contained fixture map.
- `C:\Program Files\Git\bin\bash.exe -n ods/scripts/llmfit-advisor.sh ods/tests/test-llmfit-advisor.sh`: passed.
- `python -m json.tool ods/config/llmfit-model-map.json`: passed.
- `git diff --check origin/main...origin/pr/1672`: passed.
- `gh pr checks 1672` reports no checks for this branch.

Thanks again for pushing this model-selection idea forward.

IronicRayquaza:
Addressed all three items from the advisory review:

- **Fix 1 — GGUF alignment:** Updated `llmfit-model-map.json` to use only filenames present in `model-library.json` (`Qwen3.5-2B-Q4_K_M.gguf`, `Qwen3.5-9B-Q4_K_M.gguf`, `Qwen3-30B-A3B-Q4_K_M.gguf`, and `Qwen3.6-35B-A3B-UD-Q4_K_M.gguf`).
- **Fix 2 — Catalog regression test:** Added `_test_catalog_alignment` to `test-llmfit-advisor.sh`. It loads `model-library.json` dynamically and fails if any `ods_gguf_file` in the map is absent from the catalog—so any future drift between the map and the catalog will be caught in CI.
- **Fix 3 — Removed `ods_hardware_classes`:** Removed this field from the map with a comment explaining that enforcement is deferred to the follow-up integration PR that wires the advisor into `classify-hardware.sh`.

Additionally, the test fixtures were updated to match the corrected HuggingFace model IDs. All tests now pass cleanly, including the new catalog alignment check and the existing contract/golden path tests.

Lightheartdevs:
Thanks for following through on the llmfit advisor feedback and tying the map back to the current ODS catalog.

What looks good:
- `ods/config/llmfit-model-map.json` now maps only to GGUF filenames present in `ods/config/model-library.json`, including the current Qwen tier targets used by the installer paths.
- `ods/tests/test-llmfit-advisor.sh` adds the catalog-alignment regression, so future map drift should fail in the focused suite instead of surfacing during installer integration.
- Keeping hardware-class enforcement out of this PR is a reasonable boundary since the advisor is still a pure, fallback-safe layer and the installer integration is deferred.

What I checked:
- Updated diff, prior advisory thread, and current model-library/tier-map conventions.
- Git Bash: `bash -n ods/scripts/llmfit-advisor.sh ods/tests/test-llmfit-advisor.sh`; `bash ods/tests/test-llmfit-advisor.sh` (5 fixtures plus catalog alignment passed).
- `python -m json.tool ods/config/llmfit-model-map.json` and `git diff --check origin/main...HEAD` passed.
- `gh pr checks 1672`: no checks reported for this branch.

Thanks again for pushing the selector work forward.

PR #1675, opened 2026-07-06
fix(health): parse disk capacity from POSIX df output

tang-vu:
## Summary

`scripts/health-check.sh` read disk usage with `df -h "$INSTALL_DIR" | tail -1 | awk '{print $5}'`. `df -h` wraps a **long device name** (e.g. an LVM path like `/dev/mapper/very--long--vg--name-root`) onto a second line, which shifts awk's column indexes so `$5` returns the mount point (`/`) instead of the capacity. `disk_usage` then reports `/`, and the `[ "$usage" -gt 90 ]` comparison errors with `integer expression expected`.

The fix:
- Switch `df -h` → `df -P` (POSIX single-line output, no wrapping, locale-independent — also avoids `-h`'s localized units).
- Select the capacity field by its trailing `%` so the parse survives any column shift.
- Gate the success path on a numeric percentage (`[[ "$usage" =~ ^[0-9]+$ ]]`) so the threshold comparison never runs on unexpected output.

## AI Assistance

AI assistant helped locate the parsing bug, draft the fix, and write the regression test. Author reviewed the diff and validation.

## Release Lane

## Changed Surface

- [x] Tests only  

## Risk And Validation

- Risk level: Low
- Validation run:
  - [x] `git diff --check`

Commands/results:

```text
$ bash -n ods/scripts/health-check.sh ods/tests/test-health-check.sh   # clean
$ bash ods/tests/test-health-check.sh
Result: 11 passed, 0 failed

# Regression test has teeth — reverting the fix reproduces the bug:
$ (old df -h parse) bash ods/tests/test-health-check.sh
  ✗ FAIL test_disk should use df -P for stable single-line output
  ✗ FAIL test_disk mis-parsed wrapped df output; expected disk_usage=95, got: "disk_usage": "/"
Result: 9 passed, 2 failed
```

The new test mocks `df` with wrapped (column-shifted) output and asserts the JSON reports `disk_usage=95` — not the mount point the old parse picked up.

## Operational Change Check

## Notes For Reviewers

Kept the change surgical: only `test_disk` parsing changed. `df -P` is POSIX-guaranteed single-line, so the historical `-h` wrapping (long LVM/mapper device names, ZFS datasets, bind mounts) can no longer misalign the capacity column. No behavior change on normal single-line output.


Lightheartdevs:
Thanks for tightening the disk health-check parsing around long device names and POSIX `df` output.

What looks good:
- `ods/scripts/health-check.sh` switches to `df -P` and finds the capacity field by its trailing `%`, which avoids the long-device-name wrapping bug without depending on human-readable units.
- The numeric guard keeps unexpected `df` output from reaching the `-gt` comparison, so the probe should fail softer than the old mount-point parse.
- `ods/tests/test-health-check.sh` now includes a wrapped-device fixture that reproduces the old column-shift failure.

What I checked:
- Diff and surrounding `test_disk` behavior.
- Git Bash: `bash -n ods/scripts/health-check.sh ods/tests/test-health-check.sh`; `bash ods/tests/test-health-check.sh` (11 passed).
- `gh pr checks 1675`: no checks reported for this branch.

Thanks again for tightening this diagnostic.

PR #1676, opened 2026-07-06
fix(health): warn on GPU memory pressure, not utilization

tang-vu:
## Summary

`test_gpu` in `scripts/health-check.sh` warned when `utilization.gpu > 95%`, but the inline comment documents the intent as **"GPU memory > 95%"**. For a local LLM server the two signals are very different:

- A `llama-server` actively serving inference **legitimately pins GPU utilization at ~100%**. The util-based check therefore flagged that normal, healthy state as `"warn"` on every busy poll.
- Meanwhile the signal the probe *meant* to surface — memory approaching full (OOM risk) — went unwatched.

This PR computes memory usage from the already-collected `memory.used`/`memory.total` values and warns at `>95%`, matching the documented intent. The division is guarded against non-numeric/zero totals so the probe never errors on odd `nvidia-smi` output.

## AI Assistance

AI assistant flagged the comment/code mismatch, drafted the fix, and wrote the regression tests. Author reviewed the diff and validation.

## Release Lane

## Changed Surface

- [x] Tests only  

## Risk And Validation

- Risk level: Low
- Validation run:
  - [x] `git diff --check`

Commands/results:

```text
$ bash -n ods/scripts/health-check.sh ods/tests/test-health-check.sh   # clean
$ bash ods/tests/test-health-check.sh
Result: 11 passed, 0 failed

# Tests have teeth — reverting to the old util>95 check reproduces the bug:
$ (old util-based warn) bash ods/tests/test-health-check.sh
  ✗ FAIL test_gpu warned on healthy high-util GPU; got: "gpu": "warn"
  ✗ FAIL test_gpu did not warn on >95% memory; got: "gpu": "ok"
Result: 9 passed, 2 failed
```

New tests mock `nvidia-smi`: a 100%-util / 8%-memory GPU must stay `"ok"`, and a 97%-memory GPU must `"warn"`.

## Operational Change Check

## Notes For Reviewers

Behavior change is intentional and narrow: it only affects when the GPU probe reports `warn` (a soft, non-critical status). If you'd actually prefer a *separate* high-utilization notice in addition to the memory warning, happy to add that — but a busy GPU on an inference box shouldn't read as a warning by itself.

### DIFF
--- a/ods/scripts/health-check.sh
+++ b/ods/scripts/health-check.sh
@@ -195,8 +195,16 @@ test_gpu() {
             result_set "gpu_util" "${gpu_util// /}"
             result_set "gpu_temp" "${temp// /}"
 
-            # Warn if GPU memory > 95% or temp > 80C
-            if [ "$(result_get "gpu_util")" -gt 95 ] 2>/dev/null; then
+            # Warn if GPU memory > 95% (approaching OOM) or temp > 80C.
+            # Deliberately NOT based on utilization: a llama-server doing
+            # inference legitimately pins the GPU at ~100% util, so a
+            # util-based warning would fire during normal, healthy load.
+            local mem_used mem_total
+            mem_used="$(result_get "gpu_mem_used")"
+            mem_total="$(result_get "gpu_mem_total")"
+            if [[ "$mem_used" =~ ^[0-9]+$ && "$mem_total" =~ ^[0-9]+$ ]] \
+                && [ "$mem_total" -gt 0 ] \
+                && [ $(( mem_used * 100 / mem_total )) -gt 95 ]; then
                 result_set "gpu" "warn"
             fi
             if [ "$(result_get "gpu_temp")" -gt 80 ] 2>/dev/null; then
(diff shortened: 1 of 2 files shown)


Lightheartdevs:
Thanks for correcting the GPU health signal around memory pressure instead of normal inference utilization.

What looks good:
- `ods/scripts/health-check.sh` now matches the documented intent: warn on GPU memory pressure, not on high utilization from a busy but healthy `llama-server`.
- The memory calculation is guarded for non-numeric or zero totals, and the existing temperature warning path remains intact.
- `ods/tests/test-health-check.sh` mocks both the high-util/low-memory case and the high-memory-pressure case, which should prevent the old warning behavior from returning.

What I checked:
- Diff and surrounding `test_gpu` behavior.
- Git Bash: `bash -n ods/scripts/health-check.sh ods/tests/test-health-check.sh`; `bash ods/tests/test-health-check.sh` (11 passed).
- `gh pr checks 1676`: no checks reported for this branch.

Thanks again for taking on the health-check path here.

Issue #1680, opened 2026-07-07
Optional extension: sovereign Solana wallet + RPC/MCP for local agents (devnet-first)


w3-surfer:
## Summary

Propose an **optional, self-custodial Solana extension** that gives ODS's local agents a wallet + RPC — 100% local, no hosted RPC, no custody. It fits the existing extension system (like `brave-search`) and, because the bundled Hermes agent already speaks MCP, it surfaces Solana as **native agent tools**.

## Why it fits ODS

- **Sovereign / local-first** — a self-custodial keypair on the box, no hosted RPC or custodial service. Same ethos as `privacy-shield`.
- **Agent-native** — MCP tools Hermes picks up automatically (`solana_create_wallet`, `solana_get_pubkey`, `solana_get_balance`, `solana_airdrop`, `solana_transfer`), plus plain REST for n8n / Open WebUI / OpenClaw / curl.
- **First MCP-server pattern in ODS** — a reusable seam for future agent tools, using a capability the bundled Hermes already supports.
- **Optional & safe by default** — `category: optional`; **devnet by default**; mainnet writes refused unless explicitly enabled.

## Scope (v1, intentionally minimal)

Devnet: create/import wallet, balance, airdrop, transfer SOL/SPL. No DeFi. Generic Solana only.

## Security

- Secret key stored `0600`, never returned or logged.
- Host port bound to `127.0.0.1`; other containers reach it only on `ods-network`.
- `SOLANA_ALLOW_MAINNET=false` by default (mainnet writes refused); airdrop devnet/testnet only.
- Optional `SOLANA_API_KEY` bearer on REST write routes. `no-new-privileges`, resource limits.

## Shape

`extensions/services/solana/` (Node + `@solana/web3.js`): `manifest.yaml`, `compose.yaml`, `Dockerfile`, `index.mjs` (REST + `/mcp`), `mcp.mjs`, `solana.mjs`, `README.md`. Hermes wiring is an **opt-in commented** `mcp_servers` example in `cli-config.yaml.template` — default behavior unchanged.

I have a working implementation (opening a PR alongside this issue): audit-extensions passes; the image builds and serves REST + MCP; devnet wallet/airdrop/balance/transfer plus the mainnet / API-key / input-validation guards are all verified. Happy to adjust scope — would a devnet-first, optional Solana extension be welcome?

gabsprogrammer:
Thanks for the PR, this will be noted in our database of future PRs.

PR #1681, opened 2026-07-07
feat(solana): local self-custodial Solana wallet + RPC/MCP extension

w3-surfer:
Refs #1680.

## What

Adds an optional `solana` service extension (`extensions/services/solana/`) that gives ODS's local agents a **self-custodial Solana wallet + RPC**, exposed two ways from one Node process:

- **MCP over HTTP** at `/mcp` — Hermes gets native tools: `solana_create_wallet`, `solana_get_pubkey`, `solana_get_balance`, `solana_airdrop`, `solana_transfer`.
- **REST** — for n8n / Open WebUI / OpenClaw / curl.

No hosted RPC, no custody. Mirrors the `brave-search` extension shape (manifest + compose + Dockerfile, Node + `@solana/web3.js`).

## Security (devnet-first)

- Secret key stored `0600`; **never returned by any endpoint/tool, never logged**.
- Host port bound to `127.0.0.1`; other containers reach it only on `ods-network`.
- **Devnet by default.** Transfers on `mainnet-beta` are refused unless `SOLANA_ALLOW_MAINNET=true`; airdrop is devnet/testnet only.
- Optional `SOLANA_API_KEY` bearer guards the REST write routes. `no-new-privileges`, CPU/memory limits.

## Hermes wiring (opt-in, non-breaking)

A **commented** `mcp_servers` example is added to `hermes/cli-config.yaml.template`. Default behavior is unchanged; the user uncomments it after `ods enable solana`.

## Verification (all on devnet)

- `python3 scripts/audit-extensions.py --project-dir .` → **PASS solana** (0 errors, 0 warnings).
- Service registry discovers `solana` (container/health/port).
- `docker build` succeeds and the container **serves both REST and MCP**:
  - REST: `/health`, `POST /wallet` (creates + persists), `GET /balance`, airdrop reaches the faucet, transfer builds/simulates on-chain.
  - MCP: `tools/list` returns all 5 tools; `tools/call` returns the same managed wallet as REST (shared core).
- Guards verified: mainnet write-refusal, airdrop mainnet-refusal, `SOLANA_API_KEY` → 401 on writes without it, invalid pubkey → 400.

Scope is intentionally minimal (devnet, no DeFi, generic Solana). Happy to adjust.

### DIFF
--- a/ods/config/dependency-lock.json
+++ b/ods/config/dependency-lock.json
@@ -208,6 +208,12 @@
       "value": "searxng/searxng:2026.3.8-a563127a2"
     },
     {
+      "id": "solana.node",
+      "type": "image",
+      "path": "extensions/services/solana/Dockerfile",
+      "value": "node:20-alpine"
+    },
+    {
       "id": "tailscale.sidecar",
       "type": "image",
       "path": "extensions/services/tailscale/compose.yaml",
@@ -273,6 +279,11 @@
     {
       "path": "extensions/services/brave-search/compose.yaml",
       "value": "ods-brave-search:local",
+      "reason": "Local extension image built from the adjacent Dockerfile."
+    },
+    {
+      "path": "extensions/services/solana/compose.yaml",
+      "value": "ods-solana:local",
       "reason": "Local extension image built from the adjacent Dockerfile."
     }
   ],
--- a/ods/config/network-exposure-policy.json
+++ b/ods/config/network-exposure-policy.json
@@ -122,6 +122,12 @@
       "auth_required": false,
       "notes": "Metasearch service; exposure can leak query history."
     },
+    "solana": {
+      "risk": "wallet-rpc-surface",
+      "lan_exposure": "localhost-by-default",
+      "auth_required": false,
+      "notes": "Self-custodial Solana wallet + RPC. Localhost-bound by default; MCP is read-only unless SOLANA_MCP_WRITE=true; mainnet writes require SOLANA_ALLOW_MAINNET; optional SOLANA_API_KEY guards REST writes."
+    },
     "tailscale": {
       "risk": "remote-network-access",
       "lan_exposure": "host-network",
--- a/ods/extensions/services/hermes/cli-config.yaml.template
+++ b/ods/extensions/services/hermes/cli-config.yaml.template
@@ -176,3 +176,17 @@ compression:
   # (~10 user/assistant exchanges); bumped to 40 so a longer back-and-forth
   # stays granular even after a compression pass.
   protect_last_n: 40
+
+# ─── MCP servers (optional) ───────────────────────────────────────────────────
+# External Model Context Protocol servers Hermes connects to for extra tools.
+# Left commented so default behavior is unchanged. Uncomment a block after
+# enabling the matching ODS extension, then restart Hermes.
+#
+# The `solana` extension (`ods enable solana`) exposes a local, self-custodial
+# Solana wallet + RPC as agent tools (solana_create_wallet, solana_get_balance,
+# solana_airdrop, solana_transfer). It runs on ods-network — reach it by its
+# service DNS name. Devnet by default; see extensions/services/solana/README.md.
+#
+# mcp_servers:
+#   solana:
+#     url: http://ods-solana:8590/mcp
--- a/ods/extensions/services/hermes/cli-config.yaml.template
+++ b/ods/extensions/services/hermes/cli-config.yaml.template
@@ -183,9 +183,11 @@ compression:
 # enabling the matching ODS extension, then restart Hermes.
 #
 # The `solana` extension (`ods enable solana`) exposes a local, self-custodial
-# Solana wallet + RPC as agent tools (solana_create_wallet, solana_get_balance,
-# solana_airdrop, solana_transfer). It runs on ods-network — reach it by its
-# service DNS name. Devnet by default; see extensions/services/solana/README.md.
+# Solana wallet + RPC as agent tools. Read tools (solana_get_pubkey,
+# solana_get_balance) are always available; write tools (create_wallet, airdrop,
+# transfer) require SOLANA_MCP_WRITE=true. It runs on ods-network — reach it by
+# its service DNS name. Devnet by default; see
+# extensions/services/solana/README.md.
 #
 # mcp_servers:
 #   solana:
--- a/ods/extensions/services/solana/Dockerfile
+++ b/ods/extensions/services/solana/Dockerfile
@@ -10,7 +10,7 @@ WORKDIR /app
 COPY package.json package-lock.json ./
 RUN npm ci --omit=dev --omit=optional --ignore-scripts --no-audit --no-fund
 
-COPY index.mjs solana.mjs mcp.mjs ./
+COPY index.mjs solana.mjs mcp.mjs cluster.mjs ./
 
 # Keystore volume must be writable by the unprivileged runtime user. Named
 # volumes inherit this ownership on first creation.
--- a/ods/extensions/services/solana/Dockerfile
+++ b/ods/extensions/services/solana/Dockerfile
@@ -1,0 +1,22 @@
+FROM node:20-alpine
+
+WORKDIR /app
+
+# Install deps first for layer caching (reproducible via the lockfile).
+# ws pulls native accelerators (utf-8-validate/bufferutil) that need a C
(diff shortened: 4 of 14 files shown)


Lightheartdevs:
Thanks for taking on the Solana wallet/MCP extension; the extension shape is close to the existing optional-service pattern, and the devnet-first intent is clear.

Quickest path to merge:
- Add the new image references to `ods/config/dependency-lock.json`: `extensions/services/solana/Dockerfile:1` introduces `node:20-alpine`, and `extensions/services/solana/compose.yaml:6` introduces the local image `ods-solana:local`. Mirror the existing `brave-search.node` entry plus the `allow_local_images` entry used for `ods-brave-search:local`, then rerun `python tests/test-dependency-pins.py`. This is the current `Test Linux / integration-smoke` failure.
- Tighten the mainnet write guard in `ods/extensions/services/solana/solana.mjs`. The actual chain endpoint comes from `SOLANA_RPC_URL` at lines 26-27/60-63, but `assertWriteAllowed()` only checks `SOLANA_NETWORK` at lines 40-44 before `transferSol` and `transferSpl` write at lines 120 and 138. With `SOLANA_RPC_URL=https://api.mainnet-beta.solana.com` and the default `SOLANA_NETWORK=devnet`, mainnet transfers would bypass `SOLANA_ALLOW_MAINNET`. Either derive/validate the cluster from the RPC setting or refuse known mainnet RPC URLs unless the mainnet flag is explicitly enabled, and add a small regression check for that path.
- Clarify or guard MCP write access in `ods/extensions/services/solana/index.mjs` and `mcp.mjs`. REST writes honor `SOLANA_API_KEY` at `index.mjs:78-95`, but `/mcp` dispatch at `index.mjs:104-106` bypasses that guard while `mcp.mjs:48-75` exposes airdrop and transfer tools. If MCP writes are intentionally unauthenticated inside `ods-network`, document that explicitly in the README/Hermes snippet; otherwise apply the same write guard or split read/write MCP tools behind an explicit opt-in.

What I checked:
- PR diff/static review against the repo's extension, compose, Hermes-template, and high-risk validation conventions.
- GitHub checks: all visible jobs pass except `Test Linux / integration-smoke`, which fails in `tests/test-dependency-pins.py` on the two missing dependency-lock entries above.
- Local checks on the PR worktree: `python scripts/audit-extensions.py --project-dir .` passes for all 25 services including `solana`; `node --check` passes for `index.mjs`, `mcp.mjs`, and `solana.mjs`; `python tests/test-dependency-pins.py` reproduces the dependency-lock failure.

Thanks again for pushing the integration forward.

w3-surfer:
Thanks for the thorough review — all three addressed in the latest push.

**1. dependency-lock.json** — added the `solana.node` image pin (`node:20-alpine`) and the `ods-solana:local` `allow_local_images` entry, mirroring `brave-search`. `python tests/test-dependency-pins.py` now passes.

**2. Mainnet write guard** — the guard now treats a target as mainnet when `SOLANA_NETWORK` says so **or** `SOLANA_RPC_URL` resolves to a mainnet host, so `SOLANA_RPC_URL=https://api.mainnet-beta.solana.com` with the default `SOLANA_NETWORK=devnet` can no longer bypass `SOLANA_ALLOW_MAINNET`. The classification is a pure, dependency-free `cluster.mjs`, with a regression test in `test.mjs` (`node test.mjs`, no deps). Verified live: a container with the devnet label + a mainnet RPC returns `403 writes on mainnet require SOLANA_ALLOW_MAINNET=true` on `/transfer`.

**3. MCP write access** — since `/mcp` is unauthenticated within `ods-network`, write tools (`create_wallet`/`airdrop`/`transfer`) are now registered only when `SOLANA_MCP_WRITE=true`. MCP is **read-only by default** — verified `tools/list` returns 2 read tools by default and 5 with the flag. Documented the trust model in the README and the Hermes snippet.

`audit-extensions.py` still passes for all 25 services, the image builds and serves both REST and MCP, and `node --check` passes on all modules. Thanks again!

Lightheartdevs:
Thanks for tightening up the Solana extension after the first review; the dependency-lock fix, mainnet-via-RPC guard, and read-only-by-default MCP path all move this in the right direction.

Quickest path to merge:
- Add `solana` to `ods/config/network-exposure-policy.json`. `ods/extensions/services/solana/manifest.yaml:13`-`:15` gives the service a host-facing external port, and the repo contract at `ods/tests/contracts/test-network-exposure-contracts.py:47`-`:57` requires every exposed service to have `risk`, `lan_exposure`, `auth_required`, and `notes`. This is the current `Test Linux / integration-smoke` failure: `missing exposure policy entries: ['solana']`. A good entry should call out that this is a wallet/RPC surface, localhost-bound by default, with MCP read-only unless explicitly enabled.
- Pass `SOLANA_MCP_WRITE` through `ods/extensions/services/solana/compose.yaml`. The manifest documents the knob at `manifest.yaml:43`-`:46`, and `mcp.mjs:12` reads `process.env.SOLANA_MCP_WRITE`, but the compose `environment:` block at `compose.yaml:11`-`:17` never sets it. As written, an operator can set `SOLANA_MCP_WRITE=true` in `.env` and the container will still start with MCP write tools disabled. Add `SOLANA_MCP_WRITE=${SOLANA_MCP_WRITE:-false}` there and re-check the MCP tools-list path if possible.

What I checked:
- GitHub checks for the current head: all visible jobs pass except `integration-smoke`, which fails in `test-network-exposure-contracts.py` on the missing policy entry above.
- Local PR worktree checks: `python tests/test-dependency-pins.py` passed; `python scripts/audit-extensions.py --project-dir .` passed for all 25 services including `solana`; `node test.mjs` passed; `node --check` passed for the new Solana modules; `python tests/contracts/test-network-exposure-contracts.py` reproduced the policy failure.
- Static review of the Solana manifest, compose, REST/MCP split, mainnet guard, Hermes template snippet, and repo exposure-policy convention.

w3-surfer:
Both addressed in the latest push.

**1. network-exposure-policy.json** — added a `solana` entry (`risk: wallet-rpc-surface`, `lan_exposure: localhost-by-default`, `auth_required: false`, notes covering the localhost bind, MCP read-only default, the mainnet gate, and the optional REST API key). `python tests/contracts/test-network-exposure-contracts.py` now passes (`test_exposed_services_are_policy_labeled`).

**2. compose `SOLANA_MCP_WRITE`** — added `SOLANA_MCP_WRITE=${SOLANA_MCP_WRITE:-false}` to the compose `environment` block. Verified `docker compose config` renders `false` by default and `true` when set, and a compose-started container with `SOLANA_MCP_WRITE=true` exposes all 5 MCP tools (2 read-only otherwise).

`dependency-pins` and `audit-extensions` still pass. Thanks for the careful review!

Lightheartdevs:
Thanks for tightening up the Solana extension after the follow-up reviews; the latest head closes the dependency-lock and exposure-policy gaps while keeping the wallet surface opt-in.

What looks good:
- `ods/extensions/services/solana/compose.yaml` keeps the host port loopback-bound by default and carries `SOLANA_MCP_WRITE`, so MCP write tools stay disabled unless the operator opts in.
- `ods/extensions/services/solana/cluster.mjs` and `solana.mjs` now guard mainnet writes by both network label and RPC host, which closes the mainnet-RPC-with-devnet-label bypass.
- `ods/config/network-exposure-policy.json` and `ods/config/dependency-lock.json` now account for the new service surface and local image build.

What I checked:
- GitHub check rollup for this head shows 28 successful checks and 4 skipped.
- Local focused checks: `node ods/extensions/services/solana/test.mjs`, `python ods/tests/contracts/test-network-exposure-contracts.py`, `python scripts/check-dependency-pins.py`, and `python ods/scripts/audit-extensions.py --project-dir ods solana`.
- Static review of the Solana REST/MCP handlers, compose file, manifest, dependency lock, and Hermes template wiring.

w3-surfer:
Hi, checking in on this one. No rush.

Both review rounds are addressed: the mainnet write guard now derives the cluster from SOLANA_RPC_URL (with a regression test), MCP is read-only by default behind SOLANA_MCP_WRITE, plus the dependency-lock.json and network-exposure-policy.json entries. CI is green and it's currently mergeable.

Happy to keep iterating if there's anything you'd like changed. And if a Solana extension isn't something you want in the catalog right now, or you'd rather it live out-of-tree, just say the word and I'll close it. I'd rather know than leave it sitting in your queue.

PR #1683, opened 2026-07-07
fix(health): aggregate async service results into status, exit code, and JSON

tang-vu:
## Problem

`scripts/health-check.sh` documents an exit-code contract (`0=healthy, 1=degraded (some services down), 2=critical (core services down)`), but the per-service checks run in **background subshells** (`check_service_async ... &`). The `result_set`/`ANY_FAIL=true` mutations inside `test_service` die with the subshell, and the parent display loops only `log` the per-service lines. Two user-visible consequences:

1. **False HEALTHY**: a stack with every Docker service down still summarizes `Status: HEALTHY - All services operational` with **exit 0**, as long as the llama-server inference probe (run in the parent) passes. Anything scripting around the documented exit codes gets no signal.
2. **Incomplete JSON**: the `--json` services map only ever contains `llm`, `gpu`, and `disk` — every registry service is omitted, healthy or not.

A third path makes services **vanish entirely**: `check_container_state` exits non-zero for any not-running container, and the unguarded `container_state=$(check_container_state ...)` assignment at the top of `check_service_async` lets `set -e` kill the subshell **before the result file is written**. The parent's `[[ -f $result_file ]]` then skips the service completely — the container-state messages (`container not found` / `stopped` / `restarting`) were unreachable through the async path whenever docker was present.

## Fix

- Parent display loops now aggregate each result file into `result_set` (so `--json` reports every probed service) and flip `CRITICAL_FAIL` for core services / `ANY_FAIL` for extensions, per the documented exit codes. A missing result file counts as a failure instead of silence.
- Guard the container-state assignment in `check_service_async` with `|| true` so a not-running container reports its state instead of vanishing.
- Only probe **enabled** extensions (active compose fragment — same predicate as `sr_list_enabled`, filesystem analog of `ods-cli`'s own "not running, skip silently" rule). Without this, aggregating failures would flag every disabled bundled extension and report permanent DEGRADED on stock installs.

## Verification

Sandboxed the script with a mocked registry (fake core service + enabled/disabled extensions) and `curl`/`docker` PATH stubs:

| Scenario | Before | After |
|---|---|---|
| Core service down (llm ok) | HEALTHY, exit 0, absent from JSON | CRITICAL, exit 2, `"fakesvc": "fail"` |
| Docker present, container not found | service **vanishes** from output & JSON | reported `container not found`, exit 2 |
| Enabled extension down | HEALTHY, exit 0 | DEGRADED, exit 1, `"fakeext": "fail"` |
| All probed services healthy | HEALTHY, exit 0 | HEALTHY, exit 0 (unchanged) |
| Disabled extension | probed, shown "not responding" | skipped entirely |

`tests/test-health-check.sh` gained the same scenarios as mock-based regression tests (skip gracefully when python3/PyYAML unavailable): **16 passed, 0 failed** locally (Git Bash). `bash -n` clean on both files.


Lightheartdevs:
Thanks for tightening the health-check contract here; aggregating the async service result files is the right shape for making `--json` and the documented exit codes trustworthy.

Quickest path to merge:
- In `ods/tests/test-health-check.sh`, make the first sandbox scenario hermetic before the `agg_json` run around the new async aggregation test. Right now that scenario stubs `curl` but not `docker`, while `test_service` checks Docker state before hitting the health URL. On a host with Docker installed, the fake `ods-fakesvc` container is reported as `container not found`, so the test exits critical (`2`) instead of degraded (`1`) and never reports `fakesvc: ok`.
- The smallest fix is to add a sandbox `docker` stub for the first scenario too, returning `running` for `docker inspect --format '{{.State.Status}}' ods-fakesvc` / `ods-fakeext` so the test is exercising the curl success/failure paths it intends. Keep the second scenario's failing Docker stub for the `container not found` regression.

What I checked:
- Static diff review of `ods/scripts/health-check.sh` and `ods/tests/test-health-check.sh` against the current `main` architecture and `sr_list_enabled` behavior.
- `git diff --check origin/main...pr-1683 -- ods/scripts/health-check.sh ods/tests/test-health-check.sh` passed.
- `bash -n scripts/health-check.sh` and `bash -n tests/test-health-check.sh` passed using Git Bash.
- `bash tests/test-health-check.sh` failed locally: `14 passed, 2 failed`, with `failing extension should exit 1 (degraded); got 2` and `healthy core service missing from JSON services`. The local Docker CLI was visible at `/c/Program Files/Docker/Docker/resources/bin/docker`, which is the host leak causing the fake core service to fail before the curl stub.
- `gh pr checks` currently reports no checks for this branch.

Thanks again for pushing this forward; this looks close once the regression test is isolated from the host Docker environment.

tang-vu:
Thanks for catching the host-Docker leak — confirmed and fixed in 59228b82.

**Repro first:** injected a fake host `docker` (failing `inspect`, mimicking Docker Desktop with no such containers) into the outer PATH and got exactly your result: 14 passed, 2 failed, with `failing extension should exit 1 (degraded); got 2` and `healthy core service missing from JSON services`.

**Fix:** the first scenario now stubs `docker` in the sandbox `bin/` to report every container as `running`, so it exercises the curl success/failure aggregation it was written for. The second scenario keeps its failing docker stub for the container-not-found regression, as you suggested.

**Verified both ways:** with the fake host docker leaked into the outer PATH and without any docker in PATH — 16 passed, 0 failed in both environments.

Lightheartdevs:
Thanks for tightening the async health-check aggregation; this fixes a real operator contract around JSON output and exit codes.

What looks good:
- `ods/scripts/health-check.sh` now aggregates async result files in the parent process, so core failures set `CRITICAL_FAIL`, extension failures set `ANY_FAIL`, and `--json` includes the probed services instead of losing subshell state.
- The enabled-extension filter follows the active compose-fragment predicate, so disabled bundled services do not permanently degrade a stock install.
- `ods/tests/test-health-check.sh` now stubs Docker in the aggregation scenario and separately covers `container not found`, closing the host-Docker leak from the prior review.

What I checked:
- Diff/static review of `ods/scripts/health-check.sh` and `ods/tests/test-health-check.sh`.
- `git diff --check origin/main...origin/pr/1683` passed.
- Git Bash: `bash -n scripts/health-check.sh tests/test-health-check.sh` and `bash tests/test-health-check.sh` passed: 16 passed, 0 failed.
- `gh pr checks 1683` currently reports no checks on this branch.

Thanks again for following through on the regression test fix.

PR #1684, opened 2026-07-07
fix(backup): make list, retention, and delete manage the script's own backups

tang-vu:
## Problem

`ods-backup.sh` names its backups `$(date +%Y%m%d-%H%M%S)` — **one** hyphen (e.g. `20260707-103450`). But `list_backups` and `apply_retention` discovered entries with `find -name "*-*-*"`, which requires **two** hyphens. Every backup this script creates was therefore invisible to both:

- `--list` reports **"No backups found"** immediately after a successful backup.
- `apply_retention` logs *"Applying retention policy (keeping last 5 backups)"* on every run while **never deleting anything**. Full backups include `models/` (tens of GB), so the silent unbounded growth eventually exhausts the backup disk — the exact failure the retention policy exists to prevent.

Only dashboard-initiated backups (`dashboard-YYYYMMDD-HHMMSS`, two hyphens) matched the old glob, which is presumably why this went unnoticed.

Separately, `delete_backup` only handled directories — a `.tar.gz` created with `--compress` could **never be deleted by ID** (`-d <id>` → "Backup not found").

## Fix

- Extract a shared `collect_backups` helper (used by list + retention) that matches the actual ID shapes: `YYYYMMDD-HHMMSS` with optional `<prefix>-` and `.tar.gz` suffix. Retention **deletes** what it matches, so the filter stays strict — unrelated files an operator drops into `.backups/` never qualify (covered by a test).
- `delete_backup` accepts both the bare ID and the full archive name for compressed backups.
- Bash 3.2-safe: no empty-array expansions under `set -u` (script is macOS-compatible).

## Also fixes the test harness

`tests/test-backup-integrity.sh` had been failing at source time on current `main` — `ods-backup.sh` sources `lib/rsync.sh` relative to `ODS_DIR`, and the test's fake ODS dir never had it (`line 32: .../ods/lib/rsync.sh: No such file or directory`). The harness now copies the lib in, so the pre-existing checksum/verify assertions run again.

## Verification

- New lifecycle regression tests: `--list` shows own-format IDs and skips non-backup dirs; retention prunes the oldest beyond `RETENTION_COUNT` and leaves unrelated dirs; `-d` removes a compressed backup by bare ID. Full suite: **6/6 pass**.
- Negative check: the extended test run against the unfixed script fails at `--list does not show own-format backup IDs` — the regression test bites.
- `bash -n` clean on both files. (Local runs used an `rsync` PATH stub — Git Bash has no rsync; CI/dev boxes with real rsync exercise the same paths.)

### DIFF
--- a/ods/ods-backup.sh
+++ b/ods/ods-backup.sh
@@ -135,6 +135,24 @@ ensure_backup_space() {
     fi
 }
 
+# Collect backup entries (dirs or .tar.gz archives) under BACKUP_ROOT into
+# the global COLLECTED_BACKUPS array, newest first.
+#
+# Backup IDs are YYYYMMDD-HHMMSS (this script) or <prefix>-YYYYMMDD-HHMMSS
+# (dashboard/host-agent). Match that shape explicitly: retention deletes
+# whatever this collects, so an operator's unrelated files in the backup
+# directory must never qualify.
+COLLECTED_BACKUPS=()
+collect_backups() {
+    COLLECTED_BACKUPS=()
+    local entry base
+    while IFS= read -r -d '' entry; do
+        base=$(basename "$entry")
+        [[ "$base" =~ ^([A-Za-z0-9_]+-)?[0-9]{8}-[0-9]{6}(\.tar\.gz)?$ ]] || continue
+        COLLECTED_BACKUPS+=("$entry")
+    done < <(find "$BACKUP_ROOT" -maxdepth 1 \( -type d -o -name "*.tar.gz" \) -print0 2>/dev/null | sort -z -r)
+}
+
 # Show usage
 usage() {
     cat << EOF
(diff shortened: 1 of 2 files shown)


Lightheartdevs:
Thanks for fixing the backup lifecycle path here; list, retention, and delete all need to agree on the backup IDs the script actually creates.

What looks good:
- `ods/ods-backup.sh` centralizes discovery in `collect_backups` and uses a strict timestamp/prefix filter, so script-created IDs and `dashboard-...` backups are managed while unrelated directories under `.backups` are left alone.
- `delete_backup` now handles compressed archives by bare ID without relaxing the existing path-traversal checks.
- `ods/tests/test-backup-integrity.sh` covers list visibility, retention pruning, unrelated-directory preservation, compressed archive deletion, and the required `lib/rsync.sh` fixture.

What I checked:
- Diff/static review of `ods/ods-backup.sh` and `ods/tests/test-backup-integrity.sh`.
- `git diff --check origin/main...origin/pr/1684` passed.
- Git Bash: `bash -n ods-backup.sh tests/test-backup-integrity.sh` passed.
- Git Bash: `bash tests/test-backup-integrity.sh` passed with a temporary exported `rsync` function because this Windows host does not have `rsync` installed; the run covered checksum verify/tamper, list, retention, and compressed delete.
- `gh pr checks 1684` currently reports no checks on this branch.

PR #1685, opened 2026-07-07
fix(restore): keep log and UI output out of command-substituted results

tang-vu:
## Problem

Two call sites in `ods-restore.sh` capture a function's stdout, and both functions also print their log/UI lines to stdout — the captured value swallows them:

**1. Compressed backups can never be restored.** `do_restore` captures `extract_backup`'s stdout as the backup path. For `.tar.gz` backups the `[INFO] Extracting compressed backup...` log line (ANSI codes included) is captured together with the path, so validation always fails:

```
[ERROR] Backup manifest not found: [0;34m[INFO][0m Extracting compressed backup...
/…/.backups/20260615-120000/manifest.json
[ERROR] Backup validation failed
```

Every backup created with `ods-backup.sh --compress` is unrestorable through this script (uncompressed backups take the early-return path that prints only the path, which is why they work).

**2. The interactive picker is unusable.** `main` captures `select_backup`'s stdout as the backup ID. The "Available Backups" table and the prompt are swallowed into the captured value — the user stares at a blank screen, and the resolved `backup_id` is the whole table plus the basename, which can never match a real backup.

**3. `.backups` offered as a backup.** The discovery `find`s lack `-mindepth 1`, so `BACKUP_ROOT` itself is listed as a selectable entry named `.backups` in the picker.

## Fix

- `extract_backup`: log lines → stderr; `do_restore` fails the capture explicitly (`if ! backup_dir=$(...)`) instead of leaning on `set -e`.
- `select_backup`: table + prompt → stderr, keeping only the resolved ID on stdout. Same convention as `snapshot_pre_update` in `ods-update.sh` ("all log calls redirect to stderr so command-substitution callers only capture the path").
- `-mindepth 1` on both discovery `find`s.

## Also fixes the test harness

`tests/test-backup-restore-roundtrip.sh` had been failing at source time on `main`: both scripts source `lib/rsync.sh` relative to `ODS_DIR`, and the test's fake ODS dirs never contained it. The harness now copies the lib in (same fix as the integrity-test harness in #1684, different file — no overlap).

## Verification

- Repro before fix: compressed restore → `Backup manifest not found: [INFO]…` exit 1; interactive picker → blank table, garbage ID, exit 1. After fix: both succeed, `.backups` row gone.
- Round-trip suite extended: compressed backup → restore → content assertions, plus an interactive-selection dry-run that asserts the table reaches the user (stderr) — **all pass**.
- Negative check: the extended test against the unfixed script fails at `Compressed restore failed`.
- `bash -n` clean. (Local runs used an rsync PATH stub — Git Bash ships no rsync; CI/dev boxes exercise the same paths with real rsync.)

### DIFF
--- a/ods/ods-restore.sh
+++ b/ods/ods-restore.sh
@@ -128,7 +128,7 @@ list_backups() {
     local backups=()
     while IFS= read -r -d '' backup; do
         backups+=("$backup")
-    done < <(find "$BACKUP_ROOT" -maxdepth 1 \( -type d -o -name "*.tar.gz" \) -print0 2>/dev/null | sort -z -r)
+    done < <(find "$BACKUP_ROOT" -mindepth 1 -maxdepth 1 \( -type d -o -name "*.tar.gz" \) -print0 2>/dev/null | sort -z -r)
 
     if [[ ${#backups[@]} -eq 0 ]]; then
         log_error "No backups found in: $BACKUP_ROOT"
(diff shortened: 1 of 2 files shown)


Lightheartdevs:
Thanks for cleaning up the restore stdout contract; command substitution and operator-facing output are easy to cross-wire in shell scripts.

What looks good:
- `ods/ods-restore.sh` now sends the interactive backup table and prompt to stderr while returning only the selected backup ID on stdout, so the picker remains visible and `backup_id=$(select_backup)` stays clean.
- `extract_backup` now keeps log/error output on stderr and `do_restore` handles extraction failure explicitly, so compressed restores do not fold `[INFO]` text into the manifest path.
- The `-mindepth 1` discovery fix keeps `.backups` itself out of selectable entries, and `ods/tests/test-backup-restore-roundtrip.sh` covers compressed restore plus the interactive dry-run path.

What I checked:
- Diff/static review of `ods/ods-restore.sh` and `ods/tests/test-backup-restore-roundtrip.sh`.
- `git diff --check origin/main...origin/pr/1685` passed.
- Git Bash: `bash -n ods-restore.sh tests/test-backup-restore-roundtrip.sh` passed.
- Git Bash: `bash tests/test-backup-restore-roundtrip.sh` passed with a temporary exported `rsync` function because this Windows host does not have `rsync` installed; the run covered uncompressed round-trip restore, compressed restore, and interactive selection dry-run.
- `gh pr checks 1685` currently reports no checks on this branch.

Thanks again for pushing the restore path forward.

PR #1686, opened 2026-07-07
feat(extensions): add vLLM backend support

shlok-mehndiratta:
### Description
This PR introduces **vLLM** as an opt-in Extension Service, directly addressing the recent call for alternative, high-throughput inference backends. 

While the core `llama-server` is excellent for broad consumer hardware compatibility, advanced users with multi-GPU arrays or datacenter hardware are heavily bottlenecked without PagedAttention and continuous batching. This extension provides a first-class, drop-in vLLM OpenAI-compatible server that users can route their ODS traffic to by simply pointing `LLM_API_URL` to `http://vllm:8000/v1`.

### Implementation Details
Built on top of the latest `ods` directory re-architecture, this PR strictly follows **Pattern 2 (Empty Base with Full GPU Overlay)** from `EXTENSIONS.md`. It safely isolates the addition without touching any High-Risk core routing logic.

* **GPU Backends:** Provides full overlays for both `nvidia` (using `vllm/vllm-openai:latest`) and `amd` (using `rocm/vllm:latest`, mounting `/dev/kfd` and `/dev/dri`).
* **Environment Configuration:** Exposes `VLLM_MODEL`, `VLLM_TP_SIZE`, and `HUGGING_FACE_HUB_TOKEN` via `.env` variables so users can dynamically spin up gated models and scale tensor parallelism.
* **Cache Mounting:** Mounts `./data/models/huggingface` to `/root/.cache/huggingface:z` for proper persistence and strict SELinux compliance.
* **NCCL/RCCL Crash Prevention:** Explicitly allocates `shm_size: '8g'` in both compose overlays. Without this, users running `VLLM_TP_SIZE > 1` immediately hit `bus error` crashes due to Docker's default 64MB `/dev/shm` limit restricting NCCL/RCCL shared memory allocation.
* **Port Safety:** Runs internally on `8000` but binds externally to `8008` (`VLLM_PORT`) to prevent collision with `whisper` (`9000`) or `llama-server` (`8080`).

### Validation
- [x] Passed `python3 scripts/audit-extensions.py --project-dir . vllm` with 0 errors/warnings.
- [x] Manifest successfully validated against the new `ods.services.v1` schema.
- [x] `git diff --check` passes cleanly (no trailing whitespace).
- [x] Tested against the latest `upstream/main` merge to guarantee zero integration conflicts.

### DIFF
--- a/ods/config/dependency-lock.json
+++ b/ods/config/dependency-lock.json
@@ -250,6 +250,18 @@
       "type": "image",
       "path": "installers/windows/docker-compose.windows-amd.local.yml",
       "value": "busybox:1.36.1"
+    },
+    {
+      "id": "vllm.nvidia",
+      "type": "image",
+      "path": "extensions/services/vllm/compose.nvidia.yaml",
+      "value": "vllm/vllm-openai:v0.10.2"
+    },
+    {
+      "id": "vllm.amd",
+      "type": "image",
+      "path": "extensions/services/vllm/compose.amd.yaml",
+      "value": "rocm/vllm:rocm7.0.0_vllm_0.10.2_20251006"
     }
   ],
   "allow_latest": [
--- a/ods/docs/TESTING.md
+++ b/ods/docs/TESTING.md
@@ -108,12 +108,14 @@ tests/fleet-multi-distro.sh --pull
 ```
 
 The Docker and Incus distro runners take a shared host lock by default:
-`/tmp/ods-fleet-heavy.lock`, or `ODS_FLEET_HOST_LOCK` when set. The
-private release automation uses the same lock when launching heavy install
-work, so distro-lab dry-runs do not compete with full fleet installs for
-Docker/build I/O on the same host. Use `--lock-timeout SECONDS` when a CI or
-automation should fail instead of waiting, and reserve `--no-host-lock` for
-local debugging when you know no full fleet install is running.
+`/tmp/dream-fleet-heavy.lock`. Set `ODS_FLEET_HOST_LOCK` to override it;
+the older `DREAM_FLEET_HOST_LOCK` name is still honored for release automation
+compatibility. The private release automation uses the same lock when launching
+heavy install work, so distro-lab dry-runs do not compete with full fleet
+installs for Docker/build I/O on the same host. Use `--lock-timeout SECONDS`
+when a CI or automation should fail instead of waiting, and reserve
+`--no-host-lock` for local debugging when you know no full fleet install is
+running.
 
 Release-grade fleet runs should run the distro lab alongside the hardware
 fleet. The hardware fleet proves GPU and product behavior; the distro lab proves
--- a/ods/extensions/services/vllm/README.md
+++ b/ods/extensions/services/vllm/README.md
@@ -1,0 +1,10 @@
+# vLLM Extension
+
+A high-throughput, multi-GPU inference backend utilizing PagedAttention.
+
+## Usage
+1. Provide a model repository in `.env`: `VLLM_MODEL=meta-llama/Meta-Llama-3-8B-Instruct`
+2. Provide your HuggingFace token if required: `HUGGING_FACE_HUB_TOKEN=hf_...`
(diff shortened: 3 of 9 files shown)


Lightheartdevs:
Thanks for taking on the vLLM extension path here; the GPU-only overlay shape is a good fit for how ODS models optional heavy inference services.

Quickest path to merge:
- `ods/extensions/services/vllm/compose.nvidia.yaml:7` and `ods/extensions/services/vllm/compose.amd.yaml:7`: bind the host port through the repo's normal localhost-first contract, e.g. `${BIND_ADDRESS:-127.0.0.1}:${VLLM_PORT:-8008}:8000`. As written, `${VLLM_PORT:-8008}:8000` publishes vLLM on all interfaces by default, which bypasses the ODS network-exposure posture and could expose an OpenAI-compatible model endpoint on the LAN unintentionally.
- `ods/extensions/services/vllm/compose.nvidia.yaml:3`, `ods/extensions/services/vllm/compose.amd.yaml:3`, and `ods/config/dependency-lock.json`: replace the `:latest` images with retained version tags or `@sha256` digest pins and add matching dependency-lock entries. The repo has an explicit runtime image pin contract, and this head currently fails it for both `vllm/vllm-openai:latest` and `rocm/vllm:latest`.

What I checked:
- PR metadata, changed files, full diff, and existing comments/reviews; GitHub reports no checks for this branch yet.
- `python3 scripts/audit-extensions.py --project-dir . vllm` passes.
- NVIDIA and AMD compose config checks pass with the same placeholder secrets used by CI.
- `python3 tests/test-dependency-pins.py` fails on the two new unpinned `:latest` image refs above.

Lightheartdevs:
Thanks for tightening up the vLLM extension follow-up; this head addresses the two concrete blockers from the earlier advisory pass.

What looks good:
- `ods/extensions/services/vllm/compose.nvidia.yaml` and `ods/extensions/services/vllm/compose.amd.yaml` now bind through `${BIND_ADDRESS:-127.0.0.1}:${VLLM_PORT:-8008}:8000`, which keeps the optional OpenAI-compatible endpoint aligned with the repo's localhost-first exposure posture.
- The vLLM images are now pinned and recorded in `ods/config/dependency-lock.json`, so the dependency-pin contract no longer rejects the new runtime images.
- The change stays scoped to a Pattern-2 optional extension plus dependency-lock metadata; it does not alter core llama-server routing or the installer lifecycle.

What I checked:
- PR metadata, current head SHA, changed files, diff, and the prior advisory comment; the previous marker was for an older head only.
- `git diff --check origin/main...HEAD` passes.
- `python3 scripts/audit-extensions.py --project-dir . vllm` passes.
- `python3 tests/test-dependency-pins.py` passes.
- Docker compose config checks pass for base+nvidia+vLLM and base+amd+vLLM with CI-style placeholder secrets.
- `gh pr checks` currently reports no checks for this branch, so this is based on focused local validation rather than a GitHub check run.

Lightheartdevs:
Thanks for the vLLM work. The need is real, but this implementation is not safe to merge into current ODS.

Current blockers:
- It directs applications away from the stable `ods/current` Switchboard route.
- vLLM is not a recognized backend kind or router endpoint, and there is no transactional adapter for identity/completion proof, publication, rollback, unload, or delete.
- The artifact is a mutable Hugging Face repo downloaded at container start rather than an integrity-pinned managed artifact.
- The container publishes a host port without a complete ODS exposure/auth contract; current-main integration fails with `missing exposure policy entries: ['vllm']`.
- NVIDIA and AMD are claimed together without capability gates or physical ROCm/model/quantization evidence.

A safe successor should land as a series: (1) vLLM backend + immutable HF snapshot catalog contract, initially Linux/NVIDIA only; (2) internal-only digest-pinned runtime behind model-router with generated credentials; (3) transactional `VllmAdapter` and llama-server↔vLLM rollback/lifecycle; (4) existing Models UX integration while every consumer stays on `ods/current`; (5) fresh-install/upgrade plus repeated multi-app swap validation on physical single- and multi-GPU hardware, with AMD added only after ROCm proof.

Closing this branch rather than mechanically rebasing it because the present architecture would create a second inference island. The feature remains worthwhile as the scoped successor above.

PR #1690, opened 2026-07-08
fix(installer): disk free-space check now probes INSTALL_DIR filesystem, not hardcoded $HOME

### LINKED ISSUE #1688: Ask where to install
### Description

My default $HOME has limited memory. I mount another filesystem as DATA for everything other than OS/default files.
I am unable to install the app because it keeps checking free space in my $HOME. I even tried exporting INSTALL_DIR to my own location but it still does not work.

### Use Case

App fails to install because it forces install in $HOME or root fs.

### Proposed Solution

Make INSTALL_DIR configurable within the script i.e. ask the user the install location.

### Alternatives Considered

_No response_

IronicRayquaza:
Fixes #1688

## Problem

The installer always measured available disk space against `$HOME`:

```bash
DISK_AVAIL=$(df -BG "$HOME" | tail -1 | awk '{print $4}' | tr -d 'G')
```

If a user's `$HOME` is on a small or full partition and they set `INSTALL_DIR` to a separate, larger filesystem (for example, `/mnt/data/ods`), the installer would fail the disk space check even though the target filesystem had sufficient free space.

This made it impossible to install ODS to a different volume without modifying the installer.

---

## Solution

`DISK_AVAIL` is now measured against the filesystem where ODS will actually be installed (`INSTALL_DIR`, defaulting to `~/ods`).

Since `INSTALL_DIR` may not exist when the detection phase runs, the installer walks up the directory tree until it finds the nearest existing ancestor before calling `df`.

```bash
_disk_probe_path="${INSTALL_DIR:-$HOME/ods}"

while [[ -n "$_disk_probe_path" ]] && [[ ! -e "$_disk_probe_path" ]]; do
    _disk_probe_path="$(dirname "$_disk_probe_path")"
done

_disk_probe_path="${_disk_probe_path:-$HOME}"

DISK_AVAIL=$(df -BG "$_disk_probe_path" | tail -1 | awk '{print $4}' | tr -d 'G')
```

The installer log now also reports the path that was probed for improved transparency.

---

## Changes

- **`ods/installers/phases/02-detection.sh`**
  - Measure available disk space on the filesystem containing `INSTALL_DIR`.
  - Fall back to the nearest existing parent directory when `INSTALL_DIR` does not yet exist.
  - Log the probe path used during disk detection.

- **`ods/tests/test-disk-check-install-dir.sh`**
  - Added four hermetic unit tests covering all disk detection scenarios.

---

## Testing

```bash
bash ods/tests/test-disk-check-install-dir.sh
```

All tests pass:

- ✅ No `INSTALL_DIR` set → probes `$HOME` (backward compatible)
- ✅ `INSTALL_DIR` already exists → probes it directly
- ✅ `INSTALL_DIR` does not exist → walks up to the nearest existing ancestor
- ✅ `INSTALL_DIR` is on a different mount from `$HOME` → `$HOME` is never probed

---

## Backward Compatibility

No behavioral changes for existing users.

- If `INSTALL_DIR` is unset, the installer still defaults to `~/ods`.
- No changes were made to any other installer phase.
- No changes were made to the Windows installer.

### DIFF
--- a/ods/installers/phases/02-detection.sh
+++ b/ods/installers/phases/02-detection.sh
@@ -115,8 +115,16 @@ else
 fi
 
 # Disk Detection
-DISK_AVAIL=$(df -BG "$HOME" | tail -1 | awk '{print $4}' | tr -d 'G')
-log "Available disk: ${DISK_AVAIL}GB"
+# Check free space on the filesystem where ODS will actually be installed.
+# INSTALL_DIR may not exist yet, so walk up to the nearest existing ancestor
+# so df always receives a valid path. Falls back to $HOME if nothing resolves.
+_disk_probe_path="${INSTALL_DIR:-$HOME/ods}"
+while [[ -n "$_disk_probe_path" ]] && [[ ! -e "$_disk_probe_path" ]]; do
+    _disk_probe_path="$(dirname "$_disk_probe_path")"
+done
+_disk_probe_path="${_disk_probe_path:-$HOME}"
+DISK_AVAIL=$(df -BG "$_disk_probe_path" | tail -1 | awk '{print $4}' | tr -d 'G')
+log "Available disk: ${DISK_AVAIL}GB (on filesystem: $_disk_probe_path)"
 
 # GPU Detection
 if [[ "$GPU_BACKEND_FORCED_CPU" == "true" ]]; then
--- a/ods/tests/test-disk-check-install-dir.sh
+++ b/ods/tests/test-disk-check-install-dir.sh
@@ -1,0 +1,107 @@
+#!/bin/bash
+# Tests that the installer disk check probes the INSTALL_DIR filesystem,
+# not the hardcoded $HOME filesystem. Covers issue #1688.
+#
+# Strategy: stub df so we can assert which path it was called with,
+# then run just the disk-detection block extracted from 02-detection.sh.
+
+set -euo pipefail
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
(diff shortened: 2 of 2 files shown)


Lightheartdevs:
Thanks for fixing the installer disk probe to follow the target install filesystem; that is the right direction for custom `INSTALL_DIR` installs.

Quickest path to merge:
- `ods/installers/phases/02-detection.sh`: apply the same `INSTALL_DIR` ancestor probe inside the early cloud-mode branch. That branch still does `DISK_AVAIL=$(df -BG $HOME ...)` before returning, and phase 04 uses `DISK_AVAIL` for the disk gate. So `INSTALL_DIR=/mnt/data/ods ./install.sh --cloud` can still fail against a small `$HOME` even though the selected install filesystem has room.
- `ods/Makefile`: wire `ods/tests/test-disk-check-install-dir.sh` into the normal `make test` path, near the other installer/preflight contracts, so this regression runs in the repo's standard gate instead of only when someone remembers the standalone script.
- `ods/tests/test-disk-check-install-dir.sh`: either remove the unused `DETECTION_PHASE` variable or fix it to `$ROOT_DIR/installers/phases/02-detection.sh` if you intend to assert against the real phase file. It currently points at `ods/ods/installers/...`, which does not exist, but the test never uses it.

What I checked:
- Diff/static review of `ods/installers/phases/02-detection.sh`, the new disk-check test, and the phase 04 requirements path that consumes `DISK_AVAIL`.
- Local PR worktree: `git diff --check origin/main...HEAD` passed.
- Local focused test with Git Bash: `bash ods/tests/test-disk-check-install-dir.sh` passed all 4 cases.
- `gh pr checks` currently reports no checks for this branch.

Thanks again for taking on the custom install-directory path here.

PR #1694, opened 2026-07-08
feat: add MinIO S3-compatible object storage extension

bong-water-water-bong:
Adds MinIO as an optional service extension for S3-compatible object storage.

**Review feedback addressed:**
- Dependency pin: compose.yaml uses ${MINIO_IMAGE:-...} pattern; pinned image recorded in dependency-lock.json
- Env credentials: MINIO_ROOT_USER/PASSWORD now have safe defaults instead of failing on missing vars
- No unrelated workflow files in this branch

### DIFF
--- a/dream-server/extensions/services/minio/README.md
+++ b/dream-server/extensions/services/minio/README.md
@@ -1,0 +1,35 @@
+# MinIO (S3 Object Storage)
+
+S3-compatible object storage for Dream Server. Use it to store models, datasets,
+backups, and any other large files that need S3 API access from other services.
+
+## Usage
+
+Enable from the dashboard or CLI:
+
+```bash
+dream enable minio
+dream start minio
+```
+
+Requires `MINIO_ROOT_USER` and `MINIO_ROOT_PASSWORD` in `.env`.
+
+## Access
+
+| Interface | URL |
+|---|---|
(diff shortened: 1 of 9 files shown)


Lightheartdevs:
Thanks for taking on an object-storage extension; S3-compatible storage is a useful addition for models, datasets, and backups.

Quickest path to merge:
- Move the extension from `dream-server/extensions/services/minio/*` to `ods/extensions/services/minio/*`. ODS discovers manifests from `ods/extensions/services`, so the current MinIO manifest and compose file will not be loaded by the registry, dashboard, CLI, compose resolver, or extension audit.
- Update the new extension files to the ODS conventions: `schema_version: ods.services.v1`, `compatibility.ods_min`, `container_name: ods-minio`, and docs examples using `ods enable minio` / `ods start minio` instead of `dream ...`.
- Keep `ods/config/dependency-lock.json` scoped to the MinIO entry after the path move; the current diff reorders a large part of the lock file, which makes this harder to audit than the actual service addition.

What I checked:
- Diff/static review of the MinIO manifest/compose/docs and the ODS config registrations.
- Path checks confirmed `ods/extensions/services/...` is the product extension root while the new files are under `dream-server/...`.
- `git diff --check origin/main...origin/pr/1694` passed; `gh pr checks` reports no checks on this branch.

PR #1696, opened 2026-07-08
chore: add Docker image updates to Dependabot config

bong-water-water-bong:
Adds Docker image update coverage via Dependabot. No unrelated workflow files.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -100,3 +100,84 @@ updates:
       ape-python:
         patterns:
           - "*"
+
+  # ── Docker: Core services with custom Dockerfiles ─────────────────
+  - package-ecosystem: "docker"
+    directory: "/dream-server/extensions/services/comfyui"
+    schedule:
+      interval: "weekly"
+    open-pull-requests-limit: 1
+    groups:
+      docker-core:
+        patterns:
+          - "*"
+
+  - package-ecosystem: "docker"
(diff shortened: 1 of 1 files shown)


Lightheartdevs:
Thanks for trying to extend Dependabot coverage to Docker image updates; keeping those base images visible will help maintenance.

Quickest path to merge:
- `.github/dependabot.yml`: change the new Docker update directories from `/dream-server/...` to the real ODS paths. For example, `/ods/extensions/services/dashboard`, `/ods/extensions/services/dashboard-api`, `/ods/extensions/services/comfyui`, `/ods/extensions/services/brave-search`, `/ods/extensions/services/privacy-shield`, `/ods/extensions/services/token-spy`, `/ods/extensions/services/ape`, and `/ods/images/llama-sycl`.
- Include the AMD llama-server Dockerfile at its actual directory if it should be managed here: `/ods/extensions/services/llama-server`. As written, all new Docker entries point to nonexistent directories, so Dependabot will not update the intended Dockerfiles.

What I checked:
- Static review of `.github/dependabot.yml` against the current repo layout.
- Path checks confirmed the `dream-server/...` Dockerfile directories are absent and the corresponding `ods/...` directories exist.
- `git diff --check origin/main...origin/pr/1696` passed; `gh pr checks` reports no checks on this branch.

Thanks again for keeping dependency maintenance moving.

PR #1697, opened 2026-07-08
ci: add Trivy container vulnerability scan

bong-water-water-bong:
Adds Trivy container vulnerability scanning. No unrelated workflow files.

### DIFF
--- a/.github/workflows/trivy-scan.yml
+++ b/.github/workflows/trivy-scan.yml
@@ -1,0 +1,52 @@
+name: Trivy Container Scan
+
+on:
+  pull_request:
+    branches: [main]
+  push:
+    branches: [main]
+  schedule:
+    - cron: "0 6 * * 1"  # every Monday 06:00 UTC
+
+permissions:
+  contents: read
+
+jobs:
+  trivy:
+    name: Scan Dockerfiles for CVEs
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        dockerfile:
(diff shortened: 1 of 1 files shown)


Lightheartdevs:
Thanks for adding container scanning; this is a useful safety net for the service images once it is wired to the current tree.

Quickest path to merge:
- `.github/workflows/trivy-scan.yml`: update the matrix from `dream-server/...` to the real Dockerfile paths under `ods/...` (`ods/extensions/services/...` and `ods/images/llama-sycl/Dockerfile`). The current matrix points to files that do not exist in this branch, so the workflow cannot scan the intended surfaces.
- Decide whether this should scan Dockerfile configuration or built image vulnerabilities. The PR description says HIGH/CRITICAL CVEs, but `scan-type: config` on a Dockerfile is an IaC/config scan, not a built container image vulnerability scan. If the goal is CVE coverage, build or reference the images and use the Trivy mode that scans image vulnerabilities.
- Match the repo’s workflow hardening convention by pinning `aquasecurity/trivy-action` instead of using `@master`; existing workflows pin third-party actions by SHA with a version comment.

What I checked:
- Static review of the new workflow against the current Dockerfile layout and existing workflow pinning style.
- Path checks confirmed the `dream-server/...` matrix paths are absent while the corresponding `ods/...` paths exist.
- `git diff --check origin/main...origin/pr/1697` passed; `gh pr checks` reports no checks on this branch.

Thanks again for adding this security coverage.

PR #1704, opened 2026-07-09
fix(windows-cli): keep other services' compose overlays on enable/disable

tang-vu:
## Summary

On Windows, `ods.ps1 enable <svc>` / `disable <svc>` silently corrupts `.compose-flags`. `Update-ComposeFlags` rebuilds the entire extension section by re-scanning `extensions/services/*/compose.yaml`, which **throws away every per-extension GPU overlay** and **reorders the stack**.

Reproduced against the upstream function, using a `.compose-flags` shaped exactly the way `install-windows.ps1` writes one:

```
ORIGINAL (as written by install-windows.ps1):
  --env-file .env -f docker-compose.base.yml -f docker-compose.nvidia.yml
  -f extensions/services/comfyui/compose.yaml -f extensions/services/comfyui/compose.nvidia.yaml
  -f extensions/services/n8n/compose.yaml
  -f extensions/services/qdrant/compose.yaml
  -f extensions/services/whisper/compose.yaml -f extensions/services/whisper/compose.nvidia.yaml
  -f docker-compose.tier0.yml -f docker-compose.override.yml

AFTER 'ods disable n8n' (upstream code):
  --env-file .env -f docker-compose.base.yml -f docker-compose.nvidia.yml
  -f docker-compose.tier0.yml -f docker-compose.override.yml
  -f extensions/services/comfyui/compose.yaml
  -f extensions/services/qdrant/compose.yaml
  -f extensions/services/whisper/compose.yaml

  FAIL  whisper NVIDIA overlay survives
  FAIL  comfyui NVIDIA overlay survives
  PASS  tier0 overlay survives
  PASS  override overlay survives
  PASS  n8n fragment removed
```

Two distinct defects, one root cause (rebuild instead of edit):

1. **Per-extension GPU overlays are deleted.** Disabling an unrelated service drops `-f extensions/services/whisper/compose.nvidia.yaml` and ComfyUI's. The next `ods start` runs Whisper and ComfyUI **on CPU**. On AMD hosts the same happens to `compose.amd.yaml`.
2. **Merge order is inverted.** `docker-compose.tier0.yml` and `docker-compose.override.yml` get hoisted *ahead of* the extension fragments. The installer appends them last precisely so they win the merge; after a toggle, a user's `docker-compose.override.yml` no longer overrides the extensions.

### Why the rebuild is wrong

The old docstring justified itself like this:

> The fallback intentionally mirrors only what the Windows installer writes: base + GPU overlay + enabled extension compose.yaml entries. It does NOT add GPU-specific per-extension overlays (compose.nvidia.yaml etc.) because those are the canonical resolver's responsibility…

That premise is false. `install-windows.ps1` **does** write per-extension backend overlays next to each fragment (`installers/windows/install-windows.ps1`, the `compose.nvidia.yaml` / `compose.amd.yaml` blocks), and it gates Whisper's CUDA overlay on the detected driver (skipped below driver 575). A filesystem re-scan cannot see either decision, so it deletes them.

The fix edits the file instead of regenerating it: remove only the `-f` pairs under the toggled service's directory, insert a newly enabled fragment ahead of the `tier0`/`override` tail, and leave every other token byte-for-byte. `Update-ComposeFlags` now takes `-ServiceId` and `-Action`, so it knows what changed.

### Why the resolver delegation is removed

`Update-ComposeFlags` preferred `scripts/resolve-compose-stack.sh` and called the rebuild a "fallback". That preferred path never executes, and would be wrong if it did:

- **It cannot run.** On the supported Windows setup (WSL2), `Get-Command bash` resolves to `C:\Windows\system32\bash.exe`. WSL's bash cannot execute a script named by a Windows path. Measured on a real WSL2 box, running the exact invocation from `ods.ps1`:

  ```
  bash    = C:\Windows\system32\bash.exe
  LASTEXITCODE = 127
  output empty? = True
  ---- would Path 1 be taken? ----
  NO - silently falls back to lossy minimal swap
  ```

  A vestigial `$wslInstallDir` was computed two lines above and never used — the path translation was intended and lost.

- **It would be wrong if it did run.** `resolve-compose-stack.sh` emits neither `docker-compose.tier0.yml` (memory limits on low-RAM hosts) nor `installers/windows/docker-compose.windows-amd.yml`; it knows only the Linux layout. It also defaults to `ODS_MODE=local` and `Update-ComposeFlags` never passed `--ods-mode` or `--gpu-count` — the very omission `ods-cli`'s own `_regenerate_compose_flags` warns about: *"without it, toggling any extension would silently strip multi-GPU plumbing from the persisted cache."*

So the resolver's output is not a superset of the Windows stack, and delegating to it would trade one silent loss for another. `ods.ps1` is Windows-only, so the branch is removed rather than repaired.

## AI Assistance

Claude Code was used to investigate the code path, build the PowerShell reproduction harness, draft the fix and the regression test, and run the validation below. I read the final diff, ran the reproduction against upstream and against the fix, and confirmed the WSL `exit 127` measurement on my own machine.

## Release Lane

Stable hotfix reason:

```text
n/a — mainline.
```

## Changed Surface

- [ ] Docs only
- [ ] Tests only
- [ ] Dashboard UI

## Risk And Validation

- Risk level: **High** — `docs/HIGH_RISK_CHANGE_MAP.md` maps compose stack generation and lifecycle commands to High.
- Validation run:
  - [x] `git diff --check`
  - [x] Not required because: the change rewrites one PowerShell function that edits a text file. No compose file, manifest, installer phase, or container is touched.

Commands/results:

```text
# Regression proof — upstream Update-ComposeFlags against an installer-shaped
# .compose-flags (harness lifts the function out of ods.ps1 and runs it):
  FAIL  whisper NVIDIA overlay survives
  FAIL  comfyui NVIDIA overlay survives
  Result: 3 passed, 2 failed

# Same fixture, with this PR:
  PASS  whisper NVIDIA overlay survives
  PASS  comfyui NVIDIA overlay survives
  PASS  tier0 overlay survives
  PASS  override overlay survives
  PASS  --env-file survives
  PASS  n8n fragment removed
  PASS  n8n fragment restored
  PASS  whisper overlay still intact
  PASS  n8n inserted before tier0/override
  PASS  round-trip == original token set
  Result: 10 passed, 0 failed

# Shell suites
bash tests/test-windows-cli-enable-disable.sh   -> PASS (incl. new pwsh behavioral test)
bash tests/test-windows-service-plan.sh         -> PASS
bash tests/test-windows-compose-failure-report.sh -> PASS
bash tests/test-uninstall-compose-flags.sh      -> PASS

# Lint
bash -n over all 269 shell files                 -> 0 failures
Invoke-ScriptAnalyzer -Settings ./PSScriptAnalyzerSettings.psd1 -Severity Error,Warning
  on installers/windows/ods.ps1                  -> no findings (same as upstream baseline)
python -c "yaml.safe_load(open('.github/workflows/test-linux.yml'))" -> OK
git diff --check                                 -> clean
```

## Operational Change Check

Narrower equivalent, rather than a full fleet run: `.compose-flags` is a cached token list. This PR changes how one function edits it and adds a test that asserts the edited result is token-for-token identical to the installer's output after a disable+enable round trip. Nothing about install, bootstrap, image pulls, or host mutation changes.

## Notes For Reviewers

- **`tests/test-windows-cli-enable-disable.sh` was never executed anywhere** — it is referenced by no Makefile target and no workflow. This PR wires it into `test-linux.yml`. That is why the defect shipped with a green build.
- The new behavioral test runs under `pwsh` (present on GitHub's `ubuntu-latest`) and falls back to `powershell` for Windows contributors; it skips cleanly when neither is on PATH.
- **Four static assertions were rewritten**, not deleted for convenience: they asserted the old contract (`Update-ComposeFlags` delegates to the resolver, prepends `--env-file`, passes `GPU_BACKEND`). That contract is what this PR removes, so they are replaced with assertions on the new one.
- **Deliberately out of scope**: `enable` re-adds only `compose.yaml`, not a per-backend overlay, for the newly enabled service — the same as today's behavior for that service. Adding one would mean guessing the driver gate that `install-windows.ps1` applies to Whisper, and the CLI has no access to that decision. Happy to follow up if you want `enable` to consult the manifest instead.
- **Rollback**: revert the commit. `.compose-flags` is a regenerable cache; nothing persists.

### DIFF
--- a/.github/workflows/test-linux.yml
+++ b/.github/workflows/test-linux.yml
@@ -72,6 +72,9 @@ jobs:
       - name: Windows Compose Failure Report Tests
         run: bash tests/test-windows-compose-failure-report.sh
 
+      - name: Windows CLI Enable/Disable Tests
+        run: bash tests/test-windows-cli-enable-disable.sh
+
       - name: Windows Missing Service Hint Tests
         run: bash tests/test-windows-missing-service-hints.sh
 
--- a/ods/installers/windows/ods.ps1
+++ b/ods/installers/windows/ods.ps1
@@ -1727,116 +1727,76 @@ Start-Process -FilePath $_pythonLiteral -ArgumentList `$agentArgs -WorkingDirect
 function Update-ComposeFlags {
     <#
     .SYNOPSIS
-        Regenerate .compose-flags after an enable/disable operation.
+        Update .compose-flags in place after an enable/disable operation.
 
-        Strategy (in priority order):
-        1. If scripts/resolve-compose-stack.sh exists and bash is available,
-           delegate entirely to the canonical resolver (preserves backend
-           overlays, multi-GPU overlays, user-extension overlays, and
-           docker-compose.override.yml -- exactly the same stack the installer
-           built). This is the safe path.
-        2. Otherwise fall back to a minimal in-process swap: keep every token
-           in the existing .compose-flags that is NOT an extension service -f
-           entry, then re-scan extensions/services for enabled compose.yaml
-           fragments and append them. This preserves all backend and GPU
-           overlays (--env-file, -f docker-compose.base.yml,
-           -f docker-compose.nvidia.yml, etc.) because those paths never
-           match 'extensions/services' and are kept verbatim.
+        Only the toggled service's -f entries are rewritten. Every other token
+        is preserved verbatim and in order: --env-file, docker-compose.base.yml,
+        the backend overlay, installers/windows/docker-compose.windows-amd.yml,
+        docker-compose.tier0.yml, docker-compose.override.yml, and the compose
+        fragments of every other extension.
 
-        The fallback intentionally mirrors only what the Windows installer
-        writes: base + GPU overlay + enabled extension compose.yaml entries.
-        It does NOT add GPU-specific per-extension overlays (compose.nvidia.yaml
-        etc.) because those are the canonical resolver's responsibility and
-        we must not silently diverge from it.
+        Editing rather than regenerating is deliberate. The Windows installer
+        records a per-extension GPU overlay (compose.nvidia.yaml /
+        compose.amd.yaml) next to each compose.yaml, and gates some of them on
+        the detected driver -- Whisper's CUDA overlay is skipped below driver
+        575. Rebuilding the extension entries by re-scanning the filesystem
+        cannot see those decisions and would silently drop or resurrect them.
+
+        scripts/resolve-compose-stack.sh is intentionally not consulted here.
+        Its output is not a superset of the Windows stack: it emits neither
+        docker-compose.tier0.yml nor the Windows AMD overlay, and it selects
+        compose.local.yaml overlays from its own --ods-mode default.
     #>
+    param(
+        [Parameter(Mandatory=$true)][string]$ServiceId,
+        [Parameter(Mandatory=$true)][ValidateSet("enable", "disable")][string]$Action
+    )
+
     $flagsFile = Join-Path $InstallDir ".compose-flags"
     if (-not (Test-Path $flagsFile)) {
-        Write-AIWarn "No .compose-flags file found -- skipping regeneration."
+        Write-AIWarn "No .compose-flags file found -- skipping update."
         return
     }
 
-    # ── Path 1: delegate to the canonical resolver ────────────────────────────
-    $resolverScript = Join-Path (Join-Path $InstallDir "scripts") "resolve-compose-stack.sh"
-    $bashExe = Get-Command bash -ErrorAction SilentlyContinue
-    if ((Test-Path $resolverScript) -and $bashExe) {
-        # Read GPU_BACKEND and TIER from .env so the resolver uses the same
-        # parameters that the installer originally selected.
-        $gpuBackend = "nvidia"
-        $tier = "1"
-        try {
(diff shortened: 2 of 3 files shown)


Lightheartdevs:
Thanks for taking on the Windows enable/disable compose-cache path; preserving the installer-recorded overlay order matters a lot here.

What looks good:
- `Update-ComposeFlags` now edits only the toggled service's `-f` entries and preserves the rest of `.compose-flags` byte-for-byte in order, which matches the Windows installer's backend/tier/override assumptions better than regenerating from the filesystem.
- The new behavioral test covers the important failure mode: disabling an unrelated service no longer drops other services' per-extension GPU overlays, and tier0/override stay at the tail where they can win the merge.
- The previously unrun Windows CLI enable/disable test is now wired into `.github/workflows/test-linux.yml`.

What I checked:
- Diff/static review of `ods/installers/windows/ods.ps1`, the Windows CLI test, and the CI wiring.
- Local Git Bash syntax check for `ods/tests/test-windows-cli-enable-disable.sh` passed.
- `gh pr checks 1704` is green, including the new test path, PowerShell lint on Ubuntu and Windows, shell lint, matrix smoke, integration smoke, dashboard checks, and env validation.

Thanks again for narrowing this to the cache mutation that was actually losing overlays.

PR #1705, opened 2026-07-09
fix(bootstrap): pass GPU_COUNT and ODS_MODE when recovering compose flags

tang-vu:
## Summary

When `.compose-flags` is missing, `scripts/bootstrap-upgrade.sh` re-resolves the compose stack and **persists** the result. It called the resolver with only `--tier` and `--gpu-backend`:

```bash
"$INSTALL_DIR/scripts/resolve-compose-stack.sh" \
    --script-dir "$INSTALL_DIR" \
    --tier "$_tier" \
    --gpu-backend "${_gpu_backend:-cpu}" \
    --env
```

Two inputs are missing, and each corrupts the persisted cache in a different way.

**`--gpu-count` — multi-GPU overlays are dropped.** `resolve-compose-stack.sh` sets `GPU_COUNT="1"` as a plain literal, not `${GPU_COUNT:-1}`. It is the only resolver input that ignores the environment entirely, so a caller that omits the flag always resolves a single-GPU stack. Measured against the real resolver on this repo:

```
A) exactly what bootstrap-upgrade.sh runs (no --gpu-count, no --ods-mode)
   multigpu fragments: 0

B) the call every other caller makes (--gpu-count 2 --ods-mode local)
   docker-compose.multigpu-nvidia.yml
   extensions/services/comfyui/compose.multigpu-nvidia.yaml
   extensions/services/embeddings/compose.multigpu-nvidia.yaml
   extensions/services/whisper/compose.multigpu-nvidia.yaml
   multigpu fragments: 4
```

**`--ods-mode` — local-mode overlays are forced onto cloud installs.** The resolver *does* read `ODS_MODE` from the environment, but `bootstrap-upgrade.sh` greps individual keys out of `.env` instead of sourcing it, so `ODS_MODE` is unset in that process and the resolver falls back to `local`:

```
bootstrap default (ODS_MODE unset -> 'local'): local-mode overlays = 6
explicit --ods-mode cloud                    : local-mode overlays = 0
```

Both wrong results are written straight to `$INSTALL_DIR/.compose-flags` and stick until something else regenerates them.

### Reachability

This is not a dead branch. `ods-cli` deletes `.compose-flags` in normal operation:

- when the cache is stale or corrupt (`rm -f "$INSTALL_DIR/.compose-flags"` before falling through to dynamic resolution);
- when `_regenerate_compose_flags` runs and the resolver fails or is absent.

A subsequent model upgrade then hits the recovery branch and re-creates the file — incorrectly.

### Why this shape of fix

`ods-cli` already carries the reason, twice, next to its own resolver calls:

> `--gpu-count` gates the multigpu-{backend}.yml overlay; without it, toggling any extension would silently strip multi-GPU plumbing from the persisted cache.

`bootstrap-upgrade.sh` is the **only** one of the eleven resolver call sites that passes no `--gpu-count` at all. `GPU_COUNT` and `ODS_MODE` are both written to `.env` by the installer (`installers/phases/06-directories.sh`), so the fix reads them exactly the way `TIER` and `GPU_BACKEND` are already read a few lines above, and forwards them.

`scripts/ods-preflight.sh` and `scripts/validate.sh` also omit `--ods-mode`, but both call `load_env_file` first, so `ODS_MODE` reaches the resolver through the environment. They are correct as written and are left alone.

## AI Assistance

Claude Code was used to enumerate the resolver call sites, run the before/after resolver measurements quoted above, draft the fix and the regression test, and run the validation below. I read the final diff and confirmed the new test fails on upstream and passes with the fix.

## Release Lane

Stable hotfix reason:

```text
n/a — mainline.
```

## Changed Surface

- [ ] Docs only
- [ ] Tests only
- [ ] Dashboard UI

## Risk And Validation

- Risk level: **High** — `docs/HIGH_RISK_CHANGE_MAP.md` maps compose stack generation and bootstrap/lifecycle to High.
- Validation run:
  - [x] `git diff --check`
  - [x] Not required because: the change adds two already-persisted `.env` values to one resolver invocation. No compose file, manifest, or installer phase is modified.

Commands/results:

```text
# Regression proof — the new test on upstream bootstrap-upgrade.sh:
$ bash tests/test-bootstrap-upgrade-compose-flags-recovery.sh
[FAIL] resolver called without --gpu-count 2; multi-GPU overlays would be
       stripped from the persisted cache
[resolver] --script-dir /tmp/.../install --tier 3 --gpu-backend nvidia --env

# Same test with this PR:
[PASS] compose-flags recovery forwards GPU_COUNT and ODS_MODE to the resolver

# Neighbouring bootstrap-upgrade suites (no regressions)
test-bootstrap-upgrade-docker-rollback           PASS
test-bootstrap-upgrade-hotswap-contract          PASS
test-bootstrap-upgrade-sha-retry                 PASS
test-bootstrap-upgrade-resume-status             PASS
test-bootstrap-upgrade-download-finalization     PASS
test-bootstrap-upgrade-close-inherited-fds       PASS
test-uninstall-compose-flags                     PASS

# Lint (shellcheck run with the repo's .shellcheckrc and CI's flags)
shellcheck --exclude=SC1091,SC2034 --severity=error   -> NO ERRORS
shellcheck --exclude=SC1091,SC2034 --severity=warning -> NO WARNINGS (new test)
                                                      -> no new warnings in the changed hunk
bash -n on both changed shell files                   -> OK
python -c "yaml.safe_load(open('.github/workflows/test-linux.yml'))" -> OK
git diff --check                                      -> clean
```

## Operational Change Check

Narrower equivalent, rather than a full fleet run: the diff forwards two existing `.env` values into one resolver call. The resolver's own behaviour is unchanged, and the new hermetic test asserts the exact argv it receives plus the contents of the persisted `.compose-flags`. Reproducing the fault needs a multi-GPU or cloud-mode host, which the fleet lane would cover; the argv assertion covers it without one.

## Notes For Reviewers

- The new test drives the real `bootstrap-upgrade.sh` end to end with a fake `docker`/`curl`/`uname` on `PATH` and a stub resolver that records its argv — the same harness style as `tests/test-bootstrap-upgrade-docker-rollback.sh`.
- **Most `test-bootstrap-upgrade-*.sh` suites are not wired into CI or `make test`** (only `close-inherited-fds` and the OpenClaw guard are). I wired the new one into both rather than adding it to the unrun pile. Happy to wire the remaining six in a separate PR if you want them running.
- The fix keeps `bootstrap-upgrade.sh`'s existing "grep one key at a time" style rather than sourcing `.env`, to avoid pulling the whole environment into a script that runs `docker compose`. The script uses `set -uo pipefail` (no `-e`), so a missing key yields an empty string and the explicit `[[ -n ... ]] ||` defaults apply.
- **Rollback**: revert the commit. `.compose-flags` is a regenerable cache.

### DIFF
--- a/.github/workflows/test-linux.yml
+++ b/.github/workflows/test-linux.yml
@@ -133,6 +133,9 @@ jobs:
       - name: Update Rollback Contract
         run: bash tests/test-update-rollback-contract.sh
 
+      - name: Bootstrap Upgrade Compose Flags Recovery Tests
+        run: bash tests/test-bootstrap-upgrade-compose-flags-recovery.sh
+
       - name: Support Bundle Evidence Tests
         run: bash tests/test-support-bundle.sh
 
--- a/ods/Makefile
+++ b/ods/Makefile
@@ -62,6 +62,9 @@ test: ## Run unit and contract tests
 	@echo "=== bootstrap-upgrade spawn closes inherited FDs ==="
 	@bash tests/test-bootstrap-upgrade-close-inherited-fds.sh
 	@echo ""
+	@echo "=== bootstrap-upgrade compose-flags recovery ==="
+	@bash tests/test-bootstrap-upgrade-compose-flags-recovery.sh
+	@echo ""
 	@echo "=== macOS ods-host-agent bootstrap verification ==="
 	@bash tests/test-macos-host-agent-verification.sh
 	@echo ""
--- a/ods/scripts/bootstrap-upgrade.sh
+++ b/ods/scripts/bootstrap-upgrade.sh
@@ -1360,14 +1360,26 @@ elif [[ -n "$DOCKER_CMD" ]] && $DOCKER_CMD ps --filter name=ods-llama-server --f
         read -ra COMPOSE_ARGS <<< "$(cat "$INSTALL_DIR/.compose-flags")"
     elif [[ -x "$INSTALL_DIR/scripts/resolve-compose-stack.sh" ]]; then
         _tier="1"
+        _gpu_count="1"
+        _ods_mode="local"
         if [[ -f "$ENV_FILE" ]]; then
             _tier=$(grep -E '^TIER=' "$ENV_FILE" | cut -d= -f2 | tr -d '"\047\r')
             [[ -n "$_tier" ]] || _tier="1"
+            _gpu_count=$(grep -E '^GPU_COUNT=' "$ENV_FILE" | cut -d= -f2 | tr -d '"\047\r')
+            [[ -n "$_gpu_count" ]] || _gpu_count="1"
+            _ods_mode=$(grep -E '^ODS_MODE=' "$ENV_FILE" | cut -d= -f2 | tr -d '"\047\r')
+            [[ -n "$_ods_mode" ]] || _ods_mode="local"
         fi
+        # --gpu-count gates the multigpu-{backend} overlays and --ods-mode gates
+        # the compose.local.yaml overlays. The resolver reads neither from the
+        # environment, and the flags below are persisted to .compose-flags, so
+        # omitting them writes a single-GPU local-mode stack over the real one.
         _resolved_env=$("$INSTALL_DIR/scripts/resolve-compose-stack.sh" \
             --script-dir "$INSTALL_DIR" \
             --tier "$_tier" \
             --gpu-backend "${_gpu_backend:-cpu}" \
+            --gpu-count "$_gpu_count" \
+            --ods-mode "$_ods_mode" \
             --env 2>/dev/null || true)
         _resolved_flags=$(printf '%s\n' "$_resolved_env" | sed -n 's/^COMPOSE_FLAGS="\([^"]*\)".*/\1/p')
         if [[ -n "$_resolved_flags" ]]; then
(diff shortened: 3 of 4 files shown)


Lightheartdevs:
Thanks for tightening the bootstrap-upgrade compose-flags recovery path; this is a small fix on a high-impact lifecycle edge.

What looks good:
- `ods/scripts/bootstrap-upgrade.sh` now forwards the same persisted inputs the resolver needs: `GPU_COUNT` for multi-GPU overlays and `ODS_MODE` for local/cloud overlay selection.
- The new regression test drives the missing `.compose-flags` recovery path with a stub resolver and asserts the exact argv plus the persisted multigpu flag output.
- The test is wired into both `ods/Makefile` and `.github/workflows/test-linux.yml`, so this does not become another unrun bootstrap-upgrade regression.

What I checked:
- Diff/static review against the repo's resolver/compose-cache conventions and existing `ods-cli` resolver call comments.
- Local Git Bash syntax checks for `ods/scripts/bootstrap-upgrade.sh` and `ods/tests/test-bootstrap-upgrade-compose-flags-recovery.sh` passed.
- `gh pr checks 1705` is green across shell lint, Python lint, dashboard API/frontend, matrix smoke, integration smoke, PowerShell lint, and env validation.

Thanks again for making the recovery behavior explicit and tested.

PR #1708, opened 2026-07-09
fix(rootless): auto-fix data-directory ownership for Docker rootless mode 

IronicRayquaza:
Fixes #1702

## Summary

Fixes permission issues when running ODS with Docker in rootless mode.

In rootless Docker, container UIDs are remapped through the host user's subordinate UID range (typically starting at `100000`). As a result, data directories created by the installer are owned by the host user (for example, UID `1000`) and become unwritable by containers running as non-root users.

This PR adds automatic detection and repair of rootless Docker ownership during installation, provides a standalone repair command, surfaces warnings in preflight and doctor diagnostics, and adds comprehensive test coverage.

#### `lib/rootless-fix.sh` (new)

- Added `ods_is_rootless_docker` to detect Docker rootless mode using `docker info`.
- Added `_ods_rootless_chown_dir` to repair ownership through a short-lived Alpine container without requiring `sudo`.
- Added `ods_fix_rootless_ownership`, an idempotent repair routine that:
  - Runs only when Docker is operating in rootless mode.
  - Iterates the `ODS_ROOTLESS_UID_MAP`.
  - Repairs ownership for all affected service directories:
    - n8n
    - hermes
    - whisper
    - tts
    - token-spy
    - privacy-shield
    - ape
    - langfuse
    - openclaw
  - Repairs the OpenClaw workspace configuration directory.
  - Continues even if individual directory repairs fail.
- Added `ods_warn_rootless_docker` to display a user-facing advisory.

#### `installers/phases/06-directories.sh`

- Sources `rootless-fix.sh` during installation.
- Runs `ods_fix_rootless_ownership` after all data directories are created and the existing Token Spy ownership adjustment completes.

#### `ods-preflight.sh`

- Detects Docker rootless mode after daemon validation.
- Displays a warning describing affected services.
- Suggests the recovery command:

```bash
ods repair rootless-ownership
```

#### `scripts/ods-doctor.sh`

- Detects rootless Docker (`DOCKER_ROOTLESS`).
- Includes the value in the generated JSON report as:

```json
runtime.docker_rootless
```

- Increments `runtime_warnings` when rootless mode is detected.
- Adds an autofix hint recommending:

```bash
ods repair rootless-ownership
```

#### `ods-cli`

- Added:

```bash
ods repair rootless-ownership
```

- Added alias:

```bash
ods rootless
```

- Updated CLI help and examples.

#### `tests/test-rootless-docker-ownership.sh` (new)

Adds 30 tests covering:

[...]

PR #1709, opened 2026-07-09
Support Docker rootless mode for non-root container services

AniketBajaj2024:
#Summary

Non-root container services (n8n, hermes, whisper, tts, token-spy, privacy-shield, ape, langfuse) fail with EACCES: permission denied on Docker rootless installs. In rootless mode, container UIDs are remapped — container UID 1000 maps to host UID 100999, not host UID 1000 — so directories created by the installer (owned by the host user) are inaccessible to non-root containers.

Fix by detecting rootless mode and using a helper container (--user 0:0, which maps to the host user in rootless) to chown data directories to the correct container UID before services start. Adds ROOTLESS_SERVICE_UIDS to constants.sh as a single source of truth for the service→UID map.

Affected files: installers/lib/constants.sh, installers/phases/06-directories.sh, ods-cli, ods-preflight.sh, extensions/services/langfuse/hooks/post_install.sh

AI Assistance

Gemini helped analyse the rootless UID remapping mechanics, identify all affected services and their container UIDs.

#Release Lane

- [x] Stable hotfix targeting release/2.5.x
- [ ] Mainline change targeting main

#Stable hotfix reason:

Docker rootless is increasingly adopted as a security best practice.
Affected users cannot start n8n, hermes, whisper, tts, token-spy,
privacy-shield, ape, or langfuse at all — core stack is broken on
install. No workaround without manual docker run commands per service.

#Changed Surface

- [ ] Docs only
- [ ] Tests only
- [ ] Dashboard UI

#Risk And Validation

- Risk level: Low
- Validation run:
  - [x] git diff --check
  - [ ] Stable-lane patch validation, if targeting release/2.5.x
  - [ ] Not required because:

#Commands/results:

# Non-rootless path — verify no regression
./install.sh --dry-run
ods start
ods restart
ods-preflight.sh

# Rootless path — verify fix
dockerd-rootless.sh
ods start   # all non-root services should start without EACCES
ods restart # ownership fix re-applied cleanly
ods-preflight.sh # should pass rootless ownership check

#Operational Change Check

Notes For Reviewers

- The fix is a no-op on non-rootless installs — _is_rootless_docker returns early, nothing else runs
- ods-cli and ods-preflight.sh cannot source constants.sh (standalone scripts), so they carry inline copies of the UID map with a comment to keep them in sync
- Langfuse has 3 non-root users (nextjs/1001, postgres/70, clickhouse/101) — all three are covered
- The helper container approach (alpine sh -c "chown -R uid:uid /data") is the only reliable method in rootless mode since host-side chown and sudo chown cannot reach remapped UIDs

### DIFF
--- a/ods/extensions/services/langfuse/hooks/post_install.sh
+++ b/ods/extensions/services/langfuse/hooks/post_install.sh
@@ -58,6 +58,21 @@ if [[ "$(id -u)" -ne 0 ]]; then
     fi
 fi
 
+# In rootless Docker mode, sudo chown cannot reach container UIDs.
+# Use a helper container running as root instead (root in container = host user).
+_is_rootless_docker() {
+    docker info --format '{{.SecurityOptions}}' 2>/dev/null | grep -q "rootless"
+}
+
+_rootless_chown_dir() {
+    local dir="$1" owner="$2"  # owner format: uid:gid
+    [[ -d "$dir" ]] || return 0
+    docker run --rm --user 0:0 \
+        -v "$dir:/data" \
+        alpine sh -c "chown -R ${owner} /data" 2>/dev/null || \
+        log "WARNING: rootless chown failed for $dir (non-fatal)"
+}
+
 is_container_running() {
     # Returns:
     #   0 = container is running
@@ -142,10 +157,18 @@ chown_dir() {
 }
 
 # postgres official image: uid 70 (postgres user baked into image)
-chown_dir "$POSTGRES_DIR" "70:70" "ods-langfuse-postgres"
+if _is_rootless_docker; then
+    _rootless_chown_dir "$POSTGRES_DIR" "70:70"
+else
+    chown_dir "$POSTGRES_DIR" "70:70" "ods-langfuse-postgres"
+fi
 
 # clickhouse-server image: uid 101 (clickhouse user)
-chown_dir "$CLICKHOUSE_DIR" "101:101" "ods-langfuse-clickhouse"
+if _is_rootless_docker; then
+    _rootless_chown_dir "$CLICKHOUSE_DIR" "101:101"
+else
+    chown_dir "$CLICKHOUSE_DIR" "101:101" "ods-langfuse-clickhouse"
+fi
 
 log "done"
 exit 0
--- a/ods/installers/lib/constants.sh
+++ b/ods/installers/lib/constants.sh
@@ -84,4 +84,5 @@ ROOTLESS_SERVICE_UIDS=(
     "langfuse/postgres:70"
     "langfuse/clickhouse:101"
     "hermes:10000"
+    "comfyui:1000"
 )
--- a/ods/installers/lib/constants.sh
+++ b/ods/installers/lib/constants.sh
@@ -66,3 +66,22 @@ _sed_i() {
         sed -i '' "$@"
     fi
 }
+
+
+#=============================================================================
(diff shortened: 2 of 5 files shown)


Lightheartdevs:
Thanks for digging into rootless Docker ownership; that is exactly the kind of install edge case that can leave users with opaque container EACCES failures.

Quickest path to merge:
- `ods/ods-cli:3224-3233`: restore the missing `fi` in `cmd_backup`. With the current head, `bash -n` fails with `syntax error near unexpected token '}'`, so the CLI will not parse at all.
- `ods/installers/phases/06-directories.sh:57-75`: keep `--dry-run` side-effect-free. This block is inside `if $DRY_RUN; then` but now calls `_is_rootless_docker`, `_rootless_chown_data_dir`, `docker run`, `sudo chown`, and `sudo chmod`. The dry-run branch should log what would happen only; move real ownership repair into the non-dry-run path.
- Add a focused regression test and wire it into CI. This is installer/lifecycle behavior and the PR targets a stable hotfix lane, but `gh pr checks 1709` currently reports no checks for the branch.
- Consider consolidating the UID map/repair approach with the other open rootless Docker fix so the repo does not land two divergent sources of truth for the same service ownership contract.

What I checked:
- Diff/static review of phase 06, `ods-cli`, `ods-preflight.sh`, and the Langfuse hook.
- Local syntax check via Git Bash on the PR copy of `ods/ods-cli`; it failed at the unclosed `cmd_backup` conditional.
- GitHub check status: no checks reported for this branch.

Thanks again for taking on the rootless path.

AniketBajaj2024:
Thanks for the review 

Fixed in the latest push:

1. Added the missing `fi` in cmd_backup — confirmed `bash -n` passes clean now.
2. Moved all real ownership operations out of the DRY_RUN path in 06-directories.sh.
   The dry-run branch now only logs what would happen; the rootless/non-rootless
   hermes handling is in the else branch where it belongs.

Lightheartdevs:
Thanks for circling back on the rootless ownership branch and fixing the parse/dry-run problems from the earlier review.

Quickest path to merge:
- `ods/Makefile:19-72` / `ods/tests`: add a focused regression and wire it into `make test` or CI. This PR changes installer/lifecycle ownership behavior, but it currently adds no tests and `gh pr checks 1709` still reports no checks.
- `ods/installers/phases/06-directories.sh:40-125`, `ods/installers/lib/constants.sh:76-88`, and `ods/ods-cli`: reconcile this implementation with #1708 before merging. Right now there are two active rootless ownership approaches with separate UID maps/helpers; landing both would create duplicate sources of truth for the same service ownership contract. Either rebase onto the shared helper approach or make the case that this narrower inline implementation should replace it.
- If this PR is intended to be the selected rootless fix, include the host-agent networking side of the rootless issue (`ODS_AGENT_BIND` / `ODS_AGENT_HOST`) or explicitly scope it out so maintainers know dashboard-api-to-host-agent reachability still needs a separate PR.

What I checked:
- Updated diff/static review of phase 06, constants, `ods-cli`, preflight, and the Langfuse hook.
- Local Git Bash syntax: `bash -n` passed for the changed shell files.
- GitHub check status: no checks reported for this branch.

Thanks again for pushing on the rootless install edge case.

gabsprogrammer:
Superseded by #2221, rebased on current main with active-service scoping, service-specific nested UID contracts, install/start/restart/late-enable coverage, fail-closed safety checks, and regression tests.

PR #1710, opened 2026-07-09
feat: sync existing models from LM Studio and Ollama during setup

AniketBajaj2024:
### Summary

During ODS setup, the installer unconditionally downloads a bootstrap LLM (~1.5 GB) and a full tier-appropriate model even when the user already has identical GGUF files in LM Studio or Ollama. This PR adds a sync layer that checks both external stores before any download, and symlinks the existing file into $INSTALL_DIR/data/models/ so the installer's existing file-presence checks short-circuit the download.

### What changed:
- ods/scripts/sync-external-models.sh (new): searches LM Studio model directories and the Ollama blob store for a matching GGUF by exact filename then fuzzy base-name match. On a match it creates a symlink (falls back to copy if symlinks are unavailable). Outputs a machine-readable result line (synced:lmstudio:<path>, synced:ollama:<path>, already_present, not_found) consumed by callers.
- ods/installers/phases/11-services.sh: calls the sync script before both the bootstrap model download and the background full-model download, so either or both can be skipped.
- ods/installers/lib/bootstrap-model.sh: adds sync_bootstrap_if_available() so the fast-start bootstrap model is also checked against external stores before bootstrap_needed() is evaluated.
- ods/ods-cli: adds ods model sync [file.gguf] subcommand for on-demand syncing after initial install.
- ods/bin/ods-host-agent.py: adds POST /v1/model/sync endpoint so the dashboard API (inside Docker) can trigger syncs that require host filesystem access.
- ods/extensions/services/dashboard-api/routers/models.py: adds POST /api/models/sync endpoint that proxies to the host agent.

### Release Lane

- [ ] Stable hotfix targeting release/2.5.x
- [x] Mainline change targeting main

### Changed Surface

- [ ] Docs only
- [ ] Tests only
- [ ] Dashboard UI

### Risk And Validation

- Risk level: Medium
- Validation run:
  - [x] git diff --check
  - [ ] Stable-lane patch validation, if targeting release/2.5.x
  - [ ] Not required because:

### Commands/results:

# Shell syntax validation — all 4 modified/new shell files passed
$ bash -n ods/scripts/sync-external-models.sh && echo OK   # OK
$ bash -n ods/installers/lib/bootstrap-model.sh && echo OK # OK
$ bash -n ods/installers/phases/11-services.sh && echo OK  # OK
$ bash -n ods/ods-cli && echo OK                           # OK

# Python files were visually reviewed; python3 not available on this machine
# for automated syntax check

### Operational Change Check

The sync script is purely additive — it only runs before existing download logic and uses || true so any failure is non-fatal and falls through to the normal download path. The installer's behavior is unchanged when LM Studio/Ollama are absent or the model is not found. Full fleet validation on Linux (NVIDIA + AMD) and macOS is recommended before release.

### DIFF
--- a/ods/Makefile
+++ b/ods/Makefile
@@ -70,6 +70,9 @@ test: ## Run unit and contract tests
 	@echo ""
 	@echo "=== fleet-multi-distro keep-going on docker pull failure ==="
 	@bash tests/test-fleet-multi-distro-keep-going.sh
+	@echo ""
+	@echo "=== External model sync (LM Studio / Ollama) ==="
+	@bash tests/test-sync-external-models.sh
 
 bats: ## Run BATS unit tests for shell libraries
 	@echo "=== BATS unit tests ==="
--- a/ods/bin/ods-host-agent.py
+++ b/ods/bin/ods-host-agent.py
@@ -4019,15 +4019,24 @@ class AgentHandler(BaseHTTPRequestHandler):
             json_response(self, 501, {"error": "sync-external-models.sh not found"})
             return
 
+        bash = _find_usable_bash()
+        if not bash:
+            json_response(self, 503, {"error": "Sync requires a usable Bash runtime. Install Git Bash or run ODS through WSL/Linux."})
+            return
+
         try:
             result = subprocess.run(
-                ["bash", _to_bash_path(sync_script), gguf_file],
+                [bash, _to_bash_path(sync_script), gguf_file],
                 capture_output=True,
                 text=True,
                 timeout=30,
                 env={**os.environ, "INSTALL_DIR": _to_bash_path(INSTALL_DIR)},
             )
             stdout = result.stdout.strip()
+
+            if result.returncode != 0:
+                json_response(self, 502, {"error": f"Sync script exited {result.returncode}: {result.stderr.strip() or stdout}"})
+                return
 
             if stdout.startswith("synced:lmstudio:"):
                 source = stdout[len("synced:lmstudio:"):]
@@ -4037,8 +4046,10 @@ class AgentHandler(BaseHTTPRequestHandler):
                 json_response(self, 200, {"status": "synced", "provider": "ollama", "source": source, "gguf_file": gguf_file})
             elif stdout == "already_present":
                 json_response(self, 200, {"status": "already_present", "gguf_file": gguf_file})
-            else:
+            elif stdout == "not_found":
                 json_response(self, 200, {"status": "not_found", "gguf_file": gguf_file})
+            else:
+                json_response(self, 502, {"error": f"Unexpected sync output: {stdout!r}"})
         except subprocess.TimeoutExpired:
             json_response(self, 504, {"error": "Sync script timed out"})
 
--- a/ods/bin/ods-host-agent.py
+++ b/ods/bin/ods-host-agent.py
@@ -1665,6 +1665,8 @@ class AgentHandler(BaseHTTPRequestHandler):
             self._handle_model_activate()
         elif self.path == "/v1/model/delete":
             self._handle_model_delete()
+        elif self.path == "/v1/model/sync":
+            self._handle_model_sync()
         elif self.path == "/v1/compose/invalidate-cache":
             self._handle_invalidate_compose_cache()
         elif self.path == "/v1/env/update":
@@ -3985,6 +3987,60 @@ class AgentHandler(BaseHTTPRequestHandler):
             json_response(self, 200, {"status": "deleted", "gguf_file": gguf_file})
         except OSError as exc:
             json_response(self, 500, {"error": f"Failed to delete: {exc}"})
+
+
+    def _handle_model_sync(self):
+        """Sync a GGUF model from LM Studio or Ollama into ODS's model directory.
+
+        Creates a symlink (or falls back to a copy) so the file is immediately
+        available to llama-server without a separate network download.
+
+        Body: {"gguf_file": "Qwen3.5-2B-Q4_K_M.gguf"}
+        Response: {"status": "synced"|"already_present"|"not_found", "source": ...}
+        """
+        if not check_auth(self):
+            return
+        body = read_json_body(self)
+        if body is None:
+            return
+
+        gguf_file = (body.get("gguf_file") or "").strip()
+        if not gguf_file:
+            json_response(self, 400, {"error": "gguf_file is required"})
+            return
+
+        # Path traversal prevention — only plain filenames allowed
+        if "/" in gguf_file or "\\" in gguf_file or not gguf_file.lower().endswith(".gguf"):
+            json_response(self, 400, {"error": "gguf_file must be a plain .gguf filename"})
+            return
+
+        sync_script = INSTALL_DIR / "scripts" / "sync-external-models.sh"
+        if not sync_script.is_file():
+            json_response(self, 501, {"error": "sync-external-models.sh not found"})
+            return
+
+        try:
+            result = subprocess.run(
+                ["bash", _to_bash_path(sync_script), gguf_file],
+                capture_output=True,
+                text=True,
+                timeout=30,
+                env={**os.environ, "INSTALL_DIR": _to_bash_path(INSTALL_DIR)},
+            )
+            stdout = result.stdout.strip()
+
+            if stdout.startswith("synced:lmstudio:"):
+                source = stdout[len("synced:lmstudio:"):]
+                json_response(self, 200, {"status": "synced", "provider": "lmstudio", "source": source, "gguf_file": gguf_file})
+            elif stdout.startswith("synced:ollama:"):
+                source = stdout[len("synced:ollama:"):]
+                json_response(self, 200, {"status": "synced", "provider": "ollama", "source": source, "gguf_file": gguf_file})
+            elif stdout == "already_present":
+                json_response(self, 200, {"status": "already_present", "gguf_file": gguf_file})
+            else:
+                json_response(self, 200, {"status": "not_found", "gguf_file": gguf_file})
+        except subprocess.TimeoutExpired:
+            json_response(self, 504, {"error": "Sync script timed out"})
 
 
 def _check_lemonade_health(body: str) -> bool:
--- a/ods/bin/ods-host-agent.py
+++ b/ods/bin/ods-host-agent.py
@@ -3992,7 +3992,7 @@ class AgentHandler(BaseHTTPRequestHandler):
     def _handle_model_sync(self):
         """Sync a GGUF model from LM Studio or Ollama into ODS's model directory.
 
-        Creates a symlink (or falls back to a copy) so the file is immediately
+        Hardlinks (or copies if cross-filesystem) the file so it is immediately
         available to llama-server without a separate network download.
 
         Body: {"gguf_file": "Qwen3.5-2B-Q4_K_M.gguf"}
--- a/ods/extensions/services/dashboard-api/routers/models.py
+++ b/ods/extensions/services/dashboard-api/routers/models.py
@@ -707,6 +707,32 @@ async def _run_current_model_benchmark(model_id: str, max_tokens: int) -> dict:
     }
 
 
+@router.post("/api/models/sync")
+def sync_models(body: dict[str, Any] | None = None, api_key: str = Depends(verify_api_key)):
+    """Sync a GGUF model from LM Studio or Ollama into ODS's local model store.
+
+    Avoids downloading a model that already exists on the device in another tool.
+    If no gguf_file is specified, the currently configured model is synced.
+
+    Returns:
+      {"status": "synced", "provider": "lmstudio"|"ollama", "source": "<path>", "gguf_file": "..."}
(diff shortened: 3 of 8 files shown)


Lightheartdevs:
Thanks for taking on model reuse during setup; avoiding duplicate GGUF downloads is a useful install-time improvement.

Quickest path to merge:
- `ods/scripts/sync-external-models.sh:109`, `:125`, and `:288`: avoid symlinking external provider paths into `data/models` for Docker-managed inference. `ods/docker-compose.base.yml` mounts only `./data/models:/models` and starts llama-server with `/models/${GGUF_FILE}`, so a symlink to `~/.lmstudio/...` or the Ollama blob store can look present to the host installer while resolving to a non-existent path inside the container. Copy into `data/models`, hardlink when safe, or add container-valid mounts before letting the download short-circuit.
- `ods/scripts/sync-external-models.sh:94` and `:158`: replace the Bash 4-only `mapfile` and `${base,,}` usage, or guard the script by Bash version before it is called from installer paths. The repo still keeps installer shell portable across macOS/Git Bash/WSL-style environments, and this new script advertises those paths directly.
- Add focused coverage for the new behavior and wire it into CI: no-provider no-op, LM Studio exact match, fuzzy match handling, Ollama fallback, and the important assertion that the final model path is readable from the runtime surface that consumes `/models/${GGUF_FILE}`.

What I checked:
- Diff/static review of the installer, bootstrap-model, host-agent, dashboard-api, CLI, and new sync script.
- Repo context from `ARCHITECTURE.md`, `ods/CONTRIBUTING.md`, `ods/docs/HIGH_RISK_CHANGE_MAP.md`, and CI workflow definitions.
- `gh pr checks 1710` currently reports no checks for this branch, so I did not treat the shell/Python changes as validated.

AniketBajaj2024:
Thanks for the detailed review —

Docker symlink issue: Replaced ln -sf with ln (hardlink) + cp fallback in all three write paths (LM Studio exact, LM Studio fuzzy, Ollama blob). Hardlinks share the inode, so the file is fully readable inside the container's /models mount regardless of where the source lived on the host. cp kicks in automatically when source and destination are on different filesystems.

Bash 3.2 compat: Replaced mapfile with a while IFS= read -r loop and ${base,,} with tr '[:upper:]' '[:lower:]'.

Tests + CI wiring: Added tests/test-sync-external-models.sh covering the cases you listed — no-provider no-op, LM Studio exact match, fuzzy match, Ollama dead-port fallback — plus a dedicated assertion that the synced destination is a real file and not a symlink. Wired into make test.

Lightheartdevs:
Thanks for updating the external model sync path; switching away from symlinks and adding a local regression test closes the biggest runtime concern from the first pass.

Quickest path to merge:
- `ods/ods-cli:3105` and `ods/scripts/sync-external-models.sh:354-360`: make the not-found path survive `set -e`. The sync script prints `not_found` but exits 1; `ods model sync` captures it without tolerating that nonzero exit, so the CLI exits after Searching... and never reaches the user-facing `not_found` guidance at `ods/ods-cli:3120-3124`. Either make `not_found` exit 0 for the script's contract or capture the command with a non-fatal fallback before the `case`.
- Get CI/checks to run for this branch, or have a maintainer rerun them. This touches phase 11 model selection, host-agent, dashboard API, and `ods-cli`; `gh pr checks 1710` currently reports no checks, so the high-risk install/model path is only locally validated.

What I checked:
- Diff/static review of `sync-external-models.sh`, phase 11, bootstrap model logic, host-agent, dashboard API, and `ods-cli`.
- Local Git Bash syntax passed for the changed shell files, and `bash tests/test-sync-external-models.sh` passed with 17 checks.
- Reproduced the CLI not-found issue with a minimal `ODS_HOME`: `ods model sync MissingModel-Q4_K_M.gguf` exited 1 after Searching... without printing the not-found instructions.

AniketBajaj2024:
Thanks for the detailed review.

Fix applied: Changed not_found) exit 1 → not_found) exit 0 in sync-external-models.sh (line 360). not_found is a valid, definitive result from the script — not an error condition — so exiting 1 was the wrong contract. With set -e active in ods-cli, the $() command substitution propagated that exit code through the assignment at line 3105, killing the CLI before it could reach the user-facing guidance at lines 3120–3124.

Used exit 0 rather than || true on the CLI side to stay consistent with the project's error-handling rules (no || true).

Regression test added: _test_cli_not_found_exit_code in tests/test-sync-external-models.sh runs the script as a real subprocess and asserts both that it prints not_found and exits 0 — directly covering the failure path you reproduced. Suite is now 19/19.

Lightheartdevs:
Thanks for iterating on the external model sync path; the latest head closes the earlier symlink/runtime visibility and `set -e` result-handling issues.

Quickest path to merge:
- `ods/scripts/sync-external-models.sh:128-140`: tighten or remove the fuzzy LM Studio match before it can satisfy installer download checks. That block accepts any quantization for the same base model, copies/hardlinks it to the requested filename, and the installer then treats `$INSTALL_DIR/data/models/$GGUF_FILE` as present. On Linux the later `sha256sum` check usually catches the mismatch, but `ods/installers/phases/11-services.sh:414-432` skips verification when `sha256sum` is unavailable, which is a normal macOS-style surface in this repo. That can let, for example, a Q8/F16 LM Studio file be installed as the Q4 target and skip the intended download. The smallest safe fix is exact filename only for installer sync, or require a verified hash match before a fuzzy source can short-circuit the download.
- `ods/bin/ods-host-agent.py:4023-4041`: use the existing `_find_usable_bash()` helper and handle nonzero sync-script exits explicitly. Other host-agent paths already avoid plain `bash` because Windows can have a missing or unusable launcher; this endpoint currently calls `bash` directly, only catches timeouts, and maps any unexpected stdout/nonzero failure to `not_found`. Return a clear 5xx for missing/unusable Bash or script failure so the dashboard does not silently report “not found” when sync actually failed.

What I checked:
- Diff/static review of `sync-external-models.sh`, phase 11, bootstrap-model, `ods-cli`, host-agent, dashboard-api, and the new sync tests.
- Repo context from `ARCHITECTURE.md`, `ods/CONTRIBUTING.md`, `ods/docs/HIGH_RISK_CHANGE_MAP.md`, `ods/docs/TESTING.md`, and CI workflow definitions.
- `git diff --check origin/main...HEAD` passed and Python AST parsing passed for the changed Python files.
- I could not rerun the Bash test suite in this Windows runner because the available `bash.exe` path fails through WSL (`/bin/bash` missing), and `gh pr checks 1710` still reports no checks for this branch.

AniketBajaj2024:
Both issues from this pass are addressed.

Added a SYNC_EXACT_ONLY env var guard around Pass 2. When true, the fuzzy/family match is skipped and the function falls through to not_found — a differently-quantized file can no longer satisfy the model-presence check and silently bypass the sha256 download. Installer callers opt in explicitly:

- installers/phases/11-services.sh: SYNC_EXACT_ONLY=true in the env
- installers/lib/bootstrap-model.sh: SYNC_EXACT_ONLY=true in the env

User-initiated paths (ods model sync, host-agent /model-sync) leave it unset — fuzzy matching stays available where the user is explicitly requesting a sync.

Host-agent bash handling (ods-host-agent.py:4023–4041)

- Replaced ["bash", ...] with [_find_usable_bash(), ...], consistent with every other shell invocation in the file
- Returns 503 if _find_usable_bash() finds nothing, before touching subprocess.run
- Checks result.returncode != 0 immediately after the call and returns 502 with stderr/stdout — script failures no longer reach the result-parsing block
- Added an explicit elif stdout == "not_found" arm so the catch-all else is now strictly for unexpected output and returns 502 instead of 200/not_found

Test suite is 21/21, including _test_exact_only_blocks_fuzzy (verifies the guard blocks the wrong-quantization path and leaves no file on disk) and the CLI exit-code regression from the previous pass.

Lightheartdevs:
Thanks for tightening the external model sync path again; the hardlink/copy behavior, `set -e` handling, and host-agent Bash selection are all in much better shape on this head.

Quickest path to merge:
- `ods/installers/phases/11-services.sh:1027-1029`: carry the same `SYNC_EXACT_ONLY=true` guard into the late full-model sync before the background download. The initial installer sync at `:354` and bootstrap sync in `ods/installers/lib/bootstrap-model.sh:67` now correctly block fuzzy LM Studio matches, but this later call still invokes `sync-external-models.sh` without the guard. That means a differently quantized LM Studio file can be copied as `$FULL_GGUF_FILE`, set `_BOOTSTRAP_ACTIVE=false`, and skip the intended background download in the same installer/model path this PR is protecting.
- Add a small regression around that caller or its contract: with only `TestModel-Q8_0.gguf` present and the requested full model set to `TestModel-Q4_K_M.gguf`, the installer late-sync path should leave the destination absent and continue toward the normal download.
- After that patch, please get CI/checks to run or have a maintainer rerun them; `gh pr checks 1710` still reports no checks for this branch, and this touches phase 11, host-agent, dashboard API, and model lifecycle behavior.

What I checked:
- Repo context from `README.md`, `ARCHITECTURE.md`, `CLAUDE.md`, release-validation docs, Makefile, and CI workflow definitions.
- Diff/static review of `sync-external-models.sh`, `11-services.sh`, `bootstrap-model.sh`, `ods-cli`, host-agent, dashboard API, and the sync test.
- Git Bash syntax plus `bash tests/test-sync-external-models.sh` passed: 21/21.
- `python -m py_compile ods/bin/ods-host-agent.py ods/extensions/services/dashboard-api/routers/models.py` passed.
- Local repro confirmed the missing guard matters: without `SYNC_EXACT_ONLY`, a Q8 LM Studio file syncs as a requested Q4 file; with the guard, the same fixture returns `not_found`.

AniketBajaj2024:
Thanks for the thorough review across all three rounds. Here's a summary of everything addressed in this final pass:

Stale documentation fixed:

sync-external-models.sh header and 11-services.sh comment both said "creates a symlink" — the implementation has always hardlinked/copied (symlinks were ruled out because they break inside Docker's ./data/models:/models mount). Same fix in the _handle_model_sync docstring in ods-host-agent.py.

Removed || true from both phase-11 sync call sites — the script exits 0 for all normal outcomes (synced, already_present, not_found) so the guards were redundant and silently masked any unexpected non-zero exit.

bootstrap-model.sh sync call: replaced 2>/dev/null || true with 2>>"${LOG_FILE:-/dev/stderr}" so sync errors reach the installer log instead of disappearing silently. Falls back to stderr if LOG_FILE isn't set.

_ollama_model_blob_digest: replaced 2>/dev/null || true with || _warn "..." and narrowed except Exception: pass to except (json.JSONDecodeError, KeyError, TypeError, ValueError) — the broad silent catch violated both the "no || true" rule and the "no broad catches" rule.

curl tags fetch: replaced 2>/dev/null || true with || { echo "not_found"; return 0; } — curl -s already suppresses its own error output so 2>/dev/null was redundant; explicit early return is cleaner than tolerating failure silently.

_ollama_models_dir and _lmstudio_model_dirs calls: removed bare || true — neither function can fail under normal conditions.

SIGPIPE hazard fixed (find ... | head -1):

Three occurrences replaced with find ... -print -quit. Under set -o pipefail, when head closes the read end after one line, find gets SIGPIPE which propagates as a non-zero exit and can kill the script. -print -quit stops find itself after the first match — no pipe needed.
Same fix applied to the grep ... | head -1 pipeline in _sync_from_ollama: switched to grep -m 1 with || { echo "not_found"; return 0; } replacing || true.
Late sync guard: Added SYNC_EXACT_ONLY=true to the phase-11 late full-model sync call and added two regression tests covering that path.

Lightheartdevs:
Thanks for the careful iteration on the external model sync path; the latest head addresses the earlier Docker symlink, `set -e`, fuzzy-match, and host-agent Bash-selection concerns.

Decision point:
- I did not find a new blocking code issue on this head. The remaining call is whether maintainers are comfortable merging a high-risk installer/model lifecycle change with focused local validation but no GitHub checks reported for the branch.
- This touches phase 11 model download/swap behavior, `ods-cli`, host-agent, and dashboard API control flow. The repo's high-risk map points those surfaces toward focused tests plus CI and, before release confidence, scoped hardware or release-grade validation.

What I checked:
- Repo context from `README.md`, `ARCHITECTURE.md`, `CLAUDE.md`, release-validation/testing docs, Makefile targets, and CI workflow definitions.
- Diff/static review of `ods/scripts/sync-external-models.sh`, `ods/installers/phases/11-services.sh`, `ods/installers/lib/bootstrap-model.sh`, `ods/ods-cli`, `ods/bin/ods-host-agent.py`, `ods/extensions/services/dashboard-api/routers/models.py`, and `ods/tests/test-sync-external-models.sh`.
- Detached worktree checks passed: `git diff --check origin/main...HEAD`; Git Bash `bash -n` for the changed shell files; `bash tests/test-sync-external-models.sh` with 23/23 passing; `python -m py_compile ods/bin/ods-host-agent.py ods/extensions/services/dashboard-api/routers/models.py`.
- `gh pr checks 1710` still reports no checks for this branch.

Thanks again for pushing this through the review rounds.

PR #1700, opened 2026-07-09
feat(windows-cli): add enable/disable commands to ods.ps1

IronicRayquaza:
Closes #1699

## Summary

The Linux `ods-cli` has supported `ods enable <service>` and `ods disable <service>` since the extension system was introduced. The Windows `ods.ps1` implementation was missing these commands, causing the installer's guidance (for example, `ods enable comfyui`) to fail on Windows with an `Unknown command` error.

This PR brings feature parity to the Windows CLI by implementing `enable` and `disable` commands, along with the supporting helpers required to correctly manage extension compose files and regenerate compose flags.

#### `ods/installers/windows/ods.ps1`

- Added `Invoke-Enable`
  - Renames `compose.yaml.disabled` → `compose.yaml`
  - Regenerates `.compose-flags`
  - Prevents core services from being enabled manually

- Added `Invoke-Disable`
  - Stops the running container
  - Renames `compose.yaml` → `compose.yaml.disabled`
  - Regenerates `.compose-flags`
  - Prevents core services from being disabled

- Added `Update-ComposeFlags`
  - Removes stale extension `-f` entries
  - Re-scans `extensions/services` for enabled compose fragments
  - Mirrors Linux `_regenerate_compose_flags()`

- Added `Get-ExtensionServiceDir`
  - Resolves `extensions/services/<id>` directories

- Added `Get-ExtensionCategory`
  - Reads the `category:` field from `manifest.yaml`
  - Detects core services

- Wired `enable` and `disable` into the command dispatcher

- Updated `Show-Help`
  - Added `enable` and `disable` commands
  - Added usage examples

- Updated the file header comment

#### Tests

Added `ods/tests/test-windows-cli-enable-disable.sh`

Includes **24 tests** covering:

- Static structure validation
- Filesystem simulation
- Compose flag regeneration
- Enable/disable behavior

No PowerShell, Docker, or live installation is required.

---

## Release Lane

- [ ] Stable hotfix targeting `release/2.5.x`
- [x] Mainline change targeting `main`
- [ ] Next-minor work targeting the next feature/minor release
- [ ] Not sure; reviewer should help classify

Stable hotfix reason:

```text
N/A
```

---

## Changed Surface

- [ ] Docs only
- [ ] Tests only
- [ ] Dashboard UI
- [ ] Dashboard API / host agent
- [x] Installer / bootstrap / lifecycle
- [x] Docker Compose / service manifests
- [ ] Model routing / Hermes / capabilities
[...]

PR #1712, opened 2026-07-10
fix(windows-cli): honor extension depends_on in enable/disable

tang-vu:
## Problem

`ods.ps1 enable` / `disable` rename a compose fragment without consulting the manifest `depends_on` graph. The Linux `ods-cli` resolves it in both directions — `cmd_enable` offers to enable missing dependencies, `cmd_disable` warns about reverse dependents — but the Windows CLI added in #1700 does neither.

That matters because Compose validates `depends_on` against the **merged project**. A dependent left pointing at a service the merged files no longer define fails at config-parse time with `depends on undefined service`, which takes down **every** service in the stack, not just the pair involved. This repo already guards against exactly that failure elsewhere — see `ods/bin/ods-host-agent.py:1207-1213`, which validates a narrowed flag set with `config --services` before `docker compose pull` for precisely this reason.

`extensions/services/hermes-proxy/compose.yaml` declares `depends_on: - hermes`, and `hermes` and `hermes-proxy` are both `category: recommended`, so both pass the core-service guard and can be toggled:

```powershell
.\ods.ps1 disable hermes    # succeeds; hermes-proxy stays enabled
.\ods.ps1 start             # docker compose rejects the entire project
```

The reverse breaks the same way: `.\ods.ps1 enable hermes-proxy` while `hermes` is disabled produces a stack that will not start.

Reproduced against the current `main` copy of `Invoke-Disable` (UI + Docker stubbed):

```
WARN: Docker Desktop is not running -- skipping container stop.
OK: hermes disabled.
=> hermes/compose.yaml.disabled, while hermes-proxy/compose.yaml still declares depends_on: hermes
```

## Fix

- `Invoke-Enable` enables disabled extension dependencies first, transitively, with a cycle guard (`$script:_EnableVisited`). It runs **before** the already-enabled check, so re-running `enable` on a dependent repairs an already-broken stack.
- `Invoke-Disable` refuses while an enabled extension still declares the target in `depends_on`, and names the offenders.
- `-Force` (or `--force`) keeps the old behaviour behind an explicit opt-in. This mirrors the Linux CLI's `Continue anyway? [y/N]` prompt without making the Windows CLI interactive — there is no `Read-Host` anywhere in `ods.ps1` today, and I did not add one. The default answer on Linux is *no*, so refusing by default is the parity-preserving choice.

Skipped by the dependency resolver:
- **Core services** — their compose lives in `docker-compose.base.yml`, so they own no `compose.yaml` and are always present.
- **Services with neither fragment** (`opencode` today) — a rename cannot enable them, and reporting them as disabled would abort the caller's `enable` on "No compose fragment found".

## Tests

`tests/test-windows-cli-enable-disable.sh` gains static checks plus nine behaviour cases that extract the real functions out of `ods.ps1` and run them under `pwsh`, with only the UI and Docker boundary stubbed — so the manifest parsing and the guards are exercised as written rather than re-implemented in bash. The suite skips the behaviour block gracefully when no PowerShell is on PATH.

Covered: refusal + fragment left untouched, `-Force` override, transitive enable, core deps not treated as disabled, fragment-less deps not aborting enable, cycle termination, disable with no dependents, disabled dependents ignored, `depends_on`/`category` parsed from the top-level `service:` block (not the nested `library:` one), and `-Force`/`--force` detection in either argument position.

The test file was never wired into CI when it landed, so this also adds the `Windows CLI Enable/Disable Tests` step to `test-linux.yml`.

Verified locally: suite green, `PSScriptAnalyzer` clean against `PSScriptAnalyzerSettings.psd1`, `bash -n` clean, and the nine other `ods.ps1`-touching test files still pass. (`tests/test-voice-repair-command.sh` fails identically on unmodified `main` — "help missing repair voice" — so it is pre-existing and untouched here.)

## Note for reviewers

This touches `Invoke-Enable` / `Invoke-Disable`, as does my open #1704 (which changes the `Update-ComposeFlags` call sites inside those same functions). The hunks are far apart and should merge, but whichever lands first, I will rebase the other. The `test-linux.yml` step added here is byte-identical to the one in #1704, so that hunk merges cleanly either way.


Lightheartdevs:
Thanks for tightening the Windows enable/disable path here; this closes a real compose-footgun around enabled dependents like `hermes-proxy`.

What looks good:
- `ods/installers/windows/ods.ps1` now matches the Linux CLI's dependency shape: enable repairs disabled non-core extension dependencies first, while disable refuses enabled reverse dependents unless the operator explicitly passes `-Force`.
- The implementation is scoped to the Windows extension fragment path and correctly skips core services plus fragment-less services, which matters for current manifests like `dashboard-api`, `llama-server`, and `opencode`.
- `ods/tests/test-windows-cli-enable-disable.sh` covers the regression behavior, transitive enable, cycle guard, force flag ordering, disabled dependents, and the top-level manifest parsing cases; wiring it into `test-linux.yml` closes the previous CI gap.

What I checked:
- Diff/static review of `.github/workflows/test-linux.yml`, `ods/installers/windows/ods.ps1`, and `ods/tests/test-windows-cli-enable-disable.sh` against the current main-branch architecture and Windows CLI conventions.
- Real manifest scan: current service manifests use the `depends_on: [a, b]` flow style this parser expects, including `hermes-proxy -> hermes`.
- GitHub checks are green for Test Linux, PowerShell lint on Ubuntu and Windows, ShellCheck, Matrix Smoke, dashboard frontend/API, env validation, Python lint, and secret scan.
- Local Windows PowerShell parse/help check passed, and a fake-install command-path fixture verified default `disable hermes` refuses while `hermes-proxy` is enabled, `disable hermes -Force` proceeds, and `enable hermes-proxy` re-enables the disabled `hermes` dependency. The local `bash` shim is broken in this workspace, so I could not rerun the shell wrapper locally; CI did run the exact script successfully.

PR #1715, opened 2026-07-10
fix(cli): resolve user-extension compose paths in enable/disable/purge

tang-vu:
## Summary

The Linux CLI is the third surface with the enable/disable dependency gap that #1712 (Windows CLI) and #1713 (dashboard-api) fixed — but inverted: dashboard-api only looked at `data/user-extensions/`, while `ods-cli` only looks at `extensions/services/`. The service registry loads both trees, so `SERVICE_DEPENDS`/`SERVICE_IDS` know about user extensions, but the `compose.yaml` existence checks in `cmd_enable`, `cmd_disable`, and `cmd_purge` never find them:

- **enable** — a dependency that is a dashboard-installed user extension and *already enabled* is flagged as `depends on disabled services`, prompting the user to enable it again (which then reports "already enabled"); answering `n` produces a false "may not start correctly" warning.
- **disable** — an *enabled* user extension that depends on the service being disabled is silently skipped by the reverse-dependents scan, so the stack-breakage warning never fires.
- **purge** — an *enabled* user extension passes the `still enabled` guard (its `compose.yaml` lives under `data/user-extensions/`), so `ods purge` deletes the data directory of a service that is still part of the compose stack. Reproduced in the new test: on current `main` the data dir is actually removed.

## Fix

Each check now resolves built-in first, then `data/user-extensions/` — the same fallback `cmd_enable`/`cmd_disable` already use for the target service's own `ext_dir`, and the same both-trees pattern used by `_run_hook` and `cmd_template`. Built-in still takes precedence, matching the registry's duplicate-id rule. 6 lines total in `ods-cli`.

## Tests

New `tests/test-cli-user-extension-deps.sh` (registered in `make test`): builds a throwaway `ODS_HOME` fixture with one built-in and one user extension and runs the real CLI against it — no docker needed on these paths.

- enable: enabled user-ext dependency not flagged / disabled one still flagged
- disable: enabled user-ext dependent warns + cancels / disabled dependent stays silent
- purge: enabled user extension rejected with data preserved / disabled one still purges

On unfixed `main` the test fails 5/10 (including actual data deletion by purge); with the fix 10/10 pass. `bash -n` clean; existing `test-enable-gpu-validation.sh`, `test-ods-cli-pipefail-tolerance.sh`, and `test-ods-cli-contracts.py` all pass.

## Related

- #1712 — same gap in Windows `ods.ps1` enable/disable
- #1713 — mirror image in dashboard-api (built-in extensions missed there)

Note: `preset restore` has the same single-tree assumption (`cmd_preset` restore path); left out to keep this PR scoped to the enable/disable/purge parity — can follow up separately if wanted.


Lightheartdevs:
Thanks for tightening the Linux CLI user-extension dependency path here; this closes a real stack-integrity gap around dashboard-installed extensions.

What looks good:
- `ods/ods-cli` now checks `extensions/services/<id>/compose.yaml` first and then `data/user-extensions/<id>/compose.yaml`, which matches the service-registry precedence where built-ins win on duplicate IDs.
- The change covers all three affected lifecycle paths: enable no longer flags an already-enabled user-extension dependency as missing, disable now warns on enabled user-extension dependents, and purge no longer deletes data for an enabled user extension.
- The new fixture test runs the real CLI against a throwaway `ODS_HOME` and is wired into `ods/Makefile`, so this regression should stay covered without needing Docker.

What I checked:
- Repo context from `README.md`, `ARCHITECTURE.md`, `CLAUDE.md`, `CONTRIBUTING.md`, `ods/docs/HIGH_RISK_CHANGE_MAP.md`, Makefile targets, recent `main`, and CI workflow definitions.
- Diff/static review of `ods/ods-cli`, `ods/Makefile`, and `ods/tests/test-cli-user-extension-deps.sh` against the service registry's built-in/user-extension loading model.
- Local disposable worktree: `bash -n ods/ods-cli ods/tests/test-cli-user-extension-deps.sh` and `bash tests/test-cli-user-extension-deps.sh` passed, 10/10.
- `gh pr checks 1715` reports the current head green across shell lint, Linux integration, matrix smoke, dashboard API/frontend, env validation, PowerShell lint, secret scan, and related advisory jobs.

PR #1716, opened 2026-07-10
fix(cli): restore user extensions in preset load and stop mid-restore abort

tang-vu:
## Summary

Follow-up to #1715, same root cause on the `preset` surface plus a second defect that breaks `preset load` for everyone.

**1. User extensions never restore.** `ods preset save` records dashboard-installed user extensions correctly (the registry loads both trees and `SERVICE_COMPOSE` holds their real paths), but `preset load` only looks under `extensions/services/`:

- `validate_preset_compatibility` reports existing user extensions as *"no longer available ... will be skipped during restore"*
- the restore loop's `ext_dir` check skips them, so their enabled/disabled state is never applied

**2. `preset load` aborts after the first flip — for any extension, not just user ones.** The restore loop counts with `((enabled++))` / `((disabled++))`. Bash post-increment evaluates to the *old* value, so the first increment (0→1) returns status 1, and under `set -euo pipefail` the CLI exits right there: the first extension is flipped, the rest of the list is skipped, and the `Extensions: X enabled` summary and success message never print. Reproduced in the new test — on current `main` the output stops dead after `Restored .env`.

## Fix

- Resolve both existence checks built-in first, then `data/user-extensions/` — the same fallback `cmd_enable`/`cmd_disable` already use (and the pattern #1715 applies to enable/disable/purge).
- Count with `enabled=$((enabled + 1))`, with a comment explaining the `set -e` interaction so it doesn't regress. (Other `((x++))` sites in `ods-cli` are already `|| true`-guarded except two in `cmd_update`'s verification loop — left out of scope here since that path needs docker to exercise; can follow up.)

## Tests

New `tests/test-cli-preset-restore-user-extensions.sh` (registered in `make test`), same throwaway-`ODS_HOME` fixture harness as #1715's test — real CLI, no docker needed:

- save records the user extension's state
- load recognizes the user extension as available (no false "no longer available")
- load restores both trees, both directions (enable and disable passes)
- load completes past the first flip and prints correct counts

On unfixed `main` the test fails 4/7; with the fix 7/7 pass. `bash -n` clean; existing `test-preset-diff.sh` (6/6) and `test-ods-cli-pipefail-tolerance.sh` (11/11) pass.

## Related

- #1715 — same single-tree assumption in enable/disable/purge dependency checks


Lightheartdevs:
Thanks for following the user-extension fix through to preset restore; that is the right place to close the saved-state half of the same bug.

What looks good:
- `validate_preset_compatibility` and the restore loop now resolve both built-in and `data/user-extensions` directories, so presets saved from dashboard-installed extensions can be loaded without false missing-extension warnings.
- Replacing `((enabled++))` / `((disabled++))` with arithmetic assignment is the correct `set -e` fix; it prevents the first restored extension from aborting the rest of `preset load`.
- The new test covers save, compatibility, enable restore, disable restore, and summary completion using the real CLI in a throwaway `ODS_HOME` fixture.

What I checked:
- Repo context from `README.md`, `ARCHITECTURE.md`, `CLAUDE.md`, `CONTRIBUTING.md`, `ods/docs/HIGH_RISK_CHANGE_MAP.md`, Makefile targets, recent `main`, and CI workflow definitions.
- Diff/static review of `ods/ods-cli`, `ods/Makefile`, and `ods/tests/test-cli-preset-restore-user-extensions.sh`, including how preset save already uses `SERVICE_COMPOSE` from the registry.
- Local disposable worktree: `bash -n ods/ods-cli ods/tests/test-cli-preset-restore-user-extensions.sh` and `bash tests/test-cli-preset-restore-user-extensions.sh` passed, 7/7.
- `gh pr checks 1716` reports the current head green across shell lint, Linux integration, matrix smoke, dashboard API/frontend, env validation, PowerShell lint, secret scan, and related advisory jobs.

PR #1717, opened 2026-07-10
fix(cli): stop update verification abort under set -e

tang-vu:
## Summary

`ods update` dies mid-flight on every run that has at least one enabled extension. The post-update verification loop counts services with `((total_services++))` / `((failed_services++))`; bash post-increment evaluates to the *old* value, so the first increment (0→1) returns status 1 and `set -euo pipefail` kills the CLI right after `Verifying update...`. Consequences:

- the new `ODS_VERSION` is never persisted to `.env`, so the install keeps identifying as the old version on subsequent runs (compat checks, update prompts)
- the host agent is never restarted after the update
- the CLI exits 1 with no error message, even though images were pulled and containers recreated successfully — the update *worked* but looks failed

Reproduced end-to-end in the new test: on current `main` the output stops dead after `Verifying update...` and `.env` still says the old version.

Same defect class as the `preset load` abort fixed in #1716 — these two were the only unguarded post-increments left in `ods-cli` (the others already use `|| true`).

## Fix

Count with `total_services=$((total_services + 1))` / `failed_services=$((failed_services + 1))`, with a comment on the `set -e` interaction so it doesn't regress.

## Tests

New `tests/test-cli-update-verification.sh` (registered in `make test`): throwaway-`ODS_HOME` fixture with one enabled extension plus `docker`/`curl` shims on `PATH` (same shim pattern as `test-ods-cli-pipefail-tolerance.sh`), driving the real CLI through `ods update`:

- success path: reaches `Update complete`, persists the new `ODS_VERSION`, no false verification failure
- failed-verification path (shimmed `docker ps` reports nothing running): correct `Update verification failed: 1/1` count and no success claim — exercises the `failed_services` counter past its first increment too

On unfixed `main` the test fails 3/5; with the fix 5/5 pass. `bash -n` clean; `test-ods-cli-pipefail-tolerance.sh` (11/11) passes.

## Related

- #1716 — same `((x++))` + `set -e` abort in `preset load`
- #1715 — sibling user-extension dependency fixes in enable/disable/purge

### DIFF
--- a/ods/Makefile
+++ b/ods/Makefile
@@ -53,6 +53,9 @@ test: ## Run unit and contract tests
 	@echo "=== Bash 3.2 guard contracts ==="
 	@bash tests/test-bash32-guards.sh
 	@echo ""
+	@echo "=== CLI update post-verification ==="
+	@bash tests/test-cli-update-verification.sh
+	@echo ""
 	@echo "=== ods config show secret-mask matrix ==="
 	@bash tests/test-ods-config-secret-mask.sh
 	@echo ""
--- a/ods/lib/service-registry.sh
+++ b/ods/lib/service-registry.sh
@@ -33,6 +33,7 @@ declare -A SERVICE_CATEGORIES   # service_id → core|recommended|optional
 declare -A SERVICE_DEPENDS      # service_id → space-separated dependency IDs
 declare -A SERVICE_HEALTH       # service_id → health endpoint path
 declare -A SERVICE_HEALTH_TIMEOUTS  # service_id → health check timeout in seconds
+declare -A SERVICE_STARTUP_CHECKS # service_id → "true" unless manifest sets startup_check: false
 declare -A SERVICE_PORTS        # service_id → external port (what the user hits on localhost)
 declare -A SERVICE_PORT_ENVS    # service_id → env var name for the external port
 # Services with `host_network: true` in their manifest (Docker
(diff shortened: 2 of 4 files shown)


Lightheartdevs:
Thanks for tightening the update verification path; this fixes a small shell edge case with a pretty large user-facing effect when updates already recreated containers successfully.

What looks good:
- `ods/ods-cli` changes only the two post-update verification counters from `((x++))` to `x=$((x + 1))`, which preserves the existing service-health logic while avoiding the `set -e` abort before version persistence and host-agent restart.
- `ods/tests/test-cli-update-verification.sh` drives the real CLI through `ods update` with docker/curl shims and covers both the success path and the failed-verification count path; it is also wired into `ods/Makefile`.
- The GitHub check set is green, including shell lint, Python lint, integration smoke, distro smoke matrix, env validation, PowerShell lint, frontend, and secret scan.

What I checked:
- Repo context from `README.md`, `ARCHITECTURE.md`, `CONTRIBUTING.md`, `ods/CONTRIBUTING.md`, `ods/docs/HIGH_RISK_CHANGE_MAP.md`, `ods/docs/RELEASE_VALIDATION.md`, Makefile targets, and CI workflows.
- Diff/static review of `ods/ods-cli`, `ods/tests/test-cli-update-verification.sh`, and `ods/Makefile`; `git diff --check origin/main...origin/pr/1717` is clean.
- Local focused checks with Git Bash: `bash -n ods/ods-cli` passed, and `tests/test-cli-update-verification.sh` passed 5/5.
- `gh pr checks 1717` reports the available checks passing.

Thanks again for chasing this class of update failure down.

Lightheartdevs:
Thanks for following through on the update verification path and the one-shot extension edge case; that closes the remaining gap from the older review on this PR.

What looks good:
- `ods/ods-cli` still fixes the `set -e` post-increment abort with the low-risk arithmetic assignment pattern, so successful updates can reach version persistence, host-agent restart, and `Update complete`.
- `ods/lib/service-registry.sh` now carries manifest `startup_check: false` through to a small `sr_requires_startup_check` helper, and update verification skips those services instead of treating one-shot startup checks as long-running containers.
- `ods/tests/test-cli-update-verification.sh` covers both the normal enabled-service path and the one-shot skip by expecting a failed-verification count of `1/1`, not `2/2`, when only the long-running fixture service is down.

What I checked:
- Repo context from README/architecture/contributing docs, current `main`, Makefile targets, and CI workflows before reviewing the PR.
- Diff/static review of `ods/ods-cli`, `ods/lib/service-registry.sh`, `ods/tests/test-cli-update-verification.sh`, and `ods/Makefile`, including the new commit since the older advisory (`9ab6bb883 fix(test-branch): honor one-shot service startup checks`).
- Local focused checks in a detached worktree with Git Bash: `bash -n ods/ods-cli ods/tests/test-cli-update-verification.sh ods/lib/service-registry.sh`, `git diff --check origin/main...HEAD`, `bash ods/tests/test-cli-update-verification.sh` -> 5/5 passed, and `bash ods/tests/test-ods-cli-pipefail-tolerance.sh` -> 11/11 passed.
- `gh pr checks 1717 --required=false` reports the visible check set passing, including shell lint, Python lint, Dashboard frontend/API, Linux integration smoke, distro smoke matrix, env validation, PowerShell lint, and secret scan.

PR #1718, opened 2026-07-10
fix(update): stop backup abort after first copied file under set -e

tang-vu:
## Summary

Third instance of the `((x++))` + `set -e` abort found while sweeping for the class after #1716/#1717 — this one takes out the pre-update safety net.

`cmd_backup` in `ods-update.sh` counts copied files with `((files_backed_up++))`. Bash post-increment evaluates to the *old* value, so the first increment (0→1) returns status 1 and `set -euo pipefail` kills the script right after copying the first compose file: no `metadata.json`, no `Backup created` log, exit 1. Trivially reproducible on current `main`:

```
$ ods-update.sh backup snaptest
[INFO] Creating backup: backup-snaptest-...
$ echo $?
1        # backup dir contains exactly one file, no metadata.json
```

Impact: `ods-cli`'s `cmd_update` delegates its pre-update snapshot to `ods-update.sh backup` — so the snapshot **always** fails and every update proceeds with only the warning *"Pre-update snapshot failed; proceeding without safety net"*, i.e. without the rollback backup it was designed to have. The rotation loop's `((count++))` has the same defect, so old backups were never pruned either (unreachable until now anyway, since the function died earlier).

## Fix

Count with `x=$((x + 1))`, same as #1716/#1717, with a comment on the `set -e` interaction. This was the last unguarded post-increment in a `set -e` script in the repo's shell surface — `ods-restore.sh`'s `((i++))` starts at 1 (safe), `llm-cold-storage.sh` and `test-stack.sh` don't set `-e`, and the rest are `|| true`-guarded.

## Tests

New `tests/test-update-script-backup.sh` (registered in `make test`): runs the real script with `HOME` pointed at a fixture so `BACKUP_DIR` is isolated:

- backup completes past the first file, writes `metadata.json`, logs `Backup created`
- all 4 eligible files copied and counted (`Files backed up: 4`)
- rotation prunes to `MAX_BACKUPS` and keeps the newest backup — exercises the `count` counter through and past the threshold

On unfixed `main` the test fails 5/6; with the fix 6/6 pass. `bash -n` clean.

## Related

- #1716 — same abort in `ods preset load`
- #1717 — same abort in `ods update` post-verification

### DIFF
--- a/ods/Makefile
+++ b/ods/Makefile
@@ -59,6 +59,9 @@ test: ## Run unit and contract tests
 	@echo "=== bootstrap OpenClaw compose-guard regression ==="
 	@bash tests/test-bootstrap-openclaw-compose-guard.sh
 	@echo ""
+	@echo "=== ods-update.sh backup command ==="
+	@bash tests/test-update-script-backup.sh
+	@echo ""
 	@echo "=== bootstrap-upgrade spawn closes inherited FDs ==="
 	@bash tests/test-bootstrap-upgrade-close-inherited-fds.sh
 	@echo ""
--- a/ods/ods-update.sh
+++ b/ods/ods-update.sh
@@ -570,20 +570,23 @@ cmd_backup() {
     mkdir -p "$backup_path"
     
     # Backup compose files
+    # NB: x=$((x + 1)) not ((x++)) — the post-increment form evaluates to 0
+    # on the first increment, which set -e treats as failure and aborts the
+    # backup after copying a single file.
     local files_backed_up=0
     for pattern in "docker-compose*.yml" "docker-compose*.yaml" ".env" ".env.*"; do
         for file in "${INSTALL_DIR}"/${pattern}; do
             if [[ -f "$file" ]]; then
                 cp "$file" "$backup_path/"
-                ((files_backed_up++))
+                files_backed_up=$((files_backed_up + 1))
             fi
         done
     done
-    
+
     # Backup version file
     if [[ -f "$VERSION_FILE" ]]; then
         cp "$VERSION_FILE" "$backup_path/.version"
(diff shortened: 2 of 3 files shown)


Lightheartdevs:
Thanks for taking the same `set -e` increment bug through the backup path; this restores the pre-update safety net that `ods update` depends on.

What looks good:
- `ods/ods-update.sh` replaces the unsafe post-increments in `cmd_backup` and backup rotation with assignment-form arithmetic, so the first copied file and first rotation iteration no longer return status 1 under `set -euo pipefail`.
- `ods/tests/test-update-script-backup.sh` runs the real script against an isolated fixture and verifies completion, metadata creation, the expected four-file count, copying past the first file, and pruning down to `MAX_BACKUPS`.
- The test is registered in `ods/Makefile`, and the GitHub check set is green across shell lint, integration smoke, distro smoke matrix, env validation, PowerShell lint, frontend, Python lint, and secret scan.

What I checked:
- Repo context from `README.md`, `ARCHITECTURE.md`, `CONTRIBUTING.md`, `ods/CONTRIBUTING.md`, `ods/docs/HIGH_RISK_CHANGE_MAP.md`, `ods/docs/RELEASE_VALIDATION.md`, Makefile targets, and CI workflows.
- Diff/static review of `ods/ods-update.sh`, `ods/tests/test-update-script-backup.sh`, and `ods/Makefile`; `git diff --check origin/main...origin/pr/1718` is clean.
- Local focused checks with Git Bash: `bash -n ods/ods-update.sh` passed, and `tests/test-update-script-backup.sh` passed 6/6.
- `gh pr checks 1718` reports the available checks passing.

Thanks again for closing the loop on the rollback-backup side of this.

PR #1721, opened 2026-07-10
fix(windows): preserve normal-user Lemonade install

gabsprogrammer:
## Summary

Fixes the Windows AMD installation path so ODS can keep its documented **normal-user PowerShell** workflow while installing and managing Lemonade correctly.

| Area | Change |
| --- | --- |
| Lemonade MSI | Installs into the current user's supported `%LOCALAPPDATA%\lemonade_server` location instead of forcing `ALLUSERS=1`. |
| Runtime discovery | Prefers the per-user executable, while retaining Program Files probing for existing legacy/all-users installs. |
| Lifecycle | Extends dashboard host-agent discovery to the per-user location and keeps both executable aliases supported. |
| Process cleanup | Limits installer cleanup to the resolved ODS Lemonade executable directory instead of terminating unrelated Lemonade processes by name. |
| MSI diagnostics | Writes a verbose `lemonade-msi-install.log` under the ODS runtime logs directory and prints its path on failure. |
| Docker recovery | Preserves the user's original `DOCKER_CONFIG` and promotes it consistently when the installer-scoped configuration fails. |
| Test portability | Makes Windows AMD contract assertions work on native CRLF output and gives the standalone local-mode test its required CPU-budget helper. |

## Root Cause

ODS tells Windows users to run `install.ps1` from a normal PowerShell, but the Lemonade MSI invocation forced `ALLUSERS=1`. Lemonade documents that all-users installation targets `Program Files` and requires elevation; its supported no-UAC flow is per-user. ODS also only searched Program Files, so changing only the MSI flag would have left lifecycle and dashboard activation unable to find the new runtime.

The reported installation also showed a second lifecycle break: local images recovered using the user's Docker configuration, but later image preflight and service launch reverted to the failing installer-scoped configuration.

## Behavior After This Change

1. A normal-user Windows install requests Lemonade in `%LOCALAPPDATA%\lemonade_server` and retains an MSI log if Windows Installer fails.
2. Installer startup, `ods.ps1`, and dashboard model activation resolve `LemonadeServer.exe` or `lemonade-server.exe` from the per-user root first, then retain Program Files compatibility.
3. Docker first uses the installer-scoped configuration, but preserves an existing user `DOCKER_CONFIG` before doing so. A successful recovery promotes that exact configuration for all following installer operations.
4. Recovery covers image-tag validation, local image builds, Compose image preflight, `docker compose up`, managed-container checks, and the persisted launch record.
5. The compose launch record always reports the Docker configuration that actually launched the stack.

## Regression Coverage

- MSI contract: no `ALLUSERS=1`, per-user `INSTALLDIR`, verbose MSI logging, per-user executable resolution.
- PowerShell runtime test: a mocked installer-scoped `image inspect` failure recovers using a custom original `DOCKER_CONFIG`, restores it to the process, and records it in `compose-launch.txt`.
- Process cleanup contract: an unrelated `LemonadeServer.exe` is no longer selected by name alone.
- Native Windows AMD contracts: CRLF-safe `.env` and YAML assertions, managed Lemonade routing, port remapping, and local LiteLLM configuration.

## Validation

- PowerShell parser checks:
  - `ods/installers/windows/install-windows.ps1`
  - `ods/installers/windows/lib/backend-contract.ps1`
  - `ods/installers/windows/ods.ps1`
- `python -m pytest extensions/services/dashboard-api/tests/test_model_activate.py -q` (`44 passed`)
- `python -m py_compile ods/bin/ods-host-agent.py`
- `bash tests/contracts/test-installer-hardening.sh` (passed)
- `bash tests/test-windows-compose-failure-report.sh` (`46 passed`)
- `bash tests/contracts/test-amd-lemonade-contracts.sh` (`62 passed`)
- `bash tests/test-windows-service-plan.sh` (`28 passed`)
- `git diff --check` (passed)

The Bash suites were run through the local Git Bash runtime; WSL remained disabled.

## Remaining Release Receipt

The code and simulated Windows paths are covered, but a real Windows AMD/non-admin receipt is still required before declaring the original user report fully closed. It should record: MSI exit `0`, the executable under `%LOCALAPPDATA%\lemonade_server`, health on `AMD_INFERENCE_PORT`, PID-file creation, `ods.ps1 restart`, dashboard model activation, and `ods.ps1 stop` cleanup. If Task Scheduler is blocked by policy, record the exact failure and verify the direct-process fallback through the same lifecycle.

## Scope

This intentionally does not change the existing port-conflict policy. The installer already reports conflicting owners and requires an explicit user choice before continuing; users must still free or remap occupied service ports.


Lightheartdevs:
Thanks for taking on the Windows AMD Lemonade install path here; this lines the implementation up with the documented normal-user PowerShell workflow instead of assuming an elevated all-users MSI.

What looks good:
- ods/installers/windows/lib/backend-contract.ps1 now prefers %LOCALAPPDATA%\lemonade_server\bin while retaining the existing executable variable and Program Files fallbacks, and ods/bin/ods-host-agent.py mirrors the same supported roots and executable aliases for dashboard activation.
- ods/installers/windows/install-windows.ps1 drops ALLUSERS=1, requests the per-user INSTALLDIR, keeps a verbose logs/lemonade-msi-install.log, and still falls back to Vulkan llama-server when Lemonade cannot be installed or resolved.
- The Docker fallback flow now carries a successful default-config build fallback through image preflight, compose launch, and the launch record, which addresses the lifecycle break without broadening the change beyond Windows installer startup.

What I checked:
- Repo context from README.md, ARCHITECTURE.md, CONTRIBUTING.md, ods/CONTRIBUTING.md, Makefile targets, recent main, and CI workflow definitions.
- Diff/static review of the changed Windows installer, backend contract helper, host-agent Lemonade launcher, docs, and regression tests.
- Local focused checks: Windows PowerShell parser checks for the changed .ps1 files, python -m py_compile ods/bin/ods-host-agent.py, python -m pytest tests/test_model_activate.py -q (44 passed), ash tests/contracts/test-installer-hardening.sh, ash tests/test-windows-compose-failure-report.sh (40 passed), ash tests/test-windows-service-plan.sh (19 passed), and ash tests/contracts/test-amd-lemonade-contracts.sh (62 passed). The local shell tests skipped their pwsh runtime subchecks because pwsh is not installed here.
- gh pr checks 1721 reports the current head green across Linux integration/smoke, distro matrix, shell lint, Python lint/type checks, dashboard frontend/API, env validation, PowerShell lint on Ubuntu and Windows, and secret scan.

Lightheartdevs:
Thanks for tightening the Windows AMD Lemonade recovery path again; the follow-up commits close the gap where a successful user Docker config fallback did not consistently carry through later image, compose, and inspection phases.

Decision point:
- I do not see a code-level blocker in the current diff. The open call is validation policy: this is a high-risk Windows installer path, and the PR body correctly calls out that a real normal-user Windows AMD receipt is still missing before the original report is fully closed.
- If maintainers are comfortable merging on CI plus focused simulated coverage, this looks ready. If the bar for this report is real hardware/non-admin AMD evidence before merge, the smallest remaining receipt is MSI exit 0, the executable under `%LOCALAPPDATA%\lemonade_server`, health on `AMD_INFERENCE_PORT`, PID-file creation, `ods.ps1 restart`, dashboard model activation, and `ods.ps1 stop` cleanup.

What looks good:
- `ods/installers/windows/install-windows.ps1` now snapshots the original `DOCKER_CONFIG` before creating the installer-scoped config, promotes either the original or default user config through one helper, and uses the promoted Docker args for image validation, local builds, compose image preflight, compose up, and managed-container inspection.
- The new head covers the earlier blind spot where `Resolve-ODSDockerImageOrFallback` and post-`up` inspection could still be stuck on the install-scoped config. The compose launch record is also refreshed after a compose launch fallback.
- Scope stays narrow: Lemonade cleanup remains limited to the resolved runtime bin, the troubleshooting doc points users to the actual install-dir MSI log, and the added assertions track the changed contracts rather than broadening the Windows installer surface.

What I checked:
- Repo context from `README.md`, `ARCHITECTURE.md`, `CONTRIBUTING.md`, `ods/CONTRIBUTING.md`, `ods/docs/HIGH_RISK_CHANGE_MAP.md`, `ods/docs/TESTING.md`, `ods/Makefile`, CI workflows, and recent `main` history.
- Static review of the full current diff plus the follow-up delta from the prior advisory head `5c00d9e` to `d55998f`.
- `gh pr checks 1721` reports the current head green across Linux integration/smoke, distro matrix, shell lint, Python lint/type checks, dashboard frontend/API, env validation, PowerShell lint on Ubuntu and Windows, and secret scan.
- Local focused checks on `d55998f`: Windows PowerShell parser OK for the changed `.ps1` files, `python -m py_compile ods/bin/ods-host-agent.py`, `python -m pytest tests/test_model_activate.py -q` (`44 passed`), `tests/test-windows-compose-failure-report.sh` (`45 passed`), `tests/contracts/test-amd-lemonade-contracts.sh` (`62 passed`), `tests/test-windows-service-plan.sh` (`19 passed`), `tests/contracts/test-installer-hardening.sh`, and `git diff --check`. Local `pwsh` runtime subchecks were skipped because `pwsh` is not installed here.

PR #1723, opened 2026-07-11
fix(ods): surface failed service evidence during startup failures

vaibhavsrv:
## Summary

Improve `ods start` failure diagnostics by surfacing relevant container evidence and guiding users toward recovery.

When a compose startup fails, ODS now:

- identifies failed or unhealthy containers associated with the current startup attempt
- prints the last 20 log lines for relevant failed or unhealthy containers as troubleshooting evidence
- recommends running `ods doctor` for detailed diagnostics and recovery guidance

## Why

Previously, `ods start` summarized startup failures but required users to manually inspect Docker logs to determine the underlying cause.

This change connects startup failures directly to relevant evidence and provides a clear next step while keeping `ods doctor` as the central place for system diagnostics.

## Implementation

- Reuses the existing `get_compose_flags()` helper to inspect the active compose project.
- Uses `docker compose ps -a` and `docker inspect` to validate container state.
- Limits evidence to containers relevant to the current startup attempt.
- Prints the last 20 log lines for failed or unhealthy containers.
- Preserves the existing BATS test behavior by skipping the logic under `DOCKER_STUB_MODE`.

## Validation

- Existing BATS test suite passes.
- Verified that diagnostics are only shown when startup fails.

## Context

This PR implements the startup diagnostics improvement discussed with @Gabriel in the `pull-requests` channel.

### DIFF
--- a/ods/ods-cli
+++ b/ods/ods-cli
@@ -843,6 +843,51 @@ _compose_run_with_summary() {
     else
         warn "(no error keywords matched in compose log)"
     fi
+
+    # Connect failed services to logs and recovery steps if not in test stub mode
+    if [[ -z "${DOCKER_STUB_MODE:-}" ]]; then
+        local _flags_str
+        _flags_str=$(get_compose_flags)
+        local -a _compose_args
+        read -ra _compose_args <<< "$_flags_str"
+
+        local _failed_candidates=()
+        local _c
+        while read -r _c; do
+            [[ -n "$_c" ]] && _failed_candidates+=("$_c")
+        done < <(grep -oE 'ods-[a-zA-Z0-9_-]+' "$_compose_log" | sort -u || true)
+
+        local _failed_containers=()
+        if (( ${#_failed_candidates[@]} > 0 )); then
+            local _project_containers
+            _project_containers=$(docker compose "${_compose_args[@]}" ps -a --format "{{.Name}}" 2>/dev/null || true)
+
+            for _c in "${_failed_candidates[@]}"; do
+                if grep -qxFe "$_c" <<< "$_project_containers"; then
+                    local _inspect
+                    _inspect=$(docker inspect --format '{{.State.Status}} {{if .State.Health}}{{.State.Health.Status}}{{else}}none{{end}}' "$_c" 2>/dev/null || true)
+                    local _status="" _health=""
+                    read -r _status _health <<< "$_inspect"
+                    if [[ "$_status" == "exited" || "$_status" == "dead" || "$_status" == "restarting" || "$_health" == "unhealthy" ]]; then
+                        _failed_containers+=("$_c")
+                    fi
+                fi
+            done
+        fi
+
+        for _c in "${_failed_containers[@]}"; do
+            echo ""
+            warn "Recent logs for container $_c (evidence):"
+            docker logs --tail 20 "$_c" 2>&1 | sed 's/^/  /' || true
+        done
+
+        if (( ${#_failed_containers[@]} > 0 )); then
+            echo ""
+            warn "Actionable recovery step:"
+            warn "  Run 'ods doctor' to inspect system readiness and get suggested fixes."
+        fi
+    fi
+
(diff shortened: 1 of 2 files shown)


Lightheartdevs:
Thanks for tightening the ods start failure path here; surfacing the relevant container logs is the right direction for install/startup support.

Quickest path to merge:
- Add focused regression coverage for the new branch in ods/ods-cli. ods/tests/test-ods-cli-pipefail-tolerance.sh already uses fake Docker binaries, so the smallest useful test would stub docker compose up to fail with an ods-dashboard-api unhealthy message, docker compose ps -a --format {{.Name}} to list that container, docker inspect to return 
unning unhealthy, and docker logs --tail 20 to emit a sentinel line. Assert that ods start exits non-zero, prints the sentinel evidence, and includes the ods doctor recovery step.
- In that same test, protect the existing failure contract: the diagnostic collection should stay best-effort and must not mask the original compose failure or skip the Full compose output: ... line if get_compose_flags, docker compose ps, docker inspect, or docker logs fails. Using the compose args already passed into _compose_run_with_summary for the ps -a call, or explicitly tolerating helper failures, would keep this safer under set -euo pipefail.

What I checked:
- Static diff review of ods/ods-cli around _compose_run_with_summary against the existing start/restart/update callers.
- Local checks on PR head 2d6bfe7: ash -n ods/ods-cli, ash ods/tests/test-ods-cli-pipefail-tolerance.sh, and ash ods/tests/test-compose-failure-report.sh all passed under Git Bash.
- gh pr checks / status rollup currently reports no GitHub checks for this branch, so there is no CI evidence attached to the PR yet.

vaibhavsrv:
Addressed the advisory feedback.

Added focused regression coverage for the startup failure diagnostics, including the unhealthy ods-dashboard-api path, container log evidence, and the ods doctor recovery step.

I also added best-effort failure coverage for docker compose ps, docker inspect, and docker logs. These cases verify that diagnostic collection does not mask the original compose exit code and that Full compose output: remains available.

I also reused the compose arguments already passed into _compose_run_with_summary for the diagnostic ps -a query instead of re-resolving them through get_compose_flags.

Verification:

test-ods-cli-pipefail-tolerance.sh: 15 passed, 0 failed
test-compose-failure-report.sh: 13 passed, 0 failed
bash -n ods/ods-cli: passed
git diff --check: passed

The full BATS suite was not cleanly runnable on my macOS host due to existing GNU grep -P incompatibilities in AMD topology tests and a sudo/password-interactive Docker phase, so I’m not claiming a full-suite pass.

Lightheartdevs:
Thanks for tightening this follow-up; the new tests exercise the startup diagnostics path and the best-effort failure cases from the earlier review.

What looks good:
- ods/ods-cli now derives the diagnostic docker compose ps -a scope from the compose args already passed to _compose_run_with_summary, so it stays aligned with the active stack instead of re-resolving flags after a failure.
- The log-evidence branch is best-effort: ps, inspect, and logs failures are tolerated, while the original compose exit code and Full compose output: line remain intact.
- ods/tests/test-ods-cli-pipefail-tolerance.sh now covers the unhealthy ods-dashboard-api case, the ods doctor recovery hint, and diagnostic helper failures.

What I checked:
- Static review of the diff for ods/ods-cli and ods/tests/test-ods-cli-pipefail-tolerance.sh against the repo's compose/startup failure conventions.
- Local Git Bash checks on PR head d687fb9f: ash -n ods/ods-cli, ash tests/test-ods-cli-pipefail-tolerance.sh (15 passed), and ash tests/test-compose-failure-report.sh (13 passed).
- git diff --check origin/main...origin/pr/1723 passed.
- gh pr checks reports no GitHub checks for this branch, so this advisory is based on focused local verification plus static review.

vaibhavsrv:
Thanks for the detailed follow-up review and verification. Glad the updated diagnostics and regression coverage now address the earlier concerns. I'll leave the implementation as-is and wait for maintainer review.

PR #1619, opened 2026-06-27
fix: force C locale in awk to prevent comma-as-decimal in CPU values (#1614)

bong-water-water-bong:
## Fixes #1614

**Problem:** On systems where \`LC_NUMERIC\` uses ',' as the decimal separator (French, German, Italian, etc. locales), \`awk\`'s \`printf %.1f\` outputs \`0,5\` instead of \`0.5\`. These comma-formatted values are written to the .env file (e.g. \`HERMES_CPU_RESERVATION=0,5\`), which fails schema validation since the .env.schema.json expects number-type values.

**Fix:** Prepend \`LC_NUMERIC=C\` to the \`awk\` call in \`_cap_cpu_value()\` in \`06-directories.sh\`. This forces awk to use '.' as the decimal separator regardless of the system locale, without affecting any other locale-sensitive behavior.

**Change:** One line in \`installers/phases/06-directories.sh\`

**Verification:** All 9 \`test-validate-env.sh\` tests pass. The fix matches what was independently confirmed by a user in the issue comments.

PR #1725, opened 2026-07-11
fix(bootstrap): force C locale for status percent to keep JSON valid

w3-surfer:
## Problem

`status_percent()` in `scripts/bootstrap-upgrade.sh` formats the download-progress percent with `awk printf "%.1f"`:

```sh
awk "BEGIN { printf \"%.1f\", ($display_bytes / $total) * 100 }"
```

On a host whose locale uses a comma decimal separator (de_DE, pt_BR, fr_FR, es_ES, …), `printf "%.1f"` emits `37,0` instead of `37.0`. `write_status()` then writes that straight into `data/bootstrap-status.json` as a **raw JSON number**:

```json
"percent": 37,0,
```

which is **invalid JSON** — the host-agent / dashboard consumers of `bootstrap-status.json` fail to parse it, and it breaks the `"percent": 100.0` assertion in `tests/test-bootstrap-upgrade-resume-status.sh`.

**Reproduce:**
```console
$ LC_ALL=pt_BR.UTF-8 awk 'BEGIN { printf "%.1f", 37 }'
37,0
```

## Fix

Prefix the awk with `LC_ALL=C` so the percent is always dot-decimal. This is the same locale-safe awk idiom already merged across the tree (#1619/#1635/#1636/#1641/#1666/#1667); this is the same class in a distinct, untouched location (the JSON status file rather than `.env`).

**After:**
```console
$ LC_ALL=pt_BR.UTF-8 bash -c '...; status_percent 37 100'
37.0
```

One-token change; `bash -n` clean.

### DIFF
--- a/ods/scripts/bootstrap-upgrade.sh
+++ b/ods/scripts/bootstrap-upgrade.sh
@@ -104,7 +104,7 @@ status_percent() {
     local display_bytes="$downloaded"
     [[ "$display_bytes" -lt 0 ]] && display_bytes=0
     [[ "$display_bytes" -gt "$total" ]] && display_bytes="$total"
-    awk "BEGIN { printf \"%.1f\", ($display_bytes / $total) * 100 }"
+    LC_ALL=C awk "BEGIN { printf \"%.1f\", ($display_bytes / $total) * 100 }"
 }
 
 write_failed_download_status() {


Lightheartdevs:
Thanks for catching the locale-sensitive bootstrap percentage path; keeping `bootstrap-status.json` parseable under comma-decimal locales is worth fixing in exactly this small spot.

Decision point:
- The code change itself looks right and scoped: `ods/scripts/bootstrap-upgrade.sh` now pins only the `awk` percentage formatter to `LC_ALL=C`, matching the repo's recent locale-safe fixes.
- The only red check I see is `powershell-lint (windows-latest)`, and its log fails while installing PSScriptAnalyzer because `PSGallery` is missing, before repository scripts are analyzed. I do not see evidence that this shell-only PR caused that failure.

What I checked:
- Diff/static review of `ods/scripts/bootstrap-upgrade.sh`.
- `git diff --check origin/main...origin/pr/1725` passed.
- Git Bash `bash -n ods/scripts/bootstrap-upgrade.sh` passed.
- Git awk with `LC_ALL=C` emitted `37.0` for the formatter sanity check.
- `gh run view 29159741264 --job 86562652145 --log` shows the Windows lint failure occurs during PSScriptAnalyzer/PSGallery setup.

Thanks again for tightening this bootstrap status path.

Lightheartdevs:
Thanks for keeping the bootstrap status percent locale-safe; `bootstrap-status.json` has to stay valid JSON regardless of host locale.

What looks good:
- `ods/scripts/bootstrap-upgrade.sh` scopes `LC_ALL=C` to the `awk` percentage formatter, so comma-decimal locales cannot leak into the raw JSON number.
- The fix matches the repo's recent locale-safe numeric-formatting pattern and does not mutate the caller's environment outside that one command.
- The current head has green checks; the older advisory's Windows-lint environment decision point no longer applies.

What I checked:
- Diff/static review of `status_percent()` and the status JSON writer that consumes it.
- `gh pr checks 1725` is green, including ShellCheck, Linux integration smoke, matrix smoke, env validation, dashboard jobs, PowerShell lint, and secret scan.
- `git diff --check origin/main...origin/pr/1725` passed.

Thanks again for tightening this bootstrap path.

PR #1726, opened 2026-07-11
fix(dashboard-api): emit valid ISO-8601 timestamps in updates router

w3-surfer:
## Problem

`datetime.now(timezone.utc)` is timezone-aware, so `.isoformat()` already ends with `+00:00`. `routers/updates.py` then concatenated `"Z"`:

```python
"checked_at": datetime.now(timezone.utc).isoformat() + "Z",
```

producing `2026-07-11T16:25:29.417360+00:00Z` — a timestamp with **both** an offset and a `Z`, which is **not valid ISO-8601**. Both `datetime.fromisoformat()` and JS `new Date()` reject it:

```console
$ python3 -c "from datetime import *; datetime.fromisoformat(datetime.now(timezone.utc).isoformat()+'Z')"
ValueError: Invalid isoformat string: '...+00:00Z'
```

So the dashboard renders **"Invalid Date"** for `checked_at` (GET `/api/version`) and per-release `date` (GET `/api/releases/manifest`).

## Fix

Drop the trailing `"Z"` at the five synthesized-timestamp sites, matching the bare `.isoformat()` already used elsewhere in the codebase (`setup.py`, `magic_link.py`). GitHub's own `published_at` value is untouched.

Tightened the existing `test_updates.py` assertions from `"checked_at" in data` to `datetime.fromisoformat(data["checked_at"])` so the format is verified. `py_compile` clean.

### DIFF
--- a/ods/extensions/services/dashboard-api/routers/updates.py
+++ b/ods/extensions/services/dashboard-api/routers/updates.py
@@ -165,7 +165,7 @@ def _build_version_result(current: str, payload: Optional[dict]) -> dict:
         "latest": None,
         "update_available": False,
         "changelog_url": None,
-        "checked_at": datetime.now(timezone.utc).isoformat() + "Z",
+        "checked_at": datetime.now(timezone.utc).isoformat(),
     }
     if not payload:
         return result
@@ -198,7 +198,7 @@ async def _refresh_release_cache() -> Optional[dict]:
         payload = {
             "latest": data.get("tag_name", "").lstrip("v"),
             "changelog_url": data.get("html_url"),
-            "checked_at": datetime.now(timezone.utc).isoformat() + "Z",
+            "checked_at": datetime.now(timezone.utc).isoformat(),
         }
         _version_cache = {
             "expires_at": time.monotonic() + _VERSION_CACHE_TTL,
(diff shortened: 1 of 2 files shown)


Lightheartdevs:
Thanks for tightening the update timestamps; this fixes a small API formatting issue that leaks straight into the dashboard as Invalid Date.

What looks good:
- `ods/extensions/services/dashboard-api/routers/updates.py` now emits normal timezone-aware ISO strings instead of combining `+00:00` with a trailing `Z`.
- `ods/extensions/services/dashboard-api/tests/test_updates.py` now parses the synthesized timestamps with `datetime.fromisoformat`, so the regression is pinned to the behavior that broke the UI.
- The change is narrowly scoped to synthesized timestamps and leaves GitHub `published_at` values untouched.

What I checked:
- Diff/static review of the updates router and tests.
- `python -m py_compile ods/extensions/services/dashboard-api/routers/updates.py` passed.
- `python -m pytest ods/extensions/services/dashboard-api/tests/test_updates.py -q` passed: 24 passed.
- `gh pr checks 1726` is green across lint, dashboard API/frontend, matrix smoke, env validation, and integration-smoke.

Lightheartdevs:
Thanks for tightening the update-router timestamp fix again; the regression test now checks the exact ISO parsing behavior that the dashboard depends on.

What looks good:
- `ods/extensions/services/dashboard-api/routers/updates.py` now emits timezone-aware `.isoformat()` values without appending a second `Z` designator.
- `ods/extensions/services/dashboard-api/tests/test_updates.py` validates synthesized `checked_at` values with `datetime.fromisoformat(...)`, so the old `+00:00Z` shape is covered directly.
- The change is narrow dashboard-api behavior and leaves GitHub `published_at` release dates untouched.

What I checked:
- Reviewed the diff against the existing updates router and timestamp style used elsewhere in dashboard-api.
- Ran `git diff --check origin/main...HEAD`, `python -m py_compile ods/extensions/services/dashboard-api/routers/updates.py`, and `python -m pytest ods/extensions/services/dashboard-api/tests/test_updates.py -q`; the focused test file passed 25/25.
- `gh pr checks 1726 --required=false` is green across Dashboard API/frontend, Python lint/type check, ShellCheck, Test Linux, Matrix Smoke, PowerShell lint, secret scan, and env-schema jobs.

Thanks again for keeping this precise.

PR #1727, opened 2026-07-11
fix(dashboard-api): keep GPUs with [N/A] utilization/temperature in detailed view

w3-surfer:
## Problem

`get_gpu_info_nvidia_detailed()` (`gpu.py`) parsed utilization/temperature with a bare cast:

```python
utilization_percent=int(parts[5]),
temperature_c=int(parts[6]),
```

`nvidia-smi` returns `[N/A]` for these on MIG-partitioned A100/H100, many vGPU/passthrough setups, and unified-memory parts. `int("[N/A]")` raises `ValueError`, which the row's `except (ValueError, IndexError)` catches by **dropping the entire GPU** ("Skipping unparseable nvidia-smi row") — and if all GPUs are affected the function returns `None`.

This is inconsistent within the codebase: the aggregate `get_gpu_info_nvidia()` already guards util/temp against the same `na_values` (lines 190-191), and this very function already guards memory (524) and power (518). So the aggregate `/api` view shows the GPU (util/temp = 0) while the detailed view silently loses it.

## Fix

Guard util/temp against `na_values` (defaulting to `0`), matching the sibling:

```python
utilization_percent=int(parts[5]) if parts[5] not in na_values else 0,
temperature_c=int(parts[6]) if parts[6] not in na_values else 0,
```

## Verification

Added `test_keeps_gpu_with_na_util_temp` to `tests/test_gpu_detailed.py`. The detailed-GPU test class passes **8/8**; the new test **fails on the pre-fix code** (GPU dropped) and passes after.

### DIFF
--- a/ods/extensions/services/dashboard-api/gpu.py
+++ b/ods/extensions/services/dashboard-api/gpu.py
@@ -537,8 +537,8 @@ def get_gpu_info_nvidia_detailed() -> Optional[list[IndividualGPU]]:
                 memory_used_mb=mem_used,
                 memory_total_mb=mem_total,
                 memory_percent=round(mem_used / mem_total * 100, 1) if mem_total > 0 else 0.0,
-                utilization_percent=int(parts[5]),
-                temperature_c=int(parts[6]),
+                utilization_percent=int(parts[5]) if parts[5] not in na_values else 0,
+                temperature_c=int(parts[6]) if parts[6] not in na_values else 0,
                 power_w=power_w,
                 assigned_services=uuid_service_map.get(uuid, []),
             ))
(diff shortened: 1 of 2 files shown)


Lightheartdevs:
Thanks for tightening the NVIDIA detailed GPU path here; this closes a real gap between the aggregate and detailed dashboard views for hardware that reports partial 
vidia-smi data.

What looks good:
- ods/extensions/services/dashboard-api/gpu.py now treats [N/A] utilization and temperature the same way the aggregate NVIDIA parser already does, keeping the GPU row instead of dropping it through the row-level ValueError handler.
- The change is narrowly scoped and preserves the existing behavior for memory fallback, power parsing, UUID service assignment, and normal numeric 
vidia-smi rows.
- ods/extensions/services/dashboard-api/tests/test_gpu_detailed.py adds a focused regression for the MIG/vGPU/unified-memory case that would have failed before this fix.

What I checked:
- Static diff and surrounding gpu.py parser context against origin/main.
- Local focused test: python -m pytest tests/test_gpu_detailed.py -q from the PR worktree (28 passed).
- GitHub checks for this head are green, including dashboard API tests, Ruff, mypy, shell/PowerShell lint, matrix smoke, env validation, secret scan, and integration smoke.

PR #1729, opened 2026-07-11
fix(dashboard-api): return a graceful reply when /api/chat gets empty choices

w3-surfer:
## Problem

The setup assistant (`POST /api/chat`) reads the LLM reply as:

```python
response_text = data.get("choices", [{}])[0].get("message", {}).get("content", "")
```

The `[{}]` default only covers a **missing** `choices` key. An OpenAI-compatible backend that returns HTTP **200** with `{"choices": []}` (content-filtered or empty generation) — or `{"choices": null}` — makes `[0]` raise `IndexError` / `TypeError`. The only handler is `except aiohttp.ClientError`, which catches neither, so the endpoint returns an unhandled **500** instead of a graceful reply.

```console
$ python3 -c "print({'choices': []}.get('choices',[{}])[0])"
IndexError: list index out of range
```

## Fix

```python
choices = data.get("choices") or [{}]
response_text = (choices[0] or {}).get("message", {}).get("content", "")
```

Empty/null `choices` now yields an empty response; the normal path is unchanged.

## Verification

Added `test_chat_handles_empty_choices` to `tests/test_setup.py` (mocks the LLM backend returning `200 {"choices": []}`). Passes on Python 3.11; **fails on the pre-fix code** with `IndexError`.

### DIFF
--- a/ods/extensions/services/dashboard-api/routers/setup.py
+++ b/ods/extensions/services/dashboard-api/routers/setup.py
@@ -215,7 +215,8 @@ async def chat(request: ChatRequest, api_key: str = Depends(verify_api_key)):
             async with session.post(f"{llm_url}{_api_path}/chat/completions", json=payload, headers={"Content-Type": "application/json"}) as resp:
                 if resp.status == 200:
                     data = await resp.json()
-                    response_text = data.get("choices", [{}])[0].get("message", {}).get("content", "")
+                    choices = data.get("choices") or [{}]
+                    response_text = (choices[0] or {}).get("message", {}).get("content", "")
                     # Strip thinking model tags — content may contain <think>...</think> blocks
                     response_text = re.sub(r'<think>[\s\S]*?</think>\s*', '', response_text).strip()
                     return {"response": response_text, "success": True}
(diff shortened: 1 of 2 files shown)


Lightheartdevs:
Thanks for covering the empty setup-chat completion path; a 200 response with no choices should not turn into an unhandled 500.

What looks good:
- `ods/extensions/services/dashboard-api/routers/setup.py` now treats empty or null `choices` as an empty assistant response while leaving the normal OpenAI-compatible response path unchanged.
- Non-200 backend responses and `aiohttp.ClientError` handling stay on their existing error paths.
- `ods/extensions/services/dashboard-api/tests/test_setup.py` adds a focused `/api/chat` regression test for `{choices: []}`.

What I checked:
- Diff/static review of the `/api/chat` success path and setup router tests.
- `gh pr checks 1729` is green, including Dashboard API tests, Python lint/type check, Linux integration smoke, matrix smoke, env validation, and secret scan.
- `git diff --check origin/main...origin/pr/1729` passed.

Thanks again for smoothing this setup-wizard failure mode.

PR #1730, opened 2026-07-11
fix(host-agent): reject non-object JSON bodies in read_json_body

w3-surfer:
## Problem

`read_json_body()` in `bin/ods-host-agent.py` returns `json.loads(...)` directly:

```python
try:
    return json.loads(handler.rfile.read(min(length, MAX_BODY)))
except (json.JSONDecodeError, UnicodeDecodeError):
    json_response(handler, 400, {"error": "Invalid JSON"})
    return None
```

A syntactically-valid but **non-object** body — `[]`, `"str"`, `42` — passes the `if body is None` guard that every caller relies on (`_handle_model_activate`, `_handle_model_download`, `_handle_core_recreate`, …), and the subsequent `body.get(...)` raises `AttributeError`. `do_POST` has no surrounding `try/except`, so the client gets a **dropped connection with no HTTP response** plus a server-side traceback.

The sibling `read_optional_json_body()` two functions below already guards this exact case (`if not isinstance(data, dict): ... 400`) — `read_json_body()` was just missing it.

## Fix

```python
if not isinstance(data, dict):
    json_response(handler, 400, {"error": "JSON body must be an object"})
    return None
return data
```

## Verification

Added `test_read_json_body_rejects_non_object` / `_accepts_object` to `tests/test_host_agent.py`. Both pass; the reject test **fails on the pre-fix code** (the raw `[]` is returned instead of a 400).

### DIFF
--- a/ods/bin/ods-host-agent.py
+++ b/ods/bin/ods-host-agent.py
@@ -976,10 +976,14 @@ def read_json_body(handler) -> dict | None:
         json_response(handler, 400, {"error": "Request body required"})
         return None
     try:
-        return json.loads(handler.rfile.read(min(length, MAX_BODY)))
+        data = json.loads(handler.rfile.read(min(length, MAX_BODY)))
     except (json.JSONDecodeError, UnicodeDecodeError):
         json_response(handler, 400, {"error": "Invalid JSON"})
         return None
+    if not isinstance(data, dict):
+        json_response(handler, 400, {"error": "JSON body must be an object"})
+        return None
+    return data
 
 
 def read_optional_json_body(handler) -> dict | None:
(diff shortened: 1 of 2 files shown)


Lightheartdevs:
Thanks for fixing the host-agent JSON body shape; malformed-but-valid JSON should get a 400 before callers try `body.get(...)`.

What looks good:
- `ods/bin/ods-host-agent.py` now makes `read_json_body()` mirror `read_optional_json_body()` by rejecting non-object JSON bodies with a structured 400 response.
- The change protects all required-body POST handlers that depend on `body` being a dict, without changing the existing empty-body or invalid-JSON behavior.
- `ods/extensions/services/dashboard-api/tests/test_host_agent.py` covers both the reject path (`[]`) and the accepted object path.

What I checked:
- Diff/static review around `read_json_body()`, `read_optional_json_body()`, and downstream handler assumptions.
- `gh pr checks 1730` is green, including Dashboard API tests, Python lint/type check, Linux integration smoke, matrix smoke, env validation, and secret scan.
- `git diff --check origin/main...origin/pr/1730` passed.

Thanks again for closing this API edge case.

PR #1731, opened 2026-07-11
docs(extensions): bind the example service port to loopback by default

w3-surfer:
## Problem

The "30-minute path" tutorial in `docs/EXTENSIONS.md` shows a compose example that publishes a host port with **no bind address**:

```yaml
ports:
  - "${MY_SERVICE_PORT:-9200}:8080"
```

Copy-pasting it exposes the new service on `0.0.0.0` (every interface). That contradicts:

- the shipped `extensions/templates/compose-template.yaml`, which uses `"${BIND_ADDRESS:-127.0.0.1}:${MY_SERVICE_PORT:-1234}:1234"`,
- every bundled extension (`brave-search`, etc.), and
- the installer's network-exposure convention (host-wide binds are flagged).

## Fix

Make the docs example match the template — bind to loopback by default:

```yaml
- "${BIND_ADDRESS:-127.0.0.1}:${MY_SERVICE_PORT:-9200}:8080"
```

Docs-only, security-positive.

### DIFF
--- a/ods/docs/EXTENSIONS.md
+++ b/ods/docs/EXTENSIONS.md
@@ -76,7 +76,7 @@ services:
     container_name: ods-my-service
     restart: unless-stopped
     ports:
-      - "${MY_SERVICE_PORT:-9200}:8080"
+      - "${BIND_ADDRESS:-127.0.0.1}:${MY_SERVICE_PORT:-9200}:8080"
     environment:
       - LLM_URL=http://llama-server:8080/v1
     depends_on:


Lightheartdevs:
Thanks for tightening the extension tutorial's default bind; copy-paste examples should match the safer loopback posture.

What looks good:
- `ods/docs/EXTENSIONS.md` now uses `${BIND_ADDRESS:-127.0.0.1}` in the service-authoring example, matching the shipped extension template and the repo's LAN-toggle convention.
- The docs change aligns with the extension audit logic that treats loopback-by-default port bindings as the safe pattern.
- This is docs-only and does not alter runtime compose behavior.

What I checked:
- Diff/static review of the extension authoring example.
- Existing dashboard-api extension tests cover the `${BIND_ADDRESS:-127.0.0.1}` host binding pattern and reject broad defaults.
- `gh pr checks 1731` is green, including docs link checks through the Linux integration workflow.
- `git diff --check origin/main...origin/pr/1731` passed.

Thanks again for cleaning up the example.

PR #1732, opened 2026-07-11
fix(release-gate): report missing files instead of crashing version check

w3-surfer:
## Problem

`add_regex_check()` in `scripts/check-version-consistency.py` is designed to convert a per-file failure into a collected error + a clean `[FAIL]`:

```python
try:
    checks.append((label, first_match(path, pattern, label)))
except ValueError as exc:
    errors.append(str(exc))
```

But `first_match()` → `path.read_text()` raises `FileNotFoundError` when a referenced file is missing. `FileNotFoundError` is a subclass of **`OSError`, not `ValueError`**, so it escapes the handler and kills the release gate with a traceback instead of a reported failure — most plausibly for `ARCHITECTURE.md`, which is checked outside the `ods/` tree and could be renamed/moved.

```console
$ python3 -c "from pathlib import Path; Path('/no/such.md').read_text()"
FileNotFoundError  # isinstance OSError=True, ValueError=False
```

## Fix

```python
except (ValueError, OSError) as exc:
    errors.append(str(exc))
```

## Verification

`add_regex_check(..., Path('/no/such/ARCHITECTURE.md'), ...)` now appends the error instead of raising; the gate still passes clean on the real tree.

### DIFF
--- a/ods/scripts/check-version-consistency.py
+++ b/ods/scripts/check-version-consistency.py
@@ -58,7 +58,7 @@ def add_regex_check(
 ) -> None:
     try:
         checks.append((label, first_match(path, pattern, label)))
-    except ValueError as exc:
+    except (ValueError, OSError) as exc:
         errors.append(str(exc))
 
 


Lightheartdevs:
Thanks for hardening the release gate error reporting; missing version authority files should fail as collected checks rather than crashing the gate.

What looks good:
- `ods/scripts/check-version-consistency.py` now catches `OSError` alongside `ValueError` in the per-file regex check helper, which covers missing or unreadable files cleanly.
- The change is narrow: manifest parsing still has its own top-level failure path, while regex-sourced authorities continue to accumulate into the existing `[FAIL] version consistency` report.
- This matches the release-gate behavior the repo generally wants: actionable failures without Python tracebacks.

What I checked:
- Diff/static review of `add_regex_check()` and the version authority checks that call it.
- `gh pr checks 1732` is green, including Python lint/type check, Linux integration smoke, matrix smoke, env validation, dashboard jobs, and secret scan.
- `git diff --check origin/main...origin/pr/1732` passed.
- `gh pr diff 1732 --patch` confirms the patch is limited to the intended exception tuple.

Thanks again for making the gate failure mode cleaner.

PR #1733, opened 2026-07-11
fix(select-model): exit cleanly when the model catalog is empty

w3-surfer:
## Problem

`rank_models()` in `scripts/select-model.py` falls back to:

```python
fallback = min(fallback_pool, key=lambda m: float(m.get("vram_required_gb") or 999))
```

The `... or catalog` guard on `fallback_pool` only helps when the *filtered* pool is empty. If the **catalog itself** is empty — a `--catalog` JSON with `"models": []`, or every entry filtered out by `normalize_model` — then `min([])` raises an uncaught `ValueError` and `main()` exits with a traceback:

```console
$ echo '{"models": []}' > empty.json
$ python3 scripts/select-model.py --catalog empty.json --backend cpu --ram-gb 8
ValueError: min() arg is an empty sequence   # scripts/select-model.py:272
```

(The shipped `config/model-library.json` has 20 models, so this only fires on a degenerate/partial catalog — plausible during catalog authoring/testing.)

## Fix

Guard for an empty catalog in `main()`:

```python
if not catalog:
    print("error: model catalog is empty (no usable models)", file=sys.stderr)
    return 1
```

## Verification

Empty catalog now prints a clean error and exits `1`; the real catalog is unaffected. Before the fix the same input raised the `ValueError` traceback above.

### DIFF
--- a/ods/scripts/select-model.py
+++ b/ods/scripts/select-model.py
@@ -404,6 +404,9 @@ def main() -> int:
     args = parser.parse_args()
 
     catalog = load_catalog(args.catalog)
+    if not catalog:
+        print("error: model catalog is empty (no usable models)", file=sys.stderr)
+        return 1
     profile = effective_profile(normalize_profile(args.profile), args.backend, args.tier)
     capacity_gb, memory_label = usable_memory_gb(args.backend, args.memory_type, args.vram_mb, args.ram_gb)
     confidence = "high" if args.backend not in {"unknown", "none"} and capacity_gb > 0 else "medium"


Lightheartdevs:
Thanks for closing the empty-catalog edge case in the model selector; that failure mode should be a clean installer error, not a Python traceback.

What looks good:
- `ods/scripts/select-model.py` now bails out immediately after `load_catalog()` if normalization leaves no usable models, before `rank_models()` can index/min an empty list.
- The guard preserves normal catalog/ranking behavior because it only fires on an empty post-normalization catalog.
- This fits the installer convention of returning terse user-facing errors with a non-zero exit instead of a traceback.

What I checked:
- Diff/static review of `ods/scripts/select-model.py` around `main()`, catalog loading, and the ranking call.
- `gh pr checks 1733` is green, including Python lint/type check, Linux integration smoke, matrix smoke, env validation, dashboard jobs, and secret scan.
- `git diff --check origin/main...origin/pr/1733` passed.
- A local empty-catalog probe returned `error: model catalog is empty (no usable models)` and exit 1.

Thanks again for tightening this path.

PR #1738, opened 2026-07-12
fix(dashboard-api): emit valid ISO 8601 timestamps in version endpoints

tang-vu:
## Problem

`/api/version` and `/api/releases/manifest` return `checked_at` (and each release `date`) built with:

```python
datetime.now(timezone.utc).isoformat() + "Z"
```

`datetime.now(timezone.utc)` is timezone-**aware**, so `.isoformat()` already emits a `+00:00` offset. Appending `"Z"` yields a **double-designator** string:

```
2026-07-12T14:16:17.880690+00:00Z
```

That is not valid RFC 3339 / ISO 8601 (two UTC designators). Both parsers in the stack reject it:

- `datetime.fromisoformat("…+00:00Z")` → `ValueError: Invalid isoformat string`
- JS `new Date("…+00:00Z")` → `Invalid Date`

## User-facing impact

The dashboard **Settings → Updates** panel renders the last-checked time via:

```jsx
// dashboard/src/pages/Settings.jsx:311
`Last checked: ${new Date(version.checked_at).toLocaleString()}`
```

Since `new Date(...)` returns `Invalid Date`, users see **"Last checked: Invalid Date"** on every version check. Any downstream consumer that strictly parses these timestamps hits the same failure.

## Fix

Route the five `checked_at`/`date` fields through a shared helper that swaps the offset for a single `Z`, preserving the intended `Z`-suffixed UTC form while producing a valid instant:

```python
def _utc_now_z() -> str:
    return datetime.now(timezone.utc).isoformat().replace("+00:00", "Z")
```

Output: `2026-07-12T14:16:17.880690Z` — parseable by `datetime.fromisoformat` and `new Date()`.

No behavior change beyond the timestamp shape; the `Z`-suffix convention the code already intended is kept.

## Tests

Added to `tests/test_updates.py` (all assert the single-designator invariant end-to-end):

- `test_utc_now_z_is_single_designator_iso8601` — the helper.
- `test_build_version_result_checked_at_is_valid_when_no_payload` — default (offline) path.
- `test_get_version_checked_at_is_valid_iso8601` — `/api/version` (mocked GitHub).
- `test_releases_manifest_timestamps_are_valid_iso8601` — `/api/releases/manifest` happy path.
- `test_releases_manifest_error_fallback_timestamps_are_valid` — GitHub-error fallback branch (`checked_at` + release `date`).

The shared `_assert_valid_utc_z()` helper checks: trailing `Z`, no `+00:00`, and `datetime.fromisoformat`-parseable.

## Verification

- New tests: **red on the pre-fix code** (fail on the exact `…+00:00Z` string), **green after the fix**.
- Full dashboard-api suite: **1175 passed, 9 skipped, 3 failed** — the 3 failures are the pre-existing Windows baseline (`test_setup_sentinel.py` ×2 bash-dependent, `test_user_extensions.py::test_scan_symlink_skipped` symlink), unrelated to this change and present on clean `main`.
- `ruff check --select E,F,W --ignore E501,E701,E731,E741,E402` on both files: clean.

### DIFF
--- a/ods/extensions/services/dashboard-api/routers/updates.py
+++ b/ods/extensions/services/dashboard-api/routers/updates.py
@@ -30,6 +30,18 @@ _version_cache: dict[str, object] = {"expires_at": 0.0, "payload": None}
 _version_refresh_task: Optional[asyncio.Task] = None
 
 
+def _utc_now_z() -> str:
+    """Return the current UTC time as a valid ISO 8601 string with a 'Z' suffix.
+
+    ``datetime.now(timezone.utc)`` is timezone-aware, so ``.isoformat()`` already
+    carries a ``+00:00`` offset. Appending ``"Z"`` produced ``...+00:00Z`` — a
+    string with two UTC designators that both ``datetime.fromisoformat`` and the
+    dashboard's ``new Date(...)`` reject (the Settings page rendered "Invalid
+    Date" for the last-checked time). Swap the offset for a single ``Z``.
+    """
+    return datetime.now(timezone.utc).isoformat().replace("+00:00", "Z")
+
+
 def _read_utf8(path: Path) -> str:
     """Read repository text files consistently across Windows and Linux."""
     return path.read_text(encoding="utf-8")
@@ -162,7 +174,7 @@ def _build_version_result(current: str, payload: Optional[dict]) -> dict:
         "latest": None,
         "update_available": False,
         "changelog_url": None,
-        "checked_at": datetime.now(timezone.utc).isoformat() + "Z",
+        "checked_at": _utc_now_z(),
     }
     if not payload:
         return result
@@ -195,7 +207,7 @@ async def _refresh_release_cache() -> Optional[dict]:
         payload = {
             "latest": data.get("tag_name", "").lstrip("v"),
             "changelog_url": data.get("html_url"),
-            "checked_at": datetime.now(timezone.utc).isoformat() + "Z",
+            "checked_at": _utc_now_z(),
         }
         _version_cache = {
             "expires_at": time.monotonic() + _VERSION_CACHE_TTL,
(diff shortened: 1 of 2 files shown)


Lightheartdevs:
Thanks for tightening the version endpoint timestamps and pinning the dashboard-facing `Invalid Date` regression with focused tests.

Decision point:
- This is a stronger version of the same generated-timestamp fix currently open in #1726; both touch `ods/extensions/services/dashboard-api/routers/updates.py` and `ods/extensions/services/dashboard-api/tests/test_updates.py` for the `...+00:00Z` bug.
- If maintainers want these API fields to keep the intended `Z`-suffixed UTC shape, this PR looks like the better candidate: `_utc_now_z()` centralizes that convention and covers `/api/version`, `/api/releases/manifest`, and the GitHub-error fallback timestamps.
- If maintainers prefer the repo's broader existing `datetime.now(timezone.utc).isoformat()` / `+00:00` convention, #1726 is the smaller path. Either way, I would merge one of the two and close the other to avoid duplicate/conflicting fixes.

What I checked:
- Diff/static review of `routers/updates.py` and `tests/test_updates.py` against the dashboard API timestamp parsing conventions.
- Compared the overlap with #1726's current head and advisory comment.
- `python -m py_compile routers/updates.py` from this PR head passed.
- `python -m pytest tests/test_updates.py -q` from this PR head passed: 29 passed.
- `gh pr checks 1738` is green across Ruff, mypy, dashboard API/frontend, matrix smoke, env validation, integration-smoke, PowerShell lint, ShellCheck, and secret scan.

tang-vu:
Closing — superseded by #1726, which has merged.

The `Invalid Date` bug this PR targeted is gone from `main`. #1726 removed the bogus `+ "Z"` from all five call sites in `routers/updates.py`, so `checked_at` and the fallback release `date` now emit `datetime.now(timezone.utc).isoformat()` — a valid ISO-8601 string with a `+00:00` offset that both `datetime.fromisoformat` and the dashboard's `new Date(...)` parse correctly. Verified: no `+00:00Z` remains in the file.

What is left in this PR is `_utc_now_z()`, which would swap that `+00:00` for a `Z`. Both are valid ISO-8601 and both parse; `main` has now settled on `+00:00`, matching `datetime.now(timezone.utc).isoformat()` usage elsewhere in the codebase. Pushing `Z` on top would be churn against a convention that just landed, with no behavior fix behind it.

No remaining unique coverage either — the regression is pinned by #1726's tests. Thanks @w3-surfer.

PR #1739, opened 2026-07-12
fix(token-spy): stop SQLite recent-window queries over-including a full day

tang-vu:
## Problem

Token Spy's `usage.timestamp` column is stored in ISO form:

```sql
timestamp TEXT NOT NULL DEFAULT (strftime('%Y-%m-%dT%H:%M:%fZ', 'now'))
-- e.g. 2026-07-12T14:33:15.567Z   (T-separated, trailing Z)
```

But the three "last N hours" queries in the SQLite backend bound the window with `datetime('now', ?)`:

```python
# db.py — query_usage / query_summary / query_session_status
WHERE timestamp > datetime('now', ?)   # -> '2026-07-12 13:33:15'  (space, no Z)
```

`timestamp` is a `TEXT` column, so SQLite compares it **byte-wise** (BINARY collation). The bound's space separator (`0x20`) sorts *before* the stored `T` (`0x54`), so at the character right after the date, **every row on the cutoff's calendar date compares greater than the bound regardless of its time-of-day** — and leaks into the window. The `-N hours` offset is effectively ignored for same-date rows, over-including up to a full extra day.

Verified with a plain in-memory SQLite DB:

```
stored-format (default): 2026-07-12T14:33:15.567Z
query bound (datetime) : 2026-07-12 13:33:15         <-- space, no T/Z
row 2026-07-12T00:00:01.000Z (>14h old, hours=1):
  current query includes it?  YES (BUG)
  fixed   query includes it?  no (correct)
```

## Impact

- `/api/usage`, `/token-usage`, `/api/summary` (all take a `hours` param) return more than the requested window on the current day — e.g. `hours=1` can surface ~a day of rows.
- `/api/summary` feeds the dashboard's agent-monitor throughput/session counts (`agent_monitor._fetch_token_spy_metrics`), so those are inflated too.
- `query_session_status` uses the same broken 24h bound to detect the *current* session, so stale rows corrupt session-health metrics.
- **Backend divergence**: the Postgres backend (`db_postgres.py`) filters with `NOW() - INTERVAL '%s hours'` (type-correct), so the same endpoint returns different results depending on the storage backend. This fix makes SQLite agree with Postgres.

## Fix

Bound the window in the **same format the column stores**, via a shared constant:

```python
_RECENT_TS_BOUND = "strftime('%Y-%m-%dT%H:%M:%fZ', 'now', ?)"
```

Applied to all three sites (`query_usage`, `query_summary`, `query_session_status`). Now both sides of the comparison are `…T…Z`, so the byte-wise ordering matches wall-clock order. `query_report` already used matching `…T00:00:00Z` bounds and is unchanged. No schema change; existing rows compare correctly.

## Tests

Added to `tests/test_usage_report.py` (reuse the existing `load_sqlite_db` / `insert_usage` harness):

- `test_query_usage_excludes_rows_older_than_window` — a row at the start of the cutoff day is excluded from `query_usage(hours=1)`; a 10-min-old row is included.
- `test_query_summary_respects_hours_window` — `query_summary(hours=1)` aggregates only in-window agents.

Both are deterministic on the fixed code (chronological, midnight-crossing-safe) and place the stale row on the cutoff's date so they reproduce the leak at any time of day.

## Verification

- New tests: **red on the pre-fix `db.py`** (both fail: the stale same-day row leaks in), **green after the fix**.
- Full token-spy suite: **5 passed**.
- `ruff check --select E,F,W --ignore E501,E701,E731,E741,E402` on both files: clean.

### DIFF
--- a/ods/extensions/services/token-spy/db.py
+++ b/ods/extensions/services/token-spy/db.py
@@ -7,6 +7,17 @@ from datetime import date, timedelta
 
 DB_PATH = os.environ.get("DB_PATH", os.path.join(os.path.dirname(__file__), "data", "usage.db"))
 
+# "Last N hours" cutoff for recent-window queries. The timestamp column stores
+# strftime('%Y-%m-%dT%H:%M:%fZ', ...) (T-separated, trailing Z). A bare
+# datetime('now', ...) yields a space-separated 'YYYY-MM-DD HH:MM:SS' value, and
+# because the column is TEXT the comparison is a byte-wise sort: the space
+# separator (0x20) sorts before the stored 'T' (0x54), so every row on the
+# cutoff's calendar date compares greater than the bound regardless of its
+# time-of-day and gets pulled in — over-including up to a full extra day. Build
+# the bound in the stored format so the comparison matches wall-clock order, in
+# line with the Postgres backend's NOW() - INTERVAL arithmetic.
+_RECENT_TS_BOUND = "strftime('%Y-%m-%dT%H:%M:%fZ', 'now', ?)"
+
 _local = threading.local()
 
 
@@ -126,7 +137,7 @@ def log_usage(entry: dict):
 def query_usage(agent: str | None = None, hours: int = 24, limit: int = 200) -> list[dict]:
     conn = _get_conn()
     conn.row_factory = sqlite3.Row
-    sql = "SELECT * FROM usage WHERE timestamp > datetime('now', ?)"
+    sql = f"SELECT * FROM usage WHERE timestamp > {_RECENT_TS_BOUND}"
     params: list = [f"-{hours} hours"]
     if agent:
         sql += " AND agent = ?"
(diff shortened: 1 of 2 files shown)


Lightheartdevs:
Thanks for tracking down the SQLite timestamp-window mismatch in token-spy and keeping the fix constrained to the three recent-window queries.

What looks good:
- `ods/extensions/services/token-spy/db.py` now builds the SQLite cutoff with the same `strftime('%Y-%m-%dT%H:%M:%fZ', ...)` format used by `usage.timestamp`, so text comparisons line up with the stored ISO ordering instead of same-day rows leaking into `hours=N`.
- The shared `_RECENT_TS_BOUND` keeps `query_usage`, `query_summary`, and `query_session_status` consistent, which also narrows the SQLite/Postgres behavior divergence called out in the PR body.
- The new tests in `ods/extensions/services/token-spy/tests/test_usage_report.py` reproduce the same-date stale-row case for both row listing and summary aggregation.

What I checked:
- Static diff review against token-spy SQLite query construction and the existing test harness.
- `git diff --check origin/main...origin/pr/1739`
- `python -m pytest tests/test_usage_report.py -q` from `ods/extensions/services/token-spy` in a detached worktree: 5 passed.
- `gh pr checks 1739`: all non-skipped checks are passing, including Python lint, mypy, Dashboard API tests, Test Linux integration-smoke, Matrix Smoke distro lanes, macos-smoke, ShellCheck, Secret Scan, and env schema validation.

PR #1741, opened 2026-07-12
fix(migrate-config): version check aborts under set -e and inverts the comparison

tang-vu:
## Problem

`scripts/migrate-config.sh` gates its work on `compare_versions`, but calls it as a **bare statement** and reads `$?` on the next line:

```sh
compare_versions "$current_version" "$last_migrated"
local result
result=$?
if [[ $result -eq 2 ]]; then ...
```

`compare_versions` returns a non-zero *ordering* status (1 = v1>v2, 2 = v1<v2). The script runs under `set -euo pipefail`, so a bare call that returns non-zero **aborts the script before `$?` is ever read**. Reproduced with a minimal harness:

```
before compare
exit code of that block: 1     # "after" line never runs
```

So `migrate-config.sh check` and `migrate` exit silently (code 1) on **any** version difference — which is exactly the case where a migration matters. `migrate` aborts right after taking a backup, never running a single migration script.

On top of that the comparison is **inverted**. `compare_versions "$current" "$last"` returns `1` when `current > last` (an upgrade → migrations pending), but the guards check `== 2` / `!= 2` (i.e. a *downgrade*). Even with the `set -e` abort fixed, an ordinary upgrade would report "already up to date" and skip everything.

This is confirmed by the repo's **own** test — `tests/test-migrate-config.sh` "check reads version file" already fails today.

## Fix

- Capture the status with `|| rc=$?` at all four `compare_versions` call sites (the pattern the suite's own "inline exit code capture" test asks for).
- Compare against `1` (`current > last`) for "migration needed".
- Replace the `((failed++)) || true` `set -e` workaround with `failed=$((failed + 1))` (no footgun, no suppression).

## Test suite was orphaned — and destructive

`tests/test-migrate-config.sh` was **not referenced by the Makefile or any CI workflow**, so its two already-failing assertions never gated anything. Worse, its `validate` test wrote a mock over the **real, tracked** `scripts/validate-env.sh` and then `rm`-ed it — running the suite deleted a 376-line source file from the working tree (observed here).

This PR:

- Wires `tests/test-migrate-config.sh` into `make test`.
- Adds a `MIGRATE_VALIDATOR` override to `cmd_validate` so the validate test points at a temp mock instead of clobbering `scripts/validate-env.sh` (default behavior unchanged).
- Adds two behavioral tests: `check` reports a pending migration when `current > last` (exit 2 + "Migration needed"), and reports up-to-date when versions match.

## Verification

- Suite on the pre-fix script: **11 passed, 3 failed** (`check reads version file`, `no silent error suppressions`, and the new `check reports pending migration` — the last shows `exit 1` and output truncated at "Current version: 2.4.1", i.e. the `set -e` abort).
- Suite after the fix: **14/14 pass**.
- `git status` after running the suite: **clean** — `scripts/validate-env.sh` is no longer touched.
- `bash -n` clean on the script and the test; the fix is bash 3.2-safe.
- Makefile insertion placed in the bind-address gap to avoid churn with other in-flight test registrations; tab-indented recipe verified.

### DIFF
--- a/.github/workflows/test-linux.yml
+++ b/.github/workflows/test-linux.yml
@@ -136,6 +136,9 @@ jobs:
       - name: Support Bundle Evidence Tests
         run: bash tests/test-support-bundle.sh
 
+      - name: Config Migration Tests
+        run: bash tests/test-migrate-config.sh
+
       - name: Upload Installer Simulation Artifacts
         uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7
         with:
--- a/ods/Makefile
+++ b/ods/Makefile
@@ -47,6 +47,9 @@ test: ## Run unit and contract tests
 	@bash tests/test-bind-address-sweep.sh
 	@bash tests/test-n8n-cookie-policy.sh
 	@echo ""
+	@echo "=== config migration (version check under set -e) ==="
+	@bash tests/test-migrate-config.sh
+	@echo ""
 	@echo "=== ods-cli pipefail tolerance ==="
 	@bash tests/test-ods-cli-pipefail-tolerance.sh
 	@echo ""
--- a/ods/scripts/migrate-config.sh
+++ b/ods/scripts/migrate-config.sh
@@ -174,11 +174,15 @@ cmd_check() {
     log_info "Current version: $current_version"
     log_info "Last migrated: $last_migrated"
     
-    compare_versions "$current_version" "$last_migrated"
-    local result
-    result=$?
+    # compare_versions returns a non-zero *ordering* status, so capture it with
+    # `|| result=$?`. A bare call would trip `set -e` and abort before $? is
+    # read. result 1 => current_version > last_migrated => newer migrations are
+    # pending (this was previously checked as 2, i.e. a downgrade, so an ordinary
+    # upgrade silently reported "no migration needed").
+    local result=0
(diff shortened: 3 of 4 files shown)


Lightheartdevs:
Thanks for fixing the migration version gate and making the previously orphaned migration tests safe to run in CI.

What looks good:
- `ods/scripts/migrate-config.sh` now captures `compare_versions` without tripping `set -e`, and the `current > last migrated` path now checks for status `1`, matching the helper's documented ordering semantics.
- `ods/tests/test-migrate-config.sh` covers both pending-migration and up-to-date behavior, and the `MIGRATE_VALIDATOR` override keeps the validate test from overwriting the tracked `scripts/validate-env.sh`.
- The suite is wired into both `ods/Makefile` and `.github/workflows/test-linux.yml`, so this lifecycle regression is no longer a local-only check.

What I checked:
- Static diff review against the migration helper, Makefile test target, and `test-linux` workflow placement.
- `git diff --check origin/main...HEAD` in a detached worktree.
- `C:\Program Files\Git\bin\bash.exe -lc 'bash -n scripts/migrate-config.sh tests/test-migrate-config.sh && bash tests/test-migrate-config.sh'`: 14/14 migration tests passed.
- `gh pr checks 1741`: all non-skipped checks are passing, including Test Linux integration-smoke, ShellCheck, Matrix Smoke distro lanes, macos-smoke, Dashboard, Secret Scan, Python lint, PowerShell lint, and env schema validation.

Issue #1742, opened 2026-07-12
feat(doctor): add external LLM connectivity check when EXTERNAL_LLM_URL is configured


IronicRayquaza:
## Summary

PR #1653 introduced support for reusing existing Ollama and LM Studio instances during installation by writing `EXTERNAL_LLM_URL`, `EXTERNAL_LLM_PROVIDER`, and `EXTERNAL_LLM_MODEL` to `.env` and layering `docker-compose.cloud.yml` to scale down the local `llama-server`.

This works correctly at install time, but `ods doctor` was not updated to reflect this new operating mode.

When `EXTERNAL_LLM_URL` is configured, `ods doctor` still checks for a running `llama-server` container and reports it as missing or unhealthy. The installation is functioning exactly as intended, but the diagnostic output tells the user their stack is broken. This is a false failure that erodes trust in `ods doctor` as a reliable diagnostic tool.

Additionally, `ods doctor` currently has no way to confirm that the configured external LLM is actually reachable and serving a model. It neither validates the endpoint nor reports which model is active. A user whose Ollama instance has gone down after installation would see no actionable signal from `ods doctor` beyond the misleading `llama-server` failure.

## Expected Behavior

When `EXTERNAL_LLM_URL` is configured, `ods doctor` should recognize that the deployment is intentionally using an external LLM instead of the bundled `llama-server`.

It should:

- Skip the local `llama-server` health check.
- Detect and report the configured external provider.
- Verify that the configured endpoint is reachable.
- Validate that the configured model is available (where supported).
- Display clear diagnostics for connectivity or model availability issues.

## Current Behavior

- `ods doctor` still checks for a local `llama-server`.
- It reports the container as missing or unhealthy even though it was intentionally disabled.
- No validation is performed against the configured external LLM endpoint.
- Users receive misleading diagnostic output while real external LLM failures go undetected.

## Impact

- Produces false diagnostic failures for valid external LLM deployments.
- Reduces confidence in `ods doctor` as a troubleshooting tool.
- Fails to detect when the configured external LLM becomes unavailable after installation.

## Reproduction Scenario

A user installs ODS on a machine that already has Ollama running with an existing model (for example, `qwen3.5:9b`). During installation, ODS detects the existing instance and the user chooses to reuse it instead of downloading and running a separate local `llama-server`. The installation completes successfully, with services such as Open WebUI and Hermes using the external Ollama instance as intended.

[...]

PR #1653, opened 2026-07-01
feat(installer): reuse validated Ollama and LM Studio models

IronicRayquaza:
## Summary

Adds an explicit, validated Linux installer path for reusing a host-managed Ollama or LM Studio text model instead of downloading and running a duplicate ODS `llama-server`.

The implementation treats external inference as a complete routing contract rather than a URL substitution: it validates the provider and exact model, derives separate host/container URLs, profiles managed inference out, rewires every direct consumer, health-proves a real completion, and provides a transactional path back to managed inference.

### Interactive install

When a matching Ollama or LM Studio model is running locally, the installer offers to reuse it.

### Non-interactive install

Ambient services do not change the install topology unless reuse is explicitly requested:

```bash
./install.sh --reuse-external-llm --non-interactive
```

An endpoint and exact model can also be selected explicitly:

```bash
./install.sh \
  --external-llm-url http://127.0.0.1:11434 \
  --external-llm-provider ollama \
  --external-llm-model qwen3.5:9b
```

Return to ODS-managed inference with:

```bash
./install.sh --no-external-llm
```

## Runtime contract

- Validates `http`/`https` base URLs and rejects embedded credentials, query strings, fragments, unsupported paths, and invalid ports.
- Accepts only `ollama` or `lmstudio`, and requires the selected model to be present in the provider inventory.
- Converts host loopback URLs to `host.docker.internal` only for container consumers.
- Adds Linux `host-gateway` routes to every direct LLM consumer.
- Profiles `llama-server`, `model-router`, and their local readiness dependencies out of the external stack.
- Routes Open WebUI, Dashboard API/ODS Talk, Hermes, OpenClaw, Perplexica, Privacy Shield, and Token Spy to the external endpoint.
- Skips duplicate GGUF/bootstrap downloads while preserving the local model library for browsing and downloads.
- Blocks local model activation while the external backend owns inference, with a clear recovery command.
- Requires host and container model inventory plus a real chat completion before installation succeeds.
- Rejects direct local/cloud/hybrid mode changes while external routing is active so `.env` cannot become internally inconsistent.

## Invariants

| Invariant | Enforcement |
|---|---|
| Automation is deterministic | Non-interactive ambient discovery is inert without explicit opt-in |
| One active inference owner | External reuse cannot be combined with cloud, hybrid, or external Lemonade |
| Host and containers use reachable URLs | Host URL is retained separately from the normalized container URL |
| No partial success | Missing provider/model or failed completion aborts installation |
[...]

PR #1743, opened 2026-07-12
feat: add external LLM backend connectivity check to ods doctor

IronicRayquaza:
## Summary

PR #1653 introduced support for reusing existing Ollama and LM Studio instances by configuring `EXTERNAL_LLM_URL`, `EXTERNAL_LLM_PROVIDER`, and `EXTERNAL_LLM_MODEL` during installation. In this mode, the local `llama-server` container is intentionally disabled.

Prior to this change, `ods doctor` always expected a local `llama-server` to be running, causing valid external LLM deployments to be reported as unhealthy.

This PR updates `ods doctor` to detect external LLM mode and validate the configured external endpoint instead of checking for a local `llama-server`.

**Closes:** #1742  
**Follow-up to:** #1653

---

### `ods/scripts/ods-doctor.sh`

- Added `_doctor_check_llm_backend`
  - Reads `EXTERNAL_LLM_URL`, `EXTERNAL_LLM_PROVIDER`, and `EXTERNAL_LLM_MODEL` from `.env`
  - Routes diagnostics to either the external LLM check or the existing local `llama-server` check

- Added `_doctor_check_external_llm`
  - Detects and validates external LLM backends
  - Probes:
    - `/api/tags` for Ollama
    - `/v1/models` for LM Studio/OpenAI-compatible endpoints
  - Exports diagnostic variables:
    - `LLM_STATUS`
    - `LLM_PROVIDER`
    - `LLM_MODEL`
    - `LLM_URL`
    - `LLM_RECOVERY`
    - `LLM_LOCAL_WARNING`
  - Adds recovery guidance when the endpoint is unreachable
  - Warns when an unused local `llama-server` container is still running

- Added `_doctor_check_llama_server`
  - Extracted the existing `llama-server` health check into its own function
  - Executed only when no external LLM is configured
  - Existing behavior remains unchanged

### `ods/ods-cli`

- Updated the Python formatting logic to display **LLM Backend** information within the **Runtime Environment** section, including:
  - Provider
  - Endpoint
  - Active model
  - Recovery guidance (when applicable)
  - Warning if an unused local `llama-server` is still running

---

## Behavior

| Configuration | Result |
|---------------|--------|
| `EXTERNAL_LLM_URL` not configured | Existing `llama-server` health check (unchanged) |
| External endpoint configured and reachable | ✅ Reports provider, endpoint, and model |
| External endpoint configured but unreachable | ❌ Reports failure with provider-specific recovery guidance |
| External endpoint configured while local `llama-server` is still running | ✅ External backend reported as healthy with a warning about the unused local container |

---

## Testing

- [x] `bash -n` passes on `ods/scripts/ods-doctor.sh` and `ods/ods-cli`
- [x] `shellcheck -x -S warning` passes on `ods/scripts/ods-doctor.sh`
- [x] `git diff --check origin/main` passes
- [x] Verified local mode (`EXTERNAL_LLM_URL` unset) retains existing behavior
- [x] Verified external endpoint is reachable and reports provider/model information
- [x] Verified unreachable external endpoint reports an actionable recovery hint
- [x] Verified a running but unused local `llama-server` generates a warning instead of a failure

---

## What Does **Not** Change

- All existing `ods doctor` checks (Dashboard API, Open WebUI, Voice, Search, etc.) remain unchanged.
- No new dependencies are introduced (`curl` is already used throughout `ods-doctor.sh`).
- Installations using the bundled local `llama-server` experience no behavior changes.


Lightheartdevs:
Thanks for taking on the external LLM doctor path here; this is the right diagnostic gap to close after the external Ollama/LM Studio work.

Quickest path to merge:
- In `ods/scripts/ods-doctor.sh`, `_doctor_check_llm_backend` re-parses raw `.env` lines with `grep ... | cut ... | tr -d '[:space:]'` even though the script has already loaded `.env` through `load_env_safe`. That means a valid quoted value like `EXTERNAL_LLM_URL=http://127.0.0.1:11434` gets passed to curl/reporting with literal quotes, and whitespace inside values is silently mutated. Use the already-loaded `${EXTERNAL_LLM_URL:-}`, `${EXTERNAL_LLM_PROVIDER:-}`, and `${EXTERNAL_LLM_MODEL:-}` values, or a shared env-reader that strips only surrounding quotes/CR safely.
- Add a focused doctor regression in `ods/tests/test-ods-doctor.sh` or a small new contract test for the new `runtime.llm_backend` branch. A temp `.env` with quoted `EXTERNAL_LLM_URL`, stubbed `curl`, and no local `llama-server` would prove the report URL is unquoted and the expected `/api/tags` or `/v1/models` probe is used.

What I checked:
- Diff/static review of `ods/scripts/ods-doctor.sh` and `ods/ods-cli` against current `main`.
- `git diff --check origin/main...origin/pr/1743` passed.
- Git Bash `bash -n ods/scripts/ods-doctor.sh ods/ods-cli` passed.
- Git Bash `bash ods/tests/test-ods-doctor.sh` passed: 17 passed.
- Git Bash `bash ods/tests/test-doctor-command.sh` passed: 16 passed.
- `gh pr checks 1743` is green, including shell lint, Python lint, dashboard jobs, integration smoke, distro matrix, env validation, PowerShell lint, and secret scan.

Thanks again for pushing this diagnostic path forward.

IronicRayquaza:
Thanks for the review! I've updated the PR to address both points:

### Changes Made

**1. Leverage Pre-loaded Environment Variables**

Refactored `_doctor_check_llm_backend` in `ods-doctor.sh` to use the environment variables already loaded by `load_env_safe`:

* `EXTERNAL_LLM_URL`
* `EXTERNAL_LLM_PROVIDER`
* `EXTERNAL_LLM_MODEL`

The function now reads directly from:

```bash
${EXTERNAL_LLM_URL:-}
${EXTERNAL_LLM_PROVIDER:-}
${EXTERNAL_LLM_MODEL:-}
```

instead of re-parsing `.env` with `grep`/`cut`. This reuses the existing parsing logic, ensuring values are already normalized and unquoted while avoiding duplicate parsing.

**2. Added Diagnostic Regression Test**

Added a new behavioral test (`12. External LLM backend configuration and connectivity check`) to `ods/tests/test-ods-doctor.sh` that:

* Creates a temporary mock environment with quoted values (e.g. `EXTERNAL_LLM_URL="https://mock-llm.example.com"`).
* Stubs the `curl` utility to record incoming requests.
* Executes `ods-doctor.sh`.
* Verifies that the generated report JSON contains the correctly unquoted URL, provider, and model values.
* Confirms that the Ollama provider probes the expected `/api/tags` health endpoint.

### Validation

* ✅ All **18** tests in `test-ods-doctor.sh` are passing.
* ✅ All **16** tests in `test-doctor-command.sh` are passing.

Lightheartdevs:
Thanks for tightening the external LLM doctor path here; the quoted `.env` regression coverage is a useful follow-up to the last review.

Quickest path to merge:
- In `ods/scripts/ods-doctor.sh`, `_doctor_check_llm_backend` still falls back to `_doctor_check_llama_server` whenever `EXTERNAL_LLM_URL` is empty. That makes `ODS_MODE=cloud` report `runtime.llm_backend.status=fail`, `provider=llama-server`, and an autofix hint of `Local llama-server is unreachable`, even though the same report's inference contract says cloud mode expects external ownership through LiteLLM. Gate the local llama probe on modes that actually manage local inference, and route cloud/external-owner modes through the configured `LLM_API_URL`/LiteLLM/external endpoint or leave the backend check provider-neutral when no probe target exists. A focused regression with `ODS_MODE=cloud` and no `EXTERNAL_LLM_URL` should assert that doctor does not emit a local llama-server failure/hint.
- In `ods/tests/test-ods-doctor.sh`, remove the trailing whitespace on the added blank line near the external LLM test; `git diff --check origin/main...HEAD` currently fails on that file.

What I checked:
- Diff/static review of `ods/scripts/ods-doctor.sh`, `ods/ods-cli`, and `ods/tests/test-ods-doctor.sh` against current `main`.
- Local Git Bash: `bash -n ods/scripts/ods-doctor.sh ods/ods-cli` passed.
- Local Git Bash: `bash ods/tests/test-ods-doctor.sh` passed: 19 passed.
- Local Git Bash: `bash ods/tests/test-doctor-command.sh` passed: 16 passed.
- `git diff --check origin/main...HEAD` failed on trailing whitespace in `ods/tests/test-ods-doctor.sh`.
- Ad hoc cloud-mode doctor run showed the local llama-server failure/hint while `runtime.inference_contract.expected_inference_owner` was `external` and `expected_gateway` was `litellm`.
- `gh pr checks 1743` has Fedora/openSUSE distro jobs failing during GitHub Actions setup because action download metadata returned `Service Unavailable`; the other visible lint, dashboard, env, smoke, integration, PowerShell, and secret checks passed.

Thanks again for pushing this diagnostic path forward.

IronicRayquaza:
I have updated the PR to address these items:

1. **Gated Local Llama-Server Probes**: Refactored `_doctor_check_llm_backend` to respect `ODS_MODE`. When `ODS_MODE=cloud` and `EXTERNAL_LLM_URL` is empty, it either probes the configured `LLM_API_URL` (if present) or leaves the check provider-neutral (`status=ok`, `provider=cloud`) without checking local container status or generating local `llama-server` autofix hints.
2. **Added Cloud Mode Regression Test**: Appended Test 13 in `test-ods-doctor.sh` to execute the doctor script under `ODS_MODE="cloud"` (with no external URL set) and assert that it correctly marks the backend status `ok` and generates no local container warnings.
3. **Fixed Trailing Whitespace**: Removed the trailing whitespace on line 256 of `test-ods-doctor.sh`. `git diff --check origin/main...HEAD` is now fully clean.

All 19 test cases in `test-ods-doctor.sh` and all 16 command test cases are passing.

Lightheartdevs:
Thanks for tightening the external LLM doctor path again; using the loaded env values and adding cloud-mode coverage closes the main behavioral gap from the last pass.

Quickest path to merge:
- In `ods/tests/test-ods-doctor.sh`, tests 12 and 13 both mutate `ods/.env`, reuse `ods/.env.backup`, and replace the `EXIT` trap around lines 269-337. When test 13 starts, it moves test 12's fixture `.env` over the original backup, so a local run can leave `EXTERNAL_LLM_URL="https://mock-llm.example.com"` in `ods/.env` and lose the developer's original config. Use a per-test temp backup and restore immediately after each scenario, or run each scenario in a temp install root so the checkout's `.env` is never the fixture target.
- Fold the temp workspace and `/tmp/curl_calls.log` cleanup into the same single cleanup path. I reproduced the suite passing while still leaving `ods/.env` with the external fixture and a temp `ods-doctor-test-workspace.*` directory afterward, so the assertion coverage is good but cleanup is not.

What I checked:
- Diff/static review of `ods/scripts/ods-doctor.sh`, `ods/ods-cli`, and `ods/tests/test-ods-doctor.sh` against current `main`.
- `git diff --check origin/main...origin/pr/1743` passed.
- Git Bash: `bash -n ods/scripts/ods-doctor.sh ods/ods-cli` passed.
- Git Bash: `bash ods/tests/test-ods-doctor.sh` passed: 20 passed, but left the fixture `.env` behind in the detached worktree.
- Git Bash: `bash ods/tests/test-doctor-command.sh` passed: 16 passed.
- `gh pr checks 1743` is green across lint, dashboard, integration smoke, distro matrix, env validation, PowerShell lint, and secret scan.

Thanks again for pushing this diagnostic path forward.

IronicRayquaza:
I have updated the test suite with the following improvements:

1. **Fixed Environmental Backups and Nesting Traps**:
   - Refactored `test-ods-doctor.sh` to use unique per-run backup file names (`/tmp/ods-env-backup.XXXXXX`) rather than the static `.env.backup` file name.
   - Removed all nested/overridden `trap EXIT` blocks inside individual test cases. Instead, a single robust exit trap handles cleanup at the suite level.
   - Both Test 12 and Test 13 now immediately restore the developer's original `.env` file right after the doctor execution phase.
2. **Immediate Workspace Cleanup**:
   - Test 12 now immediately deletes its temporary workspace directory and `/tmp/curl_calls.log` once execution finishes, leaving no dangling files behind.

All 19 doctor tests and 16 CLI tests pass successfully and execution is fully isolated.

Lightheartdevs:
Thanks for continuing to tighten the external/cloud LLM doctor path; the test isolation and quoted-env coverage are in much better shape now.

Quickest path to merge:
- In `ods/scripts/ods-doctor.sh`, avoid probing the container-internal cloud route `LLM_API_URL=http://litellm:4000` from the host as if it were an external endpoint. With an `.env` containing `ODS_MODE=cloud` and `LLM_API_URL=http://litellm:4000`, the current head reports `runtime.llm_backend.status=fail`, `provider=openai`, and `recovery_hint=check external LLM service connectivity`, even though `ods mode cloud` uses that LiteLLM URL as the normal service-to-service gateway. Translate known LiteLLM/container routes to the host-mapped port such as `127.0.0.1:${LITELLM_PORT:-4000}` before probing, or leave the backend check provider-neutral when the configured URL is not host-probeable.
- Add a regression in `ods/tests/test-ods-doctor.sh` for `ODS_MODE=cloud` plus `LLM_API_URL=http://litellm:4000`. It should not emit a backend failure solely because the host cannot resolve the container DNS name; if the intended behavior is a host-level LiteLLM check, assert the translated host URL. Making that fixture CRLF would also cover Windows-generated `.env` files.

What I checked:
- Diff/static review of `ods/scripts/ods-doctor.sh`, `ods/ods-cli`, and `ods/tests/test-ods-doctor.sh` against current `main`.
- Git Bash: `bash ods/tests/test-ods-doctor.sh` passed: 20 passed.
- Git Bash: `bash ods/tests/test-doctor-command.sh` passed: 16 passed.
- `gh pr checks 1743` is green across lint, dashboard, integration smoke, distro matrix, env validation, PowerShell lint, and secret scan.
- Ad hoc cloud-mode doctor probe with `ODS_MODE=cloud` and `LLM_API_URL=http://litellm:4000` reproduced the false backend failure described above.

Thanks again for pushing this diagnostic path forward.

IronicRayquaza:
I've addressed the feedback and updated the PR:

- Added port translation in `_doctor_check_llm_backend` to convert known Docker-internal routes (e.g. `http://litellm:<port>`) to their host-mapped equivalents (e.g. `http://127.0.0.1:<port>`) before running host connectivity probes.
- Added a probeable hostname filter. If the configured URL uses an internal container hostname (no `.` and not `localhost`), the host connection probe is skipped and the backend is reported as `ok (provider cloud)` instead of a connection failure.
- Updated `load_env_safe` to strip trailing carriage return (`\r`) characters from environment variables, ensuring Windows-formatted CRLF `.env` files are parsed correctly.
- Added regression tests 14 and 15 to `ods/tests/test-ods-doctor.sh` to verify translated URLs and non-probeable internal container hostnames are parsed correctly from CRLF fixtures and pass as expected.

Thanks again for the review!

Lightheartdevs:
Thanks for continuing to harden the external/cloud LLM doctor path; the LiteLLM host-port translation and CRLF coverage address the last false-failure case.

Quickest path to merge:
- In `ods/tests/test-ods-doctor.sh`, the new env backup/restore blocks still create `ENV_BACKUP_PATH=$(mktemp ...)` even when `ods/.env` did not exist. Because the restore checks only `[[ -f $ENV_BACKUP_PATH ]]`, a clean checkout ends the suite with a new empty ignored `ods/.env`. Track whether the original env existed, or only create the backup when it does; when there was no original env, remove the fixture instead of moving the empty temp file into place.
- Add a final assertion for the no-original-`.env` case so future external/cloud doctor tests cannot leave an empty `.env` behind again.

What I checked:
- Diff/static review of `ods/scripts/ods-doctor.sh`, `ods/ods-cli`, and `ods/tests/test-ods-doctor.sh` against current `main`.
- Git for Windows Bash: `bash -n ods/scripts/ods-doctor.sh ods/ods-cli` passed.
- Git for Windows Bash: `bash ods/tests/test-ods-doctor.sh` passed: 23 passed, but left `ods/.env` present with `length=0` in an initially clean worktree.
- `gh pr checks 1743` is green across lint, dashboard, integration smoke, distro matrix, env validation, PowerShell lint, and secret scan.

Thanks again for pushing this diagnostic path forward.

IronicRayquaza:
Thanks for the review! I've updated the PR to address the `.env` leak issue.

### Changes Made

* **Conditional Temporary Backup**

  * Updated tests **12, 13, 14, and 15** to first check whether `ods/.env` exists before creating a backup.
  * `ENV_BACKUP_PATH` is now only set when an actual `.env` file exists, preventing the creation of unnecessary empty backup files.

* **Fixture State Tracking**

  * Added a `FIXTURE_ACTIVE` state variable to explicitly track when a temporary test fixture is present on disk.

* **Robust Cleanup Trap**

  * Updated the `cleanup()` exit trap to inspect both `ENV_BACKUP_PATH` and `FIXTURE_ACTIVE`.
  * On interruption or test failure:

    * If an original `.env` existed, it is restored from the backup.
    * Otherwise, any temporary fixture created by the test is removed, ensuring no files are left behind.

* **Leak Assertion**

  * Added a final assertion at the end of the test suite.
  * If the repository did **not** contain a `.env` file when the suite started, the test now verifies that no `.env` file exists after execution, preventing regressions.

### Validation

* ✅ All tests pass successfully.
* ✅ Cleanup assertion passes in a fresh worktree.
* ✅ No temporary `.env` files are left behind after the test suite completes.

Lightheartdevs:
Thanks for sticking with the external/cloud LLM doctor path; the current version closes the quoted env, cloud-mode, LiteLLM host-probe, and cleanup issues from the prior passes.

What looks good:
- `ods doctor` now records an explicit `runtime.llm_backend` block and avoids reporting a local `llama-server` failure for valid external or cloud-owned inference paths.
- The LiteLLM/container URL translation and non-host-probeable fallback keep host-level doctor checks aligned with the compose/runtime split.
- The added doctor tests cover quoted external URLs, cloud mode without an external URL, CRLF `LLM_API_URL=http://litellm:4000`, non-probeable container hostnames, and `.env` cleanup.

What I checked:
- Diff/static review of `ods/scripts/ods-doctor.sh`, `ods/ods-cli`, and `ods/tests/test-ods-doctor.sh`.
- Git Bash: `bash ods/tests/test-ods-doctor.sh` passed: 23 passed, 0 failed, and the worktree stayed clean afterward.
- Git Bash: `bash ods/tests/test-doctor-command.sh` passed: 16 passed.
- `gh pr checks 1743` is green across lint, dashboard, integration smoke, distro matrix, env validation, PowerShell lint, and secret scan.

Thanks again for pushing this diagnostic path to a clean state.

PR #1744, opened 2026-07-13
test(dashboard-api): cover GGUF metadata parser end-to-end

tang-vu:
## What

Adds `tests/test_gguf_inspector.py` — 24 unit tests for `inspect_gguf` in `dashboard-api/gguf_inspector.py`.

## Why

`inspect_gguf` parses GGUF model files and drives model fit/performance reporting in `performance_oracle` (`performance_oracle.py:602,943`), yet had **no direct coverage**. Every existing suite that touches `.gguf` only writes placeholder text files whose magic bytes fail the header check, so the parser degrades to `unknown` immediately — the actual metadata-parsing paths (scalars, strings, arrays, normalization) were never exercised.

The parser also consumes untrusted binary input (user-supplied model files), so its graceful-degradation contract ("degrading to `unknown` on failure") deserves explicit pinning.

## Coverage

Tests build synthetic, **spec-accurate** GGUF byte streams (a small in-test encoder mirrors the parser's expected little-endian layout):

- **Happy path** — architecture, quantization label (`Q4_K_M`), model name, context length, block count, embedding length, attention head counts, tensor count, `size_bytes`.
- **Scalar types** — parametrized `uint8/int8/uint16/int16/uint32/int32/uint64/int64`, plus `float32/float64/bool` preserved in metadata.
- **Arrays** — small array returned inline; large array (>64) sampled into `{type,item_type,length,sample}`; exact 64-element boundary stays inline.
- **file_type normalization** — known label, unknown int → stringified, missing → `unknown`.
- **Degradation** — non-GGUF magic, truncated metadata, read-cap overrun (`max_metadata_bytes`), missing file, directory path, non-string architecture, invalid UTF-8 (`errors="replace"`).

## Notes

- Test-only change — no production code touched.
- Uses only stdlib (`struct`) + `pytest`; no new dependencies.
- Runs in the existing `pytest tests/` job (dashboard workflow); 24 passed locally in ~0.4s.

### DIFF
--- a/ods/extensions/services/dashboard-api/tests/test_gguf_inspector.py
+++ b/ods/extensions/services/dashboard-api/tests/test_gguf_inspector.py
@@ -1,0 +1,281 @@
+"""Unit tests for the GGUF metadata parser in ``gguf_inspector``.
+
+The parser drives model fit/performance reporting in ``performance_oracle`` but
+had no direct coverage — existing suites only write placeholder ``.gguf`` files
+whose magic bytes fail the header check, so the actual metadata-parsing paths
+(scalars, strings, arrays, graceful degradation) were never exercised.
+
+These tests build synthetic-but-spec-accurate GGUF byte streams so the happy
+path and every failure mode are pinned down.
+"""
+
+import struct
+
+import pytest
+
+from gguf_inspector import inspect_gguf
+
+# GGUF value type ids (see gguf_inspector._GGUF_VALUE_TYPES).
+U8, I8, U16, I16, U32, I32, F32, BOOL, STR, ARR, U64, I64, F64 = range(13)
+
+_SCALAR_FMT = {
+    U8: "<B", I8: "<b", U16: "<H", I16: "<h", U32: "<I",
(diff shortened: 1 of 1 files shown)


Lightheartdevs:
Thanks for adding direct GGUF parser coverage; this pins an untrusted binary metadata path that previously only had placeholder-file failure coverage.

What looks good:
- The in-test GGUF encoder mirrors the parser's little-endian metadata layout closely enough to exercise real scalar, string, array, quantization, and normalization paths without adding dependencies.
- Degradation cases cover non-GGUF input, truncated metadata, read-cap overrun, missing files, directories, non-string architecture, and invalid UTF-8.
- The change is test-only and lands in the existing dashboard-api pytest job.

What I checked:
- Diff/static review of `ods/extensions/services/dashboard-api/tests/test_gguf_inspector.py`.
- `python -m pytest tests/test_gguf_inspector.py -q` passed: 24 tests.
- `git diff --check origin/main...HEAD` passed; `gh pr checks 1744` is green, including Dashboard API, Ruff, mypy, integration smoke, env validation, and lint jobs.

Thanks again for making this path easier to maintain.

PR #1745, opened 2026-07-13
fix(dashboard-api): bound GGUF array nesting depth to preserve degrade contract

tang-vu:
## Summary

`inspect_gguf` (dashboard-api/`gguf_inspector.py`) documents that it degrades to `unknown` **on any failure** and catches `OSError/UnicodeDecodeError/ValueError/struct.error` to that end. But GGUF array values may themselves be arrays (item type `9` = "array"), and the parser recurses on nested arrays. A crafted or corrupt model file can chain arrays deep enough to exhaust the interpreter stack, raising a **`RecursionError`** — which is a subclass of `RuntimeError`, **not** in the caught tuple, so it escapes the parser.

`inspect_gguf` is called from model fit/performance reporting (`performance_oracle.py:602,943`), which backs dashboard endpoints. An escaped `RecursionError` turns "degrade gracefully" into a **500** on a user-supplied model file.

## Reproduction (verified empirically)

A GGUF whose single metadata value is a 5000-level nested array:

```
before fix:  UNCAUGHT RecursionError escaped inspect_gguf
after  fix:  readable=False, error='GGUF array nesting too deep'
```

## Fix

Thread a `depth` counter through `_read_value` / `_read_array` / `_skip_value` and raise a `ValueError` (already handled) once nesting passes `_MAX_ARRAY_DEPTH = 64`. This degrades **deterministically** without relying on the interpreter's recursion limit (which may already be partly consumed by an async request stack). Real GGUF exporters emit flat arrays, so legitimate metadata is unaffected.

## Tests

Adds two regression tests to `test_gguf_inspector.py`:
- shallow nested array (depth 4) still round-trips (`[[[[0]]]]`);
- deeply nested array (depth 500) degrades to `readable=False` with a bounded-depth error and **no** `RecursionError`.

Full file: 26 passed locally in ~0.4s; ruff clean.

## Note on stacking

⚠️ **Stacked on #1744** (which introduces `test_gguf_inspector.py`). This PR's diff currently includes that new test file; once #1744 merges, the diff here collapses to just the `gguf_inspector.py` change + the two regression tests. Recommend merging #1744 first. (Both are independent in intent — #1744 is coverage-only, this is the behavior fix the coverage made easy to spot.)

### DIFF
--- a/ods/extensions/services/dashboard-api/gguf_inspector.py
+++ b/ods/extensions/services/dashboard-api/gguf_inspector.py
@@ -45,6 +45,12 @@ _STRUCTS = {
     12: "<d",
 }
 
+# GGUF arrays may nest (item type 9 is itself "array"), so a crafted file can
+# chain arrays deep enough to exhaust the interpreter stack. Real exporters emit
+# flat arrays, so bound the depth and degrade gracefully instead of letting a
+# RecursionError escape this parser's failure contract.
+_MAX_ARRAY_DEPTH = 64
+
 _FILE_TYPE_LABELS = {
     0: "F32",
     1: "F16",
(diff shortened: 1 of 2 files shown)


Lightheartdevs:
Thanks for hardening the GGUF parser against nested-array input; bounding recursion is the right shape for a user-supplied model metadata parser.

Decision point:
- The behavior fix itself looks sound: arrays now carry a depth counter through read and skip paths, then degrade through the existing `ValueError` handling before `RecursionError` can escape dashboard-api endpoints.
- This PR is intentionally stacked on #1744. It can merge after #1744, or maintainers can choose to fold the coverage and fix together; merge order is the only reason I am not marking it straight merge-ready. If this lands first, #1744 will likely need to be rebased or closed as duplicate coverage.

What I checked:
- Static review of `ods/extensions/services/dashboard-api/gguf_inspector.py` and `tests/test_gguf_inspector.py`.
- `python -m py_compile gguf_inspector.py` passed.
- `python -m pytest tests/test_gguf_inspector.py -q` passed: 26 tests.
- `git diff --check origin/main...HEAD` passed; `gh pr checks 1745` is green, including Dashboard API, Ruff, mypy, integration smoke, env validation, and lint jobs.

Thanks again for pushing the parser failure contract forward.

PR #1746, opened 2026-07-13
refactor(dashboard-api): consolidate duplicated .env single-key readers

tang-vu:
## Why this matters

`gpu.py` and `performance_oracle.py` independently scan the live install `.env` at production request boundaries. Keeping two low-level parsers invites drift in quote handling, empty-value semantics, encoding, and first-match behavior; a future fix in only one reader can make model recommendations and GPU state disagree about the same persisted configuration.

## Root cause and invariant

Root cause: identical file I/O and parsing lived in private helpers owned by two consumers.

Invariant: both consumers use one dependency-free reader while preserving first matching assignment, split-on-first-`=`, exact key matching, one matching quote-pair removal, explicit-empty versus missing state, and empty-on-missing-file behavior.

## What changed

- Extend the existing upstream `env_values.py` leaf module with value-only and state-returning `.env` readers.
- Delegate GPU live-state and performance-oracle reads to that shared boundary.
- Keep process-environment precedence and every public helper signature unchanged.
- Extend the existing `test_env_values.py` suite instead of introducing a competing `env_file.py` abstraction.

## Overlap check

Searched open and closed PRs for dashboard env readers and scanned all open PRs changing the production files. #3030 extends `env_values.strip_matching_quotes` with supported double-quote escape decoding; #3055 changes unrelated GPU capability booleans; both heads merge cleanly with this branch in Git's synthetic merge tree. #2807 changes decode-error tolerance in the old performance-oracle helper and remains a separate behavior decision that will need rebasing onto the shared reader if retained. #2022/#2017 alter quote/whitespace semantics rather than consolidating ownership. No merged or open PR makes both production consumers share one reader.

## Regression coverage and validation

- `python -m pytest -q tests/test_env_values.py tests/test_gpu.py tests/test_performance_oracle.py` — 121 passed.
- `python -m py_compile env_values.py gpu.py performance_oracle.py` — passed.
- Synthetic integration in merge order #1746 → #3030 → #3055 at `d863204c2ec91234273ecec53997126431a0c449` — 128 passed across the focused env/GPU/oracle suite; Python compile and diff checks passed. Inspectable branch: `tang-vu:validation/triage-env-20260824`.
- `git diff --check` — passed.
- Local Ruff was unavailable in this checkout; GitHub's Python lint job is the authoritative lint gate.

The tests exercise the shared public helper and both nearest production consumers, including embedded `=`, key prefixes, unmatched quotes, explicit empty values, missing files, and string/Path inputs.

## Tradeoffs and rollback

The shared reader intentionally catches file-system `OSError` only; it does not silently reinterpret undecodable bytes, preserving the previous consumers' strict decode behavior. `config.py` keeps its specialized reader because it has different resilience and last-value semantics. A normal revert restores duplicated helpers without changing persisted data or external API shapes.


Lightheartdevs:
Thanks for consolidating the dashboard-api `.env` reader; this removes repeated low-level parsing without pulling extra dependencies into startup config.

What looks good:
- `env_file.py` is a stdlib-only leaf module, and existing callers still check process environment where they did before, so there is no new import cycle or environment-precedence change.
- The helper preserves first-`=` splitting, first-match wins, quote trimming, and missing-file fallback, with focused tests for prefix collisions, base64 padding, and `str` install dirs.
- The touched modules stay consistent with the dashboard-api's existing top-level import style.

What I checked:
- Diff/static review of `config.py`, `gpu.py`, `performance_oracle.py`, `env_file.py`, and `tests/test_env_file.py`.
- `python -m py_compile config.py env_file.py gpu.py performance_oracle.py` passed.
- `python -m pytest tests/test_env_file.py -q` passed: 8 tests.
- `git diff --check origin/main...HEAD` passed; `gh pr checks 1746` is green, including Dashboard API, Ruff, mypy, integration smoke, env validation, and lint jobs.

tang-vu:
Rebased onto current `main`.

Scope narrowed while resolving the conflict: `config._read_env_from_file` has since been reworked upstream into `_find_env_file_value`, which returns a found-flag (so `read_live_env_value` can tell an empty value from a missing key) and tolerates undecodable bytes. Folding that into the shared helper would change behaviour, so `config` keeps its own reader and this PR now consolidates only the two still-identical helpers (`gpu._read_env_var_from_file`, `performance_oracle.read_env_file_value`). The rationale is recorded in the `env_file` module docstring.

Still behaviour-preserving. Full dashboard-api suite: 1568 passed, 2 failures that also fail on a clean `main` checkout here (Windows-only: host-agent lemonade persistence, user-extension symlink).

Lightheartdevs:
Thanks for narrowing and rebasing the `.env` reader consolidation; keeping `config` separate after the upstream found-flag change is the right boundary.

What looks good:
- `ods/extensions/services/dashboard-api/env_file.py` now consolidates only the two behaviorally identical readers in `gpu.py` and `performance_oracle.py`, while documenting why `config` keeps its specialized parser.
- The helper stays stdlib-only and leaf-like, preserving the low-level import safety that matters for dashboard-api startup.
- `ods/extensions/services/dashboard-api/tests/test_env_file.py` covers first-`=` splitting, quote stripping, prefix non-collision, first-match behavior, missing file/key fallback, and `str` install dirs.

What I checked:
- Current diff/static review after the rebase and scope change.
- `python -m py_compile env_file.py gpu.py performance_oracle.py` passed.
- `python -m pytest tests/test_env_file.py -q` passed: 8 tests.
- `git diff --check origin/main...pr-1746` passed; current GitHub checks are green, including Dashboard frontend/API, Ruff, mypy, integration smoke, matrix smoke, env validation, and lint jobs.

tang-vu:
Conflict and overlap triage receipt (2026-08-24)

Merged current upstream/main and adapted the refactor to extend upstream's existing env_values.py leaf instead of adding a competing env_file.py module. This preserves the new matching-quote behavior and keeps the PR's original single-reader invariant.

- New head: 8ec66945`n- focused env/GPU/oracle suite: 121 passed
- Python compile and git diff --check: passed
- synthetic merge trees with #3030 and #3055 are clean

The PR body now reflects the current files, overlap comparison, validation gap (local Ruff unavailable), and rollback tradeoff.

tang-vu:
Batch compatibility receipt: merged the current heads in order #1746 (`8ec66945`) → #3030 → #3055 on top of current `upstream/main`. The synthetic integration head is `d863204c2ec91234273ecec53997126431a0c449`; the focused env/GPU/performance-oracle suite passed 128 tests, and Python compile plus `git diff --check` passed. The exact combined head is available at `tang-vu:validation/triage-env-20260824` for reviewer inspection.

Issue #1747, opened 2026-07-13
Manual rollback does not resolve layered Compose stack


vaibhavsrv:
## Description

Manual rollback through `ods-update.sh rollback` invokes Docker Compose without resolving the active layered Compose stack.

ODS relies on dynamically resolved Compose files and overlays rather than a default `docker-compose.yml`, so the rollback path can fail during service orchestration.

## Current behavior

`cmd_rollback()` runs bare Compose commands for stopping and restarting services:

```bash
docker compose down
docker compose up -d

### Steps to Reproduce

1. Use an ODS installation that relies on layered Compose files and does not have a default `docker-compose.yml`.

2. Configure the current stack with an active overlay, for example:
   `docker-compose.base.yml` + `docker-compose.nvidia.yml`.

3. Create/use a pre-update rollback snapshot whose restored configuration resolves to a different stack, for example:
   `docker-compose.base.yml` + `docker-compose.cpu.yml`.

4. Run:
   `./ods-update.sh rollback <timestamp>`

5. Observe the Docker Compose invocations made by the manual rollback path.

A hermetic reproduction with a stubbed `docker` binary records the current behavior as:

compose down
compose up -d

The active layered Compose flags are not passed to either command.

### Expected Behavior

The manual rollback path should resolve the current Compose stack and use those flags when stopping the active services.

After restoring the snapshot, it should resolve the Compose stack again because the restored `.env` and Compose files may select a different backend or overlay.

For example:

docker compose -f docker-compose.base.yml -f docker-compose.nvidia.yml down

# restore snapshot and re-resolve stack

docker compose -f docker-compose.base.yml -f docker-compose.cpu.yml up -d

### Actual Behavior

`cmd_rollback()` invokes bare Docker Compose commands:

docker compose down
docker compose up -d

On a layered ODS installation without a default `docker-compose.yml`, this can fail with:

no configuration file provided: not found

The rollback path also does not account for the Compose stack potentially changing after snapshot restoration.

### Operating System

macOS

### GPU

None (not GPU-specific)

### Docker Version

N/A — reproduced hermetically with a stubbed Docker CLI

### VRAM

N/A

### Logs

```shell
=== DOCKER LOG ===
compose down
compose up -d
info
compose version
compose -f docker-compose.base.yml -f docker-compose.cpu.yml ps --services
compose -f docker-compose.base.yml -f docker-compose.cpu.yml ps --format json placeholder

[...]

PR #1748, opened 2026-07-13
fix(update): resolve compose flags in manual rollback path

vaibhavsrv:
## Summary

Fixes manual rollback so `ods-update.sh rollback` uses the resolved layered Compose stack when stopping and restarting ODS services.

`cmd_rollback()` previously invoked bare `docker compose down` and `docker compose up -d` commands. ODS relies on dynamically resolved Compose files rather than a default `docker-compose.yml`, so manual rollback could fail with `no configuration file provided: not found`.

The rollback path now resolves the current Compose stack before stopping services, restores the snapshot, then re-resolves the stack before restarting services. Re-resolving after restore is required because the restored `.env` or Compose configuration may select a different backend or overlay.

Fixes #<ISSUE_NUMBER>

## AI Assistance

AI tools assisted with code-path auditing, edge-case review, and drafting the initial regression-test approach. I reviewed the implementation and diff, reproduced the failure with a hermetic test, selected the validation scope, and verified the final behavior locally.

## Release Lane

Stable hotfix reason:

```text
N/A — this PR targets main.

### DIFF
--- a/ods/ods-update.sh
+++ b/ods/ods-update.sh
@@ -765,12 +765,19 @@ cmd_rollback() {
         log_info "Snapshot time    : ${btime}"
     fi
 
+    local compose_flags=""
+    local -a compose_args=()
+    compose_flags=$(resolve_compose_flags 2>/dev/null || true)
+    if [[ -n "$compose_flags" ]]; then
+        read -ra compose_args <<< "$compose_flags"
+    fi
+
     # Stop services
     log_info "Stopping services..."
     cd "$INSTALL_DIR"
-    if ! docker compose down; then
+    if ! docker compose "${compose_args[@]}" down; then
         log_warn "docker compose v2 down failed, trying v1..."
-        docker-compose down
+        docker-compose "${compose_args[@]}" down
     fi
 
     # Restore — use _restore_snapshot for pre-update snapshots (they include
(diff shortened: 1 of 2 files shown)


vaibhavsrv:
The issue is addressed in #1748 by @vaibhavsrv

Lightheartdevs:
Thanks for taking on manual rollback compose resolution; using the active layered stack before restore and resolving again after restore matches how ODS actually runs.

Decision point:
- The code path looks right: `cmd_rollback()` now passes resolved `-f` flags to `docker compose down`, restores the snapshot, and re-resolves before `up -d`, so a restored `.env` or backend overlay can change the restart stack.
- I would not call this merge-ready until the normal checks run or someone executes the rollback contract on a Linux/macOS environment with `rsync`; `gh pr checks 1748` currently reports no checks, and the local Git Bash run stopped before the new assertion with `[FAIL] rsync is required`.

What I checked:
- Static diff review of `ods/ods-update.sh` and `ods/tests/test-update-rollback-contract.sh`.
- `git diff --check origin/main...HEAD` passed in a PR worktree.
- `bash -n ods-update.sh tests/test-update-rollback-contract.sh` passed.
- Attempted `bash tests/test-update-rollback-contract.sh`; blocked by missing `rsync` in this environment before it reached the rollback scenario.

vaibhavsrv:
Thanks for the review.

I ran the rollback contract locally on macOS with `rsync` available. The test reaches and passes the new manual rollback regression scenario.

Environment and focused validation:

```text
ProductName:            macOS
ProductVersion:         26.5.1
BuildVersion:           25F80
openrsync: protocol version 29

[PASS] backup created before update mutation: 20260713-103310
[PASS] update mutation changed config and data
[PASS] rollback restored baseline hash
[PASS] update rollback evidence written to artifacts/update-rollback/evidence.json
[PASS] rollback dynamically resolves distinct compose flags before and after restore

vaibhavsrv:
For visibility, this PR is the existing implementation for #1747 and was opened before the later overlapping PR.

It includes regression coverage for the stack-transition case where the active pre-rollback Compose configuration differs from the configuration restored from the snapshot.

PR #1752, opened 2026-07-13
fix(dashboard): match Lemonade GGUF stem model ids

Lightheartdevs:
## Summary
- Add GGUF filename stems to dashboard-api model aliases so Lemonade/runtime stem IDs can match catalog metadata.
- Keeps this as a narrow Tier-1 main PR separate from the larger model-management UI stack.

## Validation
- `python -m pytest extensions/services/dashboard-api/tests/test_performance_oracle.py -q`

### DIFF
--- a/ods/extensions/services/dashboard-api/performance_oracle.py
+++ b/ods/extensions/services/dashboard-api/performance_oracle.py
@@ -118,6 +118,10 @@ def _model_aliases(model: dict[str, Any]) -> set[str]:
     for part in model.get("gguf_parts", []) or []:
         if isinstance(part, dict):
             aliases.add(str(part.get("file") or ""))
+    for alias in tuple(aliases):
+        path_name = re.split(r"[\\/]", alias)[-1]
+        if path_name.lower().endswith(".gguf"):
+            aliases.add(path_name[:-5])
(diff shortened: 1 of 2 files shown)


Lightheartdevs:
Thanks for tightening the Lemonade/runtime model-id matching without pulling in the larger model-management UI work.

What looks good:
- The alias expansion stays inside `ods/extensions/services/dashboard-api/performance_oracle.py` and feeds the existing normalized exact-match path, so it fixes GGUF stem identities without loosening hardware/runtime evidence matching.
- The new `test_published_exact_matches_gguf_stem_identity` covers the regression path: evidence keyed as `Qwen3.5-9B-Q4_K_M` now matches the catalog GGUF `Qwen3.5-9B-Q4_K_M.gguf`.
- CI is green across dashboard API, frontend, lint, type check, matrix smoke, and integration smoke.

What I checked:
- Diff/static review of `performance_oracle.py` and `tests/test_performance_oracle.py`.
- `python -m pytest tests/test_performance_oracle.py -q` in the PR worktree: 14 passed.
- `gh pr checks 1752`: all reported checks pass.

Thanks again for keeping this one focused.

PR #1754, opened 2026-07-13
fix(preflight): resolve shared Python command path

### LINKED ISSUE #1753: preflight-engine skips shared Python command resolver due to incorrect root path
### Description

`ods/scripts/preflight-engine.sh` fails to load the shared Python command resolver from `ods/lib/python-cmd.sh` when invoked with `--script-dir` set to the ODS root directory, which is how the existing installer, doctor, simulator, and preflight fixture callers invoke it.

`preflight-engine.sh` currently computes `ROOT_DIR` as `${SCRIPT_DIR}/..`. Since callers already pass the ODS root as `SCRIPT_DIR`, this points one directory above ODS and causes the resolver lookup to check the wrong path.

The missing resolver is silently skipped, causing the preflight engine to fall back directly to `python3` or `python` from `PATH`. This bypasses shared Python command resolution, including configured command overrides and handling for non-functional Python aliases.

### Steps to Reproduce

`preflight-engine.sh` should locate and source `ods/lib/python-cmd.sh` when `--script-dir` points to the ODS root.

The Python command selected by the shared resolver should be used by the preflight capability checks.

### Expected Behavior

`preflight-engine.sh` should locate and source `ods/lib/python-cmd.sh` when `--script-dir` points to the ODS root.

The Python command selected by the shared resolver should be used by the preflight capability checks.

### Actual Behavior

`preflight-engine.sh` calculates `ROOT_DIR` as `${SCRIPT_DIR}/..`.

Existing callers pass the ODS root as `--script-dir`, so this calculation points one directory above ODS and checks for:

[ODS_PARENT]/lib/python-cmd.sh

instead of:

[ODS_ROOT]/lib/python-cmd.sh

The resolver is silently skipped and the engine falls back directly to `python3` or `python` from PATH.

In a hermetic reproduction where the shared resolver selects a working mock interpreter and the direct PATH Python commands fail, the preflight engine exits with code 99 from the failing PATH command.

### Operating System

macOS 26.5.1

### GPU

None (not GPU-specific)

### Docker Version

N/A — not Docker-specific

### VRAM

N/A

### Logs

```shell
Running preflight-engine.sh with --script-dir <temp>/ods
Exit code: 99
Output: FAIL: System python called
FAILURE: Shared resolver was NOT loaded/used!
```

### Screenshots

_No response_

vaibhavsrv:
## Summary

Fixes `preflight-engine.sh` so it correctly loads the shared Python command resolver when `--script-dir` points to the ODS root.

Existing callers pass the ODS root through `--script-dir`, but the preflight engine previously moved one directory above that path before looking for `lib/python-cmd.sh`. This silently skipped the shared resolver and fell back directly to `python3` or `python` from `PATH`.

Fixes #1753 
## AI Assistance

AI tools assisted with code-path auditing, reproduction design, edge-case review, and drafting the regression-test approach. I reviewed the implementation and diff, reproduced the failure hermetically, selected the validation scope, and verified the final behavior locally.

## Release Lane

Stable hotfix reason:

```text
N/A — this PR targets main.

### DIFF
--- a/.github/workflows/test-linux.yml
+++ b/.github/workflows/test-linux.yml
@@ -118,6 +118,7 @@ jobs:
         run: |
           bash tests/contracts/test-installer-contracts.sh
           bash tests/contracts/test-preflight-fixtures.sh
+          bash tests/test-preflight-resolver-bug.sh
           python3 tests/contracts/test-network-exposure-contracts.py
 
       - name: Linux install preflight tests
--- a/ods/Makefile
+++ b/ods/Makefile
@@ -25,6 +25,7 @@ test: ## Run unit and contract tests
 	@echo "=== Installer contracts ==="
 	@bash tests/contracts/test-installer-contracts.sh
 	@bash tests/contracts/test-preflight-fixtures.sh
+	@bash tests/test-preflight-resolver-bug.sh
 	@bash tests/test-linux-cloud-mode.sh
 	@echo ""
 	@echo "=== Linux install preflight ==="
--- a/ods/scripts/preflight-engine.sh
+++ b/ods/scripts/preflight-engine.sh
@@ -10,7 +10,7 @@ GPU_VRAM_MB="${GPU_VRAM_MB:-0}"
 GPU_NAME="${GPU_NAME:-Unknown}"
 PLATFORM_ID="${PLATFORM_ID:-linux}"
 COMPOSE_OVERLAYS="${COMPOSE_OVERLAYS:-}"
-SCRIPT_DIR="${SCRIPT_DIR:-$(pwd)}"
+SCRIPT_DIR="${SCRIPT_DIR:-$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)}"
 STRICT="false"
 ENV_MODE="false"
 
@@ -71,7 +71,7 @@ while [[ $# -gt 0 ]]; do
     esac
 done
 
-ROOT_DIR="$(cd "${SCRIPT_DIR}/.." && pwd)"
+ROOT_DIR="$SCRIPT_DIR"
 PYTHON_CMD="python3"
 if [[ -f "$ROOT_DIR/lib/python-cmd.sh" ]]; then
     . "$ROOT_DIR/lib/python-cmd.sh"
--- a/ods/tests/test-preflight-resolver-bug.sh
+++ b/ods/tests/test-preflight-resolver-bug.sh
@@ -1,0 +1,73 @@
+#!/usr/bin/env bash
+# ============================================================================
+# ODS preflight-engine.sh Shared Resolver Bug Regression Test
+# ============================================================================
+set -euo pipefail
+
+# Create temporary directory layout
+TEMP_WORKSPACE="$(mktemp -d)"
+trap 'rm -rf "$TEMP_WORKSPACE"' EXIT
+
+# Create temporary ODS root directory
+TEMP_ODS="$TEMP_WORKSPACE/ods"
+mkdir -p "$TEMP_ODS/lib"
+mkdir -p "$TEMP_ODS/scripts"
+
+# Create a mock python-cmd.sh
+cat > "$TEMP_ODS/lib/python-cmd.sh" <<'EOF'
+ods_detect_python_cmd() {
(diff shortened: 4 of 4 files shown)


Lightheartdevs:
Thanks for tightening the preflight Python resolver path; treating `--script-dir` as the ODS root matches the installer callers in `ods/installers/phases/04-requirements.sh`, `ods/installers/macos.sh`, and `ods/scripts/simulate-installers.sh`.

Quickest path to merge:
- `ods/tests/test-preflight-resolver-bug.sh`: make the fixture copy independent of the launch cwd. The repo's CI and docs run tests from `ods/` (`working-directory: ods`, and existing tests use `bash tests/...`), but this test currently does `cp ods/scripts/preflight-engine.sh ...`, so `cd ods && bash tests/test-preflight-resolver-bug.sh` fails with `cp: cannot stat 'ods/scripts/preflight-engine.sh'`. Derive `ROOT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"` and copy `"$ROOT_DIR/scripts/preflight-engine.sh"` instead.
- Wire the new regression into an existing gate, likely `.github/workflows/test-linux.yml` near the preflight/installer checks and `ods/Makefile`'s `test` target. As written, the new test passes from the repository root but would not catch this regression in the normal PR checks.

What I checked:
- Static review of the two-file diff and the current callers that pass `--script-dir "$SCRIPT_DIR"` / `"$ROOT_DIR"` as the ODS root.
- `git diff --check origin/main...HEAD`: pass.
- `C:\Program Files\Git\bin\bash.exe -lc 'bash -n scripts/preflight-engine.sh tests/test-preflight-resolver-bug.sh tests/test-preflight.sh'`: pass.
- `C:\Program Files\Git\bin\bash.exe -lc 'bash tests/test-preflight.sh'` from `ods/`: pass, 22 passed.
- `C:\Program Files\Git\bin\bash.exe -lc 'bash tests/test-preflight-resolver-bug.sh'` from `ods/`: fails on the relative `cp` path.
- `C:\Program Files\Git\bin\bash.exe -lc 'bash ods/tests/test-preflight-resolver-bug.sh'` from the repository root: pass.
- `gh pr checks 1754`: no checks reported yet.

Thanks again for pushing this preflight fix forward.

vaibhavsrv:
Thanks for the review. Addressed both findings.

- Made `test-preflight-resolver-bug.sh` independent of the launch cwd by deriving the ODS root from `${BASH_SOURCE[0]}`.
- Registered the regression in the Makefile test target and the Linux installer contract checks.

Verified the focused test from both invocation contexts:

- `bash ods/tests/test-preflight-resolver-bug.sh` — PASS
- `cd ods && bash tests/test-preflight-resolver-bug.sh` — PASS

Also re-ran the preflight fixture contracts, Bash syntax checks, ShellCheck, and `git diff --check`; all pass.

Lightheartdevs:
Thanks for turning around the preflight resolver follow-up; the latest revision now matches the repo's normal `ods/` test working directory and keeps the shared Python resolver on the exercised path.

What looks good:
- `ods/scripts/preflight-engine.sh` now treats `--script-dir` as the ODS root, which matches the installer, doctor, support-bundle, host-agent, and test callers that pass the install/root directory directly.
- `ods/tests/test-preflight-resolver-bug.sh` masks `python3` and `python`, so the regression only passes when `lib/python-cmd.sh` is actually loaded; it also derives its root from `${BASH_SOURCE[0]}` and passes from both repo-root and `ods/` invocation styles.
- The regression is wired into both `.github/workflows/test-linux.yml` and `ods/Makefile`, next to the existing installer/preflight contract checks.

What I checked:
- Static review of the diff and the current `preflight-engine.sh --script-dir` callers.
- `git diff --check origin/main...HEAD`: pass.
- `bash -n ods/scripts/preflight-engine.sh ods/tests/test-preflight-resolver-bug.sh`: pass.
- `bash ods/tests/test-preflight-resolver-bug.sh`: pass.
- `cd ods && bash tests/test-preflight-resolver-bug.sh`: pass.
- `cd ods && bash tests/contracts/test-preflight-fixtures.sh`: pass, all preflight fixtures green.
- `gh pr checks 1754`: no checks reported yet.

PR #1773, opened 2026-07-14
fix(windows): abort installer when a phase hits a fatal error (exit 1 in dot-sourced phases is a no-op)

### LINKED ISSUE #1771: bug(windows): installer continues past fatal Phase 2 driver check, crashes in Phase 3 on null $tierConfig
## Summary

When Phase 2 (hardware detection) fails the NVIDIA driver minimum check, the Windows installer prints the error but does **not** abort. Phase 3 then dereferences the never-assigned `$tierConfig` and crashes with a null method call.

## Repro

Windows 11, GTX 1650 with driver 516.40 (< required 570):

```powershell
.\install.ps1 -DryRun
```

```
  [XX] NVIDIA driver 516.40 is below the required minimum (570).
  >   Update your driver: https://www.nvidia.com/Download/index.aspx
  >   After updating, re-run this installer.

  ODSGATE SEQUENCE [00:00:13]  PHASE 3/13 -- FEATURE SELECTION
  ...
  [!!] Hermes enabled: increasing llama context from  to 65536 (64K floor).
You cannot call a method on a null-valued expression.
```

(Note the empty value in "from  to 65536" — `$tierConfig.MaxContext` is null.)

## Root cause

- Phase 2's driver-too-old path emits the error and returns without setting `$tierConfig`, and the orchestrator keeps sourcing subsequent phases.
- `ods/installers/windows/phases/03-features.ps1` (Hermes context floor block, ~line 154): `[int]$tierConfig.MaxContext` evaluates on `$null`, then `$tierConfig.ContainsKey("RecommendationReason")` throws `InvokeMethodOnNull`.

## Expected

Fatal Phase 2 check aborts the install with the driver guidance as the final message — the user should never see an unrelated PowerShell stack trace after an already-diagnosed problem.

Same crash occurs with and without `-DryRun`.

## Environment

ODS v2.5.3 installer, Windows 11 Home 10.0.26200, PowerShell 5.1, GTX 1650 (driver 516.40).

patil2001:
## Summary

Fixes #1771.

`exit 1` inside a dot-sourced phase script does **not** stop the calling orchestrator — PowerShell resumes at the next statement. Verified with a minimal repro on PowerShell 5.1: a dot-sourced file executing `exit 1` returns control to the caller, which continues and exits 0. This affects **all 16 fatal-exit sites** across phases 01/02/04/06, not just the driver check.

Observed effect: with an outdated NVIDIA driver, Phase 2 prints the fatal driver error, then Phase 3 runs anyway and crashes with `You cannot call a method on a null-valued expression` on the never-assigned `$tierConfig`.

## Change

- Phases 01/02/04/06: replace every fatal `exit 1` with `throw "ODS_INSTALL_ABORTED"` (the phase has already printed its guidance via `Write-AIError` at these sites).
- `install-windows.ps1`: wrap the phase dot-source block in `try/catch`; on the sentinel, `exit 1` cleanly with no stack trace. Any other exception rethrows unchanged.
- Updated the modder note in `01-preflight.ps1` accordingly.

No behavior change for `-Force` paths or successful installs.

## Test

- `[System.Management.Automation.PSParser]::Tokenize` clean on all 5 edited files (0 syntax errors).
- Sentinel pattern verified standalone: fatal phase → orchestrator stops before next phase, exit code 1, no PowerShell error spam.
- Real-hardware repro of the original bug (GTX 1650 + driver 516.40, v2.5.3): error then Phase 3 crash — with this change the install stops at the driver message.

### DIFF
--- a/ods/installers/windows/install-windows.ps1
+++ b/ods/installers/windows/install-windows.ps1
@@ -191,6 +191,11 @@ Write-ODSBanner
 #  Phase 06 → $envResult (SearxngSecret, OpenclawToken)
 #  Phase 07 → (no output -- tools installed to $env:USERPROFILE)
 
+# Phases signal a fatal, already-explained failure by throwing the
+# ODS_INSTALL_ABORTED sentinel. `exit 1` inside a dot-sourced file does NOT
+# stop this orchestrator (PowerShell resumes at the next statement), which
+# previously let later phases run against half-initialized state.
+try {
 . (Join-Path $PhasesDir "01-preflight.ps1")
 . (Join-Path $PhasesDir "02-detection.ps1")
 . (Join-Path $PhasesDir "03-features.ps1")
@@ -210,6 +215,10 @@ if ($gpuInfo.Backend -eq "amd") {
 }
 . (Join-Path $PhasesDir "06-directories.ps1")
 . (Join-Path $PhasesDir "07-devtools.ps1")
+} catch {
+    if ($_.FullyQualifiedErrorId -eq "ODS_INSTALL_ABORTED") { exit 1 }
+    throw
+}
 
 # ============================================================================
 # PHASE 8 -- LAUNCH (download model, start Docker services)
--- a/ods/installers/windows/phases/01-preflight.ps1
+++ b/ods/installers/windows/phases/01-preflight.ps1
@@ -17,7 +17,8 @@
 #
 # Modder notes:
 #   Add new pre-flight checks (e.g., minimum Windows build) here.
-#   All fatal exits use `exit 1`. Non-fatal issues use Write-AIWarn.
+#   All fatal exits use `throw "ODS_INSTALL_ABORTED"` (caught by the
+#   orchestrator, which exits 1). Non-fatal issues use Write-AIWarn.
 # ============================================================================
 
 Write-Phase -Phase 1 -Total 13 -Name "PRE-FLIGHT CHECKS" -Estimate "~30 seconds"
@@ -50,7 +51,7 @@ Write-InfoBox "PowerShell:" "$($_psVer.Version)"
 if (-not $_psVer.Sufficient) {
     Write-AIError "PowerShell 5.1 or later is required."
     Write-AI "  Download: https://github.com/PowerShell/PowerShell/releases"
-    exit 1
+    throw "ODS_INSTALL_ABORTED"
 }
 Write-AISuccess "PowerShell $($_psVer.Version) OK"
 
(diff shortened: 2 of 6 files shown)


Lightheartdevs:
Thanks for taking on the Windows phase-abort path here; replacing fatal `exit 1` with an orchestrator-visible sentinel is the right direction for dot-sourced phases.

Quickest path to merge:
- In `ods/installers/windows/phases/01-preflight.ps1`, the Windows-build fatal path still does not abort. The new `throw ODS_INSTALL_ABORTED` at line 65 is inside the existing `try` whose `catch` at line 69 treats every exception as “Could not determine Windows build -- continuing”, so the sentinel is swallowed before `install-windows.ps1` can catch it. Re-throw the sentinel from that local catch, or move the fatal build check outside the broad probe catch.
- Add a focused regression for this shape: a dot-sourced phase throws `ODS_INSTALL_ABORTED` from inside a local `try`, the phase-local catch must not swallow it, and the orchestrator must stop before the next phase.

What I checked:
- Diff/static review of `install-windows.ps1` and phases 01/02/04/06.
- PowerShell parser checks passed for all five modified files.
- Minimal PowerShell sentinel check confirmed `throw ODS_INSTALL_ABORTED` is catchable by FQID, but also confirmed a local `catch { ... }` swallows it unless it explicitly rethrows.
- `gh pr checks 1773` currently reports no checks for this branch.

Thanks again for closing the installer continuation bug; this looks close once the sentinel cannot be swallowed locally.

patil2001:
Fixed the swallowed sentinel: the preflight Windows-build catch now rethrows `ODS_INSTALL_ABORTED` before the 'continuing' fallback. Added `ods/tests/test-windows-phase-abort.sh` covering the exact shape flagged in review (sentinel thrown from inside a local try, local catch must rethrow, orchestrator exits 1 before the next phase runs). All 4 checks pass locally; parser check clean.

Lightheartdevs:
Thanks for closing the swallowed-sentinel path in the Windows preflight phase; the follow-up fixes the exact local-catch shape that let a fatal phase continue.

Quickest path to merge:
- Wire `ods/tests/test-windows-phase-abort.sh` into the normal validation surface, for example `ods/Makefile`'s `test` target and `.github/workflows/test-linux.yml` or another CI lane that has PowerShell available. Right now the regression is present in the tree but is not invoked by the repo's standard test entry points, and `gh pr checks 1773` reports no checks for this branch.
- If the test runs on Linux CI, make sure the behavioral portion actually has `pwsh` available rather than silently covering only the static grep checks. This change is in the Windows installer's fatal-abort path, so the useful signal is proving the orchestrator exits before the next dot-sourced phase.

What I checked:
- Diff/static review of `install-windows.ps1`, phases 01/02/04/06, and the new `ods/tests/test-windows-phase-abort.sh`.
- PowerShell parser checks passed for all five edited `.ps1` files.
- A direct PowerShell repro of the phase-local catch/orchestrator contract exited 1 and did not run the next phase.
- I could not run the new shell test through `bash` locally because this Windows host resolves `bash` to a broken WSL shim with no `/bin/bash`.

Thanks again for tightening this installer failure mode.

PR #1774, opened 2026-07-14
fix(windows): make enable/disable work without a usable WSL bash

### LINKED ISSUE #1770: bug(windows): 'ods.ps1 enable/disable' crashes via WSL bash without a Linux distro, fallback is dead code, leaves stack half-modified
## Summary

On a standard Windows machine where the only WSL distro is `docker-desktop` (i.e. Docker Desktop installed, no Ubuntu/other distro), `.\ods.ps1 disable <service>` stops the container and then **crashes** while regenerating `.compose-flags`, leaving the install in an inconsistent state. Follow-up to #1699 / PR #1700.

## Repro

Windows 11 Home, Docker Desktop (WSL2 backend), no other WSL distro. Fresh v2.5.3 install.

```powershell
cd $HOME\ods
.\ods.ps1 disable token-spy
```

Output:

```
  > Stopping token-spy...
bash.exe : <3>WSL (92811 - Relay) ERROR: CreateProcessCommon:818: execvpe(/bin/bash) failed: No such file or directory
At C:\Users\<user>\ods\ods.ps1:1779 char:29
+         $resolvedFlagsRaw = & $bashExe.Source "$resolverScript" `
```

Exit code 1.

## Three stacked problems

1. **Insufficient bash detection.** `Get-Command bash` finds `C:\Windows\System32\bash.exe` (the WSL relay), which exists whenever WSL is enabled — but with only `docker-desktop` registered there is no `/bin/bash`, so `execvpe` fails. The check needs to verify bash actually runs (e.g. `& $bashExe.Source -c "exit 0"` and test `$LASTEXITCODE`), or prefer Git Bash.

2. **The fallback is unreachable.** The call site redirects stderr (`2>$null`). Under PowerShell 5.1 with `$ErrorActionPreference = "Stop"`, redirecting a native command's stderr wraps the output in a `NativeCommandError` and **terminates the script** — so "Path 2: minimal in-process swap" (the non-bash fallback a few lines below) is dead code on exactly the machines that need it.

3. **No rollback on failure.** The container is already stopped when the crash happens; `.compose-flags` is left unmodified and no disabled marker is written. `status` still lists the service as part of the stack while it is down.

## Environment

- ODS v2.5.3, Windows 11 Home 10.0.26200, PowerShell 5.1
- Docker Desktop 29.6.1, WSL distro list: `docker-desktop` only

patil2001:
## Summary

Fixes #1770. On a standard Docker Desktop setup (only WSL distro is `docker-desktop`), `.\ods.ps1 disable <service>` crashed mid-command and left the stack half-modified. Three fixes to `Update-ComposeFlags`:

1. **Probe bash before trusting it.** `Get-Command bash` resolves to the System32 WSL relay whenever WSL is enabled, but it fails with `execvpe(/bin/bash) failed` when no full Linux distro is registered. Now a quick `bash -c "exit 0"` probe demotes it and routes to the fallback with a clear warning.

2. **Stop the NativeCommandError kill.** Under `$ErrorActionPreference = "Stop"`, PowerShell 5.1 wraps redirected native stderr in a terminating error — the resolver call died before reaching the fallback, making "Path 2" dead code. Native bash calls now run under a scoped non-terminating preference.

3. **Fallback applies a single-service delta.** The old fallback re-scanned `extensions/services` and re-added every directory with a `compose.yaml` — but installer-skipped services also ship a plain `compose.yaml`, so one `disable` rewrote `.compose-flags` from 8 services to 19. `Update-ComposeFlags` now takes `-AddService`/`-RemoveService` and edits only that entry, preserving everything else verbatim.

## Test

Real Windows 11 Home install (v2.5.3, Docker Desktop, `docker-desktop` distro only):

- Before: `disable token-spy` → `execvpe(/bin/bash) failed`, exit 1, container stopped but `.compose-flags` untouched; once the fallback was reachable it rewrote flags to all 19 services.
- After: `enable` → `disable` → `enable` round-trip completes; `.compose-flags` gains/loses exactly `extensions/services/token-spy/compose.yaml`; all other entries byte-identical. PSParser tokenize: 0 syntax errors.

### DIFF
--- a/ods/installers/windows/ods.ps1
+++ b/ods/installers/windows/ods.ps1
@@ -1960,7 +1960,17 @@ function Invoke-Enable {
     $disabledPath = Join-Path $svcDir "compose.yaml.disabled"
 
     if (Test-Path $composePath) {
-        Write-AISuccess "$ServiceId is already enabled."
+        # Installer-skipped services also carry a plain compose.yaml but are
+        # absent from .compose-flags -- "already enabled" would be misleading
+        # and the service would never join the stack. Add the missing entry.
+        $flagsFile = Join-Path $InstallDir ".compose-flags"
+        $relPath = "extensions/services/$ServiceId/compose.yaml"
+        if ((Test-Path $flagsFile) -and ((Get-Content $flagsFile -Raw) -notmatch [regex]::Escape($relPath))) {
+            Update-ComposeFlags -AddService $ServiceId
+            Write-AISuccess "$ServiceId enabled (added to compose stack)."
+        } else {
+            Write-AISuccess "$ServiceId is already enabled."
+        }
         Write-AI "Run '.\ods.ps1 start $ServiceId' to launch it."
         return
     }
--- a/ods/installers/windows/ods.ps1
+++ b/ods/installers/windows/ods.ps1
@@ -1735,20 +1735,23 @@ function Update-ComposeFlags {
            overlays, multi-GPU overlays, user-extension overlays, and
            docker-compose.override.yml -- exactly the same stack the installer
            built). This is the safe path.
-        2. Otherwise fall back to a minimal in-process swap: keep every token
-           in the existing .compose-flags that is NOT an extension service -f
-           entry, then re-scan extensions/services for enabled compose.yaml
-           fragments and append them. This preserves all backend and GPU
-           overlays (--env-file, -f docker-compose.base.yml,
-           -f docker-compose.nvidia.yml, etc.) because those paths never
-           match 'extensions/services' and are kept verbatim.
+        2. Otherwise fall back to a minimal in-process swap: apply only the
+           single-service delta (-AddService / -RemoveService) to the existing
+           .compose-flags. Every other token -- --env-file, base compose, GPU
+           overlays, and all other extension entries -- is kept verbatim.
 
-        The fallback intentionally mirrors only what the Windows installer
-        writes: base + GPU overlay + enabled extension compose.yaml entries.
-        It does NOT add GPU-specific per-extension overlays (compose.nvidia.yaml
-        etc.) because those are the canonical resolver's responsibility and
-        we must not silently diverge from it.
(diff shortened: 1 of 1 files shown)


patil2001:
Pushed a second commit covering edge cases found while testing on a live Tier 1 install:

- **`enable` of an installer-skipped service** (e.g. `n8n`): these carry a plain `compose.yaml` but were never in `.compose-flags`, so the CLI said "already enabled" while the service could never join the stack. Now the missing entry is added and the message says "enabled (added to compose stack)".
- **Prefix collision**: verified `disable hermes` leaves `hermes-proxy` entries untouched (removal regex requires a trailing path separator after the service id).
- **Idempotency**: repeated `enable` keeps a single flags entry (no duplicates).

Full round-trip re-verified: enable → disable → enable restores byte-identical flags.

Lightheartdevs:
Thanks for tightening the Windows enable/disable path around the no-usable-WSL case; preserving `.compose-flags` as a single-service delta is the right fix for this failure mode.

What looks good:
- `Update-ComposeFlags` now probes `bash` before using the resolver and lets resolver stderr fall back instead of terminating the CLI under `$ErrorActionPreference = Stop`.
- The fallback no longer re-adds every bundled `compose.yaml`; it removes/adds only the target service and preserves backend overlays plus unrelated extension entries.
- The installer-skipped-service path now adds a missing compose entry instead of reporting “already enabled.”

What I checked:
- Diff/static review of `ods/installers/windows/ods.ps1`.
- PowerShell parser check passed for `ods/installers/windows/ods.ps1`.
- Git Bash: `bash ods/tests/test-windows-cli-enable-disable.sh` passed.
- Temp-install PowerShell exercise with unusable `bash`: `disable token-spy` removed only token-spy, `disable hermes` preserved `hermes-proxy`, and `enable n8n` added exactly one compose entry.
- `gh pr checks 1774` currently reports no checks for this branch.

Thanks again for pushing this Windows CLI reliability fix forward.

PR #1777, opened 2026-07-14
installer(amd): select gfx target by VRAM, not lexicographic order

timgilmour:
installer(amd): select gfx target by VRAM, not lexicographic order

The AMD gfx target was chosen with:

    [.gpus[]?.gfx_version] | unique | .[0]

jq's `unique` sorts ascending, so on a mixed iGPU + dGPU host the weakest device
wins by lexicographic accident: ["gfx1036","gfx1201"] -> gfx1036, the integrated
Raphael display GPU. There is no "pick the best GPU" logic anywhere in the path.

Consequences on such a host (2x Radeon AI PRO R9700 + Raphael iGPU):

  - llama.cpp is compiled for gfx1036, while docker-compose.amd.yml schedules
    inference onto the gfx1201 cards. Wrong-ISA binary on the target GPUs.
  - ROCm ships no rocBLAS/hipBLASLt kernels for gfx1036 at all, so the image also
    ends up with no usable BLAS kernels (see the companion Dockerfile.amd fix).
  - The install still reports green. The build succeeds; it is simply the wrong
    target.

Select the gfx of the highest-VRAM GPU instead (31.9GB R9700 over the 2.0GB iGPU),
which is the card inference actually runs on. The failed-probe fallback to gfx1151
is preserved.

---

### DIFF
--- a/ods/installers/phases/06-directories.sh
+++ b/ods/installers/phases/06-directories.sh
@@ -734,7 +734,18 @@ COMFYUI_CPU_RESERVATION=${COMFYUI_CPU_RESERVATION}
 $(if [[ "$GPU_BACKEND" == "amd" ]]; then
     # Read gfx target from topology detection. Falls back to gfx1151 (Strix Halo)
     # if the topology probe failed — preserves prior behavior for the OG target.
-    _amd_gfx_detected=$(echo "${GPU_TOPOLOGY_JSON:-{\}}" | jq -r '[.gpus[]?.gfx_version] | unique | .[0] // "gfx1151"' 2>/dev/null || echo "gfx1151")
+    #
+    # Select the gfx of the GPU with the most VRAM, which is the card inference will
+    # actually run on. Do NOT use `unique | .[0]`: unique sorts ascending, so on a
+    # mixed iGPU + dGPU box the weakest device wins on a lexicographic accident
+    # (["gfx1036","gfx1201"] -> gfx1036, the Raphael display iGPU). That silently
+    # compiles the entire inference engine for a GPU it is never scheduled onto, and
+    # ROCm ships no BLAS kernels for those iGPU targets at all.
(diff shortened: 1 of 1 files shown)


Lightheartdevs:
Thanks for tightening the AMD gfx-target selection around the GPU that will actually run inference.

Decision point:
- The change in `ods/installers/phases/06-directories.sh` is directionally right for mixed iGPU + dGPU hosts: it ignores unknown gfx values, sorts by `memory_gb`, and keeps the `gfx1151` fallback. Because this writes install-time `.env` values, I would still gate merge on an AMD installer dry-run or real host check showing `AMDGPU_TARGET` follows the intended compute card and `HSA_OVERRIDE_GFX_VERSION` remains Strix-only.

What I checked:
- Static diff against `origin/main` for the phase-06 jq selection logic.
- I could not run shell syntax/installer tests in this Windows environment because the available `bash.exe` has no working WSL distro.

Thanks again for pushing this AMD path forward.

timgilmour:
Real-host evidence for the gate you asked for. This fix came off a mixed host: 2× Radeon AI PRO R9700 (gfx1201, 32 GB) plus a Raphael display iGPU (gfx1036). Plain main's installer selected the iGPU on this box (lexicographic accident); with this branch's phase-06 logic it selects the compute card.

Current state of that install, re-verified 2026-07-19:

- `.env` has `AMDGPU_TARGET=gfx1201` — chosen by VRAM, not name order.
- `HSA_OVERRIDE_GFX_VERSION` is not written at all on this host (the `.env` carries a comment documenting the deliberate unset; the compose pass-through leaves it unset in the container too, preserving the unset-vs-empty distinction from #1779). The override remains Strix-only, as intended.

Installer/AMD contract suites are green on this branch standalone (`make test`, ~400 checks; BATS 405/406 — the single failure is a pre-existing host-env leak in `detect_gpu`, verified failing identically on plain main on the same box).

PR #1778, opened 2026-07-14
llama-server(amd): build for the requested AMDGPU_TARGET, and fail loudly if it cannot

timgilmour:
llama-server(amd): build for the requested AMDGPU_TARGET, and fail loudly if it cannot

Dockerfile.amd silently produced a broken image for any target other than gfx1151.

1. Empty BLAS kernel directories, reported as a successful build.

   The rocBLAS/hipBLASLt kernels are copied with a glob on ${AMDGPU_TARGET}. When
   ROCm ships no kernels for the target, the glob matches nothing and cp fails --
   but the trailing `|| true` rescues the whole && chain, because

       (A && B) || true && C

   swallows a failure in A. The image then ships with EMPTY rocblas/library and
   hipblaslt/library, and the build reports green.

   Not hypothetical: an installer host resolved AMDGPU_TARGET=gfx1036 (an iGPU),
   ROCm 7.2 ships no gfx1036 BLAS kernels at all, and the resulting image had empty
   kernel dirs. Now the target is validated up front and the build fails with the
   list of targets the ROCm image actually supports. The genuinely-optional
   hipblasltTransform copy is parenthesized so its `|| true` cannot mask an earlier
   failure, and the redundant TensileLibrary_lazy_${AMDGPU_TARGET}.dat copy is
   dropped -- the preceding glob already matches it.

2. The Strix Halo MMQ patch was applied to every target.

   The mmq_x 64->48 / mmq_y 128->64 / nwarps 8->4 patch relieves register pressure
   specific to gfx1151's RDNA 3.5 register file. On architectures without that
   pressure it is a straight throughput loss. Gated to gfx1151.

Also re-declares ARG ROCM_VERSION inside the builder stage; the pre-FROM declaration
is global and was not visible to RUN, so it interpolated empty.

---

### DIFF
--- a/ods/extensions/services/llama-server/Dockerfile.amd
+++ b/ods/extensions/services/llama-server/Dockerfile.amd
@@ -23,6 +23,8 @@ FROM rocm/dev-ubuntu-24.04:${ROCM_VERSION}-complete AS builder
 
 ARG AMDGPU_TARGET=gfx1151
 ARG LLAMA_CPP_REF=b8763
+# Re-declared after FROM: the pre-FROM ARG is global and not visible inside this stage.
+ARG ROCM_VERSION
 
 RUN apt-get update && apt-get install -y --no-install-recommends \
         cmake ninja-build git ca-certificates \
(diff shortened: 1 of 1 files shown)


Lightheartdevs:
Thanks for making the AMD llama-server image fail loudly when the requested target is not actually present.

Decision point:
- The `ods/extensions/services/llama-server/Dockerfile.amd` changes look coherent: the `gfx1151` MMQ patch is gated, `ROCM_VERSION` is correctly re-declared after `FROM`, and the ROCm kernel-copy step now validates target-specific rocBLAS/hipBLASLt files instead of letting empty globs pass. Because this is a build/runtime path, I would want an actual image build for `gfx1151` and at least one native target such as `gfx1201` before treating it as merge-ready.

What I checked:
- Static Dockerfile review against `origin/main`.
- I did not run a Docker image build locally; that is the main remaining validation gap.

Thanks again for making this failure mode much easier to catch.

timgilmour:
Both builds you asked for now exist:

- **gfx1201 (native, on metal):** the llama-server image on a 2× Radeon AI PRO R9700 box was built with this Dockerfile for `AMDGPU_TARGET=gfx1201`, passed the kernel-dir validation, and serves Lemonade completions daily (`LEMONADE_LLAMACPP=rocm` in the running container). The validation step exists because the pre-fix Dockerfile "built green" on this exact box while shipping empty target-specific rocBLAS dirs, failing only at runtime.
- **gfx1151 (cross-compile, 2026-07-19):** a fully uncached build (0 cached layers) with `--build-arg AMDGPU_TARGET=gfx1151` completed successfully: the Strix MMQ patch gated on and applied, cmake compiled with `-DAMDGPU_TARGETS=gfx1151`, and the new validation step found and bundled ROCm 7.2's gfx1151 rocBLAS + hipBLASLt kernels into the image (3.8 GB). Had the kernels been absent it would have failed loudly with the supported-target listing — the behavior this PR adds.

No Strix hardware here, so runtime-on-metal for gfx1151 still needs a maintainer with the part.

PR #1779, opened 2026-07-14
compose(amd): never define HSA_OVERRIDE_GFX_VERSION as an empty string

timgilmour:
compose(amd): never define HSA_OVERRIDE_GFX_VERSION as an empty string

This silently disables the GPU on every AMD host that is not Strix Halo.

Both compose files inject the variable with

    HSA_OVERRIDE_GFX_VERSION=${HSA_OVERRIDE_GFX_VERSION:-}

which does not leave it unset -- it defines it as an EMPTY STRING. ROCm's HSA
runtime checks whether the variable is *defined*, not whether it is non-empty. It
then tries to parse "" as a gfx version, fails, and enumerates ZERO devices:

    ggml_cuda_init: failed to initialize ROCm: no ROCm-capable device is detected

llama.cpp falls back to CPU, or Lemonade falls back to its Vulkan build on radv.
Nothing errors out; you just quietly lose every GPU.

The irony is that leaving the override unset is exactly right for natively-supported
parts -- phase 06 deliberately does so to avoid HSA_STATUS_ERROR_INVALID_ISA. But
empty-but-defined breaks enumeration just as hard, and far more quietly. The existing
comment warns only about the non-empty case.

Isolated in-container against the same binary, changing nothing else:

    both HSA vars empty (as shipped)  -> no ROCm devices
    HSA_XNACK empty only              -> 2 devices  (empty HSA_XNACK is harmless)
    HSA_OVERRIDE_GFX_VERSION empty    -> no ROCm devices
    both truly unset                  -> 2 devices, gfx1201, 65248 MiB

Fixed with Compose's pass-through forms, which omit the variable when it is not
defined: bare `- HSA_OVERRIDE_GFX_VERSION` in list form, null-valued
`HSA_OVERRIDE_GFX_VERSION:` in map form. Strix Halo hosts, which do define it, are
unaffected.

Also stops defaulting HSA_XNACK to 1 for all AMD hosts: it is a unified-memory APU
feature, so phase 06 now emits it only for integrated targets. With the old
`${HSA_XNACK:-1}` default, unsetting it in .env had no effect anyway.

Affects gfx1100, gfx1200, gfx1201, gfx942 -- everything that is not gfx1151.

Also updates the HSA_XNACK descriptions in .env.example and .env.schema.json,
which still called it "extended memory access for large KV caches" — it is a
unified/pageable-memory APU feature, matching the phase 06 gating above.

---


Lightheartdevs:
Thanks for closing the empty-HSA-override trap in the AMD compose path.

What looks good:
- `ods/docker-compose.amd.yml` and `ods/docker-compose.multigpu-amd.yml` now use bare pass-through for `HSA_OVERRIDE_GFX_VERSION`, which preserves the important unset-vs-empty distinction for ROCm.
- `ods/installers/phases/06-directories.sh` now only emits `HSA_XNACK=1` for integrated/APU-style gfx targets, while keeping discrete cards free of that env default.
- The schema/docs updates match the behavior change and make the APU-only intent clearer.

What I checked:
- `docker compose -f docker-compose.base.yml -f docker-compose.amd.yml config --quiet`
- `docker compose -f docker-compose.base.yml -f docker-compose.amd.yml -f docker-compose.multigpu-amd.yml config --quiet`

PR #1780, opened 2026-07-14
gpu: stop scheduling services onto the display iGPU next to discrete cards

timgilmour:
gpu: stop scheduling services onto the display iGPU next to discrete cards

On a desktop AMD box with a display iGPU (e.g. Raphael gfx1036, 2GB) alongside real
compute cards, ODS assigns whisper, comfyui AND embeddings to the iGPU. It has no
meaningful compute and shares system RAM. Two independent causes.

1. amd-topo.sh mislabels discrete GPUs as unified.

   amd_memory_type() classified on GTT size:

       if [[ $gtt_gb_int -ge 16 && $vram_gb_int -le 4 ]] || [[ $gtt_gb_int -ge 32 ]]

   GTT is *system* memory the GPU may map -- roughly half of host RAM -- so a discrete
   card on a large-RAM host reports a large GTT too. The `gtt >= 32` clause therefore
   brands EVERY discrete AMD GPU on a host with >= 64 GB of RAM as "unified". Observed
   on 2x Radeon AI PRO R9700 (31 GB VRAM each, GTT 61 GB, 123 GB host): both discrete
   cards labelled unified.

   That silently disables the "prefer discrete GPUs for the LLM" logic in
   assign_gpus.py (~L517), which can only fire when something is labelled discrete.

   Fixed by checking mem_info_vram_vendor first: a discrete card has dedicated VRAM
   from a real vendor and says so (samsung/hynix/micron/gddr6/hbm); an APU has no
   dedicated VRAM and reports nothing. The existing GTT heuristics are kept as the
   fallback for the no-vendor case, so Strix Halo -- including a large VRAM carve-out --
   still resolves to unified exactly as before.

2. assign_gpus.py treats the iGPU as assignable.

   The iGPU is a GPU by every structural test, so it lands in the pool. llama_server
   takes both discrete cards, `remaining == 1`, and assign_services() hands that one
   remaining GPU -- the display iGPU -- all three services.

   Fixed by dropping integrated GPUs from the assignable pool when discrete GPUs are
   present. Guarded on discrete GPUs actually existing, so a pure-APU host (Strix Halo,
   where the integrated GPU *is* the compute device) keeps every GPU and is unaffected.

Both causes are regression-tested:

  - Cause 1 (mislabel): five bats cases in tests/bats-tests/amd-topo.bats cover the
    vendor-first branch (real vendor + large GTT stays discrete), the unknown/N-A
    fallthrough, the no-vendor Strix Halo fallback, and mem_info_vram_vendor wiring
    through detect_amd_topo.
  - Cause 2 (assignability): a fixture for the 2-discrete + 1-display-iGPU topology
    and four scheduler tests. They fail against the current code (all three services
    land on the iGPU) and pass after. Full python suite: 89 passed.

Also updates docs/HARDWARE-GUIDE.md, which still listed AMD discrete GPUs under
"What NOT to Buy" with a blanket "ROCm support limited".

---

### DIFF
--- a/ods/docs/HARDWARE-GUIDE.md
+++ b/ods/docs/HARDWARE-GUIDE.md
@@ -191,7 +191,7 @@ Rule: 2x your model size minimum
 ## What NOT to Buy
 
 ❌ **GTX 16xx/10xx** — No FP16 tensor cores
-❌ **AMD discrete GPUs (RX 7900 etc.)** — ROCm support limited; AMD Strix Halo APUs are fully supported (see README)
+⚠️ **AMD discrete GPUs** — supported via the ROCm path and scheduled like any discrete card (validated end-to-end on Radeon AI PRO R9700 / RDNA4); consumer cards (RX 7900 etc.) are not yet benchmarked. AMD Strix Halo APUs are fully supported (see README)
 ❌ **Intel Arc** — Driver problems, limited support
 ❌ **Cloud GPUs (H100/A100)** — Can't buy, rental only
 ⚠️ **8GB cards** — Limited to smaller models (2B-7B) but functional with Tier 0
--- a/ods/installers/lib/amd-topo.sh
+++ b/ods/installers/lib/amd-topo.sh
@@ -47,11 +47,29 @@ _amd_sort_card_dirs() {
 amd_memory_type() {
     local vram_bytes="${1:-0}"
     local gtt_bytes="${2:-0}"
+    local vram_vendor="${3:-}"
     local gtt_gb_int=$(( gtt_bytes / 1073741824 ))
     local vram_gb_int=$(( vram_bytes / 1073741824 ))
 
-    # GTT is the reliable unified-memory signal for AMD APUs. VRAM alone is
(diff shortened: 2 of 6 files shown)


Lightheartdevs:
Thanks for taking on the mixed AMD iGPU/dGPU scheduling issue here.

What looks good:
- `ods/installers/lib/amd-topo.sh` now checks `mem_info_vram_vendor` before using GTT size as a unified-memory signal, which addresses the large-host-RAM false positive directly.
- `ods/scripts/assign_gpus.py` removes integrated GPUs from the scheduling pool only when discrete GPUs are present, preserving pure-APU hosts while avoiding accidental service placement on a display iGPU.
- The new fixture and tests cover both topology classification and downstream service assignment.

What I checked:
- `python -m pytest ods/tests/test-assign-gpus.py -q` passed: 89 tests.
- Static review of `amd-topo.sh`, `assign_gpus.py`, and the new AMD topology fixture.

Thanks again for making the scheduler behavior line up with the actual compute hardware.

PR #1781, opened 2026-07-14
dashboard-api: stop reporting a display iGPU's GTT as its GPU memory

timgilmour:
dashboard-api: stop reporting a display iGPU's GTT as its GPU memory

The dashboard GPU page showed a 2 GB display iGPU (Raphael, next to 2x
Radeon AI PRO R9700) as having ~111 GB of memory. Two stacked causes in
gpu.py, both instances of the same GTT-heuristic mistake that the
companion branch fix/amd-igpu-assigned-real-work fixes for scheduling in
amd-topo.sh / assign_gpus.py:

1. Unified/discrete was classified by ratio alone (gtt > vram * 4). GTT is
   *system* memory the GPU may map -- roughly half of host RAM -- so the
   ratio brands modest-VRAM discrete cards on large-RAM hosts as unified
   too (a 32 GB R9700 flips on a >=256 GB host, and the memory panel with
   it). A real mem_info_vram_vendor (samsung/hynix/...) now wins first;
   the ratio remains the fallback for APUs, which report no vendor.

2. Even correctly classified, an APU next to discrete cards should not
   present GTT as "its" memory: it is a display adapter there, and its GTT
   is just host RAM (113817 MB rendered as the iGPU's memory pool).
   get_gpu_info_amd_detailed() now presents an APU's GTT only when no
   discrete GPU exists -- a pure-APU host (Strix Halo) keeps the unified
   GTT view exactly as before.

Also makes _find_amd_gpu_sysfs() (the single-GPU summary behind
/api/preflight/gpu and /gpu) prefer a discrete card over an APU, so a
desktop where the display iGPU happens to sort first does not report
"111 GB Unified" as the machine's GPU.

Observed on 2x R9700 + Raphael iGPU (123 GB host): /api/gpu/detailed
reported index 2 as 63/113817 MB. With the fix, run against the same live
sysfs: 538/2048 MB -- matching `ods gpu status` on a host that also
carries fix/amd-igpu-assigned-real-work, which makes the topology that
command reads vendor-aware in the same way. Nothing schedules from
gpu.py's numbers, so this is a display-only bug.

Adds 10 tests: vendor-first classification, the mixed-topology regression
(iGPU shows its 2 GB carve-out, R9700s their VRAM), the Strix Halo
pure-APU view preserved, and discrete-preferred summary selection.
Full dashboard-api suite: 1192 passed.

---

### DIFF
--- a/ods/extensions/services/dashboard-api/gpu.py
+++ b/ods/extensions/services/dashboard-api/gpu.py
@@ -60,13 +60,44 @@ def _read_meminfo_mb() -> Optional[tuple[int, int]]:
 
 
 def _find_amd_gpu_sysfs() -> Optional[str]:
-    """Find the sysfs base path for an AMD GPU device."""
+    """Find the sysfs base path for an AMD GPU device.
+
+    Prefers a discrete card (one reporting a real VRAM vendor) so that on a
+    desktop where the display APU happens to sort first, the summary reflects
+    the compute GPU rather than the iGPU. Falls back to the first AMD card,
+    which on an APU-only host (Strix Halo) is the compute device.
+    """
     import glob
+    first_amd = None
     for card_dir in sorted(glob.glob("/sys/class/drm/card*/device")):
         vendor = _read_sysfs(f"{card_dir}/vendor")
-        if vendor == "0x1002":
+        if vendor != "0x1002":
+            continue
+        if first_amd is None:
+            first_amd = card_dir
+        vram_vendor = (_read_sysfs(f"{card_dir}/mem_info_vram_vendor") or "").strip().lower()
+        if vram_vendor and vram_vendor not in ("unknown", "n/a"):
             return card_dir
-    return None
+    return first_amd
+
+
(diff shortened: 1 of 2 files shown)


Lightheartdevs:
Thanks for fixing the dashboard-side view of AMD memory on mixed systems.

What looks good:
- `ods/extensions/services/dashboard-api/gpu.py` now prefers a discrete AMD card for the summary view when one exists, so the dashboard does not summarize the display iGPU by accident.
- The unified-memory classifier uses the same vendor-first idea as the installer path, which keeps discrete cards from reporting host GTT as GPU memory.
- Detailed GPU reporting preserves GTT-as-memory for APU-only hosts while showing the iGPU's VRAM carveout when discrete cards are present.

What I checked:
- `python -m pytest tests/test_gpu_amd.py -q` from `ods/extensions/services/dashboard-api` passed: 27 tests.
- Static review of the AMD sysfs selection and memory reporting paths.

Thanks again for making the dashboard match what operators actually need to see.

PR #1782, opened 2026-07-14
cli: make ods model swap actually reach LiteLLM

timgilmour:
cli: make `ods model swap` actually reach LiteLLM

In lemonade mode `ods model swap <tier>` silently does nothing. It writes LLM_MODEL /
GGUF_FILE / CTX_SIZE to .env and stops -- but every client talks to LiteLLM, and LiteLLM
pins the model BY NAME in config/litellm/lemonade.yaml:

    - model_name: default
      litellm_params:
        model: openai/extra.<OLD_GGUF>
    - model_name: "*"
      litellm_params:
        model: openai/extra.<OLD_GGUF>

That file names the model explicitly on both entries -- Lemonade exposes models as
"extra.<GGUF_FILE>" and 404s on anything else, so a wildcard passthrough is not an
option. It was written in exactly two places: phase 06 at install, and
bootstrap-upgrade.sh after a hot-swap. Nothing regenerates it on a manual swap.

So the user swaps the model, `ods restart llama-server`, and Open WebUI / Hermes /
OpenClaw / Perplexica all keep serving the OLD model -- Lemonade loads on demand by
requested name, so it happily serves whatever LiteLLM asks for. .env says one thing,
every client gets another, and nothing errors.

Observed: .env asked for Qwen3.6-35B-A3B while LiteLLM served the 2B bootstrap model on
a 64GB-VRAM box.

`ods model swap` now regenerates lemonade.yaml through the same renderer contract
bootstrap-upgrade.sh:1557 uses, and restarts ods-litellm. Non-lemonade modes are
unaffected (the helper returns early). If the renderer is missing or fails, it says so
loudly and explains that the swap will NOT take effect, rather than reporting success.

Verified: `ods model swap T0` on an install serving Qwen3.6-35B-A3B rewrote both entries
to extra.Qwen3.5-2B-Q4_K_M.gguf, restarted LiteLLM, and a completion through LiteLLM
came back from Qwen3.5-2B-Q4_K_M.gguf -- where previously it would have stayed on the 35B.

Unlike bootstrap-upgrade.sh, there is deliberately no inline fallback writer when
the renderer is unavailable: a second hand-rolled writer would drift from the
renderer's output, and failing loudly beats succeeding differently. Renderer and
docker-restart stderr are surfaced on failure, not swallowed.

Docs: adds ods model swap to INSTALLER-ARCHITECTURE.md's config-writer table
(lemonade.yaml row), documents the new behavior in MODEL-MANAGEMENT.md (BYO-GGUF
step 5 + the stale-model troubleshooting entry) and MODE-SWITCH.md.

---

### DIFF
--- a/ods/docs/INSTALLER-ARCHITECTURE.md
+++ b/ods/docs/INSTALLER-ARCHITECTURE.md
@@ -123,7 +123,7 @@ done. This is the most common way install-time surprises survive a patch.
 |----------------|--------------|--------------|----------------|------------------------|
 | `.env` and core ports/secrets | `phases/06-directories.sh` | `installers/macos/lib/env-generator.sh` | `installers/windows/lib/env-generator.ps1` | `ods config`, `ods update`, installer re-runs |
 | OpenCode config | `phases/07-devtools.sh` | `installers/macos/install-macos.sh` | `installers/windows/lib/opencode-config.ps1` | `scripts/update-windows-opencode-config.ps1`, `scripts/bootstrap-upgrade.sh` |
-| LiteLLM Lemonade config | `phases/06-directories.sh` | n/a | n/a | `scripts/bootstrap-upgrade.sh`, `bin/ods-host-agent.py` |
+| LiteLLM Lemonade config | `phases/06-directories.sh` | n/a | n/a | `scripts/bootstrap-upgrade.sh`, `bin/ods-host-agent.py`, `ods model swap` (`ods-cli`) |
 | Perplexica config | `phases/12-health.sh`, `phases/13-summary.sh` | `installers/macos/lib/env-generator.sh`, `installers/macos/install-macos.sh` | `installers/windows/lib/env-generator.ps1`, `installers/windows/install-windows.ps1` | `scripts/bootstrap-upgrade.sh`, `scripts/repair/repair-perplexica.sh` |
 | Hermes config | `phases/11-services.sh`, `scripts/patch-hermes-config.py` | `installers/macos/install-macos.sh` | `installers/windows/phases/06-directories.ps1` | `scripts/bootstrap-upgrade.sh`, `bin/ods-host-agent.py` |
 
--- a/ods/docs/MODE-SWITCH.md
+++ b/ods/docs/MODE-SWITCH.md
@@ -151,6 +151,10 @@ ods model list
 ods model swap T3
 ```
 
+In lemonade mode, `ods model swap` also regenerates `config/litellm/lemonade.yaml`
+(which pins the model by name) and restarts `ods-litellm`, so clients pick up the
+new model without manual config edits.
+
 For Dashboard downloads, loading catalog models, and manual GGUF swaps, see
 [MODEL-MANAGEMENT.md](MODEL-MANAGEMENT.md).
 
--- a/ods/docs/MODEL-MANAGEMENT.md
+++ b/ods/docs/MODEL-MANAGEMENT.md
@@ -178,6 +178,14 @@ waiting too long to compact.
 
 5. For AMD/Lemonade installs, verify `config/litellm/lemonade.yaml`.
 
+This file pins the model by name on both its `default` and `"*"` entries, and
+LiteLLM is what every client talks to — if it still names the old GGUF, the swap
+has not taken effect regardless of what `.env` says. Tier swaps via
+`ods model swap` regenerate it and restart `ods-litellm` automatically; after a
+manual `.env`/GGUF edit like this one, verify it yourself (re-run
+`scripts/render-runtime-configs.py --surface litellm-lemonade` rather than
+hand-editing).
+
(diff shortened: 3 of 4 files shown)


Lightheartdevs:
Thanks for wiring `ods model swap` through the Lemonade/LiteLLM config path instead of stopping at `.env`.

Decision point:
- The new `_model_refresh_litellm_lemonade` helper in `ods/ods-cli` matches the repo's generated-config-writer guidance and uses `scripts/render-runtime-configs.py` rather than adding another heredoc writer. Because this affects a live lifecycle command, I would gate merge on an AMD/Lemonade run of `ods model swap <tier>` that verifies `config/litellm/lemonade.yaml` changes to `extra.<GGUF_FILE>` and `ods-litellm` restarts cleanly.

What I checked:
- Static review of `ods/ods-cli` plus the docs updates in `ods/docs/MODE-SWITCH.md`, `ods/docs/MODEL-MANAGEMENT.md`, and `ods/docs/INSTALLER-ARCHITECTURE.md`.
- Confirmed `ods/lib/python-cmd.sh` and `scripts/render-runtime-configs.py` exist on main.
- I could not run shell syntax tests here because this Windows environment has no working POSIX shell.

Thanks again for closing a real generated-config gap.

timgilmour:
Ran the exact validation you gated on, on a live AMD/Lemonade box (2× Radeon AI PRO R9700, 2026-07-19) — and it caught something; details below.

The run: `ods model swap T0` on a box serving `extra.Qwen3.5-27B-Q4_K_M.gguf`:

- `.env` updated (LLM_MODEL/GGUF_FILE/TIER/CTX)
- `config/litellm/lemonade.yaml` regenerated via the renderer → `openai/extra.Qwen3.5-2B-Q4_K_M.gguf` on the `default`/`lemonade`/`"*"` routes
- `ods-litellm` restarted cleanly by the CLI; `ods restart llama-server` applied the engine side
- completions through LiteLLM on both `default` and `lemonade` returned the new model, then the box was swapped back and re-verified the same way (the restore render was byte-identical to the pre-swap file)

The catch: doing this on a box with #1786/#1787's hipfire engine enabled exposed a cross-PR interaction. This PR's `_model_refresh_litellm_lemonade` invokes the renderer without hipfire flags (correctly — they don't exist in this PR's world), but #1787's renderer only emitted hipfire routes when flagged, so a CLI swap re-render would have dropped the named `hipfire`/`lemonade` routes. Fixed on #1787 (commit 7466e894): the renderer now derives hipfire state from `--output-root/.env` when the flags are absent, so flag-less callers — this PR and any future one — can't wipe engine routes. Nothing changes in this PR; the verified run above was made with that renderer in place, and the hipfire route survived the swap in both directions.

timgilmour:
Closing as superseded: `1bdf9cbe` (fix(cli): route model swaps through host agent) solves this same bug with a better shape than this PR — the CLI now goes through the host agent's `/v1/model/activate`, so the dashboard and CLI share one activation path (env update, LiteLLM config regeneration, service restart, and downstream-route verification all in one place) instead of the CLI carrying its own renderer invocation. Nice work @gabsprogrammer.

Two small leftovers from this PR that may still be worth carrying somewhere: the doc updates to `docs/MODE-SWITCH.md`/`docs/INSTALLER-ARCHITECTURE.md` describing the generated-config flow, and the loud renderer-unavailable warnings ("LiteLLM is what every client talks to, so the swap will NOT take effect") — happy to fold either into another PR if wanted. The related hipfire-route protection this PR's review surfaced already landed on #1787 (renderer .env fallback).

My stacked branches (#1787/#1895/#1896) will adopt the host-agent-routed swap in their next rebase.

gabsprogrammer:
thank you ^^ 

> Fechando como substituído: `1bdf9cbe`(fix(cli): trocas de modelo de rota através do agente host) resolve este mesmo bug com uma solução melhor do que este PR — a CLI agora passa pelo agente host `/v1/model/activate`, então o painel e a CLI compartilham um único caminho de ativação (atualização de ambiente, regeneração da configuração do LiteLLM, reinicialização do serviço e verificação da rota downstream, tudo em um só lugar) em vez da CLI carregar sua própria invocação de renderizador. Ótimo trabalho!@gabsprogrammer.
> 
> Duas pequenas alterações remanescentes deste PR que ainda podem ser úteis: as atualizações da documentação que `docs/MODE-SWITCH.md`descrevem `docs/INSTALLER-ARCHITECTURE.md`o fluxo de configuração gerada e os avisos claros de indisponibilidade do renderizador ("O LiteLLM é o que todos os clientes usam para se comunicar, então a troca NÃO terá efeito") — posso incluir qualquer uma delas em outro PR, se necessário. A proteção relacionada à rota hipfire, mencionada na revisão deste PR, já foi implementada na issue #1787 (fallback para .env do renderizador).
> 
> Meus branches empilhados ( #1787 / #1895 / #1896 ) adotarão a troca host-agent-routed em seu próximo rebase.

PR #1783, opened 2026-07-14
llama-server(amd): make Lemonade actually use the custom ROCm binary

timgilmour:
llama-server(amd): make Lemonade actually use the custom ROCm binary

On any non-gfx1151 AMD host the custom ROCm llama-server that Dockerfile.amd spends
the whole install compiling was never used, and inference ran on Vulkan or failed to
start. Three separate reasons, all fixed here.

1. Phase 06 only set LEMONADE_LLAMACPP_ROCM_BIN on gfx1151, on the stated theory that
   other hosts fall back to "Lemonade's bundled ROCm-aware binary". No such binary
   exists -- the Lemonade image ships no ROCm userspace at all (no rocminfo, no HIP
   runtime). Left unset, Lemonade reports "no ROCm-capable device is detected", pulls
   the Vulkan llama.cpp build, and runs on radv. /opt/llama-custom is the only ROCm
   stack in the container, it is built for the detected target, and its gfx1151 MMQ
   patch is gated, so point Lemonade at it on every AMD target.

2. Lemonade reads its settings from a cached config.json and ignores env on every
   start after the first, so LEMONADE_LLAMACPP_ROCM_BIN never reached it: rocm_bin
   stayed "builtin". Lemonade then tries to DOWNLOAD a llama-server matching the gfx
   arch it detects -- and its detection maps GPU *marketing names* to gfx targets
   (lemon::identify_rocm_arch_from_name), from a table that knows "Radeon RX 9000
   series" but not "AMD Radeon AI PRO R9700". It resolved the iGPU and fetched

       llama-b1231-ubuntu-rocm-gfx1036-x64.zip   -> HTTP 404

   and the model failed to load. The wrapper entrypoint already existed to reconcile
   ctx_size for exactly this "cached config wins over env" reason; it now also syncs
   llamacpp.rocm_bin and llamacpp.backend, and seeds config.json from Lemonade's
   shipped defaults when absent so that even a first start on a fresh volume gets the
   right binary rather than attempting the bad download.

3. LEMONADE_LLAMACPP_BACKEND, set in docker-compose.amd.yml, is not a variable
   Lemonade reads at all -- confirmed against the daemon's own string table:

       grep -aoE 'LEMONADE_[A-Z_]+' /opt/lemonade/lemond

   It was a no-op, so the comment explaining how it kept the backend on a
   schema-valid "auto" described code that never ran. The selector Lemonade actually
   reads is LEMONADE_LLAMACPP (-> --llamacpp), now emitted as "rocm" by phase 06.
   Contrary to that comment, "rocm" is still a valid backend value in v10.2.0: the
   server accepts it and logs `"backend": "rocm"`.

The entrypoint also defensively unsets empty HSA_OVERRIDE_GFX_VERSION / HSA_XNACK.
Compose should already omit them (see the companion compose fix), but ROCm treats a
defined-but-empty HSA_OVERRIDE_GFX_VERSION as a malformed override and enumerates
zero devices, so a single stray "VAR=${VAR:-}" anywhere in the merge chain silently
kills every GPU. Cheap to make unreachable.

Verified on 2x Radeon AI PRO R9700 (gfx1201): 25/25 layers offloaded across ROCm0 and
ROCm1, ~262 tok/s. Reverting config.json to its broken state and restarting shows the
entrypoint repairing both keys.

---

cli(amd): stop `ods gpu reassign` from deleting LEMONADE_LLAMACPP_ROCM_BIN

Same gfx1151-only gate as phase 06, duplicated in ods-cli. Re-pinning GPUs should
not be able to kill inference, and today it does.

_gpu_reassign_update_amd_arch_env() sets HSA_OVERRIDE_GFX_VERSION and
LEMONADE_LLAMACPP_ROCM_BIN together, keyed on gfx1151:

    if [[ "$gfx_ver" == "gfx1151" ]]; then
        _env_set   "HSA_OVERRIDE_GFX_VERSION" "11.5.1"
        _env_set   "LEMONADE_LLAMACPP_ROCM_BIN" "/opt/llama-custom/llama-server"
    else
        _env_unset "HSA_OVERRIDE_GFX_VERSION"
        _env_unset "LEMONADE_LLAMACPP_ROCM_BIN"     # <-- silently kills inference
    fi

The HSA_OVERRIDE half is correct and stays. The LEMONADE_LLAMACPP_ROCM_BIN half is
not arch-conditional at all: /opt/llama-custom is the only ROCm stack in the Lemonade
image, it is built for the detected AMDGPU_TARGET, and the gfx1151 MMQ patch is gated
to gfx1151 -- so it is right for every target. Unsetting it sends Lemonade back to
rocm_bin="builtin", where it tries to DOWNLOAD a llama-server for the arch it guesses
from the GPU's marketing name and 404s on anything its table does not recognise.

So on any non-Strix-Halo AMD host, `ods gpu reassign` -- a command about *GPU
placement* -- silently deletes the setting that decides whether inference runs at all.
It does not even fail loudly: the value survives in Lemonade's cached config.json, so
everything looks fine until that volume is reset, at which point the stack dies with no
obvious cause.

Verified: a real `ods gpu reassign --auto` on 2x Radeon AI PRO R9700 now preserves
LEMONADE_LLAMACPP_ROCM_BIN, and inference keeps working.

---

test(contracts): reassign must preserve LEMONADE_LLAMACPP_ROCM_BIN, not delete it

The installer contract still encoded the behavior the previous commit removed:
it required a literal `_env_unset "LEMONADE_LLAMACPP_ROCM_BIN"` in ods-cli,
i.e. the exact line that silently killed inference on non-Strix AMD hosts.
Invert it: ods-cli must NEVER unset the custom binary on reassign, and must
still pin /opt/llama-custom/llama-server.

This failure was invisible until now: make test aborted earlier on an
unrelated schema-parity check (HIPFIRE_PORT, fixed on the hipfire branch),
so the reassign contract never ran on this branch's stack.

---

### DIFF
--- a/ods/.env.schema.json
+++ b/ods/.env.schema.json
@@ -994,8 +994,8 @@
     },
     "LEMONADE_LLAMACPP": {
       "type": "string",
-      "description": "Lemonade inner llama.cpp backend selector for AMD multi-GPU. Applied by docker-compose.multigpu-amd.yml. Lemonade's schema accepts only auto|vulkan|cpu; 'auto' plus LEMONADE_LLAMACPP_ROCM_BIN picks up the custom ROCm-built binary.",
-      "enum": ["auto", "vulkan", "cpu"],
+      "description": "Lemonade inner llama.cpp backend selector for AMD multi-GPU. Applied by docker-compose.multigpu-amd.yml. 'rocm' pins the ROCm llama.cpp build (phase 06 writes this on AMD installs — 'auto' would re-run Lemonade's marketing-name gfx detection, which misses cards like the Radeon AI PRO R9700 and silently falls back to Vulkan); combined with LEMONADE_LLAMACPP_ROCM_BIN it selects the custom ROCm-built binary.",
+      "enum": ["auto", "vulkan", "cpu", "rocm"],
       "default": "auto"
     },
     "LEMONADE_LLAMACPP_ROCM_BIN": {
--- a/ods/docker-compose.amd.yml
+++ b/ods/docker-compose.amd.yml
@@ -61,15 +61,18 @@ services:
       - HSA_OVERRIDE_GFX_VERSION=${HSA_OVERRIDE_GFX_VERSION:-}
       - HSA_XNACK=${HSA_XNACK:-1}
       - ROCBLAS_USE_HIPBLASLT=${ROCBLAS_USE_HIPBLASLT:-1}
-      # Lemonade's backend selector accepts auto|vulkan|cpu only; "rocm" is no
-      # longer a valid value. Setting auto + LEMONADE_LLAMACPP_ROCM_BIN points
-      # Lemonade at the custom ROCm-built llama-server binary while keeping the
-      # field schema-valid.
-      - LEMONADE_LLAMACPP_BACKEND=auto
-      # Custom binary at /opt/llama-custom is gfx1151-only (Strix Halo patches +
-      # AMDGPU_TARGET=gfx1151 at build time). Phase 06 sets this env var only
-      # when the detected gfx target is gfx1151; on any other AMD architecture
-      # it stays unset and Lemonade uses its bundled ROCm-aware binary.
+      # NOTE: the backend selector is LEMONADE_LLAMACPP (-> --llamacpp), set in
+      # docker-compose.multigpu-amd.yml. LEMONADE_LLAMACPP_BACKEND, which used to be set
+      # here, is NOT a variable Lemonade reads (its recognised names are visible with
+      # `grep -aoE 'LEMONADE_[A-Z_]+' /opt/lemonade/lemond`) — it was a no-op, so the
+      # comment claiming it kept us on "auto" was describing something that never ran.
+      # "rocm" IS still a valid backend value in v10.2.0, contrary to that comment;
+      # the server accepts it and logs `"backend": "rocm"`.
+      #
+      # The custom binary at /opt/llama-custom is built for ${AMDGPU_TARGET} and is the
+      # only ROCm stack in this image — the Lemonade base image ships no ROCm userspace.
+      # Left empty, Lemonade detects no ROCm device, falls back to its Vulkan llama.cpp
+      # build on radv, and may fail to load models outright.
       - LEMONADE_LLAMACPP_ROCM_BIN=${LEMONADE_LLAMACPP_ROCM_BIN:-}
       - LEMONADE_CTX_SIZE=${CTX_SIZE:-131072}
     healthcheck:
--- a/ods/docker-compose.amd.yml
+++ b/ods/docker-compose.amd.yml
@@ -30,7 +30,7 @@ services:
       - "0.0.0.0"
       - --no-tray
       - --llamacpp
-      - auto
+      - "${LEMONADE_LLAMACPP:-rocm}"
       - --llamacpp-args
       - "--metrics --host 0.0.0.0 -fa on -b 2048 -ub 256"
       - --extra-models-dir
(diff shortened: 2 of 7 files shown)


Lightheartdevs:
Thanks for chasing the Lemonade custom-binary path all the way through install, entrypoint, and GPU reassign.

Decision point:
- The change is internally consistent: `lemonade-entrypoint.sh` reconciles cached `config.json`, phase 06 always pins `/opt/llama-custom/llama-server`, and `ods gpu reassign` no longer removes `LEMONADE_LLAMACPP_ROCM_BIN`. This is still high-risk AMD runtime behavior, so I would want real AMD validation that the image starts on the target card, Lemonade reports the ROCm backend, and a model completion works after both fresh install and `ods gpu reassign`.

What I checked:
- Static review of `ods/docker-compose.amd.yml`, `lemonade-entrypoint.sh`, phase 06, `ods-cli`, and the installer contract update.
- `docker compose -f docker-compose.base.yml -f docker-compose.amd.yml config --quiet` passed.

Thanks again for making this path explicit instead of relying on Lemonade's cached defaults.

timgilmour:
Real AMD validation from a 2× Radeon AI PRO R9700 (gfx1201) box, 2026-07-19:

- **Fresh install:** this install was brought up with the fixed phase-06 (it pins `/opt/llama-custom/llama-server`), so the install path is exercised on metal.
- **`ods gpu reassign`, actually applied** — not just the contract test: full run through the "apply now" path, i.e. `docker compose down --remove-orphans && up -d` across all 26 containers. After the recreate: `LEMONADE_LLAMACPP_ROCM_BIN=/opt/llama-custom/llama-server` survived in both `.env` and the recreated container's environment alongside `LEMONADE_LLAMACPP=rocm`; Lemonade loaded the custom ROCm binary and a completion through LiteLLM returned from the R9700s. (The reassign also corrected a stale `LLAMA_SERVER_GPU_INDICES` value left by earlier manual configuration — the algorithm reproduced the intended assignment exactly.)
- The regression this PR fixes (reassign deleting the pin) stays locked by the updated installer contract, which asserts the var is never unset.

Lightheartdevs:
I reproduced a single-GPU gap in this PR on Strix Halo and validated a focused fix on metal.

The installer correctly wrote:

```text
LEMONADE_LLAMACPP=rocm
LEMONADE_LLAMACPP_ROCM_BIN=/opt/llama-custom/llama-server
```

but the base `docker-compose.amd.yml` still launched Lemonade with a literal `--llamacpp auto` and did not pass `LEMONADE_LLAMACPP` into the container. The multi-GPU overlay overrides both, which explains why the 2× R9700 validation passed while a single-GPU host silently kept Lemonade's bundled Vulkan llama.cpp.

On Strix Halo, that path loaded the bundled runtime and deterministically rejected a Cohere2-MoE GGUF as an unknown `cohere2moe` architecture. Pointing only `LEMONADE_LLAMACPP_ROCM_BIN` at the custom binary was insufficient because the live process still selected the bundled Vulkan runtime.

I tested this focused base-overlay change:

```diff
-      - auto
+      - "${LEMONADE_LLAMACPP:-rocm}"
...
+      - LEMONADE_LLAMACPP=${LEMONADE_LLAMACPP:-rocm}
       - LEMONADE_LLAMACPP_ROCM_BIN=${LEMONADE_LLAMACPP_ROCM_BIN:-}
```

and added contract assertions for both lines. Results:

- full installer contract: PASS
- AMD/Lemonade contract suite: 65/65 PASS
- exact Strix Halo install at remediation SHA `b1a9af7a5d4ee8ebcdf4bf1dd0565af9c3ccfac6`: PASS
- live container command: `--llamacpp rocm`
- live environment named the custom ROCm binary
- custom binary version: llama.cpp b9641
- exact 30B/3B-active Cohere2-MoE rotation: download/SHA, 131K load, ODS Talk, signed LiteLLM/Open WebUI/OpenCode/Perplexica routes, baseline restore, artifact deletion, and environment cleanup all PASS
- final Qwen 3.5 2B route restored with exact runtime identity and completion proof

Fleet evidence:

`fleet-test/runs/2026-07-28T03-36-36Z-model-ui-product-b1a9af7a5d4e-harness-19d43e6f9f25-hosts-strix-halo-north-mini-code-strix-halo-remediation-r3`

The model-specific b9641 build bump belongs with the model/runtime-floor work, but selecting and exporting `LEMONADE_LLAMACPP` in the base AMD overlay appears necessary for this PR's custom-binary promise to hold on single-GPU systems too.

timgilmour:
Adopted — thank you for running this down on metal. The multi-GPU masking explains our blind spot exactly: the 2× R9700 validation host always composes `docker-compose.multigpu-amd.yml`, which overrides both the flag and the env, so the base overlay's literal `auto` never actually executed there.

The branch now carries your focused change at `a55de214`:

- `docker-compose.amd.yml` passes `--llamacpp "${LEMONADE_LLAMACPP:-rocm}"` and exports `LEMONADE_LLAMACPP=${LEMONADE_LLAMACPP:-rocm}` into the container — rocm default matching what phase 06 writes, mirroring the multi-GPU overlay's pattern.
- Contract assertions cover both lines (written red-first: 61 passed + 2 failed before the compose change, 63/63 after).

Verification on this head:

- rendered config, base+amd, no `.env` value: command shows `--llamacpp rocm`, container env carries `LEMONADE_LLAMACPP: rocm`
- rendered config with the multi-GPU overlay stacked: unchanged behavior (its own default still wins there; real installs resolve from `.env`)
- rendered config with explicit `LEMONADE_LLAMACPP=vulkan`: propagates to both flag and env
- full `make lint` + `make test` pass; AMD/Lemonade + installer contract suites green; ruff (CI flags) clean, dashboard-api pytest 1182/1182, vitest 124/124, shellcheck clean at CI severity

Agreed on scope: the b9641 runtime-floor bump stays with the model/runtime-floor work.

PR #1784, opened 2026-07-14
host-agent: rebind when ods-network's gateway moves; stop hiding agent failures

timgilmour:
host-agent: rebind when ods-network's gateway moves; stop hiding agent failures

Model downloads from the dashboard fail with HTTP 503 "Host agent unreachable:
[Errno 111] Connection refused" after any `ods stop && ods start`, and stay broken
until someone restarts the host agent by hand.

The agent binds ods-network's gateway (_resolve_agent_bind_addr). `ods stop` runs
`docker compose down`, which DESTROYS the compose networks; `ods start` recreates them
and Docker may allocate a different subnet. Observed: ods-network moved from
172.19.0.0/16 to 172.18.0.0/16, and 172.19.0.1 was then handed to a *different* compose
network. The agent -- a long-lived systemd service that nothing restarts -- kept
listening on 172.19.0.1, where no container on ods-network can reach it.

Note the asymmetry: the client end already re-resolves on every container start
(dashboard-api reads its own default gateway from /proc/net/route), so only the agent
end went stale. The existing comment in main() acknowledges the agent must be restarted
"after ods-network exists" for first install, but nothing handles the network being
RECREATED later.

Fixed by having the agent heal itself: a watchdog re-resolves the bind address every 30s
and, if it has moved, stops the server so the (new) serve loop rebinds to the current
gateway. A _SHUTTING_DOWN event distinguishes a rebind from a real SIGTERM, so the loop
does not spin back up after a shutdown, and a failing resolve is swallowed rather than
killing the agent. No sudo, no compose-network changes, no new failure modes.

Also stops `ods agent` from lying, which is what hid this:

  - `status` probed 127.0.0.1, but on Linux the agent binds ods-network's gateway, so it
    reported "not responding" for a perfectly healthy agent. It now probes the real bind
    address, and if it finds a listener on a *different* address it says so explicitly --
    that is precisely the stale-bind symptom -- and points at `ods agent restart`.
  - `start` and `stop` ran `sudo systemctl ... 2>/dev/null`. With no TTY, sudo cannot
    prompt for a password and fails, and swallowing stderr turned that into a bare
    "Agent start failed" / "Agent not running" with no way to tell why. They now surface
    the real error.

Verified: the watchdog is unit-tested in tests/test-host-agent-rebind.py — stable
gateway (no rebind), moved gateway (rebind requested without setting the shutdown
flag), resolver errors (agent survives), empty resolve (not a move), and SIGTERM
winning over the rebind loop; 5 passed. `ods agent status` now reports
"running (172.18.0.1:7710, systemd)" where it previously claimed "not responding"; and a
dashboard model download returns 200 and streams bytes.

---

test: SIGTERM contract follows the rebind-aware shutdown wrapper

The previous commit registers _request_shutdown (set _SHUTTING_DOWN, then
call the async-safe helper) as the SIGTERM handler, but the stop contract
still grepped for a direct signal.SIGTERM -> _request_server_shutdown
registration and failed. Assert the actual chain instead: SIGTERM must
register the wrapper, and the wrapper must route through
_request_server_shutdown.

Masked until now: make test aborted on earlier contract failures
(HIPFIRE_PORT schema parity, reassign custom-binary), so this check never
ran to failure on this stack.

---

cli: use sed -n '1p' in the stale-bind listener probe

The `ods agent status` stale-bind diagnostic filtered ss output through
a bare `| head -1`, which the ods-cli flags contract bans: under
pipefail, head exiting after its first line can SIGPIPE the producer.
Use the sed -n '1p' idiom the rest of ods-cli uses.

Caught by tests/bats-tests/test-ods-cli-flags.bats ("no bare '| head -1'
pipelines in ods-cli") — the BATS gate had not been run on this stack
before today.

---


Lightheartdevs:
Thanks for fixing the host-agent stale-bind failure and the CLI diagnostics that made it hard to see.

What looks good:
- `ods/bin/ods-host-agent.py` keeps the bind-address recheck inside the agent, which matches the current architecture where containers re-resolve their side but the long-lived host service did not.
- The SIGTERM wrapper keeps real shutdown distinct from a rebind-triggered `serve_forever()` exit, and the new unit tests cover stable, moved, empty, failing, and SIGTERM paths.
- `ods agent status/start/stop` now probes the actual bind address and surfaces systemd/sudo errors instead of hiding the useful failure text.

What I checked:
- `python -m py_compile ods/bin/ods-host-agent.py` passed.
- `python -m pytest ods/tests/test-host-agent-rebind.py -q` passed: 5 tests.
- Static review of `ods/ods-cli` and `ods/tests/test-linux-host-agent-stop.sh`.

Thanks again for pushing this lifecycle fix forward.

Lightheartdevs:
Thanks for updating the host-agent stale-bind fix after main moved under it.

Decision point:
- The original fix still fits the architecture: the long-lived host agent owns rebinding to the recreated `ods-network` gateway, while dashboard-api already re-resolves its client side. The current head also appears to have adopted main's newer `_create_host_agent_server` and `_agent_probe_host` direction instead of preserving older inline probe logic.
- I would want a maintainer to require a visible green check run or a fresh focused local receipt before final merge. `gh pr checks 1784` currently reports no checks, and this path touches `ods/bin/ods-host-agent.py` plus `ods agent status/start/stop`, which are high-risk lifecycle/host-mutation surfaces.

What I checked:
- Reviewed the current PR head, prior advisory/discussion, current file list, and diff against `main`.
- Rechecked `gh pr checks 1784`: no checks reported for this branch.
- Static review focused on `ods/bin/ods-host-agent.py`, `ods/ods-cli`, and `ods/tests/test-host-agent-rebind.py` after the main-merge commits.

Thanks again for keeping this rebased; the remaining call is mostly validation visibility on a sensitive lifecycle path.

PR #1785, opened 2026-07-14
comfyui(amd): stop spoofing RDNA4 cards as gfx1151 -- wrong-arch kernels fault the GPU

timgilmour:
comfyui(amd): stop spoofing RDNA4 cards as gfx1151 — wrong-arch kernels fault the GPU

HSA_OVERRIDE_GFX_VERSION=11.5.1 told ROCm to report gfx1201 (RDNA4) cards as
gfx1151 (Strix Halo, RDNA3.5), so PyTorch dispatched gfx1151 kernels onto RDNA4
silicon. That does not degrade to slow — it faults the card:

    HW Exception by GPU node-1 ... reason: GPU Hang

The image name (comfyui-gfx1151) is a red herring: the torch wheel it ships is a
stock ROCm 7.2 build whose arch list already includes gfx1201, so RDNA4 needs no
spoofing at all. Delete the override rather than blanking it — ROCm checks whether
the variable is defined, not whether it is non-empty, and fails parsing an empty
string (same trap as fix/amd-hsa-empty-env-kills-rocm).

Verified on 2x Radeon AI PRO R9700: boots native gfx1201, fp16 matmul correct,
all three SDPA backends work, SDXL generates.

---

comfyui(amd): cut first-generation time from 40+ min to seconds on fresh arches

On GPUs without a shipped tuning database (e.g. gfx1201), first use was unusable:

- PYTORCH_TUNABLEOP_ENABLED=1 exhaustively benchmarks every GEMM shape before the
  first sampling step completes: ~8 minutes at full board power.
- MIOpen's default find mode then compile-and-races conv kernels for the SDXL VAE
  decode: 20+ minutes more, mostly single-threaded.
- Neither cache left the container layer (TunableOp writes its CSV to the process
  cwd; MIOpen's user DB lives in /root/.config/miopen), so every recreate re-paid
  the whole cost. The user experience is a GPU pinned at its power cap for half an
  hour with no image — indistinguishable from a hang.

Drop TunableOp (hipBLASLt's built-in heuristics are close enough; re-enabling it
properly means persisting PYTORCH_TUNABLEOP_FILENAME, documented in a comment),
select conv kernels heuristically with MIOPEN_FIND_MODE=FAST, and bind-mount the
MIOpen cache so residual JIT survives recreates.

Measured on a Radeon AI PRO R9700 (gfx1201), SDXL Lightning 4-step 1024x1024:
cold first generation 10 s end-to-end, steady state 2.1 s/image (previously the
VAE decode alone had not finished after 30+ minutes).

---

comfyui(amd): update docs stale-quoted by the gfx-spoof removal

EXTENSIONS.md quoted the pre-fix compose.amd.yaml verbatim — including the
HSA_OVERRIDE_GFX_VERSION=11.5.1 line this branch removes — and its AMD notes
recommended setting that exact spoof. Anyone following the docs would
reintroduce the bug the branch fixes: wrong-arch kernels do not run slowly,
they fault the card ("HW Exception ... GPU Hang"), and defining the variable
as an empty string makes ROCm enumerate zero GPUs because it checks whether
the var is defined, not non-empty.

The service README also claimed compose.amd.yaml "targets gfx1151 (RX 9000
series)" — wrong on both counts (gfx1151 is Strix Halo, not RX 9000, and with
the spoof gone the service runs each card's native arch; the bundled ROCm 7.2
torch ships kernels for gfx1151 and gfx1201 among others) — and its AMD volume
table omitted the new persistent MIOpen cache mount.

---


Lightheartdevs:
Thanks for cleaning up the ComfyUI AMD ROCm settings and the stale docs around them.

What looks good:
- `ods/extensions/services/comfyui/compose.amd.yaml` removes the forced `HSA_OVERRIDE_GFX_VERSION=11.5.1` instead of blanking it, which preserves native RDNA4 behavior and avoids the empty-env ROCm failure mode.
- Dropping `PYTORCH_TUNABLEOP_ENABLED` and persisting `/root/.config/miopen` are narrow changes aimed at first-generation latency without changing the service contract.
- The README and extension docs no longer recommend reintroducing the gfx spoof this PR removes.

What I checked:
- `docker compose -f docker-compose.base.yml -f docker-compose.amd.yml -f extensions/services/comfyui/compose.yaml -f extensions/services/comfyui/compose.amd.yaml config --quiet` passed.
- Static review of `compose.amd.yaml`, `compose.yaml`, `README.md`, and `docs/EXTENSIONS.md`.

Thanks again for making the RDNA4 behavior explicit.

gabsprogrammer:
Please update this branch to the latest version so it can be merged.

Lightheartdevs:
Rebased and repaired at 4695128b. The branch now keeps native gfx1151/gfx1201 execution, removes exhaustive TunableOp startup tuning, uses bounded MIOpen FAST selection, persists the MIOpen cache, and pre-creates it with safe owner/traversal permissions. A new rendered Compose contract proves both architectures plus NVIDIA/base isolation and recreation persistence; it passes locally. The original dual-R9700 SDXL receipt remains relevant. I will merge only after the refreshed full CI is green and the branch remains exactly this SHA.

Lightheartdevs:
Superseded by #2076, which preserves the original commits/evidence, updates onto current main, adds executable gfx1151/gfx1201 and cache-recreation contracts, and passed the full protected CI matrix before merge. Thank you for the root-cause analysis and real R9700 validation.

PR #1786, opened 2026-07-14
feat(hipfire): add RDNA-native inference engine as an optional AMD extension

timgilmour:
feat(hipfire): add RDNA-native inference engine as an optional AMD extension

Adds `extensions/services/hipfire/` — an AMD-only, ROCm-direct inference engine
that runs ALONGSIDE llama-server/Lemonade rather than replacing it, per the
side-by-side pattern docs/VLLM-SETUP.md:42 already prescribes for vLLM. Every
`depends_on: llama-server` and the ODS Talk vision path are untouched.

Measured on 2x Radeon AI PRO R9700 (gfx1201, ROCm 7.2.4), same weights
(Qwen3.6-35B-A3B), same prompt, same harness, 256 tokens:

  llama.cpp / Lemonade  Q4_K_M   2 GPUs    64.7 tok/s
  hipfire               MQ4      1 GPU    104.3 tok/s   (1.61x, half the hardware)

Ships as Pattern 2 (GPU-only): `services: {}` stub + compose.amd.yaml, gated by
ENABLE_HIPFIRE (default false) through _sync_extension_compose, so it can be
turned off. Passes validate-manifests.sh and audit-extensions.py --strict.

Three things worth knowing, all learned the hard way:

- HSA_OVERRIDE_GFX_VERSION is passed through BARE (no "="). The `${VAR:-}` form
  defines it as an empty string, which ROCm treats as a malformed override and
  answers by enumerating zero devices.
- The Dockerfile gates on a kernel COUNT, not on compile-kernels.sh's exit code:
  69 of 528 gfx1201 kernels (all attention_dflash_wmma_* spec-decode variants)
  fail to compile upstream, reproducibly on the bare host toolchain too, and the
  engine serves fine without them.
- HIPFIRE_REF pins a master commit, deliberately NOT the v0.2.1 tag: v0.2.1's
  quantizer silently F16-dumps MoE expert tensors, turning a 22 GB GGUF into an
  unloadable 66 GB model, and the gfx1201 kernel work landed after the tag.

hipfire's model is NOT interchangeable with ODS's GGUFs (MQ4 format, and its
GGUF->MQ4 path cannot handle MoE experts at all), so the weights must be built
from safetensors out of band. See the extension README.

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>

---

hipfire: declare its .env vars in .env.schema.json

[...]

PR #1787, opened 2026-07-14
fix(amd): make hipfire's model dashboard-controlled and clobber-proof

timgilmour:
> **Stacked on #10** (`feat/hipfire-rdna-inference-engine`). This branch is built on
> top of that one, so this diff currently includes #10's two commits as well as this
> branch's own three. Please merge #10 first — once it lands, this PR will shrink to
> just its own changes after a rebase. The only hard ordering constraint in this
> delivery: everything else here is independently mergeable in any order.

---

fix(amd): make hipfire's model dashboard-controlled and clobber-proof

The dashboard's model-activate path (POST /api/models/{id}/load ->
host-agent /v1/model/activate) regenerated config/litellm/lemonade.yaml
from a stock template that knows nothing about hipfire: one dashboard
Load silently wiped the hipfire routing and sent all text back to
llama-server/Lemonade while hipfire idled on GPU0 holding ~30 GB.
hipfire's own model was a static HIPFIRE_MODEL pin in .env with no
control surface at all ("hipfire" appeared 0 times in ods-host-agent.py).

Renderer: render_litellm_lemonade gains --hipfire-{enabled,active,model,
api-base}. When enabled it always emits named routes for both engines
(hipfire / lemonade) and points default + "*" at whichever engine is
active — so a re-render can never orphan either engine. Output is
byte-identical to before when the flags are absent (verified against
the old renderer), and the host-agent fallback heredoc mirrors the
same semantics.

Host-agent: model-library entries may carry engine: "hipfire" with a
model_file instead of a GGUF. Activating one pins HIPFIRE_MODEL +
HIPFIRE_ACTIVE=true in .env, recreates the container via compose
stop/up (docker restart would keep the stale env baked at create
time — the entrypoint re-reads HIPFIRE_MODEL and pulls weights if
missing), health-gates on /health returning 200 (503 until the model
is resident; cold MQ4 loads take minutes), then restarts ods-litellm.
Failure rolls back .env + lemonade.yaml and recreates hipfire on the
old pin. Activating a GGUF sets HIPFIRE_ACTIVE=false: an explicit,
visible engine switch instead of a silent clobber, with hipfire still
reachable via its named route. Containerized agents get a loud 501
(recreating hipfire needs host-resolved bind mounts).

Dashboard: normalize_catalog_entry accepts engine entries (model_file
required, no GGUF), and build_models_payload hides them entirely when
ENABLE_HIPFIRE is off, lists them as "downloaded" (= loadable — the
engine pulls its own weights, so "available"-to-download would be a
lie) when on, and "loaded" when hipfire owns the default route.
Download/delete endpoints 400 on engine entries instead of KeyError.
Zero UI changes needed: status "downloaded" already renders a Load
button. Ships one catalog entry: qwen3.6-35b-a3b-hipfire-mq4 (the
validated gfx1201 model, 104.3 tok/s measured).

Also adds the four HIPFIRE env keys to .env.schema.json —
tests/contracts/test-port-contracts.sh already requires HIPFIRE_PORT
there (parity with ports.json, which the hipfire feature branch
predates), so make test was failing before this commit.

Tests: renderer byte-identity + routing matrix (10 pass), host-agent
fallback + env-key writer, oracle catalog gating (dashboard-api suite:
1191 pass), make lint + make test clean.

---

hipfire: close the two remaining re-render doors that wipe its routing

The dashboard-control fix made host-agent re-renders hipfire-safe, but
two other callers of render-runtime-configs.py --surface litellm-lemonade
still knew nothing about hipfire:

- scripts/bootstrap-upgrade.sh (bootstrap-model hot-swap)
- installers/phases/06-directories.sh (install / re-install)

Either one regenerated lemonade.yaml without hipfire routes — the exact
clobber the dashboard-control commit closes, through a different door.
Both now read ENABLE_HIPFIRE / HIPFIRE_MODEL / HIPFIRE_ACTIVE and pass
the renderer flags, and both inline fallback writers gained the
hipfire-aware variant (named routes for both engines, default follows
HIPFIRE_ACTIVE), mirroring the host-agent fallback.

The installer path had a second layer: phase 06 regenerates .env and
dropped all three keys entirely (nothing in the installer writes them —
they are dashboard/host-agent state). They are now preserved on
re-install with the same rule as LANGFUSE_ENABLED: existing .env wins,
CLI flag is the fresh-install default. HIPFIRE_MODEL/HIPFIRE_ACTIVE are
only written when a model pin exists, so a fresh install's .env is
unchanged.

---

hipfire: make failed activations restore container state, not just files

Two gaps in _do_hipfire_activate's failure handling, both found in
pre-push review:

- The generic exception path restored .env and lemonade.yaml but left
  the container wherever the failure caught it — possibly down, possibly
  already recreated on the NEW pin while the restored files said the old
  one. It now recreates hipfire on the restored pin and gives LiteLLM a
  best-effort restart (a half-applied `docker restart` from the commit
  path could otherwise leave it stopped), and reports rolled_back
  honestly.

- The commit-path `docker restart ods-litellm` return code was ignored:
  a failed restart still answered {"status": "activated"} while every
  client kept talking to the old routing — the exact silent-failure
  class this control surface exists to eliminate. It now raises into the
  rollback path.

Adds TestHipfireActivateRollback: health-gate failure restores files and
recreates on the old pin without touching LiteLLM; a compose failure
still restores container state; a LiteLLM restart failure is a 500 with
rollback, not a claimed activation. test_model_activate.py: 52 passed.

---


Lightheartdevs:
Thanks for carrying the hipfire integration through the dashboard and generated-runtime-config paths.

Decision point:
- The incremental diff on top of `#1786` looks well covered: host-agent activation rolls back files and container state, `render-runtime-configs.py` keeps named `hipfire` and `lemonade` routes, and dashboard-api treats hipfire entries as loadable engine models rather than GGUF downloads. Since this PR is stacked on `#1786` and includes that base extension in its full diff, I would treat it as a sequencing decision: land/rebase the base extension first, then review this smaller control-surface diff as the merge candidate.
- One small fallback-policy callout: if maintainers care about renderer-unavailable parity, the fallback branch in `ods/bin/ods-host-agent.py::_write_lemonade_config` should keep the stock `default` route as well as `"*"`, or add a test showing wildcard-only is intentional.

What I checked:
- Reviewed `origin/pr/1786..origin/pr/1787` separately from the stacked full diff.
- `python ods/tests/test-render-runtime-configs.py` passed all renderer tests, including hipfire route cases.
- `python -m pytest tests/test_model_activate.py tests/test_performance_oracle.py -q` from `ods/extensions/services/dashboard-api` passed: 69 tests.

Thanks again for tightening the clobber-proofing around this path.

timgilmour:
On sequencing: agreed — #1786 lands first, then this rebases to its own commits and the diff shrinks to the control-surface change.

On the fallback-parity callout: good catch, with one clarification. The wildcard-only shape in the hipfire-disabled arm is inherited byte-for-byte from main's current `_write_lemonade_config` fallback — this PR didn't change that arm, it only added the hipfire-enabled arm (which does write the named `default` route, matching the renderer). The remaining renderer/fallback delta is exactly the named `default` route in the disabled case, and it's behavior-neutral: LiteLLM resolves a request for `default` through `"*"` to the identical litellm_params. Happy to either append a commit adding the named `default` route to the fallback arm for full parity, or add a test pinning wildcard-only as intentional — your call.

Also: this branch gained one commit since your review (7466e894). Running #1782's requested live-swap validation on a hipfire-enabled box showed that any renderer caller which never learned the hipfire flags would silently wipe the hipfire routes from lemonade.yaml — `ods model swap` being the concrete case. The renderer now falls back to reading `ENABLE_HIPFIRE`/`HIPFIRE_MODEL`/`HIPFIRE_ACTIVE` from `--output-root/.env` when `--hipfire-enabled` is absent; explicit flags stay authoritative, and every flag-passing caller derives its flags from that same `.env`, so the two paths agree by construction. Four new renderer tests pin the fallback (env-driven active/inactive routing, env absent, flags beat a disagreeing env). Full branch gate re-run green standalone; live-verified on the RDNA4 box with a real swap in both directions.

Lightheartdevs:
Thanks for following through on the remaining hipfire route-clobber path; moving the flagless caller protection into the renderer is the right place to catch `ods model swap` and future re-render callers.

Quickest path to merge:
- Rebase/retarget this after the stacked hipfire base lands, then resolve the current dirty merge state against `main`. The useful review unit is the control-surface layer in `ods/bin/ods-host-agent.py`, `ods/scripts/render-runtime-configs.py`, `ods/installers/phases/06-directories.sh`, and `ods/scripts/bootstrap-upgrade.sh`; right now the PR still carries the base hipfire extension commits too.
- Get GitHub checks running on the rebased head. `gh pr checks 1787` currently reports no checks, and this touches high-risk runtime routing, host-agent model activation, installer re-renders, and bootstrap hot-swap config generation.
- Keep the new `ods/scripts/render-runtime-configs.py` `.env` fallback and the coverage in `ods/tests/test-render-runtime-configs.py`; that incremental commit addresses the previous wildcard/fallback concern by preserving hipfire routes when explicit flags are absent and `--output-root/.env` has `ENABLE_HIPFIRE=true`.

What I checked:
- Reviewed the PR body, discussion, current head, stacked commit list, and the incremental diff from `8ecb20d` to `7466e894`.
- Ran `python ods/tests/test-render-runtime-configs.py` on the PR worktree: 14 renderer tests passed, including the new env-fallback and explicit-flags-win cases.
- Ran `python -m pytest tests/test_model_activate.py tests/test_performance_oracle.py -q` from `ods/extensions/services/dashboard-api`: 69 passed.
- Ran `git diff --check origin/main...HEAD`: clean.
- Confirmed PR-head renderer callers pass `--output-root "$INSTALL_DIR"`, so the new fallback reads the install tree `.env` the author intended.

Thanks again for tightening this path; after the stack is flattened and CI is visible, this should be much easier to make a final merge call on.

Lightheartdevs:
Thanks for rebasing the hipfire model-control work across the latest model-activation changes.

Quickest path to merge:
- Keep the previous sequencing requirement: land/rebase the stacked hipfire base first, then shrink this PR to the control-surface layer. The current diff still includes the base `ods/extensions/services/hipfire/*` extension plus the dashboard/host-agent model-control work.
- Get checks running on the rebased head. `gh pr checks 1787` currently reports no checks, and this touches high-risk model routing, LiteLLM/Lemonade config generation, host-agent activation rollback, installer re-rendering, and bootstrap hot-swap behavior.
- Preserve the latest conflict-resolution intent in `ods/bin/ods-host-agent.py` and `ods/scripts/render-runtime-configs.py`: main's atomic lemonade config writes and `LEMONADE_MODEL` threading should remain alongside this branch's hipfire-aware fallback/routes.

What I checked:
- Reviewed the current head, prior advisory/discussion, stacked commit list, current changed files, and the post-`7466e894` main-merge notes.
- Rechecked `gh pr checks 1787`: no checks reported for this branch.
- Static review focused on the merged host-agent/render-runtime-config/bootstrap-upgrade surfaces and whether the previous route-clobber concern stayed addressed.

Thanks again for tightening the routing path; flattening the stack and getting validation visible are the pieces that would make this ready to judge cleanly.

Lightheartdevs:
Thanks for keeping the hipfire model-control branch moving with the latest config-lock and network-policy fixes from the stack.

Quickest path to merge:
- Resolve the current upstream merge state and flatten after the base hipfire extension lands. GitHub reports this head as `CONFLICTING` / `DIRTY`, and the PR still includes the base `ods/extensions/services/hipfire/*` extension plus the dashboard/host-agent control-surface layer.
- Get checks visible on the upstream PR head after the rebase. `gh pr checks 1787 --repo Light-Heart-Labs/DreamServer` still reports no checks, while this touches high-risk model routing, LiteLLM/Lemonade config generation, host-agent activation rollback, installer re-rendering, and bootstrap hot-swap behavior.
- Preserve the prior conflict-resolution intent while rebasing: main's atomic lemonade config writes and current model-activation flow should remain alongside this branch's hipfire-aware renderer fallback/routes and `HIPFIRE_ACTIVE` switching.

What I checked:
- Reviewed the current head `f8f1e956`, PR body, prior advisory thread, commit list, changed-file list, and mergeability state.
- Confirmed the upstream PR still has no reported checks and is currently conflicting/dirty against `main`.
- `git diff --check origin/main...refs/remotes/pr/1787` is clean.

Thanks again for tightening the hipfire routing path; flattening the stack and making validation visible are still the pieces needed before a clean merge call.

timgilmour:
Advisory follow-up: head advanced append-only to `204e8f2d`, merging current `main` (c9e4e4d1, 2026-07-27) — GitHub now reports this PR **mergeable** again. The resolution intent you called out is preserved: main's atomic lemonade config writes and current model-activation flow are intact alongside this branch's hipfire-aware renderer fallback/routes and `HIPFIRE_ACTIVE` switching (the renderer now also carries main's remote-llm plumbing next to the hipfire logic).

Fork CI receipts for this exact head (upstream doesn't run fork-PR workflows) — **all workflows green**, including Test Linux:
- Test Linux: https://github.com/timgilmour/ODS/actions/runs/30315463615
- All runs on this branch: https://github.com/timgilmour/ODS/actions?query=branch%3Afix%2Famd-hipfire-dashboard-model-control

Transparency note: the CI wave caught the hipfire-rollback test fixtures needing the now-required renderer stubbed in this branch's copy (`204e8f2d`, matching the sibling branches' resolution and upstream's own activation-test practice).

Flattening the stack still has the structural constraint noted earlier: cross-repo PRs can't base on fork branches, so the diff narrows once the base hipfire extension (#1786) lands.

PR #1791, opened 2026-07-15
Harden OAuth callback flow with cryptographic state validation

### LINKED ISSUE #1790: OAuth callback (/api/oauth/callback) does not validate 'state' nonce — unauthenticated authorization-code injection / login-CSRF
## Summary
The unauthenticated OAuth redirect endpoint `GET /api/oauth/callback` performs no validation of the `state` parameter against a server-issued, per-flow nonce. It trusts the returned `state` directly as the skill identifier and writes the attacker-supplied `code`/`state` to a single fixed file, `data/persona/oauth_callback.json`, which the agent then consumes to finalize an OAuth connection. This is the missing-`state`-validation (OAuth CSRF / authorization-code injection) pattern: there is nothing tying a callback to a flow the server actually initiated.

## Evidence
`ods/extensions/services/dashboard-api/routers/oauth_passthrough.py`:

- Endpoint is intentionally public (no `Depends(verify_api_key)`), lines 201-208.
- `state` is not checked against any stored value — it is used as the skill name (line ~244): `skill = state.strip() or "google-workspace"`.
- The captured `code` is written verbatim to a fixed path for later consumption:
  ```python
  payload = {"code": code, "state": skill, "captured_at": int(time.time())}
  target = _callback_dir() / "oauth_callback.json"
  ... tmp.replace(target)
  ```
- Per the endpoint docstring, "The agent ... runs the relevant skill's `setup.py --auth-code` to finalise" using this file — so the injected `code` is actively exchanged.
- The docstring describes `state` as "the opaque state token the agent issued," but the handler never compares it to what was issued; there is no server-side store of pending states.

(HTML reflection is safe — `error`, `return_url`, and `skill` are `html.escape`d and `_safe_return_path` restricts redirects to same-origin relative paths, so this is not an XSS/open-redirect report.)

## Why this matters
Without `state` binding, the endpoint cannot distinguish a genuine provider redirect from a forged request. A single fixed output filename means any request can overwrite a pending legitimate callback. The agent then exchanges whatever `code` is present. Depending on flow direction this yields either OAuth login-CSRF (the victim's assistant is connected to the attacker's account, so the victim's actions/data flow into an attacker-controlled account) or injection of an attacker-obtained code — the standard consequences of a missing `state` check (CWE-352 in the OAuth context).

## Attack or failure scenario
A user is on the same LAN as the ODS box (the documented deployment) or is induced to load a crafted URL. The attacker issues `GET /api/oauth/callback?code=<attacker_code>&state=google-workspace`, overwriting `oauth_callback.json` while the victim has a connect flow in progress (or preempting one). The agent picks up the file and runs the skill's `setup.py --auth-code <attacker_code>`, linking the victim's assistant to the attacker's Google/Workspace account. Because the endpoint is unauthenticated and the filename is fixed, no credentials or race precision beyond timing are required.

## Root cause
The OAuth flow does not implement the `state` anti-CSRF nonce: no server-side record of initiated flows, and no comparison of the returned `state` to an issued value before acting on `code`.

## Recommended fix
- On auth-URL generation, mint a high-entropy `state` nonce, store it server-side (with the target skill and a short TTL), and set it as the flow identifier.
- In the callback, reject any request whose `state` does not match a stored, unexpired, unconsumed nonce; consume it single-use. Only then persist `code`.
- Key the callback file by the validated flow/nonce (not a single fixed filename) so concurrent or forged callbacks cannot clobber a legitimate one.
- Derive the skill from the stored flow record, not from the untrusted `state` string.

## Acceptance criteria
- A callback with an unknown/absent/expired/reused `state` is rejected without writing any file or invoking the skill.
- `state` values are single-use and bound to a specific initiated flow and skill.
- The persisted callback cannot be overwritten by an unrelated request.

## Suggested labels
security, dashboard

## Priority
P2

## Severity
Medium — unauthenticated authorization-code injection / OAuth login-CSRF against the assistant's connected accounts; constrained by requiring an in-progress/for-thcoming flow and network reachability to the box.

## Confidence
Likely — the absence of `state` validation and the agent's consumption of the written `code` are confirmed in-repo; full impact depends on the per-skill `setup.py` exchange semantics.

vaibhavsrv:
## Summary

This PR hardens the OAuth callback handoff by introducing deterministic state validation before authorization codes are persisted for Hermes.

Previously, `/api/oauth/callback` accepted caller-supplied `state` and `code` query parameters and wrote them directly to `oauth_callback.json`. The callback path had no executable validation that the state belonged to an OAuth flow initiated by ODS.

The existing state check in `SOUL.md.template` was a natural-language agent instruction rather than an enforced backend security control.

Fixes #1790.

## Root Cause

The OAuth callback endpoint did not maintain a trusted record of initiated OAuth flows.

As a result:

- callback state was not validated against an expected server-issued value
- the callback skill was derived from untrusted request input
- callback artifacts could be overwritten by later callbacks
- state validation depended on agent prompt instructions rather than executable code

The external Hermes `setup.py --auth-code` handlers are stateless and accept only the authorization code, so they cannot independently validate callback state.

## Fix

This PR adds an authenticated pending-flow registration step to `dashboard-api`.

`POST /api/oauth/pending` now:

- requires dashboard API authentication
- validates the requested skill identifier
- generates a cryptographically secure state nonce with `secrets.token_urlsafe(32)`
- binds the state nonce to the intended skill
- stores the pending flow for 15 minutes
- protects pending-flow mutation and consumption with a lock

`/api/oauth/callback` now:

- requires a state parameter
- atomically validates and consumes the pending state
- rejects missing, unknown, expired, and reused state values
- derives the skill from the trusted pending-flow record instead of callback query input
- writes callback data only after successful state validation

State values are single-use.

## Callback Artifact Isolation

Validated callbacks are persisted to skill-specific files:

`/data/persona/oauth_callback_<skill-id>.json`

For example:

`/data/persona/oauth_callback_google-workspace.json`

This prevents OAuth flows for different skills from clobbering each other's callback artifacts.

The legacy `oauth_callback.json` artifact is also written for backward compatibility.

Skill-specific paths are derived from the validated skill identifier stored during authenticated registration rather than raw callback input.

## Hermes Handoff

`SOUL.md.template` now instructs Hermes to:

1. register the intended OAuth skill with `dashboard-api`
2. obtain the server-generated state nonce
3. replace the authorization URL's state value with that nonce
4. wait for the provider callback
5. read the skill-specific callback artifact
6. pass the captured code to the existing `setup.py --auth-code` flow

The external Hermes setup scripts remain unchanged and continue to receive authorization codes through their existing stateless CLI interface.

## Compatibility

The callback continues writing the legacy `oauth_callback.json` artifact for backward compatibility.

No changes are required to the external Hermes `setup.py` OAuth handlers.

The pending-flow store is process-local and in memory. The current dashboard-api deployment runs as a single Uvicorn process. A dashboard-api restart during the short OAuth flow invalidates the pending state and requires the user to restart the OAuth setup flow.

## Security Impact

This change moves OAuth state validation from a prompt-level instruction into deterministic backend enforcement.

It prevents an unauthenticated callback request from injecting an arbitrary authorization code into the handoff path without possession of a valid, server-issued pending-flow state.

It also:

- binds callback state to the intended skill
- prevents state replay through single-use consumption
- rejects expired state values
- prevents raw callback input from selecting callback artifact paths
- isolates callback artifacts by skill

## Test Coverage

The existing OAuth passthrough regression coverage is preserved and adapted to the new state-validation contract.

Additional regression tests cover:

- authenticated pending-flow registration
- high-entropy server-generated state
- unique state generation
- state-to-skill binding
- successful valid-state callbacks
- missing state rejection
- unknown state rejection
- expired state rejection
- single-use state enforcement
- trusted skill derivation
- independent skill callback artifacts
- same-state concurrent consumption allowing exactly one success
- expired pending-flow cleanup

Existing coverage for callback file permissions, atomic writes, missing codes, provider errors, provider configuration, HTML safety, and return URL validation remains in place.

## Validation

```text
pytest tests/test_oauth_passthrough.py

25 passed


Lightheartdevs:
Thanks for taking on deterministic OAuth state validation; moving this out of persona instructions and into dashboard-api is the right security shape.

Quickest path to merge:
- In `ods/extensions/services/dashboard-api/routers/oauth_passthrough.py`, merge/rebase behind #1798 or include the same `return_url` backslash guard and tests here. This head still accepts paths like `/\evil.com` in `_safe_return_path`, so the public callback success page can still render an off-origin Back link after the state hardening.
- In `ods/extensions/services/hermes/compose.yaml`, document a maintainer decision on passing the full `DASHBOARD_API_KEY` into Hermes. If Hermes only needs OAuth-flow registration, a narrower registration token or scoped route would better match least privilege.
- Get checks to run/report for this head, especially dashboard-api `pytest tests/test_oauth_passthrough.py`; `gh pr checks` currently reports no checks.

What I checked:
- Static review of the dashboard-api router/tests plus Hermes `SOUL.md.template` and compose changes.
- The new `Field(pattern=...)` usage is compatible with the repo's pinned `pydantic>=2.5` stack and existing dashboard-api model style.

Thanks again for pushing this security path forward.

vaibhavsrv:
Thanks for the review. I addressed the _safe_return_path bypass by rejecting /\... paths and added regression tests covering both //evil.com and /\evil.com.

I also investigated the Hermes DASHBOARD_API_KEY usage. Hermes currently only uses it for POST /api/oauth/pending during OAuth registration. A narrower registration token or scoped endpoint appears feasible, but that would be a broader authentication/API design change, so I left the current behavior unchanged pending maintainer guidance on the preferred direction.

Lightheartdevs:
Thanks for carrying the OAuth callback hardening forward; the server-issued state path and the `/\evil.com` return-link regression are both important fixes.

Decision point:
- The implementation now enforces single-use, server-issued OAuth state in `ods/extensions/services/dashboard-api/routers/oauth_passthrough.py`, derives the skill from the trusted pending-flow record, and covers missing/unknown/expired/replayed state plus backslash return URLs in `tests/test_oauth_passthrough.py`.
- The remaining design call is whether `ods/extensions/services/hermes/compose.yaml` should pass the full `DASHBOARD_API_KEY` into Hermes for `POST /api/oauth/pending`, or whether maintainers want a narrower OAuth-registration token/route before this lands. That is an auth-boundary decision rather than a small local patch.
- Checks still need to run or be triggered for this head; `gh pr checks 1791` reports no checks, so the dashboard-api pytest evidence is not visible on GitHub yet.

What I checked:
- Static diff review of `routers/oauth_passthrough.py`, `tests/test_oauth_passthrough.py`, `hermes/SOUL.md.template`, and `hermes/compose.yaml`.
- `git diff --check origin/main...origin/pr/1791` passed.
- Local pytest was not run because this environment does not have a working Bash/WSL setup, and the PR's GitHub checks are not currently reported for this SHA.

Thanks again for moving this from prompt guidance into enforceable backend behavior.

vaibhavsrv:
Resolved merge conflicts with `main` and pushed the updated branch. All unit tests (`pytest test_oauth_passthrough.py`: 55/55 passed) and linters passed.

Lightheartdevs:
﻿
Thanks for rebasing the OAuth hardening and getting the callback-state contract onto a green head; this moves the risky part out of prompt guidance and into executable backend checks.

What needs a maintainer call:
- `ods/extensions/services/dashboard-api/routers/oauth_passthrough.py` now enforces server-issued, single-use OAuth state for both the existing disk-backed `/api/oauth/init` path and the new authenticated `/api/oauth/pending` registration path, with skill-specific callback artifacts written only after validation.
- The previous return-link issue is covered: `_safe_return_path()` rejects both `//...` and `/\...`, and `tests/test_oauth_passthrough.py` includes regression coverage for those callback success-page links.
- The remaining decision is auth scope in `ods/extensions/services/hermes/compose.yaml`: this passes the full `DASHBOARD_API_KEY` into Hermes so it can register OAuth flows. That may be acceptable for the current trusted Hermes posture, but maintainers should explicitly accept it or split out a narrower OAuth-registration token/route before merge.

What I checked:
- Reviewed the current head diff, PR description, existing discussion, and changed files for `routers/oauth_passthrough.py`, `tests/test_oauth_passthrough.py`, and `hermes/compose.yaml`.
- Ran `git diff --check origin/main...origin/pr/1791`: passed.
- Ran `python -m pytest tests/test_oauth_passthrough.py -q` from the #1791 worktree: 53 passed, 2 skipped.
- Confirmed GitHub checks are green, including dashboard API/frontend, Python lint/type check, ShellCheck, PowerShell lint, compose validation, Linux integration smoke, distro matrix, macOS smoke, env validation, and secret scan.

Thanks again for pushing this security fix forward.

PR #1797, opened 2026-07-15
fix(cold-storage): verify archive/restore moves and create cold dir

### LINKED ISSUE #3141: scripts/llm-cold-storage.sh: --execute logs ARCHIVED but never creates COLD_DIR — every archive silently fails
## Description

`scripts/llm-cold-storage.sh` never creates `COLD_DIR`, and the archive step does not check the `mv`/`ln -s` results. On a default first run of `--execute`, every archive fails (the target directory does not exist), but the script still logs `ARCHIVED: ...` and increments its archived counter, so the log claims success while nothing was archived.

The script runs under `set -uo pipefail` (no `-e`, `llm-cold-storage.sh:10`), so the failed `mv` at line 125 does not stop execution:

```bash
# line 26 — only the log directory is ever created; COLD_DIR is not
mkdir -p "$(dirname "$LOG_FILE")"
...
# lines 124-128
log "ARCHIVING: $name ($size, idle ${idle_days}d)"
mv "$model_dir" "$COLD_DIR/$name"          # fails: $COLD_DIR does not exist
ln -s "$COLD_DIR/$name" "${model_dir%/}"   # fails: destination exists
log "ARCHIVED: $name -> $COLD_DIR/$name"   # logged unconditionally
```

`((archived++))` on line 130 then reports `N archived` in the final summary.

## Steps to Reproduce

1. On any host, without pre-creating `$HOME/llm-cold-storage`:
   ```bash
   ./scripts/llm-cold-storage.sh --execute
   ```
2. Have at least one model in `$HF_CACHE/models--*/` idle for 7+ days.
3. Check the log / output.

One-liner that shows the failure mode deterministically:

```bash
HF_CACHE=$(mktemp -d) COLD_DIR=$(mktemp -d)/nonexistent-subdir LOG_FILE=$(mktemp -d)/x.log ./scripts/llm-cold-storage.sh --execute
```

## Expected Behavior

Either `COLD_DIR` is created up front (`mkdir -p "$COLD_DIR"` alongside the log-dir creation) and a failed `mv`/`ln -s` is reported as a failure (`SKIP`/error, not counted as archived), or the script refuses to run `--execute` when the cold-storage root cannot be used.

## Actual Behavior

```
[timestamp] ARCHIVING: models--Org--Model (4.2G, idle 12d)
mv: cannot move '...' to '/home/u/llm-cold-storage/models--Org--Model': No such file or directory
ln: failed to create symbolic link ...: File exists
[timestamp] ARCHIVED: models--Org--Model -> /home/u/llm-cold-storage/models--Org--Model
========== Scan complete: 1 archived, 0 skipped ==========
```

The model remains in the hot cache (unarchived), no error exit code is produced, and the summary/log claim it was archived.

## Impact

The tool's entire purpose (freeing NVMe space by archiving idle models) silently never works on a stock setup, and the false `ARCHIVED` entries make the log useless for diagnosing why. A user who trusts the log may delete what they believe is a redundant cache copy.

## Environment

- Verified against current `main` (`6ff9b4fc`), `scripts/llm-cold-storage.sh` lines 10, 26, 118-131.
- Any OS with Bash; repro is filesystem-level, no GPU/Docker involved.

tang-vu:
## Problem

`scripts/llm-cold-storage.sh` runs without `set -e` and never checks the results of its destructive operations:

1. The first `--execute` run fails but reports success because nothing creates `$COLD_DIR`.
2. A failed `ln -s` after a successful `mv` can strand the model outside the Hugging Face cache.
3. Moving into an existing destination nests directories instead of replacing or refusing the collision.
4. Integer-only age calculation unnecessarily depends on `bc`; when it is unavailable, archiving is silently disabled.

## Fix

- Create `$COLD_DIR` before real archive operations and fail visibly if it is unusable.
- Refuse destination collisions on archive and restore.
- Check every `mv`; roll the model back when symlink creation fails.
- Count failures and return non-zero if any operation fails.
- Use Bash arithmetic for the age calculation.

## Why this matters

Issue #3141 reproduces the production failure on a stock first run: the script logs `ARCHIVED`, increments the success count, and exits successfully even though the model never moved. False success on a destructive storage path can leave operators with neither reclaimed space nor a trustworthy recovery record.

Fixes #3141.

## Overlap check

Refreshed 2026-08-25. Searched open and closed PRs and issues for `llm-cold-storage`, `COLD_DIR`, archive/restore moves, and the changed production path `ods/scripts/llm-cold-storage.sh`. No merged equivalent or competing implementation was found. PR #1797 predates issue #3141 and its boundary tests reproduce the same first-run failure, so this PR is the canonical implementation for that report.

## Test plan

`tests/test-llm-cold-storage.sh`, wired into `make test`, exercises the public script boundary:

- first-run archive creates the cold directory, moves the model, and leaves a symlink;
- a second run is idempotent;
- restore completes a round trip;
- archive collision returns non-zero without a false `ARCHIVED` receipt;
- restore onto a real directory is refused and preserves the cold copy.

Validation already recorded on this head:

- all five focused tests pass on Linux/WSL (the first fails against upstream `main` with the reported missing-directory error);
- ShellCheck passes for the production script and regression test;
- the full GitHub Actions matrix is green, including Linux integration, distro and macOS smoke, PowerShell lint, dashboard, env validation, and secret scan.

## Compatibility and rollback

Dry-run behavior is unchanged. Existing archive layouts and restore syntax remain compatible. Reverting the PR restores the previous best-effort behavior but also restores the false-success and collision risks.

### DIFF
--- a/ods/Makefile
+++ b/ods/Makefile
@@ -148,6 +148,9 @@ test: ## Run unit and contract tests
 	@echo ""
 	@echo "=== session cleanup lifecycle ==="
 	@bash tests/test-session-cleanup.sh
+	@echo ""
+	@echo "=== llm-cold-storage archive/restore safety ==="
+	@bash tests/test-llm-cold-storage.sh
 
 bats: ## Run BATS unit tests for shell libraries
 	@echo "=== BATS unit tests ==="
--- a/ods/scripts/llm-cold-storage.sh
+++ b/ods/scripts/llm-cold-storage.sh
@@ -77,15 +77,22 @@ get_last_access_days() {
     fi
     local now
     now="$(date +%s)"
-    local age_secs
-    age_secs="$(echo "$now - ${newest_atime%.*}" | bc)"
-    echo "$(( age_secs / 86400 ))"
+    # Epoch math is integer-only; bash arithmetic avoids a bc dependency
+    # (bc is absent on minimal images, and an empty result here would make
+    # every model look recent and silently disable archiving).
+    echo "$(( (now - ${newest_atime%.*}) / 86400 ))"
 }
 
 do_archive() {
     local dry_run="${1:-true}"
     local archived=0
     local skipped=0
+    local failed=0
+
+    if [[ "$dry_run" != "true" ]] && ! mkdir -p "$COLD_DIR"; then
+        log "ERROR: cannot create cold storage directory: $COLD_DIR"
+        exit 1
+    fi
 
     log "========== LLM cold storage scan started (dry_run=$dry_run) =========="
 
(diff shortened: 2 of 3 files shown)


Lightheartdevs:
Thanks for taking on the cold-storage archive/restore safety path; these are exactly the file-move failures that can quietly strand models.

What looks good:
- `ods/scripts/llm-cold-storage.sh` now creates `COLD_DIR` before real archive moves and treats failed archive/restore operations as non-zero failures.
- Archive and restore both refuse destination collisions instead of allowing nested `models--X/models--X` layouts.
- The new `ods/tests/test-llm-cold-storage.sh` covers first-run archive, idempotent rerun, restore, archive collision, and restore-onto-real-dir behavior, and it is wired into `make test`.

What I checked:
- Diff/static review of the shell script, test fixture, and Makefile wiring.

Thanks again for making the destructive path fail visibly.

Lightheartdevs:
Thanks for rebasing the cold-storage safety fix; the archive/restore path still fails visibly instead of silently stranding model files.

What looks good:
- `ods/scripts/llm-cold-storage.sh` creates `COLD_DIR` before real archive moves, checks every destructive `mv`, and rolls back if symlink creation fails.
- Archive and restore now refuse destination collisions instead of allowing nested `models--X/models--X` layouts, and failures are counted with a non-zero exit.
- `ods/tests/test-llm-cold-storage.sh` covers first-run archive, idempotent rerun, restore round-trip, archive collision, and restore-onto-real-dir refusal, with the suite wired into `ods/Makefile`.

What I checked:
- Current diff/static review of the script, test, and Makefile wiring.
- `git diff --check origin/main...pr-1797` passed.
- Current GitHub checks are green, including ShellCheck, Test Linux/integration-smoke, matrix smoke, env validation, Dashboard, and lint jobs. I did not count a local Git Bash run as execution evidence because this Windows host's `ln -s` emulation creates directory copies instead of real symlinks, which fails the test's `[[ -L ... ]]` assertion independently of this PR.

Thanks again for making this destructive path safer.

PR #1799, opened 2026-07-15
fix(release-gate): report cleanly when a version file is missing

tang-vu:
﻿﻿## Problem

`scripts/check-version-consistency.py` is the release-gate step that fails when ODS version authorities drift. It is meant to fail **cleanly** — `main()` even wraps the manifest read with a `# noqa: BLE001 - gate should report cleanly` comment. But two paths bypass that intent:

1. **Missing file crashes the gate.** `add_regex_check` catches only `ValueError`:

   ```python
   try:
       checks.append((label, first_match(path, pattern, label)))
   except ValueError as exc:
       errors.append(str(exc))
   ```

   `first_match` → `read_text` → `path.read_text()` raises `FileNotFoundError` (an `OSError`, **not** a `ValueError`) when a version-authority file is absent — e.g. a refactor renames `installers/windows/lib/constants.ps1`, or `ARCHITECTURE.md` isn't present in a sparse checkout. The exception escapes and the gate dies with a raw traceback instead of a `[FAIL]` summary listing the missing file.

2. **`relative_to(ROOT)` raises for `ARCHITECTURE.md`.** On the "could not find version" path, `first_match` builds its message with `path.relative_to(ROOT)`. The `ARCHITECTURE.md` check uses `ROOT.parent / "ARCHITECTURE.md"`, which is **not** under `ROOT`, so `relative_to` itself raises `ValueError` — masking the real "pattern not found" message.

## Fix

- Wrap the read in `first_match` and re-raise `OSError` as a `ValueError` with a clean message.
- Add a `_display_path` helper that falls back to the absolute path when `relative_to(ROOT)` can't apply (paths outside `ROOT`).
- Guard the `CHANGELOG.md` read in `latest_changelog_release` the same way.

## Tests

New `tests/test-version-consistency-error-reporting.py` (wired into the Linux CI job next to the other `python3 tests/test-*.py` steps):

- `first_match` on a missing file raises `ValueError`, not `OSError`
- `add_regex_check` records a clean error string and never propagates
- `_display_path` tolerates a path outside `ROOT`
- an unmatched pattern on an outside-`ROOT` file reports "could not find version" instead of a `relative_to` error

Verified:
- With the fix: **[PASS]**, and the real gate still passes (`[PASS] version consistency (2.5.3)`).
- Reverting only the source change: the missing-file test **fails** with `first_match leaked OSError instead of ValueError`, pinning the exact bug.

## Why this matters

Release validation is an operator-facing gate. A missing version authority should produce one actionable validation error rather than a Python traceback that obscures the broken release input.

## Overlap check

Triage refresh (2026-08-08): searched open and closed Osmantic/ODS PRs by semantic keywords (version consistency missing file), inspected the changed production files, and checked patch equivalence against current upstream/main. No strict duplicate or merged equivalent was found.

### DIFF
--- a/.github/workflows/test-linux.yml
+++ b/.github/workflows/test-linux.yml
@@ -187,6 +187,9 @@ jobs:
       - name: Offline Model Validation Tests
         run: python3 tests/test-offline-model-validation.py
 
+      - name: Version Consistency Error Reporting Tests
+        run: python3 tests/test-version-consistency-error-reporting.py
+
       - name: Runtime Config Renderer Tests
         run: python3 tests/test-render-runtime-configs.py
 
--- a/ods/scripts/check-version-consistency.py
+++ b/ods/scripts/check-version-consistency.py
@@ -17,15 +17,34 @@ def read_text(path: Path) -> str:
     return path.read_text(encoding="utf-8")
 
 
+def _display_path(path: Path) -> str:
+    """Repo-relative path for messages, tolerant of paths outside ROOT.
+
+    ARCHITECTURE.md lives at ROOT.parent, so a plain path.relative_to(ROOT)
+    would raise ValueError and mask the real error.
+    """
+    try:
+        return str(path.relative_to(ROOT))
+    except ValueError:
+        return str(path)
+
+
 def first_match(path: Path, pattern: str, label: str) -> str:
-    match = re.search(pattern, read_text(path), re.MULTILINE | re.DOTALL)
+    try:
+        text = read_text(path)
+    except OSError as exc:
+        raise ValueError(f"{label}: cannot read {_display_path(path)}: {exc}") from exc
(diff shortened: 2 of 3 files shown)


Lightheartdevs:
Thanks for tightening the version-consistency gate; this keeps a missing authority file from turning a release check into a traceback hunt.

What looks good:
- `ods/scripts/check-version-consistency.py` now routes missing/read errors through `ValueError`, matching `add_regex_check`'s existing error aggregation path.
- `_display_path` handles the root/`ARCHITECTURE.md` boundary without changing the actual version authorities.
- `ods/tests/test-version-consistency-error-reporting.py` covers the missing-file and outside-root cases and is wired into `test-linux.yml`.

What I checked:
- Diff/static review of the script, new regression test, and CI wiring.

Thanks again for making the release gate fail cleanly.

PR #1800, opened 2026-07-15
fix(service-registry): stop cache counters aborting set -e callers

tang-vu:
﻿## Problem

`sr_compose_flags` in the shared `lib/service-registry.sh` library bumps its cache counters with `((counter++))`:

```bash
if [[ "$_SR_COMPOSE_FLAGS_CACHED" == "true" ]]; then
    ((_SR_CACHE_HITS++))
    ...
fi
((_SR_CACHE_MISSES++))
```

`((counter++))` is a post-increment: it evaluates to the value **before** the increment. On the first bump (`0 -> 1`) that value is `0`, so the arithmetic command returns **exit 1**. Under `set -e`, a caller that invokes `sr_compose_flags` directly aborts mid-function.

The two current callers happen to escape this because they wrap the call in `$(...)` (`ext_flags=$(sr_compose_flags)`), and a command substitution masks the intermediate failure. But `service-registry.sh` is a **shared library** sourced by `install-core.sh`, the install phases, `ods-cli`, and `build-capability-profile.sh`. Leaving a `set -e` landmine for the next direct caller is fragile, and every other counter in the repo already avoids it (`first-boot-demo.sh`, `validate.sh`, `migrate-config.sh`, … use guards; the in-flight `ods-update.sh` fix uses the same assignment form adopted here).

## Reproduction

```bash
set -e
inc() { ((c++)); }   # same shape as the registry counters, c starts at 0
c=0
inc                  # direct call, not in $()
echo "REACHED"       # never prints — the shell aborted at ((c++))
```

## Fix

Use the arithmetic-assignment form, which always returns exit 0:

```bash
_SR_CACHE_HITS=$((_SR_CACHE_HITS + 1))
_SR_CACHE_MISSES=$((_SR_CACHE_MISSES + 1))
```

## Tests

New `tests/test-service-registry-set-e.sh` (wired into the Linux CI job) sources the library under `set -euo pipefail`, stubs out `sr_load` (so it needs neither PyYAML nor real manifests), and calls `sr_compose_flags` **directly** (not via `$(...)`), asserting the caller survives and the counters land at `hits=1 misses=1`.

Verified on Linux (WSL):
- With the fix: **passes**.
- Reverting only the library change: **fails** with `sr_compose_flags aborted its set -e caller`.

`shellcheck -x` reports no errors on the changed files; the existing `tests/test-service-registry-cache.sh` still passes.

## Why this matters

The service registry is sourced by `set -e` installer and CLI paths. A cache counter must never terminate a caller on its first increment.

## Overlap check

Triage refresh (2026-08-08): searched open and closed Osmantic/ODS PRs by semantic keywords (service registry cache counter set -e), inspected the changed production files, and checked patch equivalence against current upstream/main. No strict duplicate or merged equivalent was found.


Lightheartdevs:
Thanks for tightening up the service-registry cache path here; this is exactly the sort of small shared-library fix that prevents the next direct caller from inheriting a hard-to-debug set -e abort.

What looks good:
- ods/lib/service-registry.sh switches only the two cache counter bumps from ((counter++)) to assignment arithmetic, preserving the cache behavior while avoiding the first-increment exit status trap under set -e.
- ods/tests/test-service-registry-set-e.sh exercises the direct-call path that command substitution currently masks, and it checks both the miss and hit counters land at 1.
- The Linux CI wiring in .github/workflows/test-linux.yml puts this regression beside the existing service-registry tests, which matches the shared-library risk surface.

What I checked:
- Reviewed the PR diff, PR body, changed files, and existing discussion for head $sha.
- Compared the change against the current service-registry implementation and the repo's shell guidance around visible failures and set -euo pipefail.
- GitHub checks are green for ShellCheck, Test Linux/integration-smoke, Matrix Smoke, Dashboard, Python lint/type paths, env validation, secret scan, and PowerShell lint. I could not run the new bash test locally because this Windows automation host only exposes WSL ash.exe, and that WSL context has no /bin/bash.

Lightheartdevs:
Thanks for rebasing the service-registry counter fix; this keeps a shared library path from surprising the next direct `set -e` caller.

What looks good:
- `ods/lib/service-registry.sh` changes only the cache hit/miss increments from post-increment arithmetic to assignment arithmetic, preserving behavior while avoiding the first-increment exit-status trap.
- `ods/tests/test-service-registry-set-e.sh` exercises the direct-call path that command substitution currently masks and verifies both counters land at `hits=1 misses=1`.
- `.github/workflows/test-linux.yml` places the regression beside the existing service-registry coverage, which matches the shared-library risk surface.

What I checked:
- Current diff/static review of the library, test, and Linux CI wiring.
- `C:\Program Files\Git\bin\bash.exe tests/test-service-registry-set-e.sh` passed locally.
- `git diff --check origin/main...pr-1800` passed; current GitHub checks are green, including ShellCheck, Test Linux/integration-smoke, matrix smoke, env validation, Dashboard, and lint jobs.

Thanks again for closing this sharp edge.

PR #1801, opened 2026-07-15
fix(host-agent): keep .env mode 0600 after a settings save

w3-surfer:
## Problem (local secret disclosure)

The host-agent's `env_update` handler (dashboard **Settings → save**) writes the new `.env` atomically:

```python
tmp_path = env_path.with_name(".env.tmp")
tmp_path.write_text(payload_text, encoding="utf-8")
os.replace(str(tmp_path), str(env_path))
```

`.env.tmp` is created fresh at the umask default (typically **0644**). Because `os.replace` swaps in that new inode, the live `.env` **inherits 0644**, discarding the `0600` it was created with (`installers/phases/06-directories.sh` does `chmod 600 .env`).

`.env` holds the service secrets — `DASHBOARD_API_KEY`, `ODS_AGENT_KEY`, `JWT_SECRET`, OAuth client secrets, etc. So after **any** dashboard Settings save, they become readable by every local user on the host.

```console
# reproduction of the umask/replace behavior
BEFORE fix: .env after os.replace = 0o644
AFTER fix:  .env after os.replace = 0o600
```

## Fix

`chmod` the tmp file to `0600` before the atomic replace, so the live `.env` keeps `0600` with no intermediate 0644 window. This mirrors the sibling secret writers that already do it: `oauth_passthrough.py:241`, `magic_link.py:253`, `security.py:19`.

One-line change; `py_compile` clean.

### DIFF
--- a/ods/bin/ods-host-agent.py
+++ b/ods/bin/ods-host-agent.py
@@ -2256,6 +2256,11 @@ class AgentHandler(BaseHTTPRequestHandler):
             payload_text = raw_text if raw_text.endswith("\n") else raw_text + "\n"
             tmp_path = env_path.with_name(".env.tmp")
             tmp_path.write_text(payload_text, encoding="utf-8")
+            # .env holds service secrets (DASHBOARD_API_KEY, ODS_AGENT_KEY,
+            # JWT_SECRET, OAuth secrets). The fresh tmp file is created at the
+            # umask default (0644); tighten before os.replace() swaps it in so
+            # the live .env keeps mode 0600 instead of inheriting 0644.
+            os.chmod(tmp_path, 0o600)
             os.replace(str(tmp_path), str(env_path))
         except OSError as exc:
             logger.warning("env_update OSError from %s: %s", client_ip, exc)


Lightheartdevs:
Thanks for tightening the Settings save path; preserving `.env` permissions after atomic replace is the right bug to fix, but the temporary file also needs to start private.

Quickest path to merge:
- In `ods/bin/ods-host-agent.py`, avoid `tmp_path.write_text(...)` before the permission is restricted. `Path.write_text` creates `.env.tmp` using the process umask, so on common `022` hosts the secret-bearing temp file can be `0644` until `os.chmod(tmp_path, 0o600)` runs. Create/open the temp file with mode `0600` first, for example with `os.open(tmp_path, os.O_WRONLY | os.O_CREAT | os.O_TRUNC, 0o600)`, write through that fd, then `os.replace`.
- Add a host-agent regression that exercises the env update path under a permissive umask, or intercepts the replace to assert the temp file is already `0600` before it becomes the live `.env`.

What I checked:
- Static diff review of `_handle_env_update()` and the existing secret-writer patterns noted in the PR.
- GitHub checks for this head are passing; the blocker is the remaining pre-chmod exposure window.

Thanks again for narrowing this down.

PR #1802, opened 2026-07-15
fix(backup): restrict the compressed backup archive to owner-only

w3-surfer:
## Problem (local secret disclosure)

`compress_backup()` in `ods-backup.sh` creates the backup archive and never restricts its permissions:

```sh
tar czf "$parent_dir/$backup_name.tar.gz" -C "$parent_dir" "$backup_name"
```

Under the default umask the `.tar.gz` lands at **0644** — world-readable. The archive bundles the raw `.env`, which the config backup copies in verbatim (`DASHBOARD_API_KEY`, `ODS_SESSION_SECRET`, service passwords, `LANGFUSE_*` secrets, …). So on a multi-user host, any local user can read a compressed backup and extract **every** ODS secret.

The uncompressed backup keeps `.env` at `0600` (because `cp` carries the source mode) — only the compressed archive leaks.

```console
BEFORE fix: archive mode = 644
AFTER fix:  archive mode = 600
```

## Fix

`chmod 600` the archive right after `tar czf`, matching the `0600` the `.env` itself carries. Reachable via `ods-backup.sh -t full -c` / `-t config -c`. `bash -n` clean.

### DIFF
--- a/ods/ods-backup.sh
+++ b/ods/ods-backup.sh
@@ -462,6 +462,10 @@ compress_backup() {
     parent_dir=$(dirname "$backup_dir")
 
     tar czf "$parent_dir/$backup_name.tar.gz" -C "$parent_dir" "$backup_name"
+    # The archive bundles the raw .env (DASHBOARD_API_KEY, session secret, service
+    # passwords). Restrict it to the owner rather than leaving it world-readable
+    # at the umask default, matching the 0600 the .env itself carries.
+    chmod 600 "$parent_dir/$backup_name.tar.gz"
     local compressed_size
     compressed_size=$(du -sh "$parent_dir/$backup_name.tar.gz" | cut -f1)
 


Lightheartdevs:
Thanks for taking on the compressed-backup permissions issue; the fix is aimed at the right leak, but the archive needs to be private from the moment it is created.

Quickest path to merge:
- In `ods/ods-backup.sh`, create `$parent_dir/$backup_name.tar.gz` under a restrictive umask instead of only `chmod`ing after `tar czf`. With the current order, a default `umask 022` host can create the secret-bearing archive as `0644` briefly before `chmod 600`, so a local watcher can still race the disclosure. A compact pattern is `archive="$parent_dir/$backup_name.tar.gz"; ( umask 077; tar czf "$archive" -C "$parent_dir" "$backup_name" )`, followed by an optional `chmod 600 "$archive"` as a belt-and-suspenders check.
- Add a focused regression around compressed config/full backup under a permissive umask, asserting the resulting `.tar.gz` mode is `600`.

What I checked:
- Static diff review of `compress_backup()` in `ods/ods-backup.sh`.
- GitHub checks for this head are passing; the remaining issue is the pre-chmod local-read window in the behavior this PR is meant to harden.

PR #1804, opened 2026-07-15
fix(dashboard-api): require auth for /api/preflight/required-ports

w3-surfer:
## Problem

`/api/preflight/required-ports` is the **only** preflight endpoint without an auth dependency. Its four siblings all gate:

```python
@app.get("/api/preflight/docker", dependencies=[Depends(verify_api_key)])   # main.py:1058
@app.get("/api/preflight/gpu",    dependencies=[Depends(verify_api_key)])   # main.py:1083
@app.get("/api/preflight/required-ports")                                   # main.py:1100  <-- ungated
@app.post("/api/preflight/ports", dependencies=[Depends(verify_api_key)])   # main.py:1117
@app.get("/api/preflight/disk",   dependencies=[Depends(verify_api_key)])   # main.py:1137
```

Its response enumerates which services are **deployed** and their external ports:

```python
return {"ports": [{"port": ext_port, "service": ...} for ...]}
```

So any unauthenticated client that can reach dashboard-api (e.g. `api.<device>.local` on the LAN) gets a live map of running services and ports — service/port reconnaissance that the four adjacent endpoints already refuse to give up.

## Fix

Gate it the same way as its siblings:

```python
@app.get("/api/preflight/required-ports", dependencies=[Depends(verify_api_key)])
```

**Setup is unaffected.** The dashboard's `PreFlightChecks.jsx` fetches `required-ports` with the exact same bare `fetch('/api/preflight/...')` it uses for the already-gated `/docker`, `/gpu`, `/ports`, `/disk` — the dashboard nginx injects `Authorization: Bearer ${DASHBOARD_API_KEY}` into `/api/*`, so all five now work identically. Installer preflight runs its own local checks and doesn't call this HTTP route.

## Tests

- `test_preflight_required_ports_requires_auth` — no Bearer → **401**.
- `test_preflight_required_ports_authenticated` — with Bearer → **200** + port list.
- Updated `TestPreflightRequiredPorts.test_returns_ports` to send the auth header.
- Updated the README auth column (`No` → `Yes`).

Preflight test selection: **25 passed**.

### DIFF
--- a/ods/extensions/services/dashboard-api/README.md
+++ b/ods/extensions/services/dashboard-api/README.md
@@ -59,7 +59,7 @@ Environment variables (set in `.env`):
 |--------|------|------|-------------|
 | `GET` | `/api/preflight/docker` | Yes | Check Docker availability |
 | `GET` | `/api/preflight/gpu` | Yes | Check GPU availability |
-| `GET` | `/api/preflight/required-ports` | No | List service ports |
+| `GET` | `/api/preflight/required-ports` | Yes | List service ports |
 | `POST` | `/api/preflight/ports` | Yes | Check port availability conflicts |
 | `GET` | `/api/preflight/disk` | Yes | Check available disk space |
 
--- a/ods/extensions/services/dashboard-api/main.py
+++ b/ods/extensions/services/dashboard-api/main.py
@@ -1097,9 +1097,14 @@ async def preflight_gpu():
     return {"available": False, "error": "No GPU detected. Ensure NVIDIA drivers or AMD amdgpu driver is loaded."}
 
 
-@app.get("/api/preflight/required-ports")
+@app.get("/api/preflight/required-ports", dependencies=[Depends(verify_api_key)])
 async def preflight_required_ports():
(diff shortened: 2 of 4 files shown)


Lightheartdevs:
Thanks for closing the unauthenticated preflight port inventory path; this lines up the route with the sibling preflight checks.

What looks good:
- `ods/extensions/services/dashboard-api/main.py` gates `/api/preflight/required-ports` with the same `verify_api_key` dependency as docker, gpu, ports, and disk.
- The existing dashboard nginx config injects `Authorization` for `/api/*`, so `PreFlightChecks.jsx` can keep the same bare `fetch()` pattern used by the other gated preflight calls.
- `test_main.py`, `test_routers.py`, and the dashboard-api README all move to the authenticated contract.

What I checked:
- Static diff review against `PreFlightChecks.jsx`, `ods/extensions/services/dashboard/nginx.conf`, and the dashboard-api auth helper.
- GitHub checks for this head are passing, including dashboard-api `api`, frontend, Python lint/mypy, shell lint, smoke, distro, and env-validation jobs.

Thanks again for pushing this security cleanup forward.

PR #1805, opened 2026-07-15
fix(dashboard-api): cap PortCheckRequest.ports length

w3-surfer:
## Problem

`preflight_ports()` binds a socket per requested port, synchronously, on the event-loop thread:

```python
@app.post("/api/preflight/ports", dependencies=[Depends(verify_api_key)])
async def preflight_ports(request: PortCheckRequest):
    ...
    for port in request.ports:
        with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as sock:
            sock.settimeout(1)
            sock.bind(("0.0.0.0", port))
```

`PortCheckRequest.ports` was unbounded:

```python
class PortCheckRequest(BaseModel):
    ports: list[PortNumber]
```

`PortNumber` bounds each *value* to 1–65535, but nothing bounds the list *length* (and duplicates are allowed), so a single authenticated request carrying tens of thousands of entries forces that many blocking `socket()`/`bind()`/`close()` cycles on the event loop — a bind-probe amplification / self-DoS.

## Fix

Cap the list length, matching the defensive input-bounding already used elsewhere in this module (e.g. `message: str = Field(..., max_length=100000)`):

```python
ports: Annotated[list[PortNumber], Field(max_length=128)]
```

A real install exposes ~two dozen service ports, and the dashboard sends one entry per required service (`PreFlightChecks.jsx` → `ports.map(p => p.port)`), so 128 is comfortable headroom. Oversized lists now fail validation with **422** before any socket work runs.

## Tests

Added `TestPreflightPorts.test_rejects_oversized_port_list` — a 129-entry list → **422**. Existing valid-range / invalid-value / in-use tests still pass (**8 passed** in the ports selection).

### DIFF
--- a/ods/extensions/services/dashboard-api/models.py
+++ b/ods/extensions/services/dashboard-api/models.py
@@ -66,7 +66,10 @@ PortNumber = Annotated[int, Field(ge=1, le=65535)]
 
 
 class PortCheckRequest(BaseModel):
-    ports: list[PortNumber]
+    # preflight_ports binds a socket per entry synchronously on the event loop,
+    # so cap the list. A real install exposes a couple dozen service ports;
+    # 128 leaves ample headroom while preventing bind-probe amplification.
+    ports: Annotated[list[PortNumber], Field(max_length=128)]
 
 
 class PortConflict(BaseModel):
--- a/ods/extensions/services/dashboard-api/tests/test_main.py
+++ b/ods/extensions/services/dashboard-api/tests/test_main.py
@@ -648,6 +648,16 @@ class TestPreflightPorts:
         )
         assert resp.status_code == 200
 
+    def test_rejects_oversized_port_list(self, test_client):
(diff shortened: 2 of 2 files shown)


Lightheartdevs:
Thanks for tightening the preflight port probe here; bounding this request is a good fit for the dashboard-api input validation surface.

What looks good:
- `ods/extensions/services/dashboard-api/models.py` now bounds `PortCheckRequest.ports` before `preflight_ports()` can run the synchronous bind loop.
- The new `TestPreflightPorts.test_rejects_oversized_port_list` covers the oversized-list failure path while preserving valid port range behavior.
- The limit leaves headroom for the current service-port set and avoids changing the frontend request shape.

What I checked:
- Static diff review against the existing preflight model and `preflight_ports()` endpoint.
- GitHub checks for this head are passing, including dashboard-api `api`, Python lint/mypy, shell lint, smoke, distro, and env-validation jobs.

PR #1806, opened 2026-07-15
fix(extensions): close bind-mount bypasses in the compose volume scanner

w3-surfer:
## Problem

`_scan_compose_content` blocks host bind mounts in extension compose files by string-splitting each volume entry and checking whether the first `:`-segment starts with `/`:

```python
vol_str = str(vol)
...
vol_parts = vol_str.split(":")
if len(vol_parts) >= 2 and vol_parts[0].startswith("/"):
    raise HTTPException(... "absolute host path mount ...")
```

This only catches **short-form absolute strings** and misses two other host-path binds:

1. **Long-form mounts** — `{type: bind, source: /, target: /host}`. `vol` is a dict, so `str(vol)` is the dict repr and `vol_parts[0]` is `"{'type'"` — the absolute-path check never fires. An extension can bind `/` (or any host path) read-write.
2. **Relative traversal** — `../../:/host`. `vol_parts[0]` is `".."` (not absolute), so it passes. Compose resolves it **above** the project/extension directory, reaching ODS's `.env` and its siblings.

The scanner already has a sibling guard against bind-mount backdoors via top-level `driver_opts` (`type/device`), so blocking host binds here is an intended security boundary — these two forms are gaps in it.

> Scope/reachability: extensions are installed from the trusted library or existing on-disk content (there is no remote/URL upload path), so this is defense-in-depth against a malicious or compromised **library** extension, in the same spirit as the existing `driver_opts` check. Not remotely reachable.

## Fix

Resolve the host-side `source` for both short- and long-form entries, then reject an absolute source or one containing a `..` path segment:

```python
if isinstance(vol, dict):
    vol_source = str(vol.get("source", ""))
else:
    vol_source = vol_str.split(":", 1)[0]
if vol_source.startswith("/"):
    raise HTTPException(... "absolute host path mount ...")
if ".." in vol_source.split("/"):
    raise HTTPException(... "relative host path mount ... escaping the project directory ...")
```

Named volumes (`mydata:/data`) and project-relative subdir binds (`./data:/data`) are unaffected — only absolute and `..`-escaping sources are rejected.

## Tests

`TestScanComposeVolumeBypass` (direct `_scan_compose_content` calls):
- long-form `{type: bind, source: /}` → **400** absolute host path
- short-form `../../:/host` → **400** escaping the project directory
- long-form `source: ../..` → **400** escaping the project directory
- `mydata:/var/lib/data` (named volume) → **allowed**
- `./data:/data` (project-relative) → **allowed**

Full `test_extensions.py`: **206 passed**.

### DIFF
--- a/ods/extensions/services/dashboard-api/routers/extensions.py
+++ b/ods/extensions/services/dashboard-api/routers/extensions.py
@@ -397,11 +397,23 @@ def _scan_compose_content(
                         status_code=400,
                         detail=f"Extension rejected: Docker socket mount in {svc_name}",
                     )
-                vol_parts = vol_str.split(":")
-                if len(vol_parts) >= 2 and vol_parts[0].startswith("/"):
+                # Resolve the host-side source for both short-form
+                # ("source:target[:opts]") and long-form ({type: bind,
+                # source: ...}) entries, so a long-form bind can't slip an
+                # absolute/relative host path past the string-only check.
+                if isinstance(vol, dict):
+                    vol_source = str(vol.get("source", ""))
+                else:
+                    vol_source = vol_str.split(":", 1)[0]
+                if vol_source.startswith("/"):
                     raise HTTPException(
                         status_code=400,
-                        detail=f"Extension rejected: absolute host path mount '{vol_parts[0]}' in {svc_name}",
+                        detail=f"Extension rejected: absolute host path mount '{vol_source}' in {svc_name}",
+                    )
+                if ".." in vol_source.split("/"):
+                    raise HTTPException(
+                        status_code=400,
+                        detail=f"Extension rejected: relative host path mount '{vol_source}' escaping the project directory in {svc_name}",
                     )
         cap_add = svc_def.get("cap_add", [])
         if isinstance(cap_add, list):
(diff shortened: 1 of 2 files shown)


Lightheartdevs:
Thanks for tightening the extension compose scanner; closing the long-form bind and POSIX traversal gaps is exactly the right boundary to harden.

Quickest path to merge:
- In `ods/extensions/services/dashboard-api/routers/extensions.py`, extend the host-source parser to handle Windows-style sources as well as POSIX ones. The new short-form `vol_str.split(":", 1)[0]` still treats `C:\Users\me:/host` or `C:/Users/me:/host` as source `C`, and the `vol_source.split("/")` traversal check misses `..\..` in both short- and long-form volumes. Because ODS supports Windows/Docker Desktop, those are still host bind/traversal bypasses for the same scanner boundary.
- Add focused tests in `ods/extensions/services/dashboard-api/tests/test_extensions.py` for Windows drive-letter absolute binds (`C:\...` and `C:/...`) and backslash traversal (`..\..`), covering both short-form and long-form `source:` where practical.

What I checked:
- Static diff review of `_scan_compose_content()` and the added `TestScanComposeVolumeBypass` cases.
- `gh pr checks 1806` is green, including dashboard-api `api`, Python lint/mypy, shell lint, integration-smoke, distro smoke, frontend, PowerShell lint, and env-validation jobs.

Thanks again for pushing this hardening forward.

PR #1807, opened 2026-07-15
fix(dashboard-api): tolerate non-numeric context in get_model_info

w3-surfer:
## Problem

`get_model_info()` reads the context length straight from `.env`:

```python
context = int(env_values.get("MAX_CONTEXT") or env_values.get("CTX_SIZE") or 32768)
```

This sits inside a block whose only handler is `except OSError`, so a **non-numeric** value raises `ValueError` that escapes the function and 500s every endpoint that surfaces model info (status, model readiness, etc.).

`MAX_CONTEXT`/`CTX_SIZE` are user-supplied at install time, and values like `auto`, `8k`, or a line with a trailing comment (`CTX_SIZE=32768 # tokens`) all trigger it:

```
>>> get_model_info()   # .env has CTX_SIZE=auto
ValueError: invalid literal for int() with base 10: 'auto'
```

## Fix

Guard the conversion and fall back to the default — the same guard `routers/models.py` already applies to these exact two keys (`for key in ("MAX_CONTEXT", "CTX_SIZE"): ... except (TypeError, ValueError)`):

```python
try:
    context = int(env_values.get("MAX_CONTEXT") or env_values.get("CTX_SIZE") or 32768)
except (TypeError, ValueError):
    context = 32768
```

Numeric values are unaffected; only malformed ones fall back instead of crashing.

## Tests

Added to `TestGetModelInfo`:
- `test_parses_numeric_context` — `CTX_SIZE=8192` → `context_length == 8192` (valid values still parse).
- `test_non_numeric_context_falls_back_to_default` — `CTX_SIZE=auto` → `context_length == 32768` (no crash).

`test_helpers.py::TestGetModelInfo`: **10 passed**.

### DIFF
--- a/ods/extensions/services/dashboard-api/helpers.py
+++ b/ods/extensions/services/dashboard-api/helpers.py
@@ -623,7 +623,14 @@ def get_model_info() -> Optional[ModelInfo]:
             model_name = env_values.get("LLM_MODEL")
             if model_name:
                 size_gb, quant = 15.0, None
-                context = int(env_values.get("MAX_CONTEXT") or env_values.get("CTX_SIZE") or 32768)
+                # MAX_CONTEXT/CTX_SIZE come straight from .env and may be
+                # non-numeric (e.g. "auto", "8k", or a trailing comment); fall
+                # back to the default rather than 500-ing every caller. Mirrors
+                # the guard already used in routers/models.py.
+                try:
+                    context = int(env_values.get("MAX_CONTEXT") or env_values.get("CTX_SIZE") or 32768)
+                except (TypeError, ValueError):
+                    context = 32768
 
                 import re as _re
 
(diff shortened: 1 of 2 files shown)


Lightheartdevs:
Thanks for hardening `get_model_info()` against malformed context values; keeping the dashboard status path from 500ing on user-edited `.env` is the right direction.

Quickest path to merge:
- In `ods/extensions/services/dashboard-api/helpers.py`, parse `MAX_CONTEXT` and `CTX_SIZE` as ordered candidates instead of one combined `int(...)`. The current patch no longer crashes, but `.env` with `MAX_CONTEXT=auto` and `CTX_SIZE=8192` will report the default `32768` rather than the valid configured `8192`; `routers/models.py` already handles this by continuing after an invalid key.
- In `ods/extensions/services/dashboard-api/tests/test_helpers.py`, add a regression case for `MAX_CONTEXT=auto` plus `CTX_SIZE=8192` expecting `context_length == 8192`, alongside the new numeric and fallback tests.

What I checked:
- `gh pr diff 1807`
- `gh pr checks 1807` (green, including dashboard API tests, Ruff, mypy, integration smoke, distro smoke, and env validation)
- Static comparison against `ods/extensions/services/dashboard-api/routers/models.py` context parsing for the same two keys.

Thanks again for pushing this forward; the remaining change should stay small.

PR #1809, opened 2026-07-15
fix(dashboard-api): don't crash vision stream on empty/null choices

w3-surfer:
## Problem

`_stream_vision_chat` (ODS Talk's image path) parses each streamed chunk as:

```python
delta = chunk.get("choices", [{}])[0].get("delta", {})
```

The `[{}]` default only kicks in when the `choices` key is **absent**. An OpenAI-style **usage-only chunk** carries an explicit empty list — `{"choices": [], "usage": {...}}` — so `[][0]` raises `IndexError`; a `{"choices": [null]}` element raises `AttributeError`. The surrounding handler only catches `httpx` errors:

```python
except (httpx.ReadTimeout, httpx.ConnectError, httpx.HTTPError) as exc:
```

so the exception escapes, the async generator dies mid-stream, and the image reply is truncated with **no** `error`/`done` frame (the client just hangs / shows a broken response).

## Fix

Resolve `choices` defensively before indexing — the same pattern `routers/models.py` already uses for streamed chunks:

```python
choices = chunk.get("choices") or []
delta = (choices[0] or {}).get("delta", {}) if choices else {}
```

Handles absent / empty / null-element `choices`; valid chunks are unchanged.

## Tests

`test_vision_stream_tolerates_empty_and_null_choices` drives the real `_stream_vision_chat` with a mocked httpx stream whose lines include `{"choices": []}`, a valid delta, and `{"choices": [null]}`. It **fails before** the fix (IndexError tears down the generator) and **passes after**, still emitting the `complete` frame and the valid text. Full `test_talk.py`: **35 passed**.

### DIFF
--- a/ods/extensions/services/dashboard-api/routers/talk.py
+++ b/ods/extensions/services/dashboard-api/routers/talk.py
@@ -168,7 +168,12 @@ async def _stream_vision_chat(image_bytes: bytes, content_type: str, prompt_text
                         chunk = json.loads(payload_str)
                     except json.JSONDecodeError:
                         continue
-                    delta = chunk.get("choices", [{}])[0].get("delta", {})
+                    # A streaming chunk may carry an empty (usage-only) or
+                    # null-element choices list; the [{}] default only covers a
+                    # missing key, so guard the index/element to avoid killing
+                    # the generator mid-stream (mirrors routers/models.py).
+                    choices = chunk.get("choices") or []
+                    delta = (choices[0] or {}).get("delta", {}) if choices else {}
                     text = delta.get("content")
                     if isinstance(text, str) and text:
                         accumulated.append(text)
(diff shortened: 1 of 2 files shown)


Lightheartdevs:
Thanks for hardening the Talk vision stream path; empty usage chunks are a common enough streaming shape that this guard is worth having.

What looks good:
- `ods/extensions/services/dashboard-api/routers/talk.py` now handles missing, empty, and null-element `choices` without changing valid delta handling.
- The regression test drives the real `_stream_vision_chat` generator through empty choices, a valid content delta, null choices, and `[DONE]`, which covers the failure mode directly.
- The pattern matches the defensive streamed-chunk handling already used elsewhere in the dashboard-api routers.

What I checked:
- PR diff/static review of `routers/talk.py` and `tests/test_talk.py`.
- Reported checks are green, including Dashboard API tests, Python lint, mypy, Test Linux, ShellCheck, Matrix Smoke, PowerShell lint, secret scan, and env-schema jobs.

Thanks again for the focused fix and test.

PR #1811, opened 2026-07-15
fix: bound health-check curls with --max-time across install/update/preflight

w3-surfer:
Several health-check probes call `curl` with no timeout, so a service that is **listening but wedged** (mid-restart, or a model still loading into memory) makes the request block indefinitely and hangs the surrounding step.

This consolidates the fix across the three scripts that had unbounded probes (originally split as #1811/#1812/#1813 — combined here to reduce review overhead):

| Script | Probe(s) | Timeout | Rationale |
|---|---|---|---|
| `installers/macos/install-macos.sh` | host-native poll in the bounded health loop (`MAX_ATTEMPTS=90`, ~180s) | `--max-time 10` | one attempt could otherwise outlast the whole budget; matches the same file's other `/health` probe (`curl -sf --max-time 10 ...:8080/health`) |
| `ods-update.sh` | 3 post-update health probes | `--max-time 15` | matches the value the file already uses on its release-check curls (lines 824/839) |
| `ods-preflight.sh` | 5 service probes | `CURL_HEALTH_FLAGS=(--connect-timeout 3 --max-time 10)` | mirrors the already-hardened sibling `scripts/ods-preflight.sh` |

Each timeout value matches the convention already established in its own file, consistent with the `curl --max-time` requirement enforced by `tests/test-network-timeouts.sh`.

## Verify

`bash -n` passes on all three scripts; after the change no health-check `curl` in these files is unbounded.


Lightheartdevs:
Thanks for fixing the macOS health-loop timeout; this keeps the installer retry budget meaningful when a native service accepts connections but is still loading.

What looks good:
- The change is narrowly scoped to the host-native health probe in `ods/installers/macos/install-macos.sh` and leaves the loop structure unchanged.
- `--max-time 10` matches the existing macOS installer probe for `127.0.0.1:8080/health`, so it follows the file’s established timeout convention.

What I checked:
- PR diff/static review and comparison with the existing `--max-time 10` curl calls in the same installer.

Thanks again for tightening this macOS install edge case.

Lightheartdevs:
Thanks for expanding the timeout fix across the macOS installer, update health check, and preflight probes; this closes the same hang class in all three places.

What looks good:
- `ods/installers/macos/install-macos.sh` now caps host-native health probes with `--max-time 10`, preserving the bounded retry loop when a native service accepts a connection but stalls.
- `ods/ods-preflight.sh` centralizes `CURL_HEALTH_FLAGS=(--connect-timeout 3 --max-time 10)` and uses it for every service probe, matching the repo's timeout-test pattern.
- `ods/ods-update.sh` adds `--max-time 15` to Dashboard API and llama-server health checks, consistent with the script's existing GitHub release curl timeout.

What I checked:
- Reviewed the diff around the macOS health loop, `ods-preflight.sh` service probes, and `cmd_health()` in `ods-update.sh`.
- Ran `git diff --check origin/main...HEAD`, `bash -n ods/installers/macos/install-macos.sh ods/ods-preflight.sh ods/ods-update.sh`, and `bash ods/tests/test-network-timeouts.sh`; the timeout suite passed 17/17 with expected skips.
- `gh pr checks 1811 --required=false` is green across ShellCheck, Test Linux, Matrix Smoke, Dashboard, Python lint, PowerShell lint, secret scan, and env-schema jobs.

Thanks again for making these operational checks less fragile.

PR #1812, opened 2026-07-15
fix(update): bound post-update health-check curls with --max-time

w3-surfer:
## Problem

After an update, `ods-update.sh` verifies services with bare, unbounded curls:

```sh
if   curl -sf "http://127.0.0.1:${dashboard_api_port}/health"     &>/dev/null; then   # 909
elif curl -sf "http://127.0.0.1:${dashboard_api_port}/api/status" &>/dev/null; then   # 911
...
if   curl -sf "http://127.0.0.1:${llama_server_port}/v1/models"   &>/dev/null; then   # 922
```

None carries `--max-time`, so a service that is listening but wedged (e.g. mid-restart, or a model still loading) makes the request block with no ceiling and hangs the tail of `ods update`.

## Fix

Add `--max-time 15` to each probe — the value `ods-update.sh` **already** uses on its other curls (the release-check calls at lines 824 and 839), so the file stays internally consistent:

```diff
-    if curl -sf "http://127.0.0.1:${dashboard_api_port}/health" &>/dev/null; then
+    if curl -sf --max-time 15 "http://127.0.0.1:${dashboard_api_port}/health" &>/dev/null; then
```

(applied to all three health probes)

## Verify

`bash -n ods-update.sh` passes; the three probes are now bounded.

### DIFF
--- a/ods/ods-update.sh
+++ b/ods/ods-update.sh
@@ -906,9 +906,9 @@ cmd_health() {
     local dashboard_api_port="${DASHBOARD_API_PORT:-}"
     [[ -n "$dashboard_api_port" ]] || dashboard_api_port="$(env_file_value DASHBOARD_API_PORT)"
     dashboard_api_port="${dashboard_api_port:-3002}"
-    if curl -sf "http://127.0.0.1:${dashboard_api_port}/health" &>/dev/null; then
+    if curl -sf --max-time 15 "http://127.0.0.1:${dashboard_api_port}/health" &>/dev/null; then
         log_ok "Dashboard API: healthy"
-    elif curl -sf "http://127.0.0.1:${dashboard_api_port}/api/status" &>/dev/null; then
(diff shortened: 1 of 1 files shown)


Lightheartdevs:
Thanks for tightening the post-update health checks; this is a small change in a high-friction failure path.

What looks good:
- The diff only adds `--max-time 15` to the three `cmd_health` probes in `ods/ods-update.sh`, preserving the existing dashboard-api and llama-server endpoint order.
- The timeout value matches the release-check curls already used elsewhere in the same script, so the behavior is internally consistent.

What I checked:
- PR diff/static review of `ods/ods-update.sh` health verification.

Thanks again for keeping the update path from hanging on wedged services.

w3-surfer:
Consolidated into #1811 — combining the related `curl --max-time` health-probe fixes (install/update/preflight) into a single PR to reduce review overhead. This exact change (the ods-update.sh post-update probes) is included there.

PR #1813, opened 2026-07-15
fix(preflight): bound ods-preflight.sh health probes with curl timeouts

w3-surfer:
## Problem

The user-facing `./ods-preflight.sh` (documented in `QUICKSTART.md`, suggested by `ods-restore.sh` after a restore) probes every service — LLM, Whisper, TTS, embeddings, dashboard — with a bare, unbounded curl:

```sh
for ENDPOINT in "${LLM_ENDPOINTS[@]}"; do
    if curl -sf "$ENDPOINT" &> /dev/null; then          # + /health variants for the others
```

None of the five probes carries a timeout, so a service that is **listening but wedged** makes the request block with no ceiling and hangs the entire preflight.

The identically-purposed `scripts/ods-preflight.sh` already does this correctly:

```sh
CURL_HEALTH_FLAGS=(--connect-timeout 3 --max-time 10)
if curl -sf "${CURL_HEALTH_FLAGS[@]}" "http://127.0.0.1:${LLM_PORT}${LLM_HEALTH}" ...
```

## Fix

Mirror that sibling — define the same `CURL_HEALTH_FLAGS` array once and apply it to all five probes:

```sh
CURL_HEALTH_FLAGS=(--connect-timeout 3 --max-time 10)
...
if curl -sf "${CURL_HEALTH_FLAGS[@]}" "$ENDPOINT" &> /dev/null; then
```

Indexed arrays are already used throughout this script (`LLM_ENDPOINTS=(...)`), so it stays Bash-3.2-safe.

## Verify

- `bash -n ods-preflight.sh` passes.
- All five health probes now carry the timeout flags; no unbounded `curl -sf "$ENDPOINT"` remains.

### DIFF
--- a/ods/ods-preflight.sh
+++ b/ods/ods-preflight.sh
@@ -24,6 +24,10 @@ load_env_file "$ODS_DIR/.env"
 
 SERVICE_HOST="${SERVICE_HOST:-localhost}"
 
+# Bound every health probe so a listening-but-wedged service can't hang the
+# preflight (mirrors scripts/ods-preflight.sh).
+CURL_HEALTH_FLAGS=(--connect-timeout 3 --max-time 10)
+
 # Auto-detect backend from .env or hardware probing.
 # Priority: .env setting → nvidia-smi → AMD sysfs (any card).
 # On dual-GPU systems (AMD iGPU + NVIDIA dGPU) we must prefer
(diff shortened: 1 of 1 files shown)


Lightheartdevs:
Thanks for bounding the user-facing preflight probes here; this matches the timeout posture the repo already uses in the sibling script.

What looks good:
- `ods/ods-preflight.sh` is Bash and already uses indexed arrays, so centralizing `CURL_HEALTH_FLAGS=(--connect-timeout 3 --max-time 10)` fits the script style and stays Bash-3.2-safe.
- The same flags are applied consistently to the LLM, Whisper, TTS, embeddings, and dashboard probes, eliminating the unbounded `curl -sf $ENDPOINT` paths without changing endpoint selection.

What I checked:
- PR diff/static review, plus comparison with `ods/scripts/ods-preflight.sh`, which already centralizes the same timeout flags.

Thanks again for pushing this reliability fix forward.

w3-surfer:
Consolidated into #1811 — combining the related `curl --max-time` health-probe fixes (install/update/preflight) into a single PR to reduce review overhead. This exact change (the ods-preflight.sh probes) is included there.

PR #1814, opened 2026-07-15
docs: correct two commands that fail as written

w3-surfer:
Two documented commands error out when run verbatim (originally split as #1814/#1815 — combined here to reduce review overhead):

**1. `docs/VALIDATION_REPRODUCIBILITY.md`** — the "Minimum Local Audit Set" block calls:
```
python scripts/validate-dependency-pins.py
```
That script doesn't exist (`No such file or directory`). The real one is `scripts/check-dependency-pins.py` — *"Validate pinned runtime dependency references"* — invoked by `scripts/release-gate.sh:27`. The block's three other scripts all exist; only this name is wrong.

**2. `README.md`** — the "Management scripts" block lists:
```
ods mode status                        # Show current mode
```
But `ods-cli` `cmd_mode()` only accepts `local|cloud|hybrid`; `status` prints `✗ Unknown mode: status. Use: local, cloud, hybrid` and exits 1. The current mode is shown by **bare** `ods mode` — which README already documents correctly a few lines up.

Both are one-token fixes to runnable commands.

### DIFF
--- a/ods/README.md
+++ b/ods/README.md
@@ -393,7 +393,7 @@ ods start                      # Start everything
 # Management scripts
 ./scripts/session-cleanup.sh             # Clean up bloated agent sessions
 ./scripts/llm-cold-storage.sh --status   # Check model hot/cold storage
-ods mode status                        # Show current mode
+ods mode                               # Show current mode
 ```
 
 ## Comparison
--- a/ods/docs/VALIDATION_REPRODUCIBILITY.md
+++ b/ods/docs/VALIDATION_REPRODUCIBILITY.md
@@ -30,7 +30,7 @@ git diff --check
 python scripts/audit-extensions.py --project-dir .
 python scripts/validate-generated-configs.py
 python scripts/validate-golden-paths.py
-python scripts/validate-dependency-pins.py
+python scripts/check-dependency-pins.py
 ```
 
 Add dashboard checks when dashboard code or dashboard-api behavior changes.


Lightheartdevs:
Thanks for fixing the local audit command name; this is the kind of docs correction that saves reviewer time during validation.

What looks good:
- `ods/docs/VALIDATION_REPRODUCIBILITY.md` now points at `scripts/check-dependency-pins.py`, which is the script present in the repo and the one invoked by `ods/scripts/release-gate.sh`.
- The diff is docs-only and leaves the rest of the audit block unchanged.

What I checked:
- PR diff/static review and `git ls-tree`/`git grep` confirmation that `check-dependency-pins.py` exists while `validate-dependency-pins.py` does not.

Thanks again for the precise fix.

Lightheartdevs:
Thanks for folding the README command fix into the dependency-pin docs correction; both are the kind of runnable-doc cleanup that saves validation time.

What looks good:
- `ods/docs/VALIDATION_REPRODUCIBILITY.md` now points at `scripts/check-dependency-pins.py`, matching the script in the tree and the command used by `ods/scripts/release-gate.sh`.
- `ods/README.md` now uses bare `ods mode` for showing current mode, which matches `cmd_mode()` and the CLI help; the stale `ods mode status` form is gone.
- The diff is docs-only and scoped to two command examples.

What I checked:
- Reviewed the two-line docs diff against `ods/ods-cli` and `ods/scripts/release-gate.sh`.
- Ran `git diff --check origin/main...HEAD`, confirmed `check-dependency-pins.py` exists while `validate-dependency-pins.py` does not, and confirmed no `ods mode status`/`validate-dependency-pins` references remain in the touched docs paths.
- `gh pr checks 1814 --required=false` is green across ShellCheck, Test Linux, Matrix Smoke, Dashboard, Python lint, PowerShell lint, secret scan, and env-schema jobs.

Thanks again for tightening the validation docs.

Lightheartdevs (approved):
Audit: both corrected commands verified against the tree/CLI (validate-dependency-pins.py absent → check-dependency-pins.py runs bare in release-gate; 'ods mode status' errors, bare 'ods mode' is correct). Docs only.

PR #1815, opened 2026-07-15
docs: fix `ods mode status` in README management commands

w3-surfer:
## Problem

README's **Management scripts** block includes:

```bash
ods mode status                        # Show current mode
```

But `ods mode` only accepts a mode argument — running `ods mode status` fails:

```
✗ Unknown mode: status. Use: local, cloud, hybrid
```

(`cmd_mode()` in `ods-cli`: `case "$mode" in local|cloud|hybrid) ;; *) error "Unknown mode: $mode. Use: local, cloud, hybrid" ;;`, and `error()` exits 1.)

The current mode is shown by **bare `ods mode`** — which this same README already documents correctly a few lines up (`ods mode  # Show current mode (local/cloud/hybrid)`).

## Fix

```diff
-ods mode status                        # Show current mode
+ods mode                               # Show current mode
```

### DIFF
--- a/ods/README.md
+++ b/ods/README.md
@@ -388,7 +388,7 @@ ods start                      # Start everything
 # Management scripts
 ./scripts/session-cleanup.sh             # Clean up bloated agent sessions
 ./scripts/llm-cold-storage.sh --status   # Check model hot/cold storage
-ods mode status                        # Show current mode
+ods mode                               # Show current mode
 ```
 
 ## Comparison


Lightheartdevs:
Thanks for tightening the README command here; this removes a copy-paste trap from the management examples.

What looks good:
- The one-line docs change matches the actual `ods mode` behavior: `ods/ods-cli` treats only `local`, `cloud`, and `hybrid` as mode arguments and bare `ods mode` is the status display path.
- The fix stays scoped to `ods/README.md` and aligns with the nearby README example that already documents bare `ods mode`.

What I checked:
- PR diff/static review, plus `git grep` against `ods/ods-cli` for the `cmd_mode` contract.

w3-surfer:
Consolidated into #1814 — combining the two 'documented command fails as written' fixes into a single docs PR to reduce review overhead. The `ods mode status` → `ods mode` fix is included there.

PR #1817, opened 2026-07-15
fix(windows): read GPU VRAM as VramMB in the install report

w3-surfer:
## Problem

`Write-ODSInstallReport` builds the report's GPU block from `Get-GpuInfo`'s result:

```powershell
gpu = [ordered]@{
    backend = $gpu.Backend
    name = $gpu.Name
    vram_mb = $gpu.VRAM        # <-- wrong key
    memory_type = $gpu.MemoryType
```

But `Get-GpuInfo` (`detection.ps1`) returns the field as **`VramMB`**, not `VRAM`. `install-report.ps1:110` is the **only** place in the tree that reads `.VRAM`; every other consumer (`tier-map.ps1`, `phases/02-detection.ps1`) uses `.VramMB`. Accessing the nonexistent key yields `$null`, so `ods report` always produces:

- `report.json` → `"gpu": { "vram_mb": null, ... }`
- `report.txt` → `- VRAM:  MB`

even on a GPU host that detected VRAM correctly.

## Fix

```diff
-            vram_mb = $gpu.VRAM
+            vram_mb = $gpu.VramMB
```

The adjacent fields (`$gpu.Backend`, `$gpu.Name`, `$gpu.MemoryType`) already use the correct keys from the same hashtable, confirming the intended style.

## Verify

- `Get-GpuInfo` sets `VramMB` at `detection.ps1:52/120/140` (and documents it at :20); it never sets `VRAM`.
- After the change, `grep -rn '\.VRAM\b' installers/windows/` returns nothing — the only reader is fixed, and `report.txt`'s `- VRAM: $($report.gpu.vram_mb) MB` now receives a real number.

### DIFF
--- a/ods/installers/windows/lib/install-report.ps1
+++ b/ods/installers/windows/lib/install-report.ps1
@@ -107,7 +107,7 @@ function New-ODSInstallReport {
         gpu = [ordered]@{
             backend = $gpu.Backend
             name = $gpu.Name
-            vram_mb = $gpu.VRAM
+            vram_mb = $gpu.VramMB
             memory_type = $gpu.MemoryType
             native_inference_backend = $nativeBackend
         }


Lightheartdevs:
Thanks for tightening up the Windows install report path here; this is exactly the kind of small diagnostic fix that makes support bundles more useful.

What looks good:
- `ods/installers/windows/lib/detection.ps1` documents and returns `VramMB`, and the other Windows consumers (`phases/02-detection.ps1`, `lib/tier-map.ps1`) already read `.VramMB`; this fixes the only remaining `.VRAM` reader.
- The change is limited to report serialization/output, so it does not alter GPU detection, tier selection, compose flags, or install behavior.
- Current checks are green, including PowerShell lint on both Ubuntu and Windows plus the broader shell, Python, dashboard, integration, distro, and env-validation checks.

What I checked:
- `gh pr diff 1817`
- `rg -n VramMB|\.VRAM\b ods/installers/windows -g '*.ps1'`
- `gh pr checks 1817`

Thanks again for catching and isolating this.

Lightheartdevs (approved):
Audit: verified against merged #1894 — Get-GpuInfo returns VramMB (report was reading $gpu.VRAM, a nonexistent key → null vram_mb). Correct one-line alignment; units (MB) match the vram_mb field.

PR #1820, opened 2026-07-15
fix(dashboard-api): skip empty keys when parsing .env files

Ovatsugh:
## Summary
- Reject empty keys when parsing `.env` files in `get_model_info()`. A malformed line like `=value` or bare `=` passes the `"=" in line` check but inserts an empty-string key into the env dict.

## Changes
- `ods/extensions/services/dashboard-api/helpers.py`: add `if not key.strip(): continue` after `line.split("=", 1)`

## Testing
- [ ] `make lint`


Lightheartdevs:
Thanks for hardening the `.env` parsing path around malformed empty-key lines.

What looks good:
- The change in `ods/extensions/services/dashboard-api/helpers.py` skips only keys that are empty after stripping, so normal `KEY=value` parsing, quote stripping, and `LLM_MODEL` discovery stay unchanged.
- The behavior is scoped to `get_model_info()`, which is a read-only dashboard status helper and should degrade rather than fail on malformed local config.

What I checked:
- `python -m py_compile ods/extensions/services/dashboard-api/helpers.py`.
- `python -m pytest tests/test_helpers.py::TestGetModelInfo -q` (7 passed).

PR #1822, opened 2026-07-15
fix(dashboard-api): guard Prometheus metric parsing against malformed lines

Ovatsugh:
## Summary
- Wrap Prometheus metric extraction (`tokens_predicted_total`, `tokens_predicted_seconds_total`) in `try/except (ValueError, IndexError)`. Previously, `line.split()[-1]` would throw `IndexError` on empty/whitespace-only lines matching the keyword, and `float()` would throw `ValueError` on non-numeric content. Both were uncaught since the outer handler only covers `httpx` and `OSError`.

## Changes
- `ods/extensions/services/dashboard-api/helpers.py`: add per-metric `try/except` guards

## Testing
- [ ] `make lint`

### DIFF
--- a/ods/extensions/services/dashboard-api/helpers.py
+++ b/ods/extensions/services/dashboard-api/helpers.py
@@ -365,9 +365,15 @@ async def get_llama_metrics(model_hint: Optional[str] = None) -> dict:
             if line.startswith("#"):
                 continue
             if "tokens_predicted_total" in line:
-                metrics["tokens_predicted_total"] = float(line.split()[-1])
+                try:
+                    metrics["tokens_predicted_total"] = float(line.split()[-1])
+                except (ValueError, IndexError):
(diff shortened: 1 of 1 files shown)


Lightheartdevs:
Thanks for making the llama metrics parser more tolerant of malformed Prometheus lines.

What looks good:
- `ods/extensions/services/dashboard-api/helpers.py` now treats bad/missing metric values as absent samples instead of failing `get_llama_metrics()` entirely.
- The existing fallback path still returns the previous lifetime-token value and `0` TPS on broader HTTP/runtime failures, so this fits the dashboard metrics degrade behavior.

What I checked:
- `python -m py_compile ods/extensions/services/dashboard-api/helpers.py`.
- `python -m pytest tests/test_helpers.py::TestGetLlamaMetrics -q` (2 passed).

PR #1823, opened 2026-07-15
fix(dashboard-api): handle non-dict response in model error check

Ovatsugh:
## Summary
- Add `isinstance(data, dict)` check before calling `data.get("error")` in `_probe_loaded_lemonade_model()`. If the Lemonade API returns a JSON list or primitive instead of a dict, `.get()` raises `AttributeError`.

## Changes
- `ods/extensions/services/dashboard-api/routers/models.py`: add type guard before `.get()` call

## Testing
- [ ] `make lint`

### DIFF
--- a/ods/extensions/services/dashboard-api/routers/models.py
+++ b/ods/extensions/services/dashboard-api/routers/models.py
@@ -187,6 +187,8 @@ async def _probe_loaded_lemonade_model(model_name: str) -> bool:
         resp = await client.post(f"{base_url}/api/v1/chat/completions", json=payload, headers=headers)
         resp.raise_for_status()
         data = resp.json()
+        if not isinstance(data, dict):
+            return False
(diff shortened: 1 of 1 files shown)


Lightheartdevs:
Thanks for closing the non-dict Lemonade response path in the model readiness probe.

What looks good:
- `ods/extensions/services/dashboard-api/routers/models.py` now rejects non-object chat-completion payloads before calling `.get()`, preserving the existing ?backend not ready? return path instead of throwing.
- The change stays inside `_probe_loaded_lemonade_model()`, so it does not broaden the model activation contract or alter non-Lemonade readiness.

What I checked:
- `python -m py_compile ods/extensions/services/dashboard-api/routers/models.py`.
- `python -m pytest tests/test_models.py -q` (18 passed).

PR #1825, opened 2026-07-15
fix(config): use 127.0.0.1 instead of localhost in backend health URLs

Ovatsugh:
## Summary
- Replace `localhost` with `127.0.0.1` in all backend health URLs. On macOS, `localhost` can resolve to `::1` (IPv6) causing health checks to hang. The project's own lint rules already guard against this pattern in shell scripts.

## Changes
- `ods/config/backends/nvidia.json`: `localhost` → `127.0.0.1`
- `ods/config/backends/cpu.json`: same
- `ods/config/backends/apple.json`: same
- `ods/config/backends/amd.json`: same (health path `/api/v1/health` preserved)

## Testing
- [ ] `make test`
- [ ] Verify health checks pass on macOS

### DIFF
--- a/ods/config/backends/amd.json
+++ b/ods/config/backends/amd.json
@@ -3,7 +3,7 @@
   "llm_engine": "lemonade",
   "service_name": "llama-server",
   "public_api_port": 8080,
-  "public_health_url": "http://localhost:8080/api/v1/health",
+  "public_health_url": "http://127.0.0.1:8080/api/v1/health",
   "provider_name": "local-lemonade",
   "provider_url": "http://llama-server:8080/api/v1",
   "runtime": {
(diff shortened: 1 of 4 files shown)


Lightheartdevs:
Thanks for taking on the backend health URL `localhost` ambiguity.

What looks good:
- The PR changes only host-facing `public_health_url` values in `ods/config/backends/*.json` to `127.0.0.1`, while leaving container-facing `provider_url` values on service DNS (`llama-server:8080`). That separation fits the backend contract use in installer detection.
- The AMD `/api/v1/health` path and the llama-server `/health` paths are preserved, so the change is about address resolution rather than endpoint semantics.

What I checked:
- `python -m json.tool` on all four changed backend JSON files.

Lightheartdevs (approved):
Audit: all four backend health URLs are loopback probes consumed via load-backend-contract.sh; 127.0.0.1 is strictly more robust than localhost (avoids ::1 when the service binds IPv4-only). Valid JSON post-change. Follow-up (not blocking): Windows PS health probes (install-windows.ps1, llm-endpoint.ps1, ods.ps1) still use localhost — same latent issue, separate scope.

PR #1826, opened 2026-07-15
fix(dashboard-api): stop leaking system details in error responses

Ovatsugh:
## Summary
- Remove system-level error details (`str(exc)`) from HTTP error responses sent to clients. Internal paths, hostnames, and OS error messages were being exposed via the `reason` field in two settings endpoints.

## Changes
- `ods/extensions/services/dashboard-api/main.py`:
  - `api_settings_env_save`: remove `"reason": str(exc)` from 500 response, add `logger.error(...)` instead
  - `api_settings_env_apply`: replace f-string embedding `{exc}` with generic message (existing `logger.exception()` already captures details)

## Testing
- [ ] `make lint`

### DIFF
--- a/ods/extensions/services/dashboard-api/main.py
+++ b/ods/extensions/services/dashboard-api/main.py
@@ -1522,9 +1522,10 @@ async def api_settings_env_save(
             detail={"message": "ODS host agent is not reachable. Start the host agent, then try again."},
         ) from exc
     except OSError as exc:
+        logger.error("Failed to contact host agent for env update: %s", exc)
         raise HTTPException(
             status_code=500,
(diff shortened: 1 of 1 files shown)


Lightheartdevs:
Thanks for reducing what the settings endpoints disclose on host-agent failures.

What looks good:
- `ods/extensions/services/dashboard-api/main.py` no longer returns raw `OSError` text to clients from the env-save/apply paths, which avoids leaking local paths or OS-level details through API responses.
- The internal diagnostics are still preserved through `logger.error()` / existing `logger.exception()`, so operators retain server-side evidence without exposing it in the response body.

What I checked:
- `python -m py_compile ods/extensions/services/dashboard-api/main.py`.
- `python -m pytest tests/test_settings_env.py -q` (27 passed).

PR #1830, opened 2026-07-16
fix(installer): replace GNU-only df -BG with POSIX-compatible df -Pk

Ovatsugh:
## Summary
- Replace `df -BG` (GNU coreutils extension) with `df -Pk` in phase 02-detection. `df -BG` fails on macOS/BSD systems. The rest of the project already uses `df -Pk` (e.g., `path-utils.sh`).

## Changes
- `ods/installers/phases/02-detection.sh`: two call sites updated — CPU-only fallback (line 65) and main disk probe (line 126)

## Testing
- [ ] `bash -n ods/installers/phases/02-detection.sh`
- [ ] `make lint`
- [ ] Verify on macOS: `df -Pk "$HOME" | tail -1 | awk '{printf "%d", $4 / 1048576}'`

### DIFF
--- a/ods/installers/phases/02-detection.sh
+++ b/ods/installers/phases/02-detection.sh
@@ -62,7 +62,7 @@ if [[ "${ODS_MODE:-local}" == "cloud" ]]; then
         RAM_KB=$(grep MemTotal /proc/meminfo | awk '{print $2}')
     fi
     RAM_GB=$((RAM_KB / 1024 / 1024))
-    DISK_AVAIL=$(df -BG "$HOME" | tail -1 | awk '{print $4}' | tr -d 'G')
+    DISK_AVAIL=$(df -Pk "$HOME" 2>/dev/null | tail -1 | awk '{printf "%d", $4 / 1048576}')
     BACKEND_ID="cpu"
     LLM_HEALTHCHECK_URL="http://localhost:4000/health/readiness"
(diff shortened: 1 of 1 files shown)


Lightheartdevs:
Thanks for tackling BSD/macOS portability in phase 02 disk detection.

Quickest path to merge:
- Update `ods/tests/test-disk-check-install-dir.sh` to mirror the new `df -Pk ... awk '{printf "%d", $4 / 1048576}'` logic instead of still stubbing/calling `df -BG`. That test says it mirrors the exact disk-probe logic from `02-detection.sh`, so leaving the GNU-only flag there drops coverage for the portability change.
- Consider asserting that the stubbed `df` receives `-Pk`, so future changes cannot silently reintroduce `-BG` in this path.
- Re-run the disk-check regression plus `bash -n ods/installers/phases/02-detection.sh` in a bash-capable environment.

What I checked:
- PR diff, `ods/installers/lib/path-utils.sh` POSIX `df -Pk` convention, and the dedicated disk-check regression test.
- `git grep` on `origin/pr/1830` confirmed `02-detection.sh` changed to `df -Pk` while `test-disk-check-install-dir.sh` still uses `df -BG`.
- Existing `gh pr checks` output: no checks reported for this head; local bash checks could not run because `bash` resolves to a broken WSL shim here.

Thanks again for pushing the installer toward portable shell behavior.

PR #1831, opened 2026-07-16
fix(installer): use TMPDIR fallback instead of hardcoded /tmp

Ovatsugh:
## Summary
- Replace hardcoded `/tmp` with `${TMPDIR:-/tmp}` in installer constants. Some container environments or non-standard setups may not have `/tmp` writable or available.

## Changes
- `ods/installers/lib/constants.sh`: update `LOG_FILE`, `CAPABILITY_PROFILE_FILE`, `PREFLIGHT_REPORT_FILE` defaults

## Testing
- [ ] `bash -n ods/installers/lib/constants.sh`
- [ ] `make lint`

### DIFF
--- a/ods/installers/lib/constants.sh
+++ b/ods/installers/lib/constants.sh
@@ -26,9 +26,9 @@ else
     INSTALL_DIR="${INSTALL_DIR:-$HOME/ods}"
 fi
 
-LOG_FILE="${LOG_FILE:-/tmp/ods-install.log}"
-CAPABILITY_PROFILE_FILE="${CAPABILITY_PROFILE_FILE:-/tmp/ods-capabilities.json}"
-PREFLIGHT_REPORT_FILE="${PREFLIGHT_REPORT_FILE:-/tmp/ods-preflight-report.json}"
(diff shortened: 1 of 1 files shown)


Lightheartdevs:
Thanks for moving the installer default temp artifacts off hardcoded `/tmp`.

What looks good:
- `LOG_FILE`, `CAPABILITY_PROFILE_FILE`, and `PREFLIGHT_REPORT_FILE` still respect explicit environment overrides, and their default path now follows the `${TMPDIR:-/tmp}` pattern already used in nearby installer code.
- The change is limited to default artifact locations in `ods/installers/lib/constants.sh`; it does not alter install directory resolution or generated runtime config.

What I checked:
- Static review of the PR diff and surrounding constants initialization.
- Compared with existing `${TMPDIR:-/tmp}` use in service-registry and phase 07 temporary service files.
- `gh pr checks` reported no checks for this head at review time; local bash syntax checks could not run because `bash` resolves to a broken WSL shim here.

Thanks again for improving portability without broadening the change.

PR #1832, opened 2026-07-16
fix(qdrant): add missing CPU resource limit

Ovatsugh:
## Summary
- Qdrant was the only extension service with a memory limit but no CPU limit. Under load, a runaway vector search could starve other services on the host.

## Changes
- `ods/extensions/services/qdrant/compose.yaml`: add `cpus: '2.0'` alongside existing `memory: 1G`

## Testing
- [ ] `make test` (compose validation)
- [ ] `docker compose config` with Qdrant enabled

### DIFF
--- a/ods/extensions/services/qdrant/compose.yaml
+++ b/ods/extensions/services/qdrant/compose.yaml
@@ -22,6 +22,7 @@ services:
     deploy:
       resources:
         limits:
+          cpus: '2.0'
           memory: 1G
     logging:
       driver: "json-file"


Lightheartdevs:
Thanks for adding a CPU cap to the Qdrant service.

What looks good:
- `ods/extensions/services/qdrant/compose.yaml` now has a `deploy.resources.limits.cpus: '2.0'` entry alongside the existing `memory: 1G`, which matches the resource-limit style used by other built-in extension services.
- The patch does not change Qdrant ports, volumes, health checks, image, or service identity, so the runtime blast radius is narrow.

What I checked:
- Static review of the PR diff and the full Qdrant compose file.
- Compared resource-limit conventions across other extension compose files.

Thanks again for keeping the RAG service from crowding the rest of the stack.

PR #1833, opened 2026-07-16
fix(tests): assert WorkingDirectory in opencode-web plist contract

Ovatsugh:
## Summary
- The opencode-web plist section extracts `workdir_val` (WorkingDirectory) but never asserts on it, unlike the ods-host-agent section which does. Adds the missing assertion to verify WorkingDirectory uses `${INSTALL_DIR}`.

## Changes
- `ods/tests/contracts/test-plist-log-paths.sh`: add `assert_contains` for opencode-web WorkingDirectory

## Testing
- [ ] `bash ods/tests/contracts/test-plist-log-paths.sh`

### DIFF
--- a/ods/installers/macos/install-macos.sh
+++ b/ods/installers/macos/install-macos.sh
@@ -1813,7 +1813,7 @@ OPENCODE_EOF
         <string>127.0.0.1</string>
     </array>
     <key>WorkingDirectory</key>
-    <string>${HOME}</string>
+    <string>${INSTALL_DIR}</string>
     <key>EnvironmentVariables</key>
     <dict>
         <key>HOME</key>
--- a/ods/tests/contracts/test-plist-log-paths.sh
+++ b/ods/tests/contracts/test-plist-log-paths.sh
@@ -69,6 +69,7 @@ assert_contains     "StandardOutPath uses HOME/Library/Logs/ODS"   "$stdout_val"
 assert_contains     "StandardErrorPath uses HOME/Library/Logs/ODS" "$stderr_val"  '${HOME}/Library/Logs/ODS/'
 assert_not_contains "StandardOutPath does not use INSTALL_DIR"             "$stdout_val"  'INSTALL_DIR'
 assert_not_contains "StandardErrorPath does not use INSTALL_DIR"           "$stderr_val"  'INSTALL_DIR'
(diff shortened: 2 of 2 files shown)


Lightheartdevs:
Thanks for trying to close the missing plist assertion.

Quickest path to merge:
- Reconcile the new assertion with the actual macOS installer heredoc. `ods/installers/macos/install-macos.sh` currently emits the opencode-web `WorkingDirectory` as `${HOME}`, while this PR adds an assertion that it contains `${INSTALL_DIR}`. As written, the contract test should fail.
- If opencode-web really should run from `${INSTALL_DIR}`, update the installer heredoc in the same PR and explain why that is safe for launchd/opencode. If `${HOME}` is intentional, change the new assertion to protect `${HOME}` instead and name that behavior clearly.
- Run `bash ods/tests/contracts/test-plist-log-paths.sh` in a bash-capable environment after the expectation and implementation agree.

What I checked:
- PR diff plus the opencode-web and ods-host-agent plist heredocs in `ods/installers/macos/install-macos.sh`.
- I could not run the contract test locally because `bash` resolves to a broken WSL shim in this Windows environment.

Thanks again for adding coverage here; it just needs to match the shipped plist behavior.

Lightheartdevs:
Thanks for following through on the opencode-web launchd contract and aligning the installer with the new assertion.

What looks good:
- `ods/installers/macos/install-macos.sh` now emits opencode-web `WorkingDirectory` as `${INSTALL_DIR}`, matching the new contract assertion instead of leaving the test ahead of implementation.
- The log-path safety from issue #341 is preserved: both opencode-web log paths still stay under `${HOME}/Library/Logs/ODS/` and the test guards against drifting back to `INSTALL_DIR` for logs.
- The change is tightly scoped to the macOS plist heredoc plus its existing contract test.

What I checked:
- `gh pr diff 1833` and surrounding macOS launchd plist context for opencode-web and ods-host-agent.
- `C:\Program Files\Git\bin\bash.exe -lc "cd .../.codex-pr-1833/ods && bash tests/contracts/test-plist-log-paths.sh"` -> 10 passed, 0 failed.
- `git diff --check origin/main...HEAD` -> clean.
- `gh pr checks 1833` -> no checks reported for this branch.

Thanks again for closing the loop on this one.

PR #1834, opened 2026-07-16
fix(dashboard): add aria-label to troubleshooting search input

Ovatsugh:
## Summary
- Add `aria-label="Search troubleshooting issues"` to the search input in `TroubleshootingAssistant`. Screen readers announce `placeholder` text inconsistently, so an explicit `aria-label` ensures the field is always accessible.

## Changes
- `ods/extensions/services/dashboard/src/components/TroubleshootingAssistant.jsx`

## Testing
- [ ] `cd ods/extensions/services/dashboard && npm run lint && npm run build`

### DIFF
--- a/ods/extensions/services/dashboard/src/components/TroubleshootingAssistant.jsx
+++ b/ods/extensions/services/dashboard/src/components/TroubleshootingAssistant.jsx
@@ -142,6 +142,7 @@ export function TroubleshootingAssistant({ serviceStatus }) {
       <input
         type="text"
         placeholder="Search issues..."
+        aria-label="Search troubleshooting issues"
         value={search}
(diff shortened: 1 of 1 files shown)


Lightheartdevs:
Thanks for tightening the troubleshooting search accessibility.

What looks good:
- The explicit `aria-label` names the input without changing its current value, placeholder, styling, or filtering behavior.
- The scope is a single JSX attribute in `TroubleshootingAssistant.jsx`, so the risk is limited to improving screen-reader semantics for an existing control.

What I checked:
- Static review of the PR diff and surrounding search input code.
- `git grep` on the PR head to confirm the label is present at the intended input.

Thanks again for improving this path.

PR #1836, opened 2026-07-16
fix(dashboard-api): catch JSONDecodeError in host-agent response parsing

Ovatsugh:
## Summary
- `_post_agent_json()` in resources.py did not catch `json.JSONDecodeError`. A malformed or empty response from the host agent would crash with an unhandled 500 instead of returning an empty dict gracefully.
- Also removes `{exc}` from the `OSError` detail to stop leaking internal error details to the client.

## Changes
- `ods/extensions/services/dashboard-api/routers/resources.py`: add `json.JSONDecodeError` handler, log warning server-side, return `{}` as fallback. Replace f-string detail with generic message.

## Testing
- [ ] `make lint`

### DIFF
--- a/ods/extensions/services/dashboard-api/routers/resources.py
+++ b/ods/extensions/services/dashboard-api/routers/resources.py
@@ -88,7 +88,11 @@ def _post_agent_json(path: str, body: dict, timeout: int = 65) -> dict:
     )
     try:
         with urllib.request.urlopen(req, timeout=timeout) as resp:
-            return json.loads(resp.read().decode() or "{}")
+            raw = resp.read().decode()
+            return json.loads(raw) if raw else {}
+    except json.JSONDecodeError:
+        logger.warning("Host agent returned invalid JSON for %s", path)
+        return {}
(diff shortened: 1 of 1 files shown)


Lightheartdevs:
Thanks for hardening the host-agent JSON parsing and removing the raw exception detail from client responses.

Quickest path to merge:
- Add focused coverage in `ods/extensions/services/dashboard-api/tests/test_resources.py` for `_post_agent_json`: an invalid JSON response should return `{}` instead of raising, and an empty body should still return `{}`.
- Add a small assertion for the `URLError`/`OSError` path that the API response detail is the generic `Host agent unavailable`, not the underlying socket/path error.
- Keep the server-side warning log; that gives operators diagnostics without leaking host details to dashboard clients.

What I checked:
- PR diff and surrounding `routers/resources.py` code.
- Existing `test_resources.py` coverage; it exercises `_fetch_container_stats` and restart endpoints, but not `_post_agent_json` itself.

Thanks again for closing down this failure mode.

PR #1837, opened 2026-07-16
fix(installer): use TMPDIR in mktemp calls for portability

Ovatsugh:
## Summary
- Replace hardcoded `/tmp` in `mktemp` calls with `${TMPDIR:-/tmp}`. Some container runtimes and non-standard environments may not have `/tmp` writable.

## Changes
- `ods/ods-update.sh`: health log temp file
- `ods/installers/phases/03-features.sh`: GPU topology temp file

## Testing
- [ ] `bash -n ods/ods-update.sh`
- [ ] `bash -n ods/installers/phases/03-features.sh`
- [ ] `make lint`

### DIFF
--- a/ods/installers/phases/03-features.sh
+++ b/ods/installers/phases/03-features.sh
@@ -269,7 +269,7 @@ fi
 # Multi-GPU Configuration
 
 # write $GPU_TOPOLOGY_JSON into a tmpfile to use by the commands
-TOPOLOGY_FILE=$(mktemp /tmp/ods_gpu_topology.XXXXXX.json)
+TOPOLOGY_FILE=$(mktemp "${TMPDIR:-/tmp}/ods_gpu_topology.XXXXXX.json")
 trap 'rm -f "$TOPOLOGY_FILE"' EXIT
 echo "$GPU_TOPOLOGY_JSON" > "$TOPOLOGY_FILE"
 
--- a/ods/ods-update.sh
+++ b/ods/ods-update.sh
@@ -365,7 +365,7 @@ wait_for_healthy() {
     local attempt=0
     local delay=10
     local health_log
-    health_log=$(mktemp /tmp/ods-health-XXXXXX.log)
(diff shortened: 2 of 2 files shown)


Lightheartdevs:
Thanks for taking on `TMPDIR` portability in update and multi-GPU topology temp files.

Quickest path to merge:
- In `ods/installers/phases/03-features.sh`, update the cleanup trap along with the `mktemp` change. After moving `TOPOLOGY_FILE` under user-provided `TMPDIR`, `trap "rm -f $TOPOLOGY_FILE" EXIT` expands an untrusted path into shell code at trap registration time. Use a form that quotes at execution time, for example `trap 'rm -f -- "$TOPOLOGY_FILE"' EXIT`.
- Keep the explicit final `rm -f "$TOPOLOGY_FILE"` cleanup; that path is already quoted correctly.
- If feasible, add a small regression with `TMPDIR` containing a space so this portability fix is actually protected.

What I checked:
- PR diff and `git grep` on `origin/pr/1837` for the changed `mktemp` and trap sites.
- I could not run bash syntax checks locally because `bash` resolves to a broken WSL shim in this Windows environment.

Thanks again for working through the portability edge cases.

Lightheartdevs (dismissed):
Audit: mktemp TMPDIR fallback correct, preserves XXXXXX template, both touched sites (GPU topology, update health log) are runtime-path tmpfiles. Follow-up (not blocking): 4 other mktemp /tmp sites remain (05-docker, 07-devtools x2, macos opencode) — completing them needs test-tmpfile-cleanup.sh grep patterns updated too.

Lightheartdevs (approved):
Audit: mktemp TMPDIR fallback correct (both runtime-path tmpfiles). Rebased onto main, resolving the conflict with the trap-quoting fix (0baa614) by taking BOTH improvements: TMPDIR mktemp + quoted trap. Syntax verified, CI green.

PR #1838, opened 2026-07-16
fix(installer): use 127.0.0.1 in health check URLs to avoid IPv6 hangs

Ovatsugh:
## Summary
- Replace `localhost` with `127.0.0.1` in two LLM health check URL assignments in phase 02-detection. On macOS, `localhost` can resolve to `::1` (IPv6) before `127.0.0.1`, causing curl health checks to hang when the service only listens on IPv4.

## Changes
- `ods/installers/phases/02-detection.sh`: fix cloud-mode health URL (line 67) and fallback health URL (line 190)

## Testing
- [ ] `bash -n ods/installers/phases/02-detection.sh`
- [ ] `make lint`

### DIFF
--- a/ods/installers/phases/02-detection.sh
+++ b/ods/installers/phases/02-detection.sh
@@ -64,7 +64,7 @@ if [[ "${ODS_MODE:-local}" == "cloud" ]]; then
     RAM_GB=$((RAM_KB / 1024 / 1024))
     DISK_AVAIL=$(df -BG "$HOME" | tail -1 | awk '{print $4}' | tr -d 'G')
     BACKEND_ID="cpu"
-    LLM_HEALTHCHECK_URL="http://localhost:4000/health/readiness"
+    LLM_HEALTHCHECK_URL="http://127.0.0.1:4000/health/readiness"
     LLM_PUBLIC_API_PORT="4000"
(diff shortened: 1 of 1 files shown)


Lightheartdevs:
Thanks for tightening the IPv6 health-check path in detection.

Quickest path to merge:
- Update `ods/config/backends/{cpu,nvidia,apple,amd}.json` as well as `ods/installers/phases/02-detection.sh`. `load_backend_contract` populates `BACKEND_PUBLIC_HEALTH_URL`, so the current fallback change at the local-backend assignment is bypassed for normal CPU/NVIDIA/Apple/AMD paths while those JSON files still say `http://localhost:...`.
- Cover the CPU fallback rewrite in `ods/installers/phases/11-services.sh` too, either through the backend JSON updates or by changing its own localhost fallback.
- Add a small contract/static check that backend `public_health_url` values used by host curl probes do not use `localhost`, so the macOS IPv6 hang does not come back through config.

What I checked:
- PR diff plus `git grep` on `origin/pr/1838` for `http://localhost` in backend contracts and installer phases.
- I could not run bash syntax checks locally because `bash` resolves to a broken WSL shim in this Windows environment.

PR #1839, opened 2026-07-16
fix(dashboard): close Invites modal on Escape key press

Ovatsugh:
## Summary
- Add Escape key handler to the Modal component in Invites.jsx. WCAG 2.1 requires dialogs to be dismissible via keyboard. The component already had proper `role="dialog"` and `aria-modal="true"` attributes but was missing the keydown listener.

## Changes
- `ods/extensions/services/dashboard/src/pages/Invites.jsx`: add `useEffect` with `keydown` listener for Escape key in `Modal` component

## Testing
- [ ] `cd ods/extensions/services/dashboard && npm run lint && npm run build`
- [ ] Open Invites modal, press Escape — should close

### DIFF
--- a/ods/extensions/services/dashboard/src/pages/Invites.jsx
+++ b/ods/extensions/services/dashboard/src/pages/Invites.jsx
@@ -550,6 +550,12 @@ function CreateGuestModal({ onClose, onCreated }) {
 }
 
 function Modal({ title, label, onClose, children }) {
+  useEffect(() => {
+    const handleKey = (e) => { if (e.key === 'Escape') onClose() }
+    document.addEventListener('keydown', handleKey)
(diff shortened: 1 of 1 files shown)


Lightheartdevs:
Thanks for taking on the keyboard-dismiss path for the Invites modals.

What looks good:
- `Invites.jsx` already imports `useEffect`, and the new document keydown listener matches the existing Escape-close modal pattern in `Extensions.jsx`.
- The change is scoped to the shared Invites `Modal` wrapper, so the owner and guest create dialogs get the accessibility behavior without touching token generation logic.

What I checked:
- Static review of the PR diff and surrounding `Invites.jsx` modal usage.
- Compared against existing dashboard modal Escape handlers in `Extensions.jsx`.

PR #1841, opened 2026-07-16
fix(dashboard-api): emit UTC-aware timestamps from agent monitor

tang-vu:
## Problem

`agent_monitor.py` builds every timestamp with a bare `datetime.now()` — **naive**, local wall-clock:

```python
self.last_update = datetime.now()                       # AgentMetrics.__init__
"timestamp": datetime.now().isoformat()                 # ThroughputMetrics.add_sample
cutoff = datetime.now() - timedelta(...)                 # prune
agent_metrics.last_update = datetime.now()               # collect_metrics
"timestamp": datetime.now().isoformat()                 # get_full_agent_metrics
```

`to_dict()` and `get_full_agent_metrics()` serialize those with `.isoformat()`, so `GET /api/agents/metrics` returns `last_update` / `timestamp` **with no timezone designator**.

Every other dashboard-api module emits `datetime.now(timezone.utc)` (offset-aware). A browser parsing a no-offset ISO string with `new Date()` interprets it as **local** time. On the common setup — container clock in UTC, viewer in another zone — the agent-monitor times drift by the UTC offset and any "time since last update" math is wrong.

```
OLD (naive): 2026-07-16T11:28:01.676062            → new Date() reads as LOCAL
NEW (aware): 2026-07-16T04:28:01.676086+00:00      → new Date() reads as UTC
```
(same instant; a UTC+7 viewer saw a 7-hour skew)

## Fix

Switch all five `datetime.now()` call sites to `datetime.now(timezone.utc)` and import `timezone`. This matches the rest of the API and produces valid offset-aware ISO 8601.

## Verification

- `python -m py_compile` clean.
- Prune comparison stays consistent (aware vs aware): re-ran `add_sample` logic, `datetime.fromisoformat(p["timestamp"]) > cutoff` works, both retained.
- Emitted value now ends with `+00:00` and round-trips through `fromisoformat` as tz-aware.
- Existing `test_agents.py` unaffected — it sets `last_update` explicitly and asserts HTML escaping, not tz format.

### DIFF
--- a/ods/extensions/services/dashboard-api/agent_monitor.py
+++ b/ods/extensions/services/dashboard-api/agent_monitor.py
@@ -6,7 +6,7 @@ Collects real-time metrics on agent swarms, sessions, and throughput.
 import asyncio
 import json
 import logging
-from datetime import datetime, timedelta
+from datetime import datetime, timedelta, timezone
 from typing import List
 import os
 
@@ -22,7 +22,7 @@ class AgentMetrics:
     """Real-time agent monitoring metrics"""
 
     def __init__(self):
-        self.last_update = datetime.now()
+        self.last_update = datetime.now(timezone.utc)
         self.session_count = 0
         self.tokens_per_second = 0.0  # no data source located; use throughput.get_stats() for live rate
         self.error_rate_1h = 0.0
@@ -96,12 +96,12 @@ class ThroughputMetrics:
     def add_sample(self, tokens_per_sec: float):
         """Add a new throughput sample"""
         self.data_points.append({
-            "timestamp": datetime.now().isoformat(),
+            "timestamp": datetime.now(timezone.utc).isoformat(),
             "tokens_per_sec": tokens_per_sec
         })
 
         # Prune old data
-        cutoff = datetime.now() - timedelta(minutes=self.history_minutes)
+        cutoff = datetime.now(timezone.utc) - timedelta(minutes=self.history_minutes)
         self.data_points = [
             p for p in self.data_points
             if datetime.fromisoformat(p["timestamp"]) > cutoff
@@ -173,7 +173,7 @@ async def collect_metrics():
             # Update agent session count and throughput from Token Spy
             await _fetch_token_spy_metrics()
 
-            agent_metrics.last_update = datetime.now()
+            agent_metrics.last_update = datetime.now(timezone.utc)
 
         except FileNotFoundError as e:
             logger.debug("Metrics collection failed: command not found - %s", e)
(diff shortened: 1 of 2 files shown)


Lightheartdevs:
Thanks for moving the agent-monitor timestamps toward UTC-aware API output; that direction matches the rest of dashboard-api.

Quickest path to merge:
- Fix `ods/extensions/services/dashboard-api/tests/test_agent_monitor.py::TestThroughputMetrics::test_prunes_old_data`: it seeds `tm.data_points` with `datetime.now().isoformat()`, which is now offset-naive, then `ThroughputMetrics.add_sample()` compares it against an aware UTC cutoff and raises `TypeError: can't compare offset-naive and offset-aware datetimes`.
- The smallest change is likely to import `timezone` in `test_agent_monitor.py` and build that fixture with `datetime.now(timezone.utc) - timedelta(minutes=10)`, so the test matches the new runtime invariant.
- If you want `ThroughputMetrics` to tolerate legacy/manual naive samples, normalize parsed timestamps in `ods/extensions/services/dashboard-api/agent_monitor.py` before the prune comparison and add a focused test for mixed naive/aware history. Since these samples are process-local, updating the fixture may be enough.

What I checked:
- Static diff review of `ods/extensions/services/dashboard-api/agent_monitor.py` and the affected test fixture.
- Failing GitHub Dashboard API job: `pytest tests/ -v --cov=. --cov-report=term --cov-report=lcov:coverage.lcov` reported `1 failed, 1182 passed`; the failure is `tests/test_agent_monitor.py::TestThroughputMetrics::test_prunes_old_data` with the naive/aware comparison error above.
- `git diff --check origin/main...origin/pr/1841` passed.
- Other checks are green, including Python type check, Python lint, Test Linux integration-smoke, matrix smoke, ShellCheck, secret scan, env validation, and Dashboard frontend.

Thanks again for pushing the timestamp consistency forward.

tang-vu:
Fixed the failing test: `test_prunes_old_data` now seeds its stale sample with `datetime.now(timezone.utc)`, matching the runtime invariant (all real samples are aware after this change, and `data_points` is process-local so there is no legacy naive history to tolerate — went with the smallest change as suggested). `tests/test_agent_monitor.py` + `tests/test_agents.py`: 25 passed locally.

Lightheartdevs:
Thanks for the follow-up on the agent-monitor timestamp fix; the updated fixture now matches the new UTC-aware invariant.

What looks good:
- `ods/extensions/services/dashboard-api/agent_monitor.py` now emits offset-aware UTC timestamps for `last_update`, throughput samples, pruning cutoffs, and the full metrics response.
- `test_agent_monitor.py::test_prunes_old_data` now seeds an aware stale sample, so the prune comparison exercises the runtime behavior instead of mixing naive and aware datetimes.

What I checked:
- Reviewed the current two-commit diff and prior advisory thread; the new head addresses the TypeError in the test fixture.
- Checked current head `f08cc4d2ffcf0ab7cb6adf671f060f8efa86e7b3`; CI is green across `api`, `integration-smoke`, lint, smoke, distro, and env-validation jobs.
- Ran `git diff --check origin/main...origin/pr/1841` locally.

Thanks again for aligning this endpoint with the rest of dashboard-api.

PR #1842, opened 2026-07-16
fix(dashboard-api): use is_relative_to for workflow path containment

tang-vu:
## Problem

`enable_workflow()` validates the resolved workflow file with a **string prefix** check:

```python
workflow_file = workflow_file.resolve()
if not str(workflow_file).startswith(str(WORKFLOW_DIR.resolve())):
    raise HTTPException(status_code=400, detail="Invalid workflow file path")
```

`str.startswith()` is not a safe containment test: a **sibling directory that shares the base as a name prefix** slips through.

```
WORKFLOW_DIR = /data/workflows
/data/workflows-evil/steal.json  → startswith('/data/workflows')  == True   (WRONG)
                                 → is_relative_to('/data/workflows') == False (CORRECT)
/data/workflows/rag.json         → is_relative_to(...)              == True   (legit, still allowed)
```

The rest of dashboard-api already uses `Path.is_relative_to()` for this exact boundary (`helpers.py:703`, `routers/extensions.py` ×6, `routers/models.py:449`). `workflows.py` was the lone holdout.

## Fix

```python
if not workflow_file.is_relative_to(WORKFLOW_DIR.resolve()):
```

The surrounding `try/except (OSError, ValueError)` still guards `.resolve()`.

`wf_info["file"]` currently comes from the trusted workflow catalog, so this is **defense-in-depth + convention consistency** rather than a live exploit. But the code clearly treats this as a security boundary (it raises *"Invalid workflow file path"*), so it should actually hold, and match how the codebase does containment everywhere else.

## Verification

- `python -m py_compile` clean.
- Demonstrated the sibling-prefix divergence above (startswith True / is_relative_to False), and confirmed a legitimate in-dir path still passes.
- `is_relative_to` is Python 3.9+, already relied on elsewhere in this package.

### DIFF
--- a/ods/extensions/services/dashboard-api/routers/workflows.py
+++ b/ods/extensions/services/dashboard-api/routers/workflows.py
@@ -184,7 +184,7 @@ async def enable_workflow(workflow_id: str, api_key: str = Depends(verify_api_ke
     workflow_file = WORKFLOW_DIR / wf_info["file"]
     try:
         workflow_file = workflow_file.resolve()
-        if not str(workflow_file).startswith(str(WORKFLOW_DIR.resolve())):
+        if not workflow_file.is_relative_to(WORKFLOW_DIR.resolve()):
             raise HTTPException(status_code=400, detail="Invalid workflow file path")
     except HTTPException:
         raise
--- a/ods/extensions/services/dashboard-api/tests/test_workflows.py
+++ b/ods/extensions/services/dashboard-api/tests/test_workflows.py
@@ -911,3 +911,30 @@ def test_workflows_with_matching_n8n(test_client, tmp_path, monkeypatch):
     assert wf["n8nId"] == "55"
     assert wf["executions"] == 42
     assert wf["featured"] is True
+
+
+def test_workflow_enable_rejects_path_traversal_in_catalog_file(test_client, monkeypatch):
+    """A catalog 'file' that escapes WORKFLOW_DIR via .. must be refused by
+    the is_relative_to containment guard, not resolved and imported."""
+    import routers.workflows as wf_mod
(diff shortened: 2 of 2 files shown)


Lightheartdevs:
Thanks for tightening the workflow import path boundary here.

What looks good:
- `ods/extensions/services/dashboard-api/routers/workflows.py` replaces the unsafe string-prefix containment check with `Path.is_relative_to()`, which closes the sibling-prefix bypass described in the PR.
- The approach matches the existing dashboard-api convention already used in helpers, extensions, and model path checks.
- The change stays inside the resolved workflow-file validation block and preserves the existing `HTTPException`, `OSError`, and `ValueError` handling.

What I checked:
- Static diff review of `enable_workflow()` and the surrounding workflow import path.
- Existing `tests/test_workflows.py` coverage around workflow enable success, missing files, invalid IDs, dependency failure, and n8n errors.
- `git diff --check origin/main...origin/pr/1842` passed.
- GitHub checks are green for this head, including Dashboard API, Python type check, Python lint, Test Linux integration-smoke, matrix smoke, ShellCheck, secret scan, env validation, and Dashboard frontend.

Thanks again for bringing this boundary in line with the rest of the API.

Lightheartdevs (approved):
Audit: old startswith(str(...)) containment was sibling-prefix bypassable (a catalog file '../n8n-evil/x.json' resolves under config/ and passes the string check); is_relative_to with both sides resolved is correct (symlinks followed, .. normalized before the boundary test), returns bool so the except-envelope is unaffected. Added the missing traversal negative test (400 on escaping catalog file). Local: ruff clean, 45 passed.

PR #1843, opened 2026-07-16
fix(dashboard): persist dismissed update banner across reloads

tang-vu:
## Problem

`useVersion().dismissUpdate()` records the dismissed release and hides the banner:

```js
localStorage.setItem('dismissed-update', version.latest)
setVersion({ ...version, update_available: false })
```

But `dismissed-update` was **write-only** — nothing ever read it back (the only `getItem('dismissed-update')` in the tree is a test asserting the write). So on the next mount/reload, `useVersion` re-fetches `/api/version`, gets `update_available: true` again, and the update banner **reappears**. "Dismiss" effectively lasted only until the next page refresh.

## Fix

Read `dismissed-update` after each fetch and clear `update_available` when it matches the reported `latest`:

```js
if (data.update_available && data.latest &&
    localStorage.getItem('dismissed-update') === data.latest) {
  data.update_available = false
}
```

The dismissal is version-specific: a genuinely newer `latest` no longer matches, so a new release still surfaces (users don't get stuck permanently hiding all future updates).

## Tests

Added to `useVersion.test.js`:
- dismissed version stays hidden on the next fetch;
- a newer version supersedes the dismissal and re-surfaces.

```
Test Files  1 passed (1)
     Tests  6 passed (6)   # 4 existing + 2 new
```
eslint clean.

### DIFF
--- a/ods/extensions/services/dashboard/src/hooks/__tests__/useVersion.test.js
+++ b/ods/extensions/services/dashboard/src/hooks/__tests__/useVersion.test.js
@@ -61,6 +61,38 @@ describe('useVersion', () => {
     expect(result.current.version.update_available).toBe(false)
   })
 
+  test('suppresses update banner for a previously dismissed version', async () => {
+    globalThis.localStorage.setItem('dismissed-update', '2.0.0')
+    fetch.mockResolvedValue({
+      ok: true,
+      json: () => Promise.resolve({ current: '1.0.0', latest: '2.0.0', update_available: true })
+    })
+
+    const { result } = renderHook(() => useVersion())
+
+    await waitFor(() => {
+      expect(result.current.version).toBeTruthy()
+    })
+    // Dismissed 2.0.0 → banner stays hidden even though the API reports it.
+    expect(result.current.version.update_available).toBe(false)
+  })
+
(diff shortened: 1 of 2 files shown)


Lightheartdevs:
Thanks for fixing the update banner dismissal so it survives the next dashboard load.

What looks good:
- `ods/extensions/services/dashboard/src/hooks/useVersion.js` now reads the same `dismissed-update` key that `dismissUpdate()` writes, closing the write-only localStorage gap.
- The suppression is tied to the reported `latest` version, so dismissing one release does not hide a future newer release.
- `ods/extensions/services/dashboard/src/hooks/__tests__/useVersion.test.js` covers both the dismissed-version case and the newer-version resurfacing case.

What I checked:
- Static diff review of `useVersion.js` and its tests.
- `git diff --check origin/main...origin/pr/1843` passed.
- GitHub checks are green for this head, including Dashboard frontend/API, Test Linux integration-smoke, matrix smoke, ShellCheck, secret scan, env validation, Python lint, and PowerShell lint.

Thanks again for tightening this workflow.

Lightheartdevs (approved):
Audit: localStorage dismissed-update keyed on version string (no PII); newer release re-surfaces the banner (both directions tested); suppression applies on interval refetch too. Client-only, additive.

PR #1844, opened 2026-07-16
fix(dashboard): floor fractional seconds in download ETA

tang-vu:
## Problem

`useDownloadProgress`'s `formatEta()` floors the minutes but not the seconds:

```js
const mins = Math.floor(eta / 60)
const secs = eta % 60            // not floored
```

The ETA comes from bytes-remaining / rate, so it is generally a **float**. A value like `90.7` renders in the model-download progress UI as:

```
1m 30.700000000000003s
```

## Fix

Floor the seconds too, mirroring the minutes:

```js
const secs = Math.floor(eta % 60)   // → '1m 30s'
```

## Tests

Extended the existing `formatEta` test with fractional inputs (`90.7 → '1m 30s'`, `30.9 → '30s'`). Integer cases unchanged.

```
Test Files  1 passed (1)
     Tests  6 passed (6)
```
eslint clean.

### DIFF
--- a/ods/extensions/services/dashboard/src/hooks/__tests__/useDownloadProgress.test.js
+++ b/ods/extensions/services/dashboard/src/hooks/__tests__/useDownloadProgress.test.js
@@ -137,6 +137,9 @@ describe('useDownloadProgress', () => {
 
     expect(result.current.formatEta(90)).toBe('1m 30s')
     expect(result.current.formatEta(30)).toBe('30s')
+    // Fractional eta (bytes-remaining / rate) must not leak float noise.
+    expect(result.current.formatEta(90.7)).toBe('1m 30s')
+    expect(result.current.formatEta(30.9)).toBe('30s')
     expect(result.current.formatEta(null)).toBe('calculating...')
     expect(result.current.formatEta('calculating...')).toBe('calculating...')
   })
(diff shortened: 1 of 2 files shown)


Lightheartdevs:
Thanks for smoothing out the download ETA display path here.

What looks good:
- `ods/extensions/services/dashboard/src/hooks/useDownloadProgress.js` keeps the existing minute handling and floors the seconds component, which directly prevents fractional ETA noise from surfacing in the UI.
- The change is tightly scoped to formatting and leaves the existing `null` / `calculating...` behavior alone.
- `ods/extensions/services/dashboard/src/hooks/__tests__/useDownloadProgress.test.js` adds fractional ETA cases for both minute-plus-second and seconds-only output.

What I checked:
- Static diff review of the hook and its existing test file.
- `git diff --check origin/main...origin/pr/1844` passed.
- GitHub checks are green for this head, including the Dashboard frontend job, Test Linux integration-smoke, matrix smoke, ShellCheck, secret scan, env validation, Python lint, and PowerShell lint.

Thanks again for polishing this user-visible edge case.

Lightheartdevs (approved):
Audit: mirrors the existing minutes floor; ETA fractional-seconds tests cover 90.7/30.9. Trivial, no interaction with polling change above it.

PR #1845, opened 2026-07-16
fix(macos-cli): redact all sensitive keys in ods config show

tang-vu:
## Problem

macOS `ods config show` (`installers/macos/ods-macos.sh` → `cmd_config_show`) masked secrets with:

```bash
grep -qE "(SECRET|PASS|TOKEN|KEY)="
```

That only matches when the keyword is **immediately before `=`**. Keys whose sensitive word sits elsewhere in the name printed their (auto-generated) values in **cleartext**:

| Key (all `secret:true` in `.env.schema.json`) | Old result |
|---|---|
| `OPENCODE_SERVER_PASSWORD` | ❌ leaked (`PASSWORD=` ≠ `PASS=`) |
| `LANGFUSE_DB_PASSWORD`, `LANGFUSE_CLICKHOUSE_PASSWORD`, `LANGFUSE_REDIS_PASSWORD` | ❌ leaked |
| `LANGFUSE_SALT` | ❌ leaked (no `SALT` keyword) |
| `N8N_USER`, `LANGFUSE_INIT_USER_EMAIL` | ❌ leaked |

Same class of bug fixed for the Windows CLI (`ods.ps1`).

## Fix

Extract the key name first, then match it (lowercased with `tr` — Bash 3.2 / macOS safe) against the same keyword set the Linux CLI's `_cmd_config_is_secret` falls back to:

```
*secret*|*password*|*pass*|*token*|*key*|*salt*|*bearer*|*user*|*email*
```

Every `secret:true` key in the current schema matches this set. Non-`KEY=VALUE` lines and non-secret keys are unaffected.

## Test

`tests/test-macos-config-show-secret-mask.sh` (wired into `make test`) drives the real script end-to-end against a scaffold install (with a `docker` shim for the running-Docker check) and asserts secret values never appear, masked keys render as `KEY=***`, and non-secret keys stay in clear.

```
old code: 6 passed, 9 failed   (5 secret values leaked in cleartext)
fixed:   15 passed, 0 failed
```
`bash -n` clean.

### DIFF
--- a/ods/Makefile
+++ b/ods/Makefile
@@ -60,6 +60,9 @@ test: ## Run unit and contract tests
 	@echo "=== ods config show secret-mask matrix ==="
 	@bash tests/test-ods-config-secret-mask.sh
 	@echo ""
+	@echo "=== macOS ods config show secret-mask ==="
+	@bash tests/test-macos-config-show-secret-mask.sh
+	@echo ""
 	@echo "=== bootstrap OpenClaw compose-guard regression ==="
 	@bash tests/test-bootstrap-openclaw-compose-guard.sh
 	@echo ""
--- a/ods/installers/macos/ods-macos.sh
+++ b/ods/installers/macos/ods-macos.sh
@@ -584,13 +584,24 @@ cmd_config_show() {
         line_trimmed=$(echo "$line" | sed 's/^[[:space:]]*//;s/[[:space:]]*$//')
         [[ "$line_trimmed" =~ ^# ]] && echo -e "  ${DGRN}${line_trimmed}${NC}" && continue
         [[ -z "$line_trimmed" ]] && continue
-        if echo "$line_trimmed" | grep -qE "(SECRET|PASS|TOKEN|KEY)="; then
-            local key
-            key=$(echo "$line_trimmed" | cut -d= -f1)
-            echo -e "  ${DGRN}${key}=***${NC}"
-        else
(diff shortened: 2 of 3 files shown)


Lightheartdevs:
Thanks for taking on the macOS `ods config show` masking leak; this lines up an important platform-specific path with the rest of the CLI behavior.

What looks good:
- `ods/installers/macos/ods-macos.sh` now masks by parsed key name instead of requiring the sensitive word to sit immediately before `=`, which covers `OPENCODE_SERVER_PASSWORD`, `LANGFUSE_*_PASSWORD`, `LANGFUSE_SALT`, `N8N_USER`, and email-style secret keys.
- The lowercasing uses `tr`, so the implementation stays compatible with the macOS Bash 3.2 constraints called out in the contributor guide.
- `ods/tests/test-macos-config-show-secret-mask.sh` exercises the real command with a scaffolded install and checks both sides of the contract: secret values are hidden and ordinary keys remain visible.

What I checked:
- Static diff review of `ods/installers/macos/ods-macos.sh`, the new test, and the `ods/Makefile` wiring.
- `git diff --check origin/main...origin/pr/1845` passed.
- GitHub checks are green for this head, including ShellCheck, Test Linux integration-smoke, Dashboard frontend/API, matrix smoke, secret scan, env validation, and PowerShell lint.
- I did not rerun the Bash test locally because this Windows runner routes `bash` through a WSL shim without `/bin/bash`.

Thanks again for closing this platform-specific leak.

PR #1846, opened 2026-07-16
fix(support-bundle): redact user/email secret keys in shared bundle

tang-vu:
## Problem

`ods-support-bundle.sh` builds a diagnostics archive intended to be attached to **public** issues — its header promises *"Raw .env is never included; only config/env.redacted is written."* But the redaction keyword set was:

```
KEY | TOKEN | SECRET | PASSWORD | PASS | SALT | AUTH | CREDENTIAL
```

— missing **USER** and **EMAIL**. Keys marked `secret: true` in `.env.schema.json` therefore shipped in **cleartext** in two places:

| Surface | Function |
|---|---|
| `config/env.redacted` | `write_redacted_env` |
| `manifest/evidence.json` → `env_keys[..].value` | evidence generator |

Affected keys: `N8N_USER`, `LANGFUSE_INIT_USER_EMAIL` (PII), `LANGFUSE_MINIO_ROOT_USER`.

```
# old env.redacted
N8N_USER=admin-username           ← leaked
LANGFUSE_INIT_USER_EMAIL=a@b.test ← leaked (PII)
```

## Fix

Add `USER|EMAIL|BEARER` to **both env-key-scoped** patterns (`write_redacted_env` and the evidence generator), matching `.env.schema.json`'s secret set and the CLI's `config show` masking.

`redact_file` (arbitrary logs / JSON / `docker compose config`) is **intentionally left unchanged** — adding `user`/`email` to its JSON pattern would over-redact container `user: "1000:1000"` UID/GID and `*user*`-suffixed diagnostic fields, hurting debuggability. The two patterns fixed here key off env variable **names**, so there is zero over-redaction.

## Test

Extends `tests/test-support-bundle.sh` (already gated in `test-linux.yml`): the fixture `.env` now carries `N8N_USER` and `LANGFUSE_INIT_USER_EMAIL`, plus an explicit `env.redacted` redaction assertion.

```
old code: 17 passed, 3 failed   (env.redacted, evidence.json, archive all leaked the sentinel)
fixed:    20 passed, 0 failed
```
Verified end-to-end via the real script (WSL); `bash -n` clean.

### DIFF
--- a/ods/scripts/ods-support-bundle.sh
+++ b/ods/scripts/ods-support-bundle.sh
@@ -215,7 +215,11 @@ from pathlib import Path
 
 src = Path(sys.argv[1])
 dest = Path(sys.argv[2])
-secret = re.compile(r"(KEY|TOKEN|SECRET|PASSWORD|PASS|SALT|AUTH|CREDENTIAL)", re.I)
+# USER|EMAIL|BEARER cover schema secret:true keys the old pattern missed —
+# N8N_USER, LANGFUSE_INIT_USER_EMAIL, LANGFUSE_MINIO_ROOT_USER — which are
+# published in cleartext when this env.redacted is shared on a public issue.
+# Match .env.schema.json's secret set / the CLI's config-show masking.
+secret = re.compile(r"(KEY|TOKEN|SECRET|PASSWORD|PASS|SALT|AUTH|CREDENTIAL|USER|EMAIL|BEARER)", re.I)
 
 lines = []
 for line in src.read_text(encoding="utf-8", errors="replace").splitlines():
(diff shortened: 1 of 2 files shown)


Lightheartdevs:
Thanks for closing the support-bundle redaction gap around user/email-style secret keys.

What looks good:
- `ods/scripts/ods-support-bundle.sh` updates both env-key-scoped redaction paths, so `config/env.redacted` and `manifest/evidence.json` now treat `USER`, `EMAIL`, and `BEARER` keys like the rest of the schema secret set.
- The broader arbitrary-file redactor is left alone, which avoids over-redacting diagnostic fields such as Compose `user:` while still protecting env values intended for public issue attachments.
- `ods/tests/test-support-bundle.sh` adds a concrete fixture and assertion for `N8N_USER` and `LANGFUSE_INIT_USER_EMAIL`, matching the leak class described in the PR.

What I checked:
- Static diff review of `ods/scripts/ods-support-bundle.sh` and `ods/tests/test-support-bundle.sh`.
- `git diff --check origin/main...origin/pr/1846` passed.
- GitHub checks are green for this head, including ShellCheck, Test Linux integration-smoke, secret scan, Dashboard frontend/API, PowerShell lint, matrix smoke, and env validation.
- I did not rerun the Bash test locally because this Windows runner routes `bash` through a WSL shim without `/bin/bash`.

PR #1849, opened 2026-07-16
fix(windows): stop switch fall-through producing an invalid TIMEZONE

w3-surfer:
## Problem

The Windows→IANA timezone fallback in `New-ODSEnv` uses `switch -Wildcard` with **no `break`** on any arm:

```powershell
switch -Wildcard ($tzInfo.Id) {
    "*Eastern*"    { "America/New_York" }
    ...
    "*AUS Eastern*"  { "Australia/Sydney" }
    ...
}
```

PowerShell runs **every** matching arm, and as a subexpression (`$tz = $(... switch ...)`) it collects all emitted values into an **array**. So an ID like `"AUS Eastern Standard Time"` (Sydney/Melbourne/Canberra) matches both `"*Eastern*"` **and** `"*AUS Eastern*"`, and `$tz` becomes `@("America/New_York","Australia/Sydney")`. The here-string then interpolates it (`$OFS` = space):

```
TIMEZONE=America/New_York Australia/Sydney
```

— an invalid TZ written into `.env`, which flows into the containers (n8n, etc.). `"Central Pacific Standard Time"` hits the same fault via `*Central*` + `*Pacific*`.

This fallback is the **primary path on Windows PowerShell 5.1** (the default shell, which the installer supports): `TimeZoneInfo.TryConvertWindowsIdToIanaId` doesn't exist on .NET Framework 4.x, so the guarded call throws, `$ianaId` stays `$null`, and the switch runs.

## Fix

Add `break` to every arm so the **first** match wins (the switch can then only ever emit a single value), and order the more specific `"*AUS Eastern*"` ahead of `"*Eastern*"` so eastern-Australia hosts resolve to `Australia/Sydney`:

```powershell
switch -Wildcard ($tzInfo.Id) {
    "*AUS Eastern*"  { "Australia/Sydney"; break }
    "*Eastern*"    { "America/New_York"; break }
    ...
    default        { "UTC"; break }
}
```

## Notes

- `break` inside a switch (no enclosing loop here) simply exits the switch — no behavior change for single-match IDs.
- The concrete correctness win is eastern Australia (`AUS Eastern` → `Australia/Sydney` instead of a malformed value); more broadly, `TIMEZONE` can no longer be written as a space-joined multi-value string for any ID.

### DIFF
--- a/ods/installers/windows/lib/env-generator.ps1
+++ b/ods/installers/windows/lib/env-generator.ps1
@@ -325,30 +325,37 @@ function New-ODSEnv {
             if ($ok -and $outIana) { $ianaId = $outIana }
         } catch { }
         if ($ianaId) { $ianaId } else {
+            # PowerShell `switch` runs *every* matching arm, and as a
+            # subexpression it collects all emitted values into an array, so
+            # multi-matching IDs (e.g. "AUS Eastern Standard Time" hits both
+            # "*AUS Eastern*" and "*Eastern*") would write an invalid
+            # "TIMEZONE=America/New_York Australia/Sydney". `break` on every arm
+            # makes the first match win; the more specific "*AUS Eastern*" is
+            # ordered ahead of "*Eastern*" so it takes precedence.
             switch -Wildcard ($tzInfo.Id) {
-                "*Eastern*"    { "America/New_York" }
-                "*Central*"    { "America/Chicago" }
-                "*Mountain*"   { "America/Denver" }
-                "*Pacific*"    { "America/Los_Angeles" }
-                "*Alaska*"     { "America/Anchorage" }
-                "*Hawaii*"     { "Pacific/Honolulu" }
-                "*UTC*"        { "UTC" }
-                "*GMT*"        { "Europe/London" }
-                "*W. Europe*"  { "Europe/Berlin" }
-                "*Romance*"    { "Europe/Paris" }
-                "*India*"      { "Asia/Kolkata" }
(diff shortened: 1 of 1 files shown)


Lightheartdevs:
Thanks for tightening the Windows timezone fallback; this is exactly the kind of PowerShell switch behavior that is easy to miss until it writes a bad `.env`.

What looks good:
- `ods/installers/windows/lib/env-generator.ps1` now orders `*AUS Eastern*` before the generic `*Eastern*` case and breaks after every match, so the fallback emits one timezone string instead of a space-joined array.
- The newer `[TimeZoneInfo]::TryConvertWindowsIdToIanaId()` path is left unchanged; the patch only hardens the Windows PowerShell 5.1 fallback path where this matters.
- The change stays scoped to `.env` timezone generation and does not alter compose, schema, or runtime service wiring.

What I checked:
- Static diff review of the `New-ODSEnv` timezone block.
- `git diff --check origin/main...origin/pr/1849` passed.
- PowerShell parser check on the PR version of `env-generator.ps1` passed; local switch probes confirmed `AUS Eastern Standard Time` maps to a single `Australia/Sydney`, `Eastern Standard Time` maps to a single `America/New_York`, and overlapping names no longer produce multiple values.
- `gh pr checks 1849 --required=false` is green, including Dashboard API/frontend, integration-smoke, matrix smoke, ShellCheck/Ruff/PowerShell lint, secret scan, and env validation.

Thanks again for cleaning up this Windows install edge.

PR #1850, opened 2026-07-16
fix(windows): guard the native-inference PID parse against a bad PID file

w3-surfer:
## Problem

`Get-NativeInferenceStatus` parses the PID file **before** its `try` block:

```powershell
if (-not (Test-Path $script:INFERENCE_PID_FILE)) { return $result }

$savedPid = [int](Get-Content $script:INFERENCE_PID_FILE -Raw).Trim()
try {
    ...
```

`ods.ps1` sets `$ErrorActionPreference = "Stop"` at the top, so if `data/llama-server.pid` exists but is **empty** (0 bytes) or **non-numeric** — plausible after an interrupted write / partial install — `[int]""` (or `.Trim()` on the `$null` from an empty `-Raw` read) throws a *terminating* error and crashes the entire command.

`Get-NativeInferenceStatus` is called by `Invoke-Status`, `Start-NativeInferenceServer`, and `Stop-NativeInferenceServer`, so **`ods status` / `start` / `stop` / `restart` all crash** instead of treating the native server as not-running:

```
Cannot convert value "" to type "System.Int32". ...
```

## Fix

Validate the PID before casting — mirroring the `'^\d+$'` guard the **same file** already uses in its stop path (`~line 756`) — and on a bad PID file, clear it and return "not running":

```powershell
$rawPid = (Get-Content $script:INFERENCE_PID_FILE -Raw)
if (-not $rawPid) { $rawPid = "" }
if ($rawPid.Trim() -notmatch '^\d+$') {
    Remove-Item $script:INFERENCE_PID_FILE -Force -ErrorAction SilentlyContinue
    return $result
}
$savedPid = [int]$rawPid.Trim()
```

The `if (-not $rawPid)` normalization also covers a truly-empty (0-byte) file, where `Get-Content -Raw` returns `$null`. A valid numeric PID behaves exactly as before.

### DIFF
--- a/ods/installers/windows/ods.ps1
+++ b/ods/installers/windows/ods.ps1
@@ -643,7 +643,17 @@ function Get-NativeInferenceStatus {
 
     if (-not (Test-Path $script:INFERENCE_PID_FILE)) { return $result }
 
-    $savedPid = [int](Get-Content $script:INFERENCE_PID_FILE -Raw).Trim()
+    # Guard the PID parse: this runs with $ErrorActionPreference = "Stop", so
+    # casting an empty or non-numeric PID file to [int] throws a terminating
+    # error and crashes `ods status`/`start`/`stop`. Mirror the numeric guard
+    # used elsewhere and treat a bad PID file as "not running" (and clear it).
+    $rawPid = (Get-Content $script:INFERENCE_PID_FILE -Raw)
+    if (-not $rawPid) { $rawPid = "" }
+    if ($rawPid.Trim() -notmatch '^\d+$') {
+        Remove-Item $script:INFERENCE_PID_FILE -Force -ErrorAction SilentlyContinue
+        return $result
+    }
+    $savedPid = [int]$rawPid.Trim()
     try {
         $proc = Get-Process -Id $savedPid -ErrorAction SilentlyContinue
         if ($proc -and -not $proc.HasExited) {


Lightheartdevs:
Thanks for hardening the native-inference status path against stale PID files; treating a bad PID as not-running is the right behavior for `ods status`/`start`/`stop`.

Quickest path to merge:
- In `ods/installers/windows/ods.ps1`, replace the `^\d+$` check plus `[int]$rawPid.Trim()` cast in `Get-NativeInferenceStatus` with `[int]::TryParse(...)` and reject non-positive values. A corrupt PID file containing only digits but larger than `Int32` still passes the regex and throws a terminating cast before the `try`, so the crash path is not fully closed yet.
- While touching PID parsing, apply the same `TryParse` pattern to the PID-file read in `Stop-ODSLemonadeRuntime`; that path also regex-checks and then casts, so direct Lemonade cleanup can preserve the same overflow crash.
- Add or document a focused PID-file case for empty, nonnumeric, valid, negative, and overflow-sized digit-only contents so this regression stays covered.

What I checked:
- Static diff review of `Get-NativeInferenceStatus` and the neighboring Windows native-runtime stop path.
- `git diff --check origin/main...origin/pr/1850` passed.
- PowerShell parser check on the PR version of `ods.ps1` passed; local behavior probes showed empty/`abc` invalid, `123`/` 456 ` valid, `-1`/`12.3` invalid, and `999999999999999999999999` still passes the regex before `[int]` throws.
- `gh pr checks 1850 --required=false` has one red CachyOS distro job; its log fails while installing `rsync` via pacman before any Windows PID-path code runs. The rest of the visible checks, including PowerShell lint on Ubuntu/Windows, are green.

Thanks again; this should be a small follow-up to make the guard complete.

Lightheartdevs:
Thanks for continuing to harden the Windows native-inference PID handling; clearing corrupt PID files is the right user-facing behavior for status/start/stop.

Quickest path to merge:
- In `ods/installers/windows/ods.ps1`, update `Get-NativeInferenceStatus` to parse `$rawPid.Trim()` with `[int]::TryParse(...)` and reject non-positive values before calling `Get-Process`. The current regex still lets digit-only values larger than `Int32` through, so `[int]$rawPid.Trim()` can still throw a terminating error before the `try`.
- Apply the same `TryParse`/positive-PID pattern to the PID-file read in `Stop-ODSLemonadeRuntime`, where `$rawPid -match '^\d+$'` is still followed by `([int]$rawPid)`.
- Add or document a focused PID-file check for empty, nonnumeric, valid, negative, and overflow-sized digit-only contents so this stays covered.

What I checked:
- Static diff review of `Get-NativeInferenceStatus` and nearby native-runtime cleanup in `ods/installers/windows/ods.ps1`.
- Ran `git diff --check origin/main...HEAD` and parsed `ods.ps1` via `[scriptblock]::Create(...)`.
- Reproduced the remaining failure mode locally: `"999999999999999999999999" -notmatch '^\d+$'` is false, but `[int]` casting it throws `RuntimeException`.
- `gh pr checks 1850 --required=false` is green, including PowerShell lint on Ubuntu/Windows and the distro/matrix jobs.

Thanks again; this should be a small targeted follow-up.

PR #1851, opened 2026-07-16
fix(n8n): register the shipped hermes-agent-trigger workflow in the catalog

w3-surfer:
## Problem

`config/n8n/hermes-agent-trigger.json` ships in this release, but it is the **only** workflow file in `config/n8n/` that is missing from `catalog.json`.

The dashboard surfaces and imports n8n templates **solely** from the catalog — `routers/workflows.py` `load_workflow_catalog()` and the import path iterate over catalog entries only. So a workflow file that isn't in the catalog is unreachable: users can neither see it in the template list nor import it, even though it ships in the repo.

This is a wiring gap, not a feature request:
- `docs/MIGRATION-OPENCLAW-TO-HERMES.md` documents that a `hermes-agent-trigger.json` ships alongside OpenClaw's for the deprecation release.
- The workflow file's own sticky note says: *"Replaces the deprecated `openclaw-agent-trigger`. Both templates ship in the deprecation release."*
- Its sibling `openclaw-agent-trigger` **is** in the catalog (and `featured`), so the replacement being absent is an oversight.

`GET /api/workflows` currently returns 17 templates; `hermes-agent-trigger` never appears.

## Fix

Add the catalog entry next to the `openclaw-agent-trigger` sibling it replaces, mirroring its shape (category `agents`, deps `hermes` + `llama-server` — both real service ids):

```json
{
  "id": "hermes-agent-trigger",
  "file": "hermes-agent-trigger.json",
  "name": "Hermes Agent Trigger",
  "description": "Dispatch a goal to a Hermes Agent session via its OpenAI-compatible endpoint, stream the response, and store the final result.",
  "icon": "Bot",
  "category": "agents",
  "dependencies": ["hermes", "llama-server"],
  "setupTime": "3 minutes",
  "featured": true
}
```

(No `diagram` — the workflow file is a manual-trigger + sticky-note template stub like the other catalog entries that omit one.)

## Verify

- `python3 -m json.tool` parses; `id`/`name` match the workflow file's `name` and `meta.templateId`.
- `GET /api/workflows` now lists **18** templates including `hermes-agent-trigger`.

### DIFF
--- a/ods/config/n8n/catalog.json
+++ b/ods/config/n8n/catalog.json
@@ -568,6 +568,20 @@
       }
     },
     {
+      "id": "hermes-agent-trigger",
+      "file": "hermes-agent-trigger.json",
+      "name": "Hermes Agent Trigger",
+      "description": "Dispatch a goal to a Hermes Agent session via its OpenAI-compatible endpoint, stream the response, and store the final result.",
+      "icon": "Bot",
+      "category": "agents",
+      "dependencies": [
+        "hermes",
+        "llama-server"
+      ],
+      "setupTime": "3 minutes",
+      "featured": true
+    },
+    {
       "id": "openclaw-agent-trigger",
       "file": "openclaw-agent-trigger.json",
       "name": "OpenClaw Agent Trigger",


Lightheartdevs:
Thanks for wiring the shipped Hermes n8n trigger into the workflow catalog; this closes a visible template-discovery gap for the OpenClaw-to-Hermes migration path.

What looks good:
- `ods/config/n8n/catalog.json` adds only the missing `hermes-agent-trigger` entry, adjacent to the existing `openclaw-agent-trigger`, so the dashboard/API catalog enumeration can expose the shipped workflow file without changing import behavior.
- The entry's `id`, `file`, and `name` match `ods/config/n8n/hermes-agent-trigger.json`'s `name` and `meta.templateId`, and the dependencies point at the expected `hermes` + `llama-server` runtime services.
- This is a narrow config-only wiring fix that matches the migration docs and the workflow's own note that both templates ship during the deprecation window.

What I checked:
- Static diff review of the single catalog change and the surrounding n8n workflow entries.
- `git diff --check origin/main...origin/pr/1851` passed.
- Local JSON/catalog consistency check against the PR head verified the new catalog entry and shipped workflow metadata line up.
- `gh pr checks 1851 --required=false` is green, including Dashboard API/frontend, Test Linux integration-smoke, matrix smoke, ShellCheck/Ruff/PowerShell lint, secret scan, and env validation.

Thanks again for catching this missing catalog entry.

Lightheartdevs (approved):
Audit: 18 workflows on disk, 17 catalog entries, hermes the only gap; patched catalog validates (18 entries, valid JSON, schema matches siblings). No auto-import path — upgrades just see one more template, zero duplicate risk.

PR #1853, opened 2026-07-16
chore: improve contributor guidance, fix silent skip in runtime check, and add jq guard in first-boot-demo (#1852)

### LINKED ISSUE #1852: chore: three small quality gaps — ComfyUI stub comment, runtime-check skip-vs-assert, and missing jq guard in first-boot-demo
## Summary

While auditing the codebase for small gaps not covered by existing issues or PRs, I found three minor, self-contained improvements across different areas. Each is under ~10 lines to fix and doesn't warrant its own issue.

---

## 1. Improve contributor guidance in `ods/extensions/services/comfyui/compose.yaml`

**File:** `ods/extensions/services/comfyui/compose.yaml`

### Current

```yaml
# This file exists so the registry can detect comfyui as enabled.
services: {}
```

The comment is technically correct but doesn't explain:

- Why the stub exists (registry detection requires a `compose.yaml`)
- Where the real service definition lives
- Which GPU overlay files are available
- How the installer selects the appropriate overlay

New contributors inspecting this file have no indication that the actual service definitions live elsewhere, which has caused confusion in the past.

### Suggested fix

Expand the comment block to document the overlay pattern, for example:

```yaml
# ComfyUI — Image Generation
#
# This stub exists so the service registry can detect ComfyUI as installed.
# The actual container definition lives in GPU-specific overlay files that are
# merged with this stub at runtime by the installer/compose resolver:
#
#   compose.amd.yaml             — AMD ROCm (single GPU, gfx1151)
#   compose.nvidia.yaml          — NVIDIA CUDA (single GPU, builds local Dockerfile)
#   compose.multigpu-amd.yaml    — AMD ROCm multi-GPU
#   compose.multigpu-nvidia.yaml — NVIDIA multi-GPU
#
# The correct overlay is selected by the installer based on the detected GPU backend.
# Do NOT add a service definition here; add it to the appropriate overlay file.
services: {}
```

---

## 2. Assert expected output instead of skipping in `test-extension-runtime-check.sh`

**File:** `ods/tests/test-extension-runtime-check.sh`

### Current

```bash
if docker info >/dev/null 2>&1; then
    if echo "$out" | grep -qE '\[OK\]|\[BAD\]|\[INFO\]'; then
        pass "docker available — check lines present"
    else
        skip "docker up but no OK/BAD/INFO lines (minimal stack is OK)"
    fi
fi
```

When Docker is available but no ODS services are deployed (the common case in CI and many developer environments), the runtime check prints something like:

```text
[INFO] No services in registry — nothing to check
```

The test currently **skips** instead of validating this expected output, meaning regressions in the informational path could go unnoticed.

### Suggested fix

Treat the "nothing to check" path as a successful assertion:

```bash
if docker info >/dev/null 2>&1; then
    if echo "$out" | grep -qE '\[OK\]|\[BAD\]|\[INFO\]'; then
        pass "docker available — check lines present"
    elif echo "$out" | grep -q "nothing to check"; then
        pass "docker available but no services registered — script exits cleanly"
    else
        fail "docker available but output contained no expected status lines"
    fi
fi
```

---

## 3. Check for `jq` before using it in `first-boot-demo.sh`

**File:** `ods/scripts/first-boot-demo.sh`

The script uses `jq` throughout, for example:

```bash
... | jq -r '.choices[0].message.content' 2>/dev/null || echo ""
```

Because stderr is redirected and the pipeline falls back to `echo ""`, a missing `jq` installation produces confusing output:

```text
✗ No response from LLM
✗ No response from code assistant
```

instead of explaining that a required dependency is missing.

### Suggested fix

Add a dependency check near the top of the script (after argument parsing):

```bash
if ! command -v jq >/dev/null 2>&1; then
    echo "Error: 'jq' is required for first-boot-demo.sh but was not found in PATH."
    echo "Install it with: sudo apt install jq  (or brew install jq on macOS)"
    exit 1
fi
```

This provides an immediate, actionable error instead of silently masking the real issue.

### Use Case

A contributor opens `compose.yaml` to modify the ComfyUI service, sees only `services: {}`, and assumes the service definition is missing. Without guidance, they may edit the stub instead of the appropriate GPU overlay or spend time searching for the real configuration.

---

### Use Case

A developer changes the runtime check output and runs the tests on a machine with Docker but no ODS services. The test silently skips instead of validating the expected "no services" path, allowing regressions in the output to go unnoticed.

---

### Use Case

A first-time user runs `first-boot-demo.sh` on a system without `jq`. The script silently reports empty LLM responses instead of explaining that a required dependency is missing, leading the user to believe the installation is broken.

### Proposed Solution

_No response_

### Alternatives Considered

_No response_

IronicRayquaza:
## Summary

Closes #1852.

This PR includes three small, self-contained quality improvements identified during a codebase audit. Each change is low-risk, isolated, and improves either contributor experience, test coverage, or user-facing diagnostics.

### 1. `ods/extensions/services/comfyui/compose.yaml` — Expand stub comment

The existing stub comment was technically accurate but did not explain why the file exists, where the actual service definition lives, or how GPU-specific overlays are selected. The comment has been expanded to document the overlay pattern and direct contributors to the appropriate files.

---

### 2. `ods/tests/test-extension-runtime-check.sh` — Replace silent skip with explicit assertions

When Docker is available but no ODS services are deployed (the common case in CI and bare-checkout environments), the test could silently skip instead of validating the expected "nothing to check" path. This change replaces the skip with explicit pass/fail assertions so regressions in the informational output are detected.

---

### 3. `ods/scripts/first-boot-demo.sh` — Add `jq` dependency check

The script relies on `jq` to parse API responses but previously suppressed errors when it was missing, causing empty responses that appeared to indicate a broken LLM. A startup dependency check now reports a clear, actionable error if `jq` is not installed.

## Testing

- `bash -n` passes on all modified shell scripts.
- `git diff --check` reports no whitespace errors.
- Verified `test-extension-runtime-check.sh` behavior with and without Docker available.
- Verified `first-boot-demo.sh` exits with a clear error message when `jq` is not installed.

## Files Changed

| File | Change |
|------|--------|
| `ods/extensions/services/comfyui/compose.yaml` | Improve contributor documentation |
| `ods/tests/test-extension-runtime-check.sh` | Replace silent skip with explicit assertions |
| `ods/scripts/first-boot-demo.sh` | Add `jq` dependency guard |

### DIFF
--- a/ods/extensions/services/comfyui/compose.yaml
+++ b/ods/extensions/services/comfyui/compose.yaml
@@ -1,7 +1,14 @@
 # ComfyUI — Image Generation
-# This base stub is merged with a GPU-specific overlay:
-#   compose.amd.yaml   (AMD ROCm, gfx1151)
-#   compose.nvidia.yaml (NVIDIA CUDA)
-# The GPU overlay provides the full service definition.
-# This file exists so the registry can detect comfyui as enabled.
+#
+# This stub exists so the service registry can detect ComfyUI as installed.
+# The actual container definition lives in GPU-specific overlay files that are
+# merged with this stub at runtime by the installer/compose resolver:
+#
+#   compose.amd.yaml             — AMD ROCm (single GPU, gfx1151)
+#   compose.nvidia.yaml          — NVIDIA CUDA (single GPU, builds local Dockerfile)
+#   compose.multigpu-amd.yaml    — AMD ROCm multi-GPU
+#   compose.multigpu-nvidia.yaml — NVIDIA multi-GPU
+#
+# The correct overlay is selected by the installer based on the detected GPU backend.
+# Do NOT add a service definition here; add it to the appropriate overlay file.
 services: {}
--- a/ods/scripts/first-boot-demo.sh
+++ b/ods/scripts/first-boot-demo.sh
@@ -55,6 +55,14 @@ while [[ $# -gt 0 ]]; do
     esac
 done
 
+# Dependency check — jq is required for parsing API responses.
+if ! command -v jq >/dev/null 2>&1; then
+    echo "Error: 'jq' is required for first-boot-demo.sh but was not found in PATH."
+    echo "Install it with:  sudo apt install jq   (Debian/Ubuntu)"
+    echo "                  brew install jq        (macOS)"
+    exit 1
+fi
+
 #=============================================================================
 # Helpers
 #=============================================================================
--- a/ods/tests/test-extension-runtime-check.sh
+++ b/ods/tests/test-extension-runtime-check.sh
@@ -66,8 +66,10 @@ fi
 if docker info >/dev/null 2>&1; then
     if echo "$out" | grep -qE '\[OK\]|\[BAD\]|\[INFO\]'; then
         pass "docker available — check lines present"
+    elif echo "$out" | grep -q "nothing to check"; then
+        pass "docker available but no services registered — script exits cleanly"
     else
-        skip "docker up but no OK/BAD/INFO lines (minimal stack is OK)"
+        fail "docker available but output contained no expected status lines"
     fi
 else
     if echo "$out" | grep -qi docker; then


Lightheartdevs:
Thanks for tightening up the ComfyUI contributor notes and turning the runtime-check edge case into a real assertion.

What looks good:
- `ods/extensions/services/comfyui/compose.yaml` now accurately documents that `compose.yaml` is a registry stub and points contributors at the GPU-specific overlays that actually define the service.
- `ods/scripts/first-boot-demo.sh` now fails early with an actionable `jq` install message before it would turn LLM JSON parsing into empty demo output.
- `ods/tests/test-extension-runtime-check.sh` keeps the no-services Docker path non-blocking while asserting the expected "nothing to check" output instead of silently skipping.

What I checked:
- Reviewed the PR diff against the extension/compose registry pattern and the existing first-boot demo flow.
- Ran `git diff --check origin/main...HEAD`, `bash -n ods/scripts/first-boot-demo.sh ods/tests/test-extension-runtime-check.sh`, and `bash ods/tests/test-extension-runtime-check.sh` in the PR worktree; the runtime-check suite passed 5/5.
- `gh pr checks 1853 --required=false` is green across ShellCheck, Docker Compose validation, Test Linux, Matrix Smoke, Dashboard, Python lint, PowerShell lint, secret scan, and env-schema jobs.

Thanks again for cleaning up these small but useful gaps.

Lightheartdevs (approved):
Audit: comfyui overlay files named in the comment all exist; first-boot-demo jq guard sits after arg parsing (--help works), installer preflight auto-installs jq. Non-CI dev test. Noted: the 'nothing to check' test branch is unreachable (message carries [INFO]) — inert, not harmful.

PR #1858, opened 2026-07-17
feat(preflight): validate rootless Docker subordinate UID/GID ranges

### LINKED ISSUE #1719: Enhancement: Validate subordinate UID/GID range allocation for rootless Docker in preflight checks
## Problem

When Docker is running in **rootless mode**, it relies on subordinate UID/GID mappings configured in `/etc/subuid` and `/etc/subgid`. If the current user does not have a valid subordinate ID range, or the allocated range is smaller than the required minimum (typically **65536** IDs), rootless containers may fail to start or exhibit unexpected permission issues.

Currently, ODS does not validate this prerequisite during installation or runtime diagnostics, leaving users to discover the problem only after container failures.

## Proposed Solution

Add subordinate UID/GID validation to both the installer preflight checks and runtime diagnostics.

### Preflight (`ods-preflight.sh`)

- Detect when Docker is running in rootless mode.
- Verify that the current user has entries in:
  - `/etc/subuid`
  - `/etc/subgid`
- Validate that each allocated range is at least **65536** subordinate IDs.
- Display a clear warning or error if:
  - either entry is missing, or
  - the allocated range is smaller than the required minimum.
- Provide actionable remediation guidance, for example:
  ```bash
  sudo usermod \
    --add-subuids 100000-165535 \
    --add-subgids 100000-165535 \
    <username>

### Use Case

In Docker rootless mode, non-root containers (such as `n8n`, `whisper`, `tts`, `token-spy`, `privacy-shield`, `ape`, `langfuse`, and others) fail to initialize user namespace mappings if the host's subordinate UID/GID ranges are missing or insufficient.

Instead of failing later with cryptic container runtime errors (for example, `failed to register layer: uid/gid map out of range` or containers failing to start), this change validates the prerequisite during installation and runtime diagnostics, providing immediate, actionable guidance to the system administrator before deployment begins.

### Proposed Solution

_No response_

### Alternatives Considered

_No response_

tang-vu:
## Summary
Implements #1719: the Linux install preflight now validates subordinate UID/GID allocation when the Docker daemon runs in rootless mode.

- Detects a rootless daemon via `docker info --format '{{.SecurityOptions}}'` (only probed when the daemon check already passed).
- New `ROOTLESS_SUBID` check parses `/etc/subuid` + `/etc/subgid` for the current user:
  - **fail** — no allocation at all (rootless containers cannot map user namespaces; this is the "failed to register layer: uid/gid map out of range" failure mode reported in the issue)
  - **warn** — allocation exists but the summed range is below 65536 IDs
  - **pass** — ≥65536 subuids and subgids
- Handles entries keyed by numeric UID and multiple ranges per user (summed), and `/etc/subgid` is correctly treated as keyed by user, not group.
- Remediation message gives the exact `sudo usermod --add-subuids 100000-165535 --add-subgids 100000-165535 <user>` command plus the rootless daemon restart step.
- On rootful daemons the check reports an explicit "skipped" pass so the JSON contract stays stable.

## Testability
`ODS_ASSUME_ROOTLESS=1` plus `ODS_SUBUID_FILE`/`ODS_SUBGID_FILE` overrides let tests (and operators debugging a box) exercise the check without a rootless daemon.

## Tests
Extended `tests/test-linux-install-preflight.sh` with 5 fixture cases: full range → pass, missing entry → fail, small range → warn, split ranges summed → pass, numeric-UID key → pass. All pass locally; the two pre-existing failures on Windows dev hosts (msys /tmp path vs python3) are unrelated baseline issues that pass on Linux CI.

Closes #1719

### DIFF
--- a/ods/scripts/linux-install-preflight.sh
+++ b/ods/scripts/linux-install-preflight.sh
@@ -236,6 +236,59 @@ else
         "Usually fine on modern distros; if Docker fails oddly, see LINUX-TROUBLESHOOTING-GUIDE.md#cgroups."
 fi
 
+# --- Rootless Docker subordinate UID/GID ranges ---
+# Rootless Docker maps in-container users through the subordinate ID ranges
+# allocated in /etc/subuid and /etc/subgid. A missing entry, or a range
+# smaller than 65536 IDs, breaks image extraction for non-root images
+# ("failed to register layer: uid/gid map out of range") long after install.
+# Only meaningful when the daemon actually runs rootless.
+SUBUID_FILE="${ODS_SUBUID_FILE:-/etc/subuid}"
+SUBGID_FILE="${ODS_SUBGID_FILE:-/etc/subgid}"
+MIN_SUBID_RANGE=65536
+
+subid_range_total() {
+    # Sum every range allocated to the current user in a subuid/subgid file.
+    # Entries may be keyed by user name or by numeric UID (subgid is also
+    # keyed by user, not group). Multiple ranges per user are legal.
+    local file="$1" user="$2" numeric_uid="$3" total=0 owner start count
+    [[ -r "$file" ]] || { echo 0; return; }
+    while IFS=: read -r owner start count; do
+        if [[ "$owner" == "$user" || "$owner" == "$numeric_uid" ]]; then
+            [[ "$count" =~ ^[0-9]+$ ]] && total=$((total + count))
+        fi
+    done <"$file"
+    echo "$total"
+}
+
+ROOTLESS_DOCKER=false
+if [[ "${ODS_ASSUME_ROOTLESS:-0}" == "1" ]]; then
+    ROOTLESS_DOCKER=true
+elif [[ "$DOCKER_INFO_OK" == true ]] && docker info --format '{{.SecurityOptions}}' 2>/dev/null | grep -q 'rootless'; then
+    ROOTLESS_DOCKER=true
+fi
+
+if [[ "$ROOTLESS_DOCKER" == true ]]; then
+    SUBID_USER="$(id -un)"
+    SUBID_UID="$(id -u)"
+    SUBUID_TOTAL="$(subid_range_total "$SUBUID_FILE" "$SUBID_USER" "$SUBID_UID")"
+    SUBGID_TOTAL="$(subid_range_total "$SUBGID_FILE" "$SUBID_USER" "$SUBID_UID")"
+    SUBID_FIX="Allocate ranges with: sudo usermod --add-subuids 100000-165535 --add-subgids 100000-165535 $SUBID_USER, then restart the rootless Docker daemon (systemctl --user restart docker)."
+    if [[ "$SUBUID_TOTAL" -eq 0 || "$SUBGID_TOTAL" -eq 0 ]]; then
+        append_check "ROOTLESS_SUBID" "fail" \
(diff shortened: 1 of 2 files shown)


Lightheartdevs:
Thanks for taking on the rootless Docker preflight here; catching subordinate-ID problems during install is the right place to prevent the later `uid/gid map out of range` failure.

Quickest path to merge:
- In `ods/scripts/linux-install-preflight.sh`, avoid using the same `SUBID_FIX` for both the missing-range failure and the small-range warning. For the warning path, an existing allocation can already occupy `100000:1000`, so `sudo usermod --add-subuids 100000-165535 --add-subgids 100000-165535 <user>` may overlap the current range. Either compute a non-overlapping add range or change the warning remediation to tell the operator to extend/reallocate to at least 65536 IDs.
- Extend `ods/tests/test-linux-install-preflight.sh` to assert the small-range warning does not suggest the same fixed range as the missing-entry failure, so the operator hint remains valid for the fixture that already has `100000:1000`.

What I checked:
- Diff/static review of `ods/scripts/linux-install-preflight.sh` and the new `ROOTLESS_SUBID` fixture cases in `ods/tests/test-linux-install-preflight.sh`.
- Detection, JSON status shape, numeric-UID matching, and split-range summing look sound; the blocker is the remediation command for existing-but-too-small ranges.

Thanks again for improving the rootless install path.

tang-vu:
Addressed: the remediation is now split — the **missing-entry failure** keeps the exact `sudo usermod --add-subuids 100000-165535 --add-subgids 100000-165535 <user>` command (safe when nothing is allocated), while the **small-range warning** asks the operator to pick a start above every range already listed in `/etc/subuid`/`/etc/subgid` and extend to ≥65536 IDs, so it can never suggest a range that overlaps the existing allocation (like the fixture's `100000:1000`).

Tests extended per the review: the missing-entry case asserts the fixed range is present, and the small-range case asserts it is **absent** while the ≥65536 guidance is present. 7/7 ROOTLESS_SUBID fixture assertions passing locally.

Lightheartdevs:
Thanks for the follow-up on the Linux preflight rootless check; this now gives users a safer path before Docker fails during image extraction.

What looks good:
- `ods/scripts/linux-install-preflight.sh` only runs the subordinate-ID check for rootless Docker, sums multiple name/UID-keyed ranges, and keeps rootful installs as an explicit skipped pass.
- The remediation now distinguishes no allocation from small existing allocation, so the warn path asks for a non-overlapping extension instead of suggesting a range that may already be partly occupied.
- The added fixtures in `ods/tests/test-linux-install-preflight.sh` cover pass, fail, warn, split ranges, numeric UID entries, and both remediation shapes.

What I checked:
- Reviewed the current diff and prior advisory thread; the latest commit addresses the unsafe fixed-range warning.
- Checked current head `9fb0e1d868b00c090b4d528b336cc93ef57c7ab5`; CI is green across integration, shell lint, Linux/macOS smoke, distro, and env-validation jobs.
- Ran `git diff --check origin/main...origin/pr/1858` locally.

Thanks again for keeping this installer-facing change well covered.

Lightheartdevs (approved):
Audit: rootless-Docker subuid/subgid range validation. NOT install-gating — lives in the standalone linux-install-preflight report tool, not phase 01; only exits nonzero under opt-in --strict, so it cannot block an install that previously succeeded. Rootful Docker → no-op pass. Correct range summing (multi-range, name+numeric keys, inclusive 65536 boundary), sensible non-overlapping remediation. 7 new tests pass; the 2 unrelated failures (JSON contract, podman shim) are pre-existing/environmental and green in CI.

PR #1859, opened 2026-07-17
fix(dashboard-api): bound the usage report date range

tang-vu:
## Problem
`GET /api/usage/report` validates date **shape** and ordering but not **span**. A single authenticated request like `start=0001-01-01&end=9999-12-31` walks `_date_range` across ~3.65 million days: whenever Token Spy is unreachable (or returns an HTTP error — both common states), `_empty_report` materializes one dict per day, so the API builds a multi-million-element `daily` list in memory and serializes it onto the wire. That's an easy accidental-DoS from a misbehaving client or a fat-fingered date picker.

## Fix
- New `MAX_REPORT_RANGE_DAYS = 366` cap (dashboard presets top out well below a year; one full leap year stays supported).
- Oversized spans get the same contract as other bad inputs: HTTP 200 with `source.status="invalid_range"`, a clear detail message, and an **empty** `daily` list.
- `_empty_report` itself now refuses to materialize buckets for reversed/oversized spans, so no invalid_range path can ever build the huge list.

## Tests
- `test_usage_report_rejects_oversized_date_range` — the 0001→9999 request returns `invalid_range` with `daily == []`.
- `test_usage_report_accepts_full_year_range` — 366 inclusive days (leap year) still produces a normal report with 366 buckets.

Full dashboard-api suite: **1173 passed**, 3 pre-existing Windows-host failures unrelated to this change (pass on Linux CI).

### DIFF
--- a/ods/extensions/services/dashboard-api/routers/usage.py
+++ b/ods/extensions/services/dashboard-api/routers/usage.py
@@ -27,6 +27,11 @@ TOKEN_SPY_SERVICE_ID = "token-spy"
 TOKEN_SPY_URL = os.environ.get("TOKEN_SPY_URL", "http://token-spy:8080")
 TOKEN_SPY_API_KEY = os.environ.get("TOKEN_SPY_API_KEY", "")
 TOKEN_SPY_KEY_FILE = Path(os.environ.get("TOKEN_SPY_KEY_FILE", "/data/token-spy/token-spy-api-key.txt"))
+# Inclusive-day span cap for /report. Without it a single request for e.g.
+# 0001-01-01..9999-12-31 makes _empty_report materialize ~3.6 million daily
+# buckets in memory and on the wire. The dashboard presets top out well
+# below a year; one full year is a generous ceiling.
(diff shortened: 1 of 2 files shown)


Lightheartdevs:
Thanks for bounding the usage-report date range; this closes a small but real authenticated DoS path in a way that keeps the existing API contract intact.

What looks good:
- `ods/extensions/services/dashboard-api/routers/usage.py` rejects spans over 366 inclusive days before calling Token Spy, and `_empty_report()` also refuses to materialize oversized fallback buckets defensively.
- The invalid-range response stays consistent with the existing route behavior: HTTP 200, `source.status="invalid_range"`, clear detail, and `daily == []`.
- `ods/extensions/services/dashboard-api/tests/test_usage.py` covers both the pathological 0001-to-9999 range and the accepted full leap-year boundary.

What I checked:
- Diff/static review of the usage router range checks and focused test additions.
- CI is green for this head, including Dashboard API tests, Python lint/type check, `Test Linux / integration-smoke`, matrix smoke, env validation, and secret scan.

Thanks again for hardening this endpoint.

PR #1860, opened 2026-07-17
fix(dashboard-api): stop the Hermes pool guard blocking all phones during a cold connection open

tang-vu:
## Problem
`hermes_bridge._get_connection` holds the **global** `_POOL_GUARD` while calling `_open_connection`, which performs real network I/O: dashboard token fetch, WebSocket connect, and `session.create` — up to the full client timeout (~200s) when Hermes is slow, restarting, or down.

Consequence: one phone's cold open head-of-line blocks **every** other phone's pool lookup — including phones whose pooled connection is warm and ready. That directly contradicts the pool's own contract ("two different-key prompts run in parallel") and turns a single Hermes hiccup into a full ODS Talk outage for all connected phones.

## Fix
- `_POOL_GUARD` is now held only for dict bookkeeping (lookup, evict, insert).
- The open itself runs outside the guard, serialized by a **per-key opening lock**, so two concurrent first-messages from the same phone share one open instead of racing into duplicate Hermes sessions.
- After acquiring the opening lock the pool is re-checked, so the second waiter adopts the connection the first one just opened.
- Housekeeping: the idle sweeper prunes unlocked, unpooled opening locks (they're keyed by cookie hash and would otherwise grow forever); `shutdown_pool` clears them.

## Tests
New `test_hermes_bridge_slow_open_does_not_block_other_keys`:
- while `cold-key`'s open hangs on an event, `warm-key`'s pooled connection is served within 1s (this **times out against the old implementation** — verified by running the test against upstream `hermes_bridge.py`: fails there, passes here);
- a concurrent same-key call shares the in-flight open's result with exactly one `_open_connection` call.

`tests/test_talk.py`: 35 passed (all existing pool contracts — same-key serialization, transparent retry, dead-WS eviction, sweeper behavior — unchanged).

### DIFF
--- a/ods/extensions/services/dashboard-api/hermes_bridge.py
+++ b/ods/extensions/services/dashboard-api/hermes_bridge.py
@@ -202,11 +202,19 @@ class _HermesConnection:
 
 _CONNECTION_POOL: dict[str, _HermesConnection] = {}
 _POOL_GUARD = asyncio.Lock()
+# Per-key creation locks: two concurrent first-messages from the same phone
+# share one open instead of racing, without holding the global guard.
+_OPENING_LOCKS: dict[str, asyncio.Lock] = {}
 _SWEEPER_TASK: asyncio.Task | None = None
 
 
 async def _open_connection(session_key: str) -> _HermesConnection:
-    """Open one fresh WS + run session.create. Caller holds _POOL_GUARD."""
+    """Open one fresh WS + run session.create.
+
+    Caller holds the per-key opening lock, NOT _POOL_GUARD — this does
+    network I/O that can hang for the full client timeout when Hermes is
+    slow or down, and must never stall other phones' pool lookups.
+    """
     timeout_seconds = _request_timeout()
     timeout = aiohttp.ClientTimeout(total=timeout_seconds + 20)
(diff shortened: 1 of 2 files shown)


Lightheartdevs:
Thanks for taking on the Hermes pool head-of-line blocking path; the change preserves same-phone serialization without letting one cold open stall every other phone.

What looks good:
- `ods/extensions/services/dashboard-api/hermes_bridge.py` now keeps `_POOL_GUARD` scoped to dictionary bookkeeping and moves network I/O behind per-key opening locks, which matches the existing pool contract.
- The stale-connection close, idle sweeper cleanup, and `shutdown_pool()` handling keep `_OPENING_LOCKS` from becoming a long-lived leak.
- `test_hermes_bridge_slow_open_does_not_block_other_keys` covers both the warm-key fast path during a stalled cold open and the same-key single-open behavior.

What I checked:
- Diff/static review of `hermes_bridge.py` and the new async regression coverage in `tests/test_talk.py`.
- CI is green for this head, including Dashboard API tests, Python lint/type check, `Test Linux / integration-smoke`, matrix smoke, env validation, and secret scan.

Thanks again for improving the ODS Talk cold-connection behavior.

PR #1861, opened 2026-07-17
fix(windows): redact user/email env values in compose failure reports

tang-vu:
## Problem
`installers/windows/lib/compose-diagnostics.ps1` writes the install failure report that Windows users are explicitly asked to attach to public GitHub issues. Its redaction key set (`KEY|TOKEN|SECRET|PASSWORD|PASS|SALT|AUTH|CREDENTIAL`) misses keys that `.env.schema.json` marks `secret: true`:

- `N8N_USER`
- `LANGFUSE_INIT_USER_EMAIL`
- `LANGFUSE_MINIO_ROOT_USER`

Both the `docker compose config` tail and the global sensitive-value scrub let those values through, so a shared report leaks the operator's login identifiers in cleartext.

## Fix
Add `USER|EMAIL|BEARER` to both places that carry the key set:
- `Get-ODSComposeSensitiveEnvValues` (collects .env values for global scrubbing), and
- `ConvertTo-ODSComposeRedactedLine` (per-line compose-config redactor),

with comments keeping the two in sync. This mirrors the Linux support bundle's redaction set (#1846) so both platforms redact the same key classes.

## Tests
Extended `tests/test-windows-compose-failure-report.sh`: the mocked `docker compose config` and fixture `.env` now include `N8N_USER` + `LANGFUSE_INIT_USER_EMAIL`; the runtime assertion requires both to appear as `[REDACTED]` and fails if any raw value leaks anywhere in the report. Also verified both functions directly on Windows PowerShell 5.1:
```
N8N_USER: [REDACTED]
LANGFUSE_INIT_USER_EMAIL: [REDACTED]
OLLAMA_PORT: 39134        # non-secret values untouched
```
Full test file: 39 passed. PowerShell AST parse: clean.

### DIFF
--- a/ods/installers/windows/lib/compose-diagnostics.ps1
+++ b/ods/installers/windows/lib/compose-diagnostics.ps1
@@ -108,7 +108,12 @@ function Get-ODSComposeSensitiveEnvValues {
             if ($line -notmatch '^([A-Za-z_][A-Za-z0-9_]*)=(.*)$') { continue }
             $key = $Matches[1]
             $value = $Matches[2].Trim().Trim('"').Trim("'")
-            if ($value.Length -ge 4 -and $key -match '(?i)(KEY|TOKEN|SECRET|PASSWORD|PASS|SALT|AUTH|CREDENTIAL)') {
+            # USER|EMAIL|BEARER cover schema secret:true keys the shorter set
+            # missed - N8N_USER, LANGFUSE_INIT_USER_EMAIL, LANGFUSE_MINIO_ROOT_USER -
+            # whose values would otherwise ship in the shareable failure report.
+            # Keep in sync with ConvertTo-ODSComposeRedactedLine below and the
+            # Linux support bundle's redaction set.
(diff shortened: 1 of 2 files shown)


Lightheartdevs:
Thanks for tightening the Windows compose failure report redaction; this is exactly the kind of support-bundle detail that matters once users are asked to attach reports publicly.

What looks good:
- `ods/installers/windows/lib/compose-diagnostics.ps1` now applies the broader sensitive key classes in both value collection and per-line compose config redaction, keeping the two redaction paths aligned.
- The test fixture in `ods/tests/test-windows-compose-failure-report.sh` covers `N8N_USER` and `LANGFUSE_INIT_USER_EMAIL` both in `.env` and mocked compose output, and checks that raw values do not leak anywhere in the report.
- The conservative over-redaction of `USER`/`EMAIL` is appropriate for a shareable diagnostic artifact and matches the recent Linux support-bundle posture.

What I checked:
- Diff/static review of the PowerShell redaction functions and shell-driven report test.
- CI is green for this head, including PowerShell lint on Ubuntu and Windows, `Test Linux / integration-smoke`, matrix smoke, env validation, secret scan, and dashboard jobs.

Thanks again for closing this privacy gap.

PR #1862, opened 2026-07-17
feat(doctor): report rootless Docker subordinate ID allocation

tang-vu:
## Summary
Runtime-diagnostics half of #1719 (the install-preflight half is #1858 — independent PRs, either merges standalone). When `docker info` says the daemon is rootless, `ods doctor` now validates the current user's subordinate UID/GID allocation and surfaces the failure mode *before* users hit `failed to register layer: uid/gid map out of range` on non-root images.

## What the report gains
`runtime.rootless_docker` (bool) and `runtime.rootless_subid_check`:
```json
{"status": "fail|warn|pass|skipped", "user": "alice", "subuid_total": 0, "subgid_total": 65536}
```
- **fail** — no allocation in `/etc/subuid` or `/etc/subgid` (rootless containers can't map user namespaces)
- **warn** — allocation exists but the summed range is under 65536 IDs
- **pass** — ≥65536 both; **skipped** — daemon is rootful
- Entries keyed by numeric UID and multiple ranges per user are handled (summed); `/etc/subgid` is correctly treated as keyed by user, not group.
- warn/fail increments `summary.runtime_warnings` and appends an `autofix_hints` entry with the exact `sudo usermod --add-subuids 100000-165535 --add-subgids 100000-165535 <user>` command + rootless daemon restart step.

## Tests
New `tests/test-ods-doctor-rootless-subid.sh` (registered in `make test`) drives the **full doctor** with `ODS_ASSUME_ROOTLESS=1` + fixture subuid/subgid files and asserts the generated report JSON for the pass/fail/warn matrix, the rootless flag, and hint presence/absence — 8/8 passing locally end-to-end. Skips gracefully on hosts where doctor's dependencies are unavailable.

Refs #1719

### DIFF
--- a/ods/Makefile
+++ b/ods/Makefile
@@ -35,6 +35,9 @@ test: ## Run unit and contract tests
 	@bash tests/test-linux-install-preflight.sh
 	@bash tests/test-linux-host-agent-stop.sh
 	@echo ""
+	@echo "=== ods-doctor rootless subordinate IDs ==="
+	@bash tests/test-ods-doctor-rootless-subid.sh
+	@echo ""
 	@echo "=== AMD/Lemonade contracts ==="
 	@bash tests/contracts/test-amd-lemonade-contracts.sh
 	@bash tests/test-litellm-amd-auth-enforced.sh
--- a/ods/scripts/ods-doctor.sh
+++ b/ods/scripts/ods-doctor.sh
@@ -1382,13 +1382,27 @@ if runtime["docker_daemon"] and not runtime["webui_http"]:
 
 # Rootless daemon without (enough) subordinate IDs: images fail to extract
 # with "uid/gid map out of range" long after doctor reports healthy Docker.
+# Two hints: the fixed 100000-165535 range is only safe to suggest when the
+# user has NO allocation; with an existing-but-small range that command
+# could overlap what is already allocated, so ask for a non-overlapping
+# extension instead.
 _rootless_subid = report["runtime"]["rootless_subid_check"]
-if _rootless_subid["status"] in {"warn", "fail"}:
+if _rootless_subid["status"] == "fail":
     fix_hints.append(
-        "Rootless Docker: subordinate ID ranges for "
-        f"{_rootless_subid['user']} are missing or below 65536 "
+        "Rootless Docker: no subordinate ID allocation for "
+        f"{_rootless_subid['user']} "
         f"(subuids={_rootless_subid['subuid_total']}, subgids={_rootless_subid['subgid_total']}). "
         f"Run: sudo usermod --add-subuids 100000-165535 --add-subgids 100000-165535 {_rootless_subid['user']} "
+        "then restart the rootless Docker daemon (systemctl --user restart docker)."
+    )
+elif _rootless_subid["status"] == "warn":
+    fix_hints.append(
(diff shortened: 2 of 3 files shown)


Lightheartdevs:
Thanks for adding the runtime-side rootless Docker diagnosis to `ods doctor`; surfacing this before users hit layer-extraction failures is valuable.

Quickest path to merge:
- Split the remediation hint in `ods/scripts/ods-doctor.sh` for `fail` versus `warn`. The `warn` case means the user already has a subordinate-ID allocation, and the current fixed `sudo usermod --add-subuids 100000-165535 --add-subgids 100000-165535 <user>` hint can overlap an existing small range such as the test fixture's `100000:1000`. Keep the exact command for the missing-entry case if desired, but make the small-range case say to extend/reallocate to at least 65536 IDs with a non-overlapping range, or compute a safe add range.
- Add/adjust a fixture assertion in `ods/tests/test-ods-doctor-rootless-subid.sh` so the small-range warning does not emit the same potentially overlapping command as the missing-range failure.

What I checked:
- Diff/static review of `ods/scripts/ods-doctor.sh`, the new `rootless_subid_check` report fields, and `ods/tests/test-ods-doctor-rootless-subid.sh`.
- The pass/fail/warn status detection and report wiring look coherent; the concern is only the operator-facing remediation text.

Thanks again for pushing the rootless diagnostics forward.

tang-vu:
Addressed: the autofix hint is now split by status — **fail** (no allocation) keeps the exact `sudo usermod --add-subuids 100000-165535 --add-subgids 100000-165535 <user>` command, while **warn** (existing-but-small allocation, like the fixture's `100000:1000`) instructs the operator to extend to ≥65536 IDs with a start above every range already listed in `/etc/subuid`/`/etc/subgid`, so the suggested command can no longer overlap an existing allocation.

Fixture test updated per the review: it now classifies the emitted hint (`fixed-range` vs `non-overlapping`) and asserts the missing-entry failure carries the exact range while the small-range warning does **not**. 10/10 passing locally end-to-end through the generated report.

Lightheartdevs:
Thanks for the follow-up on the rootless Docker doctor check; the remediation is clearer and safer now.

What looks good:
- `ods/scripts/ods-doctor.sh` reports rootless subordinate-ID state in `runtime.rootless_subid_check` and increments runtime warnings only for warn/fail states.
- The new fix hints correctly split missing allocation from existing-but-small allocation, avoiding a potentially overlapping `100000-165535` suggestion for the warn path.
- `ods/tests/test-ods-doctor-rootless-subid.sh` exercises pass/fail/warn report content and verifies the hint classification.

What I checked:
- Reviewed the current diff and prior advisory thread; the latest commit addresses the overlap concern.
- Checked current head `02d221b8e542a4d181605e10c326f18513084055`; CI is green across integration, shell lint, API/frontend, smoke, distro, and env-validation jobs.
- Ran `git diff --check origin/main...origin/pr/1862` locally.

Thanks again for making this diagnostic actionable without over-prescribing the unsafe case.

Lightheartdevs (approved):
Audit: read-only doctor reporting of rootless subid allocation; 10/10 tests pass. Consistent detection with #1858 (docker info SecurityOptions grep rootless). Nit: reimplements subid_range_total rather than sharing #1858's helper — acceptable for two independent report tools.

PR #1863, opened 2026-07-17
fix(session-cleanup): resolve Python before deleting any session file

tang-vu:
## Problem
`scripts/session-cleanup.sh` deletes a bloated session file first and only then (inside the per-ID wipe loop) resolves the Python interpreter used to clear that session's reference from `sessions.json`. Under `set -euo pipefail`, a failed resolution (`ods_detect_python_cmd` returning 1 — e.g. Windows Git Bash where the resolver rejects WindowsApps store pythons, or a minimal host without python3) kills the script **between the two steps**: the `.jsonl` file is gone, but `sessions.json` still references it, so the OpenClaw gateway keeps pointing at a session file that no longer exists. The temporary `.bak-cleanup` backup is also stranded.

## Fix
- Resolve Python **once, before the file-processing loop**, and verify it can actually run (`-c 'import json'`). No usable interpreter → exit 1 with a clear message and **zero files touched** — fail-loud, but before the destructive step instead of in the middle of it.
- Removes the redundant re-source + re-detection previously executed for every wiped session ID.

## Tests (`tests/test-session-cleanup.sh`, now registered in `make test` — it previously never ran in CI)
- New: a bloated wipe must delete the file, clear its `sessions.json` reference, preserve healthy references, and remove the `.bak-cleanup` backup.
- New ordering-guard test: with a PATH stripped of every python (and `ODS_PYTHON_CMD` unset), cleanup must exit non-zero with **no** files deleted and `sessions.json` untouched. Against the previous script on Linux this catches the bug (files deleted, reference orphaned).
- Full suite: 15/15 passing locally.

The suite's existing CLAUDE.md-compliance checks (no `2>/dev/null`, no `|| true`, inline exit-capture pattern) still pass on the modified script.

### DIFF
--- a/ods/Makefile
+++ b/ods/Makefile
@@ -112,6 +112,9 @@ test: ## Run unit and contract tests
 	@echo ""
 	@echo "=== CLI user-extension dependency resolution ==="
 	@bash tests/test-cli-user-extension-deps.sh
+	@echo ""
+	@echo "=== session cleanup lifecycle ==="
+	@bash tests/test-session-cleanup.sh
 
 bats: ## Run BATS unit tests for shell libraries
 	@echo "=== BATS unit tests ==="
--- a/ods/scripts/session-cleanup.sh
+++ b/ods/scripts/session-cleanup.sh
@@ -35,7 +35,8 @@ usage() {
     echo "  SESSIONS_DIR   Sessions directory (default: \$OPENCLAW_DIR/agents/main/sessions)"
     echo "  MAX_SIZE       Max session file size in bytes (default: 256000)"
     echo ""
-    echo "Exit: 0 (always; missing paths are skipped with a log message)."
+    echo "Exit: 0 on success or when paths are missing (skipped with a log message);"
+    echo "      1 when no usable Python is found (refused before touching any session)."
 }
 
 case "${1:-}" in
--- a/ods/scripts/session-cleanup.sh
+++ b/ods/scripts/session-cleanup.sh
@@ -82,6 +82,23 @@ if [ "$DELETED_COUNT" -gt 0 ] || [ "$BAK_COUNT" -gt 0 ]; then
     echo "[$(date)] Cleaned up $DELETED_COUNT .deleted files, $BAK_COUNT .bak files"
 fi
 
+# ── Python (needed to update sessions.json after wipes) ───────
+# Resolved BEFORE any session file is deleted: under set -e, a failed
+# detection inside the wipe loop used to kill the script after a bloated
+# session file was already removed but before its reference was cleared
+# from sessions.json — leaving the gateway pointing at a missing file.
+PYTHON_CMD="python3"
+if [[ -f "$(dirname "$0")/../lib/python-cmd.sh" ]]; then
+    . "$(dirname "$0")/../lib/python-cmd.sh"
+    PYTHON_CMD="$(ods_detect_python_cmd)" || PYTHON_CMD=""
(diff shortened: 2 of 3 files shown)


Lightheartdevs:
Thanks for moving the Python resolution ahead of the destructive session cleanup path; that is the right failure ordering for this script.

Quickest path to merge:
- Update `ods/scripts/session-cleanup.sh` help text in `usage()`: it still says `Exit: 0 (always; missing paths are skipped with a log message).`, but this PR intentionally adds a no-usable-Python exit 1 before pruning. The behavior looks intentional, so the CLI contract should say missing paths skip, while missing Python fails before any session file is touched.

What I checked:
- Diff/static review of `ods/scripts/session-cleanup.sh`, `ods/tests/test-session-cleanup.sh`, and the `ods/Makefile` test registration.
- The new tests cover both successful bloated-session reference cleanup and the missing-Python ordering guard.

Thanks again for tightening this lifecycle path.

tang-vu:
Updated the `usage()` contract text: missing paths still skip with exit 0, and the new no-usable-Python case is documented as exit 1 before any session file is touched. Test suite re-run locally: 15/15 passing.

Lightheartdevs:
Thanks for the follow-up on the session-cleanup contract; documenting the new fail-before-delete behavior closes the previous review gap.

What looks good:
- `ods/scripts/session-cleanup.sh` resolves and validates Python before any inactive or bloated session files are removed, so a missing interpreter no longer leaves `sessions.json` pointing at deleted files.
- The usage text now matches the intentional exit behavior, and `ods/tests/test-session-cleanup.sh` covers both successful reference cleanup and the no-Python no-mutation guard.

What I checked:
- Reviewed the current two-commit diff and the prior advisory thread; the new head addresses the documented CLI contract issue.
- Checked current head `ae435dc85565a681a6b5d71f4bbb9a8c293d3d1b`; CI is green across integration, shell lint, API/frontend, smoke, distro, and env-validation jobs.
- Ran `git diff --check origin/main...origin/pr/1863` locally.

Thanks again for tightening a recovery path that users only hit when something is already fragile.

Lightheartdevs:
Thanks for rebasing the session-cleanup fix; the fail-before-delete ordering and documented exit contract still line up.

What looks good:
- `ods/scripts/session-cleanup.sh` resolves and validates Python before pruning session files, so a missing interpreter cannot leave `sessions.json` pointing at a file that was already deleted.
- The `usage()` text now matches the intentional exit 1 path for missing Python, while preserving the missing-path skip behavior.
- `ods/tests/test-session-cleanup.sh` covers the successful bloated-session reference cleanup and the no-usable-Python no-mutation guard, and `ods/Makefile` now runs it with the main test target.

What I checked:
- Current diff/static review of the script, test, and Makefile wiring.
- `C:\Program Files\Git\bin\bash.exe tests/test-session-cleanup.sh` passed locally: 15 tests.
- `git diff --check origin/main...pr-1863` passed; current GitHub checks are green, including ShellCheck, Test Linux/integration-smoke, matrix smoke, env validation, Dashboard, and lint jobs.

Thanks again for tightening this recovery path.

Lightheartdevs (approved):
Audit: hazard real — session-cleanup.sh deletes the bloated jsonl before resolving python (line 123 vs 136), so a python-resolution failure under set -e strands a .bak-cleanup the debris pass never reaps. Fix resolves+verifies python before touching any session file; pythonless host now fails loud (exit 1) with zero deletions. Strong end-to-end + ordering tests, newly wired into make test.

PR #1869, opened 2026-07-18
fix(windows): strip only matched quote pairs when parsing .env values

tang-vu:
## Problem
`Get-WindowsODSEnvMap` (`installers/windows/lib/llm-endpoint.ps1`) parses `.env` with:
```powershell
$val = $Matches[2].Trim('"').Trim("'")
```
`.Trim()` removes **all** leading/trailing occurrences of each quote character independently, so:
- `KEY="'literal'"` → `literal` (inner single quotes destroyed)
- `KEY="'"` → empty string
- `KEY=abc"` (value that legitimately ends with a quote, e.g. a password) → `abc`
- `KEY="` → empty instead of a literal `"`

This is the Windows sibling of the Linux `lib/safe-env.sh` matched-pair fix (#1740). It matters because this parser backs `Read-ODSEnv` / `Get-ODSEnvValue` in `ods.ps1` and the LLM endpoint resolver — corrupted values steer real behavior (ports, model names, API keys), not just display.

## Fix
Strip exactly **one matching pair** of surrounding quotes (`"…"` or `'…'`); keep mismatched or lone quotes verbatim — same semantics as `load_env_file` on Linux, so `.env` values now round-trip identically on both platforms.

## Tests
Extended `tests/test-windows-llm-endpoint.sh` (runs under powershell.exe/pwsh) with an 8-case quote matrix (plain, double-quoted, single-quoted, cross-nested quotes, mismatched trailing quote, lone quote, empty). Verified the matrix **fails against the previous parser** (`LONE_DQ = <> expected <">`) and passes with the fix. Existing endpoint-resolver assertions unchanged and passing.

Scope note: a few installer-side one-off value parses use the same double-Trim pattern on low-risk fields (category names, image tags); left untouched here to keep this focused on the shared parser — can follow up if wanted.

### DIFF
--- a/ods/installers/windows/lib/llm-endpoint.ps1
+++ b/ods/installers/windows/lib/llm-endpoint.ps1
@@ -32,7 +32,18 @@ function Get-WindowsODSEnvMap {
             if ($line -match "^#" -or $line -eq "") { return }
             if ($line -match "^([A-Za-z_][A-Za-z0-9_]*)=(.*)$") {
                 $key = $Matches[1]
-                $val = $Matches[2].Trim('"').Trim("'")
+                $val = $Matches[2]
+                # Strip exactly one matching pair of surrounding quotes.
+                # Trimming each quote character independently corrupts values
+                # that legitimately start or end with the other quote: a
+                # double-quoted "'literal'" loses its inner single quotes and
+                # KEY="'" collapses to empty. Mismatched quotes are kept
+                # verbatim, matching lib/safe-env.sh on Linux.
+                if ($val.Length -ge 2 -and (
+                        ($val.StartsWith('"') -and $val.EndsWith('"')) -or
+                        ($val.StartsWith("'") -and $val.EndsWith("'")))) {
+                    $val = $val.Substring(1, $val.Length - 2)
+                }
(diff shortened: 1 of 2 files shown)


Lightheartdevs:
Thanks for fixing the Windows `.env` parser in the shared LLM endpoint path; that is a good place to remove cross-platform drift.

What looks good:
- `ods/installers/windows/lib/llm-endpoint.ps1` now preserves mismatched/lone quotes and strips only one matched surrounding pair, matching `lib/safe-env.sh` semantics.
- The expanded `ods/tests/test-windows-llm-endpoint.sh` quote matrix covers the corrupted cases while keeping the existing endpoint resolver assertions intact.

What I checked:
- Reviewed the PR body, PowerShell parser change, and regression fixture.
- Checked current head `13c655f5dcc3557d8bfdbb06f8167f6325af8d39`; CI is green across PowerShell lint on Windows/Ubuntu plus the broader API, integration, smoke, distro, and env-validation jobs.
- Ran `git diff --check origin/main...origin/pr/1869` locally.

Thanks again for keeping this focused on the shared parser.

PR #1870, opened 2026-07-18
fix(macos-cli): strip only matched quote pairs when parsing .env values

tang-vu:
## Problem
`read_ods_env` in `installers/macos/ods-macos.sh` strips quotes with:
```bash
val=$(echo "$val" | sed 's/^["'\''']//;s/["'\''']$//')
```
The leading and trailing strip are independent and accept **either** quote type, so:
- `KEY=abc"` (value legitimately ending in a quote, e.g. a password) → `abc`
- `KEY="abc'` (mismatched) → `abc` (cut on both ends)
- `KEY="` → empty instead of a literal quote

These values are exported as the `ENV_*` variables the macOS CLI acts on, so the corruption affects behavior, not just display.

## Fix
Strip exactly **one matching pair** of surrounding quotes (`"…"` or `'…'`); mismatched or lone quotes stay verbatim. Same semantics as `lib/safe-env.sh` on Linux (#1740) and the Windows parser fix (#1869) — `.env` values now round-trip identically on all three platforms.

## Tests
New `tests/test-macos-env-quote-strip.sh` (registered in `make test`): extracts `read_ods_env` and asserts an 8-case quote matrix. Against the previous parser 3 cases fail (mismatched trailing quote, cross-quoted, lone quote); with the fix 9/9 pass.

### DIFF
--- a/ods/Makefile
+++ b/ods/Makefile
@@ -72,6 +72,9 @@ test: ## Run unit and contract tests
 	@echo "=== macOS model download retry contract ==="
 	@bash tests/test-macos-download-retries.sh
 	@echo ""
+	@echo "=== macOS CLI .env quote handling ==="
+	@bash tests/test-macos-env-quote-strip.sh
+	@echo ""
 	@echo "=== fleet-multi-distro keep-going on docker pull failure ==="
 	@bash tests/test-fleet-multi-distro-keep-going.sh
 	@echo ""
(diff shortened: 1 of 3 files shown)


Lightheartdevs:
Thanks for tightening the macOS CLI `.env` reader so it matches the Linux quote contract.

What looks good:
- `ods/installers/macos/ods-macos.sh` stops stripping leading and trailing quote characters independently, which preserves mismatched and lone quotes instead of silently corrupting values.
- `ods/tests/test-macos-env-quote-strip.sh` covers plain, matched single/double quote, inner quote, mismatched, lone quote, and empty quote cases, and it is wired into `make test`.

What I checked:
- Reviewed the PR body, `read_ods_env()` diff, new test fixture, and Makefile wiring.
- Checked current head `a172b26ad64f8078ce5cd096474b03e3ca42d749`; CI is green across shell lint, macOS/Linux smoke, integration, API/frontend, distro, and env-validation jobs.
- Ran `git diff --check origin/main...origin/pr/1870` locally.

Thanks again for closing the platform parity gap here.

PR #1871, opened 2026-07-18
fix(dashboard-api): strip only matched quote pairs in model info env parsing

tang-vu:
## Problem
`get_model_info` (`helpers.py`) parses `.env` with `value.strip().strip('\"\'')`. Python's `str.strip` removes **any run of either quote character from both ends**, so:
- `LLM_MODEL="'literal'"` → `literal` (inner quotes destroyed)
- a value legitimately ending in a quote is truncated
- `KEY="` collapses to empty

The parsed `LLM_MODEL` / `GGUF_FILE` / `MAX_CONTEXT` values feed the dashboard's model card, so corruption is user-visible.

## Fix
Strip exactly **one matching pair** of surrounding quotes; mismatched or lone quotes stay verbatim. This is the dashboard-api sibling of the same matched-pair contract now applied on Linux (`lib/safe-env.sh`, #1740), Windows (#1869), and macOS (#1870) — `.env` values round-trip identically everywhere.

(Related: #1746 consolidates the *single-key* env readers into `env_file.py`; that reader has the same strip pattern and can adopt this semantics when it lands — kept separate to avoid cross-PR churn.)

## Tests
Two new cases in `tests/test_helpers.py`: double-quoted value keeps inner single quotes; mismatched `"…'` value is preserved verbatim. `test_helpers.py`: **78 passed**.

### DIFF
--- a/ods/extensions/services/dashboard-api/helpers.py
+++ b/ods/extensions/services/dashboard-api/helpers.py
@@ -618,7 +618,15 @@ def get_model_info() -> Optional[ModelInfo]:
                     if "=" not in line or line.lstrip().startswith("#"):
                         continue
                     key, value = line.split("=", 1)
-                    env_values[key.strip()] = value.strip().strip('"\'')
+                    value = value.strip()
+                    # Strip exactly one matching pair of surrounding quotes.
+                    # str.strip("\"'") removes any run of either quote from
+                    # both ends, so a value legitimately ending in a quote is
+                    # truncated and "'literal'" loses its inner quotes. Keep
+                    # mismatched quotes verbatim, matching lib/safe-env.sh.
+                    if len(value) >= 2 and value[0] == value[-1] and value[0] in "\"'":
+                        value = value[1:-1]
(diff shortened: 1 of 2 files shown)


Lightheartdevs:
Thanks for carrying the matched-quote `.env` contract into the dashboard model card path.

What looks good:
- `ods/extensions/services/dashboard-api/helpers.py` now strips only one matching quote pair, so display/model metadata no longer destroys inner quotes or lone trailing quote characters.
- The added `test_helpers.py` cases cover both the normal quoted model value and the mismatched-quote preservation behavior, matching the Linux safe-env contract.

What I checked:
- Reviewed the PR body, focused diff, and current `get_model_info()` parsing context.
- Checked current head `79ab0d7641e0cabc6b0194581702100f74b69745`; CI is green across `api`, `integration-smoke`, lint, smoke, distro, and env-validation jobs.
- Ran `git diff --check origin/main...origin/pr/1871` locally.

Thanks again for keeping this scoped.

PR #1872, opened 2026-07-18
fix(dashboard-api): parse quoted ODS_VERSION in update dry-run

tang-vu:
## Problem
`GET /api/update/dry-run` re-parses `ODS_VERSION` from `.env` with a plain `.strip()` — unlike `_read_current_version` (used by `/api/version`), it never strips surrounding quotes. With `ODS_VERSION="1.3.0"` in `.env`:

- `current` stays `"1.3.0"` (with quotes),
- in `_parts()` every dotted component fails `isdigit()` (`'"1'`, `'0"'`), so the current version degrades to `[0,0,0]`,
- `update_available` reports **true against any GitHub release**, and the endpoint shows a bogus `current_version`.

Meanwhile `/api/version` (which does strip quotes) reports the correct version — the two endpoints disagree about the same install.

## Fix
Strip one matching pair of surrounding quotes in the dry-run parse, mirroring `_read_current_version`'s handling (with matched-pair semantics per the platform-wide quote contract: #1740/#1869/#1870/#1871). No contract changes — the `0.0.0` no-files fallback and the `.version`/JSON fallbacks behave exactly as the existing tests lock them.

## Tests
New `test_update_dry_run_parses_quoted_ods_version` — fails against the previous code (`current_version == '"1.3.0"'`), passes with the fix. `tests/test_updates.py`: **26 passed**.

### DIFF
--- a/ods/extensions/services/dashboard-api/routers/updates.py
+++ b/ods/extensions/services/dashboard-api/routers/updates.py
@@ -291,6 +291,12 @@ async def get_update_dry_run():
         for line in _read_utf8(env_file).splitlines():
             if line.startswith("ODS_VERSION="):
                 current = line.split("=", 1)[1].strip()
+                # Match _read_current_version's quote handling: a quoted
+                # ODS_VERSION would otherwise reach the semver comparison
+                # with its quotes attached, so every digit fails isdigit()
+                # and the update verdict is computed against 0.0.0.
+                if len(current) >= 2 and current[0] == current[-1] and current[0] in "\"'":
+                    current = current[1:-1]
                 break
     if current == "0.0.0" and version_file.exists():
         try:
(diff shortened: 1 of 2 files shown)


Lightheartdevs:
Thanks for tightening the update dry-run version path; this is a small fix in exactly the endpoint where the user-visible version verdict can drift.

What looks good:
- `ods/extensions/services/dashboard-api/routers/updates.py` now applies the same matched surrounding-quote handling that `/api/version` already uses before `_normalize_version()` and semver comparison.
- The regression in `ods/extensions/services/dashboard-api/tests/test_updates.py` covers the quoted `ODS_VERSION=1.3.0` case without changing the `.version` fallback contract.

What I checked:
- Reviewed the PR body, diff, and surrounding update-router code.
- Checked current head `5f5424297d2d2b911d56cb2c659050983d922c3e`; CI is green across `api`, `integration-smoke`, lint, smoke, distro, and env-validation jobs.
- Ran `git diff --check origin/main...origin/pr/1872` locally.

Thanks again for keeping the dashboard/version contracts aligned.

Lightheartdevs (approved):
Audit: dry-run ODS_VERSION parse now handles quoted values (double/single/none), unquoted regression covered; only two parse sites exist, none missed. Follow-up noted: share the quote-strip helper with _read_current_version for mismatched-quote parity.

PR #1877, opened 2026-07-18
fix: STRICT mode test coverage and tier-aware model in first-boot-demo

### LINKED ISSUE #1875: fix: STRICT mode untested in extension-runtime-check and hardcoded model name breaks first-boot-demo on T0/T1 tiers
## Summary

While auditing the codebase, I found two medium-priority gaps that can both result in silent failures with little or no diagnostic output. One affects test coverage, while the other impacts the first-run user experience. Neither appears to be covered by an existing issue or PR.

---

### Suggested Fix

Add a fixture-based test that:

- Creates a minimal ODS root in a temporary directory with `lib/service-registry.sh` and a manifest for a single extension that includes a health endpoint.
- Uses a mock `docker` executable that reports the extension container as running.
- Uses a mock `curl` executable that always exits with a failure status to simulate a failed health probe.
- Verifies that the script exits successfully (`0`) in the default non-strict mode.
- Verifies that the script exits with status `1` when `EXTENSION_RUNTIME_CHECK_STRICT=1` is set.

---

### Suggested Fix

Resolve the demo model from the configured environment after the LLM URL configuration:

```bash
# Use the installed model from .env if available; fall back to the T3 default.
DEMO_MODEL="${LLM_MODEL:-Qwen/Qwen2.5-32B-Instruct-AWQ}"
```

Then replace the hardcoded model name in all three API request payloads with:

```json
"model": "$DEMO_MODEL"
```

This preserves the current behavior when no configured model is available while allowing the demo to work correctly on installations that use different LLMs.

### Use Case

A developer refactors the exit logic in `extension-runtime-check.sh` and inadvertently removes the strict-mode exit block. The test suite still passes because no test exercises the `had_health_fail` code path. An operator relying on `EXTENSION_RUNTIME_CHECK_STRICT=1` in a CI pipeline begins receiving exit code `0` even when extensions are unhealthy, with no indication of when or why the behavior changed.

---

### Use Case

A user installs ODS on a 16 GB RAM laptop (Tier 0). During installation, ODS correctly selects and downloads a smaller quantized model appropriate for the hardware. Installation completes successfully, and the user runs:

```bash
./scripts/first-boot-demo.sh
```

All three demos print:

```text
✗ No response from LLM
```

making the installation appear broken. In reality, the demo is requesting a hardcoded 32B model that was never installed. The actual cause is hidden from the user, often resulting in unnecessary debugging or a support issue, even though the LLM stack is functioning correctly.

### Proposed Solution

_No response_

### Alternatives Considered

_No response_

IronicRayquaza:
## Summary

Closes #1875.

This PR addresses two medium-priority issues identified during a codebase audit. One adds regression coverage for a documented but previously untested exit path, while the other fixes the first-boot demo so it automatically uses the model configured during installation instead of assuming a high-tier deployment.

### 1. `ods/tests/test-extension-runtime-check.sh` — Add `STRICT` mode regression tests

`scripts/extension-runtime-check.sh` documents `EXTENSION_RUNTIME_CHECK_STRICT=1` as a public interface that causes the script to exit with status `1` when a running extension fails its health probe. Prior to this change, the test suite never exercised this code path.

Added a self-contained fixture that runs entirely under `/tmp` and:

- Copies `lib/service-registry.sh` into a minimal temporary ODS root.
- Creates a valid `manifest.yaml` for a single optional extension with a `health` endpoint and stub `compose.yaml`.
- Provides a mock `docker` executable that reports the extension container as running.
- Provides a mock `curl` executable that always exits with failure, simulating a failed health probe.
- Verifies that the script exits with status `0` in the default (non-strict) mode while emitting a `[BAD]` status.
- Verifies that the script exits with status `1` when `EXTENSION_RUNTIME_CHECK_STRICT=1` is enabled while still emitting a `[BAD]` status.
- Cleans up the temporary fixture using a `trap` on exit.

This adds regression coverage for both the default and strict execution paths of the runtime check.

---

### 2. `ods/scripts/first-boot-demo.sh` — Use the configured model from `.env`

The demo previously hardcoded the model name:

```text
Qwen/Qwen2.5-32B-Instruct-AWQ
```

for all three API requests. This only works on higher-tier installations where that model is deployed. On T0/T1 installations, the installer configures a different model, causing every demo request to fail with an unknown model error that was subsequently hidden by the existing error handling.

The script already loads `LLM_MODEL` from `.env` through the service registry initialization, but the value was never used.

Added after the URL configuration block:

```bash
# Use the installed model from .env if available; fall back to the T3 default.
DEMO_MODEL="${LLM_MODEL:-Qwen/Qwen2.5-32B-Instruct-AWQ}"
```

Replaced all three hardcoded model names (chat, code assistant, and streaming demos) with `${DEMO_MODEL}`.

The fallback preserves the previous behavior for users running the script outside the normal installer flow while allowing installed systems to automatically use the model that was actually configured.

## Testing

- `bash -n` passes on both modified shell scripts.
- `git diff --check` reports no whitespace issues.
- Verified `bash ods/tests/test-extension-runtime-check.sh` passes locally, including all newly added `STRICT` mode assertions.

## Files Changed

| File | Change |
|------|--------|
| `ods/tests/test-extension-runtime-check.sh` | Add fixture-based regression tests covering both default and `EXTENSION_RUNTIME_CHECK_STRICT=1` behavior |
| `ods/scripts/first-boot-demo.sh` | Introduce `DEMO_MODEL` from `.env` and replace hardcoded model names in all three demo API requests |

### DIFF
--- a/ods/scripts/first-boot-demo.sh
+++ b/ods/scripts/first-boot-demo.sh
@@ -37,6 +37,9 @@ WHISPER_URL="${WHISPER_URL:-http://localhost:${SERVICE_PORTS[whisper]:-9000}}"
 PIPER_URL="${PIPER_URL:-http://localhost:${SERVICE_PORTS[tts]:-8880}}"
 N8N_URL="${N8N_URL:-http://localhost:${SERVICE_PORTS[n8n]:-5678}}"
 WEBUI_URL="${WEBUI_URL:-http://localhost:${SERVICE_PORTS[open-webui]:-3000}}"
+# Use the installed model from .env if available; fall back to a widely-known default
+# so the demo works out-of-the-box on any tier without hardcoding a specific model name.
+DEMO_MODEL="${LLM_MODEL:-Qwen/Qwen2.5-32B-Instruct-AWQ}"
 
 QUICK_MODE=false
 ALL_MODE=false
@@ -193,12 +196,8 @@ demo "Asking your local AI a question..."
 
 RESPONSE=$(curl -sf "${LLM_URL}/v1/chat/completions" \
     -H "Content-Type: application/json" \
-    -d '{
-        "model": "Qwen/Qwen2.5-32B-Instruct-AWQ",
-        "messages": [{"role": "user", "content": "In one sentence, what makes local AI special?"}],
-        "max_tokens": 100,
-        "temperature": 0.7
-    }' 2>/dev/null | jq -r '.choices[0].message.content' 2>/dev/null || echo "")
+    -d "{\"model\": \"${DEMO_MODEL}\", \"messages\": [{\"role\": \"user\", \"content\": \"In one sentence, what makes local AI special?\"}], \"max_tokens\": 100, \"temperature\": 0.7}" \
+    2>/dev/null | jq -r '.choices[0].message.content' 2>/dev/null || echo "")
 
 if [[ -n "$RESPONSE" && "$RESPONSE" != "null" ]]; then
     echo ""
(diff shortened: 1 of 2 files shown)


Lightheartdevs:
Thanks for tightening up both the strict-mode contract and the first-boot model path here.

What looks good:
- `ods/tests/test-extension-runtime-check.sh` now exercises the documented `EXTENSION_RUNTIME_CHECK_STRICT=1` behavior with a hermetic docker/curl fixture, including both the default non-blocking exit and the strict failure exit.
- `ods/scripts/first-boot-demo.sh` now uses the installed `LLM_MODEL` alias loaded from `.env`, which matches the repo's model-management contract and avoids asking lower-tier installs for the old hardcoded 32B model.
- The PR keeps the runtime change small and preserves the previous fallback when no configured model is available.

What I checked:
- Reviewed the PR diff, linked issue #1875, changed files, and surrounding model-routing/service-registry context.
- Verified GitHub checks are green for shell lint, Linux integration smoke, matrix smoke, dashboard frontend/API, env validation, secret scan, and the review workflow.
- Ran locally on the PR head: `git diff --check origin/main...HEAD`, `bash -n ods/scripts/first-boot-demo.sh ods/tests/test-extension-runtime-check.sh`, and `bash ods/tests/test-extension-runtime-check.sh` (9 passed, 0 failed).

Lightheartdevs (approved):
Audit: first-boot-demo now uses the installed model (loads .env at line 31, DEMO_MODEL=${LLM_MODEL:-fallback} at 42) instead of a hardcoded 32B that fails on most tiers — genuinely tier-aware. Adds STRICT-mode extension-runtime-check tests (9/9 pass). Rebases clean on main (no conflict with merged #1853); MERGEABLE.

PR #1880, opened 2026-07-18
fix(macos): correct Qwen3-30B-A3B SHA256 in macOS tier map

bokiko:
## Summary

The macOS tier map's tier-3 `GGUF_SHA256` diverged from the canonical tier map despite the header's "keep values byte-identical" contract (`installers/macos/lib/tier-map.sh` line 8). The canonical value (`9f1a2470…`, `installers/lib/tier-map.sh`) matches the HuggingFace LFS oid for `unsloth/Qwen3-30B-A3B-GGUF/Qwen3-30B-A3B-Q4_K_M.gguf`; the macOS value (`84b5f7f1…`) matches nothing.

Impact: on the tier-map fallback path (catalog selector disabled via `ODS_DISABLE_CATALOG_MODEL_SELECTOR=true`, or no usable Python), a macOS tier-3 install downloads the full 18.6 GB model and then fails checksum verification.

This PR corrects the checksum and adds `tests/test-tier-map-parity.sh`, which asserts `GGUF_SHA256` and `GGUF_URL` agree between the two tier maps for every GGUF both define (15 fields today). The existing `test-model-integrity.sh` only checks that checksums exist in the canonical map, so this drift was invisible to CI. The new test fails on the previous commit with exactly this mismatch.

## AI Assistance

Claude (Claude Code) assisted with drafting the fix and the parity test. The author reviewed the final diff and ran the validation below.

## Release Lane

- [ ] Stable hotfix targeting `release/2.5.x`
- [x] Mainline change targeting `main`
- [ ] Next-minor work targeting the next feature/minor release
- [ ] Not sure; reviewer should help classify

Stable hotfix reason:

```text
n/a
```

## Changed Surface

- [ ] Docs only
- [ ] Tests only
- [ ] Dashboard UI
- [ ] Dashboard API / host agent
- [x] Installer / bootstrap / lifecycle
- [ ] Docker Compose / service manifests
- [ ] Model routing / Hermes / capabilities
- [ ] Network exposure / auth / proxy
- [ ] Dependencies / runtime wiring

## Risk And Validation

- Risk level: Low
- Validation run:
  - [x] `git diff --check`
  - [ ] Markdown/link sanity for docs
  - [x] Focused tests listed below
  - [ ] Dashboard lint/test/build
  - [ ] Extension audit / compose validation
  - [ ] Release-grade fleet or scoped hardware validation
  - [ ] Stable-lane patch validation, if targeting `release/2.5.x`

Focused tests:

[...]

PR #1884, opened 2026-07-19
fix(installer): --tier override now bounds catalog model selection

### LINKED ISSUE #1881: bug(installer): --tier override does not constrain model selection — catalog selector overwrites it
### Description

`./install.sh --tier 2 --non-interactive` on a Mac Mini M4 Pro (48GB) logs `Selected tier: 2 (Prosumer)` but installs `qwen3.5-27b` — the tier override is accepted for tier selection yet has no effect on which model is installed.

### Root cause

`installers/macos/install-macos.sh` first resolves the tier-map model for the selected tier (`resolve_tier_config`), then runs `scripts/select-model.py`, whose output overwrites the model via `load_model_selector_env_from_output`.

`select-model.py` accepts `--tier` (arg parser, ~line 400) but only consults it for cloud detection (`{"cloud","0","t0"}`, ~line 71) and the nv-ultra special case (~line 306). Numeric tiers 1–3 never constrain model size, so the catalog policy (`context-aware-largest-capable-general-v1`) picks the largest RAM-fit model regardless of the requested tier.

The installer's own usage text says `./install-macos.sh --tier 3  # Force tier 3`, which implies the opposite behavior. The only way to make `--tier` control the model is the undocumented `ODS_DISABLE_CATALOG_MODEL_SELECTOR=true`.

### Steps to reproduce

1. Apple Silicon Mac with enough RAM that catalog fit exceeds the requested tier's model (e.g. 48GB).
2. `./install.sh --tier 2 --non-interactive`
3. Observe `Selected tier: 2` in the log, but `MODEL_RECOMMENDED_MODEL=qwen3.5-27b` / `MODEL_RECOMMENDATION_SOURCE=catalog_fit_pre_download` in the generated `.env`, and the 27B GGUF downloaded.

### Expected

Either `--tier N` bounds the model selection (selector treats the tier's model/size as a ceiling), or the flag documentation states that model choice is catalog-driven and `--tier` only affects service defaults.

### Related observation

The catalog fit validates the model at 32768 context ("needs about 20GB including context/KV"), then the installer raises runtime context to 65536 for Hermes *after* the fit calculation (`MODEL_RECOMMENDED_CONTEXT=65536`), so the KV budget the fit approved is not the one the server runs with.

Happy to PR whichever direction is preferred. Found during a field install @ e100a6f, macOS 26.5.1.

vaibhavsrv:
## Summary

Fixes #1881. `--tier N` was accepted by the installer but had no effect on which
model got installed: resolve_tier_config() sets the tier's designated model, but
scripts/select-model.py's catalog pick unconditionally overwrote it afterward.
The selector only ever consulted --tier for cloud detection and one NV_ULTRA
arm64 special case — never for bounding model size — so on generous-RAM hosts
the catalog selector always upgraded past the requested tier. Same bug exists
in both the macOS and Linux installer paths.

Added an optional --max-size-mb ceiling to select-model.py's rank_models(),
wired both installer call sites to pass their resolved LLM_MODEL_SIZE_MB as
that ceiling. Default stays 0 (unbounded), so any other caller of the script
keeps today's behavior unless it opts in. The fallback path now prefers the
smallest model that still respects the ceiling, and the recommendation reason
notes when a ceiling constrained the pick.

Does not touch the separate KV-budget-vs-runtime-context mismatch mentioned in
the issue's "Related observation" — that's a different code path, left for a
follow-up.

## AI Assistance

Used Claude to trace the root cause across both installer paths, implement the
--max-size-mb fix, and write/verify the regression tests. I reviewed the full
diff, ran the test suite locally, and I'm responsible for this PR.

## Release Lane

## Changed Surface

- [x] Tests only (partially — test file also updated)

## Risk And Validation

- Risk level: High
- Validation run:

Commands/results:

\```
bash tests/test-tier-map.sh
127 passed, 0 failed
\```

- [ ] Release-grade fleet or scoped hardware validation — not run, no Mac/GPU
  hardware available to me for a real install smoke test.

## Operational Change Check

  release-grade macOS/Linux hardware install smoke. Selector-level tests
  (including 3 new regression tests covering this exact bug) pass; a maintainer
  or reviewer with hardware access should confirm an actual install run before
  this ships in a release.

## Notes For Reviewers

Reproduced the reported bug exactly as described (tier 2 on 48GB Apple unified
picked qwen3.5-27b instead of the tier's qwen3.5-9b) before writing the fix, and
confirmed the fix resolves it. Happy to adjust the tolerance logic in
size_within_ceiling() if maintainers prefer a stricter/looser match.

### DIFF
--- a/ods/installers/macos/install-macos.sh
+++ b/ods/installers/macos/install-macos.sh
@@ -677,6 +677,7 @@ if [[ "${ODS_DISABLE_CATALOG_MODEL_SELECTOR:-false}" != "true" && "$SELECTED_TIE
                 --ram-gb "${SYSTEM_RAM_GB:-0}" \
                 --profile "${MODEL_PROFILE_EFFECTIVE:-${MODEL_PROFILE:-qwen}}" \
                 --tier "$SELECTED_TIER" \
+                --max-size-mb "${LLM_MODEL_SIZE_MB:-0}" \
                 --host-arch "$(uname -m 2>/dev/null || echo unknown)" \
                 --installable-only \
                 --env 2>>"$ODS_LOG_FILE" || true)"
--- a/ods/installers/phases/02-detection.sh
+++ b/ods/installers/phases/02-detection.sh
@@ -551,6 +551,7 @@ if [[ "${ODS_DISABLE_CATALOG_MODEL_SELECTOR:-false}" != "true" && "${TIER:-}" !=
                 --ram-gb "${RAM_GB:-0}" \
                 --profile "${MODEL_PROFILE_EFFECTIVE:-${MODEL_PROFILE:-qwen}}" \
                 --tier "${TIER:-1}" \
+                --max-size-mb "${LLM_MODEL_SIZE_MB:-0}" \
                 --host-arch "${HOST_ARCH:-unknown}" \
                 --installable-only \
                 --env 2>>"$LOG_FILE" || true)"
--- a/ods/scripts/select-model.py
+++ b/ods/scripts/select-model.py
@@ -249,14 +249,33 @@ def score_model(model: dict[str, Any], capacity_gb: float, profile: str) -> floa
     return specialty_weight + family_bonus + context_bonus + capability - headroom_penalty
 
 
+def size_within_ceiling(model: dict[str, Any], max_size_mb: float) -> bool:
+    """True if `model` respects an optional tier size ceiling.
+
+    `max_size_mb` <= 0 means "no ceiling" (unbounded, current behavior).
+    A small tolerance absorbs rounding differences between the tier map's
+    declared LLM_MODEL_SIZE_MB and the catalog's size_mb for the same
+    model, so the tier's own designated model is never excluded by its
+    own ceiling.
+    """
+    if max_size_mb <= 0:
+        return True
+    size_mb = float(model.get("size_mb") or 0)
+    tolerance_mb = max(max_size_mb * 0.02, 64.0)
+    return size_mb <= max_size_mb + tolerance_mb
+
+
 def rank_models(catalog: list[dict[str, Any]], capacity_gb: float, profile: str,
                 installable_only: bool, backend: str, memory_type: str,
-                vram_mb: int, ram_gb: int, host_arch: str) -> list[dict[str, Any]]:
+                vram_mb: int, ram_gb: int, host_arch: str,
+                max_size_mb: float = 0) -> list[dict[str, Any]]:
     candidates = []
     for model in catalog:
         if installable_only and not model.get("gguf_url"):
             continue
         if not family_allowed(model, profile):
+            continue
+        if not size_within_ceiling(model, max_size_mb):
             continue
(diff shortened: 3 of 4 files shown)


Lightheartdevs:
Thanks for tightening the tier override path here; this addresses the confusing installer case where `--tier 2` was accepted but the catalog still selected a much larger model.

Human decision point:
- The implementation looks coherent for the reported bug: both macOS and Linux selector call sites now pass the tier-map `LLM_MODEL_SIZE_MB` ceiling, and `select-model.py` keeps the old unbounded behavior when `--max-size-mb` is absent.
- Because this touches high-risk installer/model-routing behavior and GitHub reports no checks for this head, I would treat merge as a maintainer call: either wait for CI/focused `tests/test-tier-map.sh` in a Bash environment, or merge with explicit scoped install/model-selection validation before release.

What I checked:
- Repo context pass across README, architecture docs, contributor guidance, high-risk change map, Makefile targets, CI workflows, source layout, and recent `main` history.
- PR title/body/diff, linked issue #1881, changed files, existing comments, and head SHA idempotency marker state.
- `git diff --check origin/main...HEAD` passed, and `python -m py_compile ods/scripts/select-model.py` passed.
- Direct selector checks on the PR head: Apple unified 48GB tier 2 with `--max-size-mb 5760` selects `qwen3.5-9b`; the same uncapped selector preserves the existing `qwen3.5-27b` auto-upgrade; tier 0 with `--max-size-mb 1500` selects `qwen3.5-2b`; NVIDIA 48GB tier 2 with `--max-size-mb 5760` selects `qwen3.5-9b`; Gemma tier 2 with `--max-size-mb 5340` selects `gemma-4-e4b-it`.
- I could not run `tests/test-tier-map.sh` or Bash syntax checks on this Windows runner because `bash` resolves to WSL and `/bin/bash` is missing. `gh pr checks` and commit check-runs both report no checks for this head.

Thanks again for pushing this fix forward.

bokiko:
Tested this branch on the machine from #1881 (Mac Mini M4 Pro 48GB, macOS 26.5.1).

The ceiling works when `--max-size-mb` receives a value, but on macOS it never does: `install-macos.sh` sources `installers/macos/lib/tier-map.sh`, which doesn't define `LLM_MODEL_SIZE_MB` for any tier (0 occurrences on this branch, vs 23 in the canonical `installers/lib/tier-map.sh`). So `--max-size-mb "${LLM_MODEL_SIZE_MB:-0}"` expands to 0 — documented in `size_within_ceiling` as "no ceiling" — and the selector still picks the 27B for the reported case:

```
python3 scripts/select-model.py --catalog config/model-library.json \
  --backend apple --memory-type unified --vram-mb 0 --ram-gb 48 \
  --profile qwen --tier 2 --host-arch arm64 --installable-only --env
→ LLM_MODEL="qwen3.5-27b"
```

The Linux path (`02-detection.sh`) looks right since the canonical map defines the sizes. Two ways to complete the macOS side: add `LLM_MODEL_SIZE_MB` entries to the macOS tier map (the parity test from #1880 could be extended to keep them in sync), or source sizes from the canonical map. Happy to re-validate a revision on this hardware.

PR #1893, opened 2026-07-19
fix(windows): check the active inference port

gabsprogrammer:
## Summary

Makes Windows preflight check the host inference port selected by the active backend instead of always probing Docker llama-server's `11434`.

## Root Cause

On Windows AMD/Lemonade, a foreign process could occupy `AMD_INFERENCE_PORT` while Phase 04 checked only `11434`. The installer then failed later at runtime health. A naive backend-aware check also creates a reinstall regression: it can mistake the ODS-managed Lemonade listener for an external conflict.

## Backend-Aware Resolution

| Mode | Port checked |
| --- | --- |
| AMD local | Process `AMD_INFERENCE_PORT`, existing `.env`, then Lemonade default |
| NVIDIA/CPU local | Process `OLLAMA_PORT` / `LLAMA_SERVER_PORT`, existing `.env`, then `11434` |
| Cloud | No local inference port |

- Accepts only TCP ports in `1..65535`; malformed values fall back to the backend default.
- Reports process name and PID for real conflicts.
- Labels the AMD endpoint as `Lemonade (LLM)`.
- Reuses a listener only when its PID matches a known ODS Lemonade process.
- Continues to reject an unrelated process on the same AMD port.

## Regression Coverage

- AMD, Docker, and cloud defaults
- contract-supplied Lemonade default
- process and persisted `.env` precedence
- invalid override fallback
- real loopback listener detection with owning PID
- managed Lemonade PID accepted
- foreign PID rejected
- free port remains non-conflicting

## Validation

- `powershell.exe ... tests/test-windows-port-preflight.ps1` - passed
- `bash tests/test-windows-port-preflight.sh` - passed
- `bash tests/test-windows-installer-flags.sh` - **30 passed**
- `bash tests/contracts/test-amd-lemonade-contracts.sh` - **63 passed**
- `bash tests/contracts/test-installer-hardening.sh` - passed
- PowerShell parser checks - passed
- `git diff --check` - passed
- GitHub Windows/Ubuntu PowerShell, dashboard, integration, distro, shell, Python, env-schema, and secret checks - passed

## Scope

This PR never terminates an unrelated process and never silently rewrites a configured port.

### DIFF
--- a/ods/installers/windows/phases/04-requirements.ps1
+++ b/ods/installers/windows/phases/04-requirements.ps1
@@ -138,6 +138,26 @@ function Get-WindowsODSLemonadeProcesses {
     }
 }
 
+function Test-WindowsODSLemonadeOwnsPort {
+    <#
+    .SYNOPSIS
+        Return true when a listening port belongs to a known Lemonade process.
+    #>
+    param(
+        [hashtable]$PortResult,
+        [object[]]$LemonadeProcesses = @()
+    )
+
+    if (-not $PortResult -or -not $PortResult.InUse -or [int]$PortResult.ProcessId -le 0) {
+        return $false
+    }
+
+    $listenerPid = [int]$PortResult.ProcessId
+    return [bool]@($LemonadeProcesses | Where-Object {
+        $_.ProcessId -and [int]$_.ProcessId -eq $listenerPid
+    }).Count
+}
+
 function Stop-WindowsODSLemonadePortConflicts {
     <#
     .SYNOPSIS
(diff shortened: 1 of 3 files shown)


Lightheartdevs:
Thanks for tightening the Windows preflight around the active inference backend instead of assuming Docker llama-server is always on `11434`.

What looks good:
- `ods/installers/windows/phases/04-requirements.ps1` now resolves the LLM port from backend, process env, persisted `.env`, and backend defaults, while skipping local LLM checks in cloud mode.
- The AMD/Lemonade reuse path is conservative: a busy Lemonade port is accepted only when the owning PID matches a known ODS Lemonade process, and unrelated listeners still fail preflight.
- `ods/tests/test-windows-port-preflight.ps1` covers defaults, overrides, persisted ports, invalid fallback, live listener PID detection, managed Lemonade ownership, and foreign-listener rejection.

What I checked:
- Reviewed the PR body, current head, changed files, and diff against `main`.
- Checked `gh pr checks 1893`: PowerShell lint on Windows/Ubuntu, shell/Python lint, dashboard API/frontend, integration smoke, distro matrix, env validation, and secret scan are green.
- Static review focused on the new port resolver, Lemonade ownership helper, and preflight test wrapper.

Thanks again for catching this before it becomes a late runtime-health failure.

Lightheartdevs (approved):
Audit: backend-aware port preflight verified against main (constants/llm-endpoint sourcing order, OLLAMA_PORT compose default, AMD host-Lemonade port). PS suite + shell contract PASS on real Windows 11. Noted for follow-up: 'managed Lemonade' matches any Lemonade-looking process, not specifically the ODS-owned task (reuse message overstates); LLAMA_SERVER_PORT/OLLAMA_PORT precedence diverges slightly from shell convention. Non-blocking.

PR #1897, opened 2026-07-20
test(windows): accept CRLF LiteLLM config

gabsprogrammer:
## Summary

- accept both LF and CRLF line endings when validating the generated Windows LiteLLM configuration
- keep the existing assertions strict about the general_settings block and generated master-key reference

## Problem

The Windows AMD/Lemonade contract read a PowerShell-generated YAML file with CRLF endings but anchored two regex assertions as LF-only. The generated configuration was valid, yet the local Windows contract reported a failure.

## Validation

- 	ests/contracts/test-amd-lemonade-contracts.sh: **63 passed, 0 failed** on Windows with Git Bash and Windows PowerShell
- git diff --check: passed

## Scope

Test-only change. Runtime configuration generation and authentication behavior are unchanged.

### DIFF
--- a/ods/tests/contracts/test-amd-lemonade-contracts.sh
+++ b/ods/tests/contracts/test-amd-lemonade-contracts.sh
@@ -635,7 +635,7 @@ if command -v pwsh >/dev/null 2>&1; then
         if ($localText -notmatch "(?m)^  request_timeout: 900\r?$" -or $localText -notmatch "(?m)^  stream_timeout: 900\r?$") {
             throw "Windows local LiteLLM config must keep long-model proxy timeouts at 900s"
         }
-        if ($localText -notmatch "(?m)^general_settings:$" -or $localText -notmatch "(?m)^  master_key: os\.environ/LITELLM_MASTER_KEY$") {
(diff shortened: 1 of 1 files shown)


Lightheartdevs:
Thanks for making the Windows LiteLLM contract accept the line endings PowerShell actually emits.

What looks good:
- The change is test-only and keeps the assertions strict about the `general_settings` block and generated master-key reference.
- Accepting optional `\r` before line anchors in `ods/tests/contracts/test-amd-lemonade-contracts.sh` fits the Windows contract surface without relaxing the YAML content being validated.
- CI is green across the normal lint, dashboard, integration, distro, env validation, PowerShell, and secret-scan jobs.

What I checked:
- Reviewed the PR body, current head, and the one-line diff.
- Checked `gh pr checks 1897`: all reported required-style checks are passing.
- Static review confirmed no runtime config-generation files changed.

Thanks again for keeping the contract strict while making it portable.

Lightheartdevs (approved):
Audit: verified the CRLF tolerance is genuinely absent on main and in-flight branches; PowerShell-written YAML uses CRLF; change is test-only, single line, no product surface. Safe.

PR #1902, opened 2026-07-20
feat(router): ODS stable-alias data plane (PR 3, observe default)

Lightheartdevs:
Model Switchboard series PR 3 — plan: `ods/docs/MODEL-SWITCHBOARD.md`. **Depends on:** #1901 (merged). **Mode:** observe by default; `enabled` renders the router LiteLLM map, exercised only on named canaries.

Internal model-router service (no host port, non-root, read-only FS, cap-drop) forwarding the OpenAI paths; rewrites request model to the active concrete route and restores the public alias in bodies + SSE chunks; strips client auth and injects backend creds; bounded body/queue/timeout with structured 503 during swaps; HMAC-probe route evidence with no prompt/credential storage. Endpoint allowlist + `enabled`-gated LiteLLM stable-alias map added to the renderer (observe/legacy byte-identical). 14 router + renderer gating/origin tests green; ruff clean.


Lightheartdevs:
Supervisor hold before merge: independent audit found three foundation P1s inherited from #1899/#1900 that PR 3 would build on. (1) dashboard-api reads INSTALL_DIR/data/model-state.json, but production mounts shared state at /data/model-state.json; (2) cloud-only startup can reconstruct a bogus local llama route from cloud LLM_MODEL; (3) adapter verification records configured expected_gguf instead of the concrete model returned by the proof response. Also keep unverified reconstructed routes out of rollback history. I am preparing focused regression-backed fixes now. Please keep this PR draft and rebase/retest after those land.

PR #1918, opened 2026-07-20
fix(installers): build the brave-search local image on every platform

tang-vu:
## Problem

`brave-search` ships a `Dockerfile` and a `build:` section, and its image tag `ods-brave-search:local` exists in no registry — only a local build can produce it. It is missing from the local-build list in **all three** installers:

- `installers/phases/11-services.sh:1028`
- `installers/macos/install-macos.sh:2380`
- `installers/windows/install-windows.ps1:1651`

Installers start the stack with `docker compose up -d --no-build --pull never`. With brave-search enabled, its image was never built and `--pull never` cannot fetch it, so compose-up fails with `No such image: ods-brave-search:local`.

This is the same failure mode as the model-router fleet regression fixed in 399bc74b — the hardcoded build lists drift from what the stack actually needs built.

`tests/test-local-image-build-coverage.py` was added to close exactly this gap, but it could not see brave-search:

1. it scanned only `docker-compose.base.yml`, not extension composes; and
2. it classified a service as build-only via `build:` **and no** `image:` — brave-search has both, so an unpullable local tag read as pull-able.

## Fix

Add brave-search to the three build lists (Windows gated on `Test-ODSWindowsServiceEnabled`, matching the ape / token-spy / privacy-shield arms), and close both contract gaps: scan `extensions/services/*/compose.yaml` too, and treat a local-only tag (`:local`, `:latest-local`, `:dev`) as build-only.

## Verification

Extended contract **fails on all three installers before the fix**, passes after:

```
[FAIL] local-image build coverage:
  - Linux 11-services.sh: build-only services not in build list: ['brave-search']
  - macOS install-macos.sh: build-only services not in build list: ['brave-search']
  - Windows install-windows.ps1: build-only services not in build list: ['brave-search']
```
```
[OK] local-image build coverage: ['ape', 'brave-search', 'dashboard', 'dashboard-api',
     'model-router', 'privacy-shield', 'token-spy'] covered by all installers
```

The widened contract also now genuinely covers ape / token-spy / privacy-shield (previously out of scope; all three already pass). `bash -n` clean on both shell installers; PowerShell `ParseFile` clean. Already wired into CI via the existing `Local Image Build Coverage` step.

### DIFF
--- a/ods/installers/macos/install-macos.sh
+++ b/ods/installers/macos/install-macos.sh
@@ -2377,7 +2377,7 @@ for service in (data.get("services") or {}).values():
     # surface unrelated Dockerfile failures and make a healthy selected stack
     # look broken.
     ai "Rebuilding local-built images..."
-    _macos_candidate_build_services=(dashboard dashboard-api model-router ape token-spy privacy-shield)
+    _macos_candidate_build_services=(dashboard dashboard-api model-router ape token-spy privacy-shield brave-search)
     if ! _macos_enabled_services="$(docker compose "${COMPOSE_FLAGS[@]}" config --services 2>>"$ODS_LOG_FILE")"; then
         ai_err "Could not resolve macOS compose services for local image rebuilds."
         ai "Inspect compose config with: cd '$INSTALL_DIR' && docker compose ${COMPOSE_FLAGS[*]} config --services"
--- a/ods/installers/phases/11-services.sh
+++ b/ods/installers/phases/11-services.sh
@@ -1025,7 +1025,7 @@ MODELS_INI_EOF
     compose_ok=false
     # Build local images individually so every failure is reported before the
     # installer refuses to launch any potentially stale image.
-    _candidate_build_services=(dashboard dashboard-api model-router ape token-spy privacy-shield)
+    _candidate_build_services=(dashboard dashboard-api model-router ape token-spy privacy-shield brave-search)
     [[ "$ENABLE_COMFYUI" == "true" ]] && _candidate_build_services+=(comfyui)
     [[ "$GPU_BACKEND" == "amd" ]] && _candidate_build_services+=(llama-server)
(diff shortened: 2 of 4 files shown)


Lightheartdevs:
Thanks for extending local image build coverage to Brave Search; this closes a fresh-host installer gap before `--no-build --pull never` can make it user-visible.

What looks good:
- `ods/installers/phases/11-services.sh`, `ods/installers/macos/install-macos.sh`, and `ods/installers/windows/install-windows.ps1` now include `brave-search` in the explicit local-build path when the service is enabled, matching the existing stale-image prevention pattern.
- `ods/tests/test-local-image-build-coverage.py` now treats extension compose files and local-only tags such as `ods-brave-search:local` as build-required, which is the contract that matters when compose up uses `--no-build --pull never`.
- The change is narrow and fits the repo?s recent direction of proving installer image-build coverage with a regression test rather than relying on compose config validation alone.

What I checked:
- Static diff review against `ods/extensions/services/brave-search/compose.yaml`, which has both `build:` and `image: ods-brave-search:local`.
- `python ods/tests/test-local-image-build-coverage.py` locally: OK, with `brave-search` included in the required build-only service set.
- Native PowerShell parser check for `ods/installers/windows/install-windows.ps1`: OK. Local Bash syntax checks were not available on this Windows host because `bash` resolves to a WSL shim without `/bin/bash`, but `gh pr checks 1918` shows ShellCheck, PowerShell lint on Ubuntu and Windows, Test Linux, Matrix Smoke, env validation, and secret scan all passing.

Lightheartdevs (approved):
Audit: adds build-only brave-search to all three installers' local-build lists (same proven pattern as merged #1915) — without it a fresh host enabling brave-search hits 'No such image' on --no-build up. Strengthened coverage contract now catches local-only image tags (ods-brave-search:local) and extension composes; contract passes covering all 7 build-only services, confirming no other gap.

PR #1919, opened 2026-07-20
fix(host-agent): write the model-state record under ODS_DATA_DIR

tang-vu:
﻿## Problem

The Model Switchboard has one writer (the host agent) and several readers. The writer hardcoded the install tree; the readers all honour `ODS_DATA_DIR`.

Writer — `bin/ods-host-agent.py`, both sites built the path inline:
```python
_switchboard_state.initialize_if_missing(INSTALL_DIR / "data" / "model-state.json", env)
_switchboard_state.record_verified_route(INSTALL_DIR / "data" / "model-state.json", ...)
```

Readers:
| Reader | Path |
|---|---|
| model-router (bind-mount, `docker-compose.base.yml:315`) | `${ODS_DATA_DIR:-./data}/model-state.json` |
| dashboard-api `/api/models/state` (`routers/model_state.py:32`) | `os.environ.get("ODS_DATA_DIR") or DATA_DIR` |
| host agent itself, every other data path | `DATA_DIR` |

`ODS_DATA_DIR` is a supported, documented setting (`docs/HOST-AGENT-API.md:30`, "Data directory root"). When it points off the install tree, the writer publishes to a path nobody reads.

The router failure is the harsh one: its mount **source** no longer exists, so Docker creates a **directory** there. The container then gets a directory where it expects a file, `STATE_PATH.read_text()` raises `IsADirectoryError`, and the router answers `503 no_active_route` for the life of the install. `/health` keeps returning 200 throughout, so nothing restarts it.

## Fix

Route both writers through a `_model_state_path()` helper anchored to `DATA_DIR`.

Also moves startup reconstruction to **after** `DATA_DIR` is assigned — it previously ran ~8 lines earlier, while `DATA_DIR` was still the module-level `Path()` default, so the helper could not simply be dropped in at the old position.

## Verification

3 new tests in `TestModelStatePath`, all **red before the fix**:
- relocated `ODS_DATA_DIR` resolves under the data dir, not the install tree
- default layout still resolves to `<install>/data/model-state.json` (no behaviour change for standard installs)
- no call site anchors the record to the install tree (regression guard on the actual bug)

`test_host_agent.py`: 207 passed. The single failure (`test_activation_never_persists_blank_lemonade_model_during_restore`) reproduces on a clean `main` checkout here and is unrelated. `py_compile` clean.

## Why this matters

Installations that relocate `ODS_DATA_DIR` need the host-agent writer, dashboard reader, and container bind mount to agree on the model-state record or model activation status becomes stale or invisible.

## Overlap check

Triage refresh (2026-08-08): searched open and closed Osmantic/ODS PRs by semantic keywords (model state ODS_DATA_DIR), inspected the changed production files, and checked patch equivalence against current upstream/main. No strict duplicate or merged equivalent was found.

## Test plan

- [x] `pytest ods/extensions/services/dashboard-api/tests/test_host_agent.py`

### DIFF
--- a/ods/bin/ods-host-agent.py
+++ b/ods/bin/ods-host-agent.py
@@ -194,6 +194,23 @@ _MODEL_TIERS = frozenset({
 _MIN_MODEL_CONTEXT = 1024
 _MAX_MODEL_CONTEXT = 9007199254740991
 
+
+def _model_state_path() -> Path:
+    """Location of the Model Switchboard state record.
+
+    Tracks DATA_DIR, not INSTALL_DIR/data: every reader resolves the record
+    under ODS_DATA_DIR — the model-router bind-mount in
+    docker-compose.base.yml and the dashboard-api state endpoint both do —
+    so a writer anchored to the install tree publishes where nothing reads.
+
+    DATA_DIR is only assigned once main() has parsed .env. Falling back to
+    the install-tree default keeps a caller that runs before that from
+    resolving a bare relative path and dropping the record into whatever
+    the working directory happens to be.
+    """
+    base = DATA_DIR if DATA_DIR != Path() else INSTALL_DIR / "data"
+    return base / "model-state.json"
+
 # Per-service locks to prevent concurrent start+stop races on the same service
 _service_locks: dict[str, threading.Lock] = collections.defaultdict(threading.Lock)
 _ALLOWED_CORE_RECREATE_IDS = frozenset({
@@ -8417,7 +8434,7 @@ class AgentHandler(BaseHTTPRequestHandler):
                             (switchboard_run or {}).get("capabilities") or {}
                         )
                         _switchboard_state.record_verified_route(
-                            INSTALL_DIR / "data" / "model-state.json",
+                            _model_state_path(),
                             catalog_id=str(model_id),
                             runtime_model_id=verified_runtime_identity,
                             backend_kind=(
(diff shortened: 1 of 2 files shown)


tang-vu:
Pushed a correction. The first revision broke three existing tests (`test_model_activate`, `test_model_state`, `test_switchboard_reconciler`) that set `INSTALL_DIR` but not `DATA_DIR` — the helper then resolved against the unset module default and produced a bare relative path.

Rather than update those tests, the helper now falls back to the install-tree default when `DATA_DIR` is unset. That is the better behaviour independent of the tests: `DATA_DIR` is only assigned once `main()` has parsed `.env`, and any caller running before that would otherwise have dropped the record into the working directory. `ODS_DATA_DIR` still wins whenever it is set, which is the actual fix.

Full dashboard-api suite now: 1564 passed, with only the two failures that also reproduce on a clean `main` checkout on this Windows box (`test_activation_never_persists_blank_lemonade_model_during_restore`, `test_scan_symlink_skipped`). All four new tests verified red against `upstream/main`.

Lightheartdevs:
Thanks for fixing the model-state record placement; relocated `ODS_DATA_DIR` is exactly the kind of install variation the switchboard path needs to honor.

What looks good:
- `ods/bin/ods-host-agent.py` now writes the switchboard state through `_model_state_path()`, aligning the writer with the router bind mount at `${ODS_DATA_DIR:-./data}/model-state.json` and the dashboard-api reader path.
- Startup initialization now runs after `DATA_DIR` is parsed from `.env`, while the helper still falls back to `INSTALL_DIR / "data"` for early callers that run before main has populated globals.
- `ods/extensions/services/dashboard-api/tests/test_host_agent.py` covers relocated data dirs, default layout, unset `DATA_DIR`, and guards against reintroducing inline `INSTALL_DIR / "data" / "model-state.json"` writer paths.

What I checked:
- Static diff review against `ods/docker-compose.base.yml` model-router mounts and the host-agent activation write path.
- `python -m pytest ods/extensions/services/dashboard-api/tests/test_host_agent.py -k ModelStatePath` locally: 4 passed.
- `gh pr checks 1919`: relevant CI is green, including Test Linux, Matrix Smoke, Python lint/type check, ShellCheck, dashboard checks, env validation, and secret scan.

PR #1920, opened 2026-07-20
fix(model-router): keep the public alias across SSE chunk boundaries

tang-vu:
## Problem

`_rewrite_sse_chunk` was applied to each chunk from `aiter_bytes()` independently. Those chunks are network-sized and carry no relationship to SSE line boundaries, so a `data:` line split across two of them is valid JSON in **neither** half — `json.loads` raises on both, the rewrite is skipped, and the upstream's concrete model id reaches the client verbatim.

Reproduced against the shipped function:

```
whole chunk  -> b'data: {"model":"ods/current","choices":[]}\n\n'
split chunk  -> b'data: {"model":"Qwen3.5-9B-Q4_K_M.gguf","choices":[]}\n\n'
```

The stable public alias is the router's whole reason to exist (`PUBLIC_ALIASES`, the `X-ODS-Requested-Model` contract), and it silently stops holding for streaming requests — which is the dominant chat path. The internal model identity leaks with it.

The existing `test_sse_chunks_restore_alias` passes because the mock upstream returns the whole SSE body as a single chunk, so the boundary case was never exercised.

## Fix

`_SSERewriter` buffers an incomplete trailing line until its newline arrives, rewrites only complete lines, and flushes a final unterminated line at end of stream. `_rewrite_sse_chunk` becomes `_rewrite_sse_lines`, unchanged, as the pure line-level rewriter.

`_MAX_PENDING` (1 MiB) bounds the hold-back so an upstream that streams without newlines cannot grow the buffer without limit — past the cap the pending bytes are forwarded unmodified, which is exactly the old behaviour. This is deliberate: buffering is new here, and it should not introduce a memory failure mode of its own.

## Verification

4 new tests, all **red before the fix**:
- exhaustive: **every** byte boundary inside a `data:` line yields the same output as the unsplit line
- a byte-at-a-time stream is byte-identical to the single-chunk result
- a final line with no trailing newline is not dropped
- the pending buffer is bounded on a newline-free stream

Full router suite: 18 passed (14 pre-existing + 4 new).

### DIFF
--- a/ods/extensions/services/model-router/app/main.py
+++ b/ods/extensions/services/model-router/app/main.py
@@ -200,10 +200,10 @@ def _sanitize_headers(request: Request) -> dict[str, str]:
     return headers
 
 
-def _rewrite_sse_chunk(chunk: bytes, alias: str) -> bytes:
-    """Stamp the requested alias into every SSE data line's model field."""
+def _rewrite_sse_lines(block: bytes, alias: str) -> bytes:
+    """Stamp the requested alias into every complete SSE data line."""
     out_lines = []
-    for line in chunk.split(b"\n"):
+    for line in block.split(b"\n"):
         if line.startswith(b"data:") and b"{" in line:
             payload = line[5:].strip()
             try:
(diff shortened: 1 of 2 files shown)


Lightheartdevs:
Thanks for tightening SSE alias preservation; this protects the stable public model contract even when upstream chunks split inside a JSON line.

What looks good:
- `ods/extensions/services/model-router/app/main.py` moves from per-chunk rewriting to a small stateful SSE rewriter, so `ods/current` survives valid network chunk boundaries instead of leaking `Concrete.gguf` mid-stream.
- The pending buffer is bounded and flushes a final unterminated line, which keeps the fix from trading alias correctness for an unbounded memory edge case or dropped tail data.
- `ods/extensions/services/model-router/tests/test_router.py` covers every byte split in one SSE line, byte-at-a-time streaming, final-line flush, and the no-newline buffer cap.

What I checked:
- Static diff review against the current model-router stable-alias route contract.
- `python -m pytest ods/extensions/services/model-router/tests/test_router.py` locally: 18 passed.
- `gh pr checks 1920`: relevant CI is green, including Test Linux, Matrix Smoke, Python lint/type check, ShellCheck, dashboard checks, env validation, and secret scan.

Lightheartdevs:
Superseded by #1914, now merged to main (`f7a578bb`). The merged router frames SSE events across transport chunk boundaries before alias rewriting — `_SSERewriter` with delimiter-regex framing (`\r\n|\r|\n` doubles, stronger than the \n-only rpartition here) plus `_rewrite_sse_event`, verified by `test_fragmented_sse_is_framed_before_rewrite` (splits a `data:` line mid-key across chunks with CRLF delimiters; asserts the public alias is restored and the concrete model id never leaks). Same defect, same fix shape, with tests, already on main.

One element of this PR not carried over: the 1MB `_MAX_PENDING` bound on the holdback buffer — the merged rewriter's buffer is unbounded until a delimiter arrives. That's queued as a small fix-forward PR against main rather than a reason to keep this open. Thank you — the every-byte-boundary sweep test here was a genuinely nice touch.

PR #1921, opened 2026-07-20
fix(model-router): hold the concurrency slot for the life of a stream

tang-vu:
## Problem

`forward` released the inflight slot in its `finally`:

```python
try:
    return await _forward_inner(request, path, payload, requested_alias, body)
finally:
    async with _inflight_lock:
        _inflight -= 1
```

That `finally` runs when the **handler** returns. For a streaming request the handler returns as soon as the upstream response headers arrive — the `StreamingResponse` body generator is pulled by the ASGI server afterwards, and the upstream request stays open the whole time.

So the slot was freed at the *start* of the work it was meant to account for. `MAX_QUEUE_DEPTH` bounded only non-streaming requests — the short ones. Streaming chat, the dominant path and the one that actually occupies the runtime for minutes, could pile up without limit; the router shed nothing and "Router queue is full" never fired for it.

## Fix

`_InflightSlot` with an idempotent `release()`. A streaming response hands the slot to its body generator, which releases it in the same `finally` that closes the upstream response; every other path (including upstream failures) still releases when the handler unwinds.

## Verification

3 new tests, the central one **red before the fix** with exactly its diagnostic:

```
>       assert held == 1, "slot freed before the stream body was consumed"
E       AssertionError: slot freed before the stream body was consumed
```

- slot is held after `forward` returns and released once the body is consumed
- non-streaming requests release their slot
- `release()` is idempotent, so dual ownership cannot double-decrement

A note on test shape: these call `forward` directly rather than going through `TestClient`. `TestClient` consumes a mocked stream eagerly, so the window between headers and body — the entire bug — is already closed by the time a test could observe it. An earlier `TestClient`-based version of these tests passed against the **unfixed** code, which is exactly why it isn't used here.

Full router suite: 17 passed (14 pre-existing + 3 new).

### DIFF
--- a/ods/extensions/services/model-router/app/main.py
+++ b/ods/extensions/services/model-router/app/main.py
@@ -287,6 +287,7 @@ async def forward(full_path: str, request: Request) -> Response:
     requested_alias = str(payload.get("model") or PUBLIC_ALIASES[0])
 
     global _inflight
+    slot = _InflightSlot()
     async with _inflight_lock:
         if _inflight >= MAX_QUEUE_DEPTH:
             return JSONResponse(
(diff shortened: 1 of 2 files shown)


Lightheartdevs:
Thanks for tightening the streaming concurrency path; this fixes the path that dominates real chat runtime rather than only the short non-streaming cases.

What looks good:
- `ods/extensions/services/model-router/app/main.py` keeps the inflight slot owned by the stream body generator, so `MAX_QUEUE_DEPTH` continues to bound upstream work after the FastAPI handler has returned.
- The idempotent `_InflightSlot.release()` keeps non-streaming, upstream-error, and stream-completion paths from double-decrementing or leaking the global counter.
- `ods/extensions/services/model-router/tests/test_router.py` directly exercises the handler-return/body-consumption window that `TestClient` would otherwise hide, plus non-streaming release and idempotent release behavior.

What I checked:
- Static diff review against the router forwarding and streaming response lifecycle.
- `python -m pytest ods/extensions/services/model-router/tests/test_router.py` locally: 17 passed.
- `gh pr checks 1921`: relevant CI is green, including Test Linux, Matrix Smoke, Python lint/type check, ShellCheck, dashboard checks, env validation, and secret scan.

Lightheartdevs:
Superseded by #1914, now merged to main (`f7a578bb`). The merged router holds the admission slot for the full life of a stream: `_forward_inner` returns `(response, stream_owns_admission)`, the handler releases only when not streaming, and the stream generator releases in its `finally` — verified by `test_stream_holds_admission_until_teardown` (asserts `_inflight == 1` while the stream is open and `0` after teardown under `MAX_QUEUE_DEPTH=1`) and `test_failed_stream_records_no_evidence_and_releases_admission` (release on the failure path). Same defect, same fix shape, with tests, already on main.

PR #1922, opened 2026-07-20
fix(model-router): retry the endpoint allowlist instead of caching a failure

tang-vu:
## Problem

`_load_endpoints()` set `loaded = True` on **every** path, including the failure path:

```python
except (OSError, ValueError) as exc:
    logger.error("endpoints allowlist unavailable: %s", exc)
_endpoints_cache["endpoints"] = endpoints   # {} after a failure
_endpoints_cache["loaded"] = True           # ...and pinned forever
```

One bad read pins an empty allowlist for the life of the process. `_active_route()` then raises `endpoint_not_allowlisted` for every request, and only a container restart clears it.

The "startup-validated" framing understates the exposure. `endpoints.json` is a bind mount that `scripts/render-runtime-configs.py` rewrites **in place** with a non-atomic `write_text` (`render-runtime-configs.py:315`), so every re-render — install, update, config change — opens a window where the file is truncated or half-written. A read landing in that window was permanent.

`/health` made it worse by returning 200 regardless. A router that answered nothing but 503 reported healthy, so nothing surfaced the failure.

## Fix

- cache only a **successful** read; a failed one retries on the next request
- `/health` reports `endpointCount` and returns 503 when the allowlist is empty

The compose healthcheck already fails correctly on 503 (`urlopen` raises `HTTPError`, so its `sys.exit` never runs) — no compose change needed. `endpoints.json` is rendered before the router starts, so normal boot is unaffected.

## Verification

5 new tests, **4 red before the fix**:
- a truncated file (the mid-rewrite window) blocks, then the next request recovers
- a missing file recovers once it appears
- `/health` is `degraded`/503 with no allowlist, `ok`/200 with one
- a good allowlist is **still cached** — passes both before and after, guarding against turning every request into a disk read

Full router suite: 19 passed (14 pre-existing + 5 new).

### DIFF
--- a/ods/extensions/services/model-router/app/main.py
+++ b/ods/extensions/services/model-router/app/main.py
@@ -95,23 +95,32 @@ class RouterError(Exception):
 
 
 def _load_endpoints() -> dict[str, dict[str, Any]]:
-    """Startup-validated allowlist: endpointId -> {baseUrl, apiKeyEnv?}."""
+    """Startup-validated allowlist: endpointId -> {baseUrl, apiKeyEnv?}.
+
+    Only a successful read is cached. The file is a read-only bind mount that
+    scripts/render-runtime-configs.py rewrites in place with a non-atomic
+    write_text, so a read can land on a partial or absent file. Caching that
+    outcome would pin an empty allowlist for the life of the process and every
+    request would answer endpoint_not_allowlisted until someone restarted the
+    container; retrying instead lets the next request recover on its own.
+    """
     if _endpoints_cache["loaded"]:
         return _endpoints_cache["endpoints"]
     endpoints: dict[str, dict[str, Any]] = {}
     try:
         raw = json.loads(ENDPOINTS_PATH.read_text(encoding="utf-8"))
-        for entry in raw.get("endpoints", []):
-            endpoint_id = str(entry.get("id") or "")
-            base_url = str(entry.get("baseUrl") or "")
-            if not endpoint_id or not base_url.startswith(("http://", "https://")):
-                continue
-            endpoints[endpoint_id] = {
-                "baseUrl": base_url.rstrip("/"),
-                "apiKeyEnv": str(entry.get("apiKeyEnv") or ""),
-            }
     except (OSError, ValueError) as exc:
-        logger.error("endpoints allowlist unavailable: %s", exc)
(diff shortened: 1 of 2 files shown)


Lightheartdevs:
Thanks for tightening the model-router allowlist recovery; this closes a failure mode that would be painful during generated config rewrites.

What looks good:
- `ods/extensions/services/model-router/app/main.py` now only caches successful `endpoints.json` reads, so a transient missing or truncated allowlist no longer pins every request in `endpoint_not_allowlisted` until restart.
- `/health` now returns 503 with `endpointCount: 0` when the router has no allowlist, which matches the compose healthcheck expectation that a 200 means the service can actually route.
- `ods/extensions/services/model-router/tests/test_router.py` covers truncated-file recovery, missing-file recovery, degraded/ok health, and the still-cached happy path.

What I checked:
- Static diff review against the current model-router/compose contract, including the read-only `endpoints.json` bind mount and `scripts/render-runtime-configs.py` rewrite path.
- `python -m pytest ods/extensions/services/model-router/tests/test_router.py` locally: 19 passed.
- `gh pr checks 1922`: relevant CI is green, including Test Linux, Matrix Smoke, Python lint/type check, ShellCheck, dashboard checks, env validation, and secret scan.

Lightheartdevs:
Superseded by #1931, now merged to main (`2d6affc4`). The merged router re-reads `endpoints.json` on mtime/size change (`_load_endpoints` keyed on `(st_mtime_ns, st_size)`): a failed stat never poisons the cache (retries next request), and a partial or invalid rewrite **retains the last good allowlist** instead of serving empty — strictly stronger than the retry here for the non-atomic-rewrite window this PR describes, and verified by `test_new_endpoint_visible_without_restart` and `test_invalid_rewrite_retains_last_good_allowlist`. Your observability half was folded in with credit: `/health` now reports `endpointCount` and body `status:"degraded"` when the allowlist is empty (`test_health_reports_allowlist_health`). The one piece deliberately not taken is the HTTP 503 flip — the compose healthcheck polls `/health`, so 503 would let a mid-render config gap cascade into container-unhealthy/dependent-ordering effects; that's tracked as an explicit owner decision rather than a silent change. The writer-side truncation itself is being addressed separately in #1927 (atomic replace). Thanks — the diagnosis of the cached-failure pin was exactly right.

PR #1923, opened 2026-07-20
fix(token-spy): stop the session manager deleting live sessions

tang-vu:
## Problem

`clean_inactive()` decides what to delete by diffing the `*.jsonl` files on disk against the `sessionId`s in `sessions.json`, and extracted those ids with a greedy sed:

```bash
sed -n 's/.*"sessionId":[[:space:]]*"\([^"]*\)".*/\1/p'
```

The leading `.*` is greedy, so on any line holding more than one entry the pattern reports only the **last** id. `sessions.json` is written by the agent framework, not by this script, and a compact document puts every entry on one line.

Driving the shipped `clean_inactive` against a compact `sessions.json` listing three active sessions:

```
sessions.json lists 3 active: aaa bbb ccc
before: 3 session files
after : 1 remaining
  kept: ccc.jsonl
```

Two live sessions deleted. This runs periodically from a systemd timer or cron, and token-spy is `category: recommended`.

`scripts/session-cleanup.sh:58` already reads the same file correctly, per occurrence with `grep -oE` — one file format, two different readings.

## Second failure in the same function

An empty extraction was treated as "nothing is active" rather than "the parse failed", so a schema change or a partial write deletes **every** session file. The SSH remote path had the same shape: its `ACTIVE_IDS` block is only emitted when the remote `sessions.json` exists, yet an empty block still queued every remote session for `rm -f`.

## Fix

- `extract_active_ids()` matches per occurrence, mirroring `session-cleanup.sh`
- both the local and the remote path refuse to clean when a present `sessions.json` yields no ids
- the same greedy pattern in the embedded remote script is corrected

## Verification

New `tests/test-token-spy-session-manager.sh` (wired into `make test`), sourcing the real functions out of the script:

1. every id is extracted from a compact single-line `sessions.json`
2. three live sessions survive cleanup, and an orphan is still collected
3. an unparseable `sessions.json` cancels cleanup instead of wiping
4. pretty-printed layout still works — no regression

`bash -n` clean.

### DIFF
--- a/ods/Makefile
+++ b/ods/Makefile
@@ -101,6 +101,9 @@ test: ## Run unit and contract tests
 	@echo "=== macOS uninstall LaunchAgent cleanup ==="
 	@bash tests/test-macos-uninstall-launchagents.sh
 	@echo ""
+	@echo "=== token-spy session manager active-id extraction ==="
+	@bash tests/test-token-spy-session-manager.sh
+	@echo ""
 	@echo "=== fleet-multi-distro keep-going on docker pull failure ==="
 	@bash tests/test-fleet-multi-distro-keep-going.sh
 	@echo ""
--- a/ods/extensions/services/token-spy/session-manager.sh
+++ b/ods/extensions/services/token-spy/session-manager.sh
@@ -89,9 +89,11 @@ clean_inactive() {
   local active_ids
   active_ids=$(extract_active_ids "$sessions_json")
 
-  # No ids out of a non-empty sessions.json means the parse failed, not that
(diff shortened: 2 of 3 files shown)


Lightheartdevs (approved):
Audit + fix: greedy sed read only the last id from a compact sessions.json, deleting every other live session; per-occurrence grep extraction is correct across compact/pretty/spaced layouts. Fixed the new guard to fail closed on a zero-byte file (partial write), matching the remote path; added zero-byte regression. Local suite passes incl. the new case.

PR #1924, opened 2026-07-20
fix(resolver): restore user-extension compose guard parity with the API

tang-vu:
﻿## Problem

`_scan_user_compose_content` in `scripts/resolve-compose-stack.sh` says of itself:

> Mirrors dashboard-api/routers/extensions.py:_scan_compose_content (without the FastAPI HTTPException dependency).

It has drifted from it in four places. This matters because the resolver is the layer that assembles the compose stack the install actually starts — anything it lets through reaches a running container, including extensions that never went through the API endpoint (dropped into `data/user-extensions/`, or installed before a check existed).

**1. Relative bind-mount escape — the significant one.** The resolver rejected only `source.startswith("/")`. Compose resolves a relative source against the project directory, so `../../../../etc:/host-etc` reaches exactly the host paths the absolute check exists to deny. The API has rejected this since `extensions.py:432`:

```python
if ".." in vol_source.split("/"):
    raise HTTPException(400, "...escaping the project directory...")
```

Against the shipped resolver the malicious extension is merged into the stack:

```
Test 1: relative bind-mount escaping the project directory (short form)
  ✗ short-form '../' escape refused — resolver merged the extension instead of rejecting it
```

**2. Long-form volumes skipped the rest of the rules.** The `isinstance(vol, dict)` branch checked the Docker socket and the absolute path, then `continue`d — so a long-form `{type: bind, source: ../..}` bypassed everything else too.

**3. `cap_add` missed the `CAP_` spelling.** The resolver compared `str(cap).upper()` against bare names, so `CAP_SYS_ADMIN` — which Docker accepts and treats as `SYS_ADMIN` — sailed past a check written to stop `SYS_ADMIN`. The API normalizes with `.removeprefix("CAP_")`.

**4. Reserved labels.** The resolver matched only `com.docker.compose.` and case-sensitively; the API matches `("com.docker.compose.", "io.docker.")` after `.lower()`.

## Fix

Both volume forms now derive one `source` and run one shared set of rules, so a long-form entry can never skip a check a short-form entry gets. Capability and label matching adopt the API's normalization.

No new policy is introduced here — every rule already exists on the API side. This only closes the drift.

## Verification

New `tests/test-resolver-compose-guard-parity.sh` (wired into `make test`) drives the **real resolver** against fixture user-extensions and asserts the extension's compose is not merged. Tests 1–4 fail on `main`, one per gap. Test 5 pins that a benign in-project relative mount (`./data/benign:/data`) still merges — it passes both before and after, so the fix cannot be read as banning relative mounts.

`bash -n` clean.

## Why this matters

User extension compose files reach the runtime through both the dashboard API and the shell resolver. Equivalent guardrails prevent a hand-restored extension from bypassing checks enforced during dashboard installation.

## Overlap check

Triage refresh (2026-08-08): searched open and closed Osmantic/ODS PRs by semantic keywords (resolver compose guard parity), inspected the changed production files, and checked patch equivalence against current upstream/main. No strict duplicate or merged equivalent was found.

## Test plan

- [x] `bash ods/tests/test-resolver-compose-guard-parity.sh`


Lightheartdevs:
Thanks for tightening resolver/API parity on user-extension compose scanning; this is the right place to keep the CLI guardrails aligned with dashboard validation.

What looks good:
- `ods/scripts/resolve-compose-stack.sh` now normalizes `CAP_` capabilities, rejects parent-directory bind escapes in short and long volume forms, and blocks `io.docker.*` labels in line with `dashboard-api`'s `_scan_compose_content` checks.
- `ods/tests/test-resolver-compose-guard-parity.sh` is wired into `ods/Makefile` and includes both rejected cases and a benign relative mount case, so the resolver parity guard is easy to run with the existing test target.
- Current GitHub checks are green across the active CI surface.

What I checked:
- Compared the resolver behavior with `ods/extensions/services/dashboard-api/routers/extensions.py`.
- `git diff --check origin/main...origin/pr/1924`.
- `gh pr checks 1924`; I could not run the local bash test on this Windows runner because bash/WSL is not available here.

Thanks again for closing this validation gap.

PR #1925, opened 2026-07-20
fix(cli): surface compose-flags resolver failures instead of silencing them

### LINKED ISSUE #1876: ods-cli: compose-flags regeneration fails silently without PyYAML, breaking 'ods start <ext>'
Found during live QA of #1868 on a clean macOS install.

**Repro:** host python3 without PyYAML → `ods enable omnigent` succeeds (with a registry warning), but `ods start omnigent` fails with compose's `no configuration file provided: not found`.

**Cause:** `_regenerate_compose_flags` in `ods-cli` runs `scripts/resolve-compose-stack.sh` with `2>/dev/null || rm -f .compose-flags`. The resolver hard-requires PyYAML and prints a clear error — but the CLI discards stderr and deletes the flags cache, so the eventual failure surfaces two commands later with no hint of the cause. The CLI's own PyYAML warning says "service registry will be incomplete," which understates the impact (lifecycle commands break).

**Suggested fix:** stop silencing the resolver (`2>/dev/null` → capture and surface on failure), and/or make the PyYAML warning say that enable/start will not work until it's installed. A preflight check for host-python PyYAML on the macOS path would also catch it at install time.

**Workaround:** `pip3 install --break-system-packages pyyaml` then rerun `ods start`.

w3-surfer:
Fixes #1876.

## Problem

`_regenerate_compose_flags` in `ods-cli` ran the resolver with stderr discarded, and removed the cache on failure:

```bash
"$INSTALL_DIR/scripts/resolve-compose-stack.sh" \
    ... \
    > "$INSTALL_DIR/.compose-flags" 2>/dev/null || rm -f "$INSTALL_DIR/.compose-flags"
```

`resolve-compose-stack.sh` already prints an actionable diagnosis when it can't run:

```
ERROR: PyYAML is required by resolve-compose-stack.sh for compose validation.
       Or install manually: <python> -m pip install pyyaml
```

but `2>/dev/null` threw it away. As reported in the issue, `ods enable <ext>` then appeared to succeed and the real failure only surfaced two commands later as compose's `no configuration file provided: not found`, with no hint of the cause.

## Fix

Capture the resolver's stderr and print it after a warning that names the consequence, so the failure is reported where it happens:

```bash
local resolver_err=""
if ! resolver_err=$("$INSTALL_DIR/scripts/resolve-compose-stack.sh" \
    ... \
    2>&1 > "$INSTALL_DIR/.compose-flags"); then
    rm -f "$INSTALL_DIR/.compose-flags"
    warn "Could not regenerate the compose stack cache. 'ods start' will fail until this is resolved:"
    [[ -n "$resolver_err" ]] && printf '%s\n' "$resolver_err" >&2
fi
```

The `2>&1 > file` ordering keeps stdout going to `.compose-flags` and captures only stderr, so successful runs behave exactly as before.

Also corrected the `lib/service-registry.sh` warning, which said only *"Service registry will be incomplete"* while lifecycle commands actually break. It now states that `ods enable` / `ods start` will fail.

## Verification

Extracted the real function and ran it against a stub resolver covering both paths:

**Resolver fails** (stub reproduces the PyYAML error):
```
WARN: Could not regenerate the compose stack cache. 'ods start' will fail until this is resolved:
ERROR: PyYAML is required by resolve-compose-stack.sh for compose validation.
       Or install manually: python3 -m pip install pyyaml
.compose-flags removed
```

**Resolver succeeds:** stderr empty, `.compose-flags` contains `-f docker-compose.base.yml -f docker-compose.nvidia.yml`.

`bash -n` passes on both files.

## Out of scope

The issue also suggests a host-Python PyYAML preflight check on the macOS install path. That is a larger, separate change and is not included here.

### DIFF
--- a/ods/.env.schema.json
+++ b/ods/.env.schema.json
@@ -1069,6 +1069,11 @@
       "description": "Hostname or IP that dashboard-api containers use to reach the ODS Host Agent. Docker Desktop uses host.docker.internal; Colima uses its private vmnet host gateway.",
       "default": ""
     },
+    "ODS_PYTHON_CMD": {
+      "type": "string",
+      "description": "Python interpreter that ODS shell helpers must use, recorded by the installer when the host python3 cannot import yaml (the compose resolver requires PyYAML). Empty means detect at runtime.",
+      "default": ""
+    },
     "ODS_MACOS_HOST_GATEWAY": {
       "type": "string",
       "description": "Private macOS host-side vmnet address used by Colima containers to reach loopback-only ODS services.",
--- a/ods/installers/macos/install-macos.sh
+++ b/ods/installers/macos/install-macos.sh
@@ -1666,6 +1666,18 @@ else
         ai_warn "ODS_AGENT_BIND=${_macos_agent_bind_raw} uses an unsupported IPv6 server socket; using ${_macos_agent_bind}"
         upsert_env_value "${INSTALL_DIR}/.env" "ODS_AGENT_BIND" "$_macos_agent_bind"
     fi
+    # Persist the interpreter that proved it can import yaml. _ensure_macos_pyyaml
+    # may have satisfied the compose resolver through an ODS-owned venv, but that
+    # choice only lives in this installer process. Without recording it, a later
+    # `ods enable` / `ods start` resolves python again, falls back to a host
+    # python3 that cannot import yaml, and scripts/resolve-compose-stack.sh fails
+    # after the install already reported success. lib/safe-env.sh exports .env
+    # keys, so the CLI and the resolver subprocess both pick this up.
+    if [[ -n "${ODS_PYTHON_CMD:-}" ]] && ! _macos_python_imports_yaml python3; then
+        upsert_env_value "${INSTALL_DIR}/.env" "ODS_PYTHON_CMD" "$ODS_PYTHON_CMD"
+        ai_ok "Recorded compose-resolver Python: ${ODS_PYTHON_CMD}"
+    fi
+
     _macos_llm_bridge_enabled="false"
     if [[ "${DOCKER_BACKEND:-unknown}" == "colima" ]]; then
         _macos_llm_bind="$(read_env_value "${INSTALL_DIR}/.env" "BIND_ADDRESS")"
--- a/ods/lib/service-registry.sh
+++ b/ods/lib/service-registry.sh
@@ -75,7 +75,7 @@ sr_load() {
             pkg_install $(pkg_resolve python3-pyyaml) 2>>"${LOG_FILE:-/dev/null}" || true
         fi
         if ! "$PYTHON_CMD" -c "import yaml" 2>/dev/null; then
-            declare -f warn &>/dev/null && warn "PyYAML not available. Service registry will be incomplete."
+            declare -f warn &>/dev/null && warn "PyYAML not available. Service registry will be incomplete and lifecycle commands ('ods enable', 'ods start') will fail."
             declare -f warn &>/dev/null && warn "Install manually: pip3 install pyyaml"
             _SR_LOADED=true  # Prevent repeated retries
             _SR_FAILED=true
--- a/ods/ods-cli
+++ b/ods/ods-cli
@@ -699,10 +699,18 @@ _regenerate_compose_flags() {
         # --gpu-count gates the multigpu-{backend}.yml overlay; without it,
         # toggling any extension would silently strip multi-GPU plumbing
         # from the persisted cache.
-        "$INSTALL_DIR/scripts/resolve-compose-stack.sh" \
+        # Capture the resolver's stderr instead of discarding it: it reports
+        # actionable causes (e.g. missing PyYAML) that would otherwise only
+        # surface commands later as compose's "no configuration file provided".
+        local resolver_err=""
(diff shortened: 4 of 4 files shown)


w3-surfer:
@nolanmak thanks for the precise diagnosis here, it made this a quick fix. If you were already planning to take it yourself, say the word and I'll close this in favor of yours.

gabsprogrammer:
Please attach the PyYAML pre-check for the Python host in the macOS installation path; you can include it here, and the PR will be merged.

Lightheartdevs:
Thanks for surfacing the compose resolver failure where it actually happens; that makes the extension enable/start path much easier to debug.

Quickest path to merge:
- Add the maintainer-requested macOS host-Python PyYAML precheck in the macOS install path, likely near the existing Python/module guard flow in `ods/installers/macos/install-macos.sh` or by reusing `ods/installers/lib/python-runtime.sh`, so macOS users fail before `ods enable`/`ods start` depends on `resolve-compose-stack.sh`.
- Include a focused verification note for the missing-PyYAML macOS path, because this PR currently improves the CLI warning but leaves the original macOS preflight gap called out in the discussion.

What I checked:
- Reviewed the PR description, existing discussion, changed files, and patch for `ods/ods-cli` and `ods/lib/service-registry.sh`.
- Checked the surrounding resolver/PyYAML usage in `ods/scripts/resolve-compose-stack.sh`, Linux installer module guard in `ods/install-core.sh`, and macOS installer entry point.
- Confirmed the existing CI checks on this head are green; I did not rerun the shell tests locally on this Windows runner.

Thanks again for taking the sharp edge off this failure path.

w3-surfer:
@gabsprogrammer thanks, added in `8e0944c`.

While wiring the precheck I found the macOS path already has one: `_ensure_macos_pyyaml` runs during preflight and, when the host `python3` cannot import yaml, creates an ODS-owned venv and installs PyYAML. So the install itself was never the failure point.

The gap is that it records the result with `export ODS_PYTHON_CMD=...`, which only lives in the installer process. Nothing persists it, so a later `ods enable` / `ods start` resolves Python again, falls back to the host `python3` that still lacks PyYAML, and `resolve-compose-stack.sh` fails after the install reported success. That matches the QA report in #1876 exactly: install fine, `ods start` broken.

So this commit persists the proven interpreter rather than adding a second check:

- `installers/macos/install-macos.sh` — after `generate_ods_env`, `upsert_env_value ... ODS_PYTHON_CMD` **only when the host `python3` cannot import yaml**. Installs whose `python3` already has PyYAML are untouched, so no new key appears on healthy hosts.
- `.env.schema.json` — declares `ODS_PYTHON_CMD`. Required: `scripts/validate-env.sh` treats unknown `.env` keys as a hard error, so persisting without this would have failed env validation on exactly the hosts the fix targets.

`lib/safe-env.sh` exports `.env` keys, so both the CLI and the `resolve-compose-stack.sh` subprocess inherit it, and `lib/python-cmd.sh` already honors `ODS_PYTHON_CMD` before falling back to detection.

## Verification

```
bash -n installers/macos/install-macos.sh          -> OK
python3 -m json.tool .env.schema.json              -> OK

# the new key survives env validation (schema addition)
scripts/validate-env.sh <env with ODS_PYTHON_CMD>  -> exit 0
scripts/validate-env.sh <env with a bogus key>     -> exit 2, "Unknown keys not defined in schema"

# .env value reaches the resolver subprocess
load_env_file .env; printenv ODS_PYTHON_CMD        -> /…/.venv/installer-python/bin/python
ods_detect_python_cmd (lib/python-cmd.sh)          -> honors ODS_PYTHON_CMD

# persist condition
host python3 HAS yaml                              -> not persisted
host python3 lacks yaml + venv resolved            -> persisted
```

Honest limitation: I verified the logic, syntax, env-validation and export propagation on macOS, but I do not have a clean macOS host without PyYAML to run a full end-to-end install against, so the live install path is unverified by me.

Lightheartdevs:
Thanks for closing the loop on the compose resolver/PyYAML failure path; keeping the CLI error visible and persisting the installer-proven Python are both directly tied to #1876.

What looks good:
- `ods/ods-cli` now preserves resolver stderr while keeping stdout pointed at `.compose-flags`, so failures no longer turn into a later opaque compose error.
- `ods/installers/macos/install-macos.sh` records `ODS_PYTHON_CMD` after env generation only when the host `python3` cannot import yaml, so healthy hosts keep detecting at runtime while venv-backed installs keep using the interpreter that passed the PyYAML check.
- `ods/.env.schema.json` declares `ODS_PYTHON_CMD`, and `ods/lib/service-registry.sh` now tells operators that missing PyYAML breaks lifecycle commands instead of only saying the registry is incomplete.

What I checked:
- Reviewed the new head, previous maintainer request, and diffs for `ods-cli`, `service-registry.sh`, `install-macos.sh`, and `.env.schema.json`.
- Ran Git Bash `bash -n` on `ods/ods-cli`, `ods/lib/service-registry.sh`, `ods/installers/macos/install-macos.sh`, and `ods/scripts/resolve-compose-stack.sh`.
- Validated `.env.schema.json` with Python JSON parsing and checked `git diff --check origin/main...origin/pr/1925`.
- GitHub checks are green for the current head, including shell/PowerShell lint, dashboard API/frontend, Python lint, matrix smoke, Linux integration smoke, env validation, and secret scan.

Thanks again for taking the issue through the macOS follow-up.

PR #1926, opened 2026-07-20
fix(token-spy): bound the report date range

tang-vu:
## Problem

`/api/report` (`main.py:1667`) passes user-supplied `start`/`end` straight into `_parse_report_dates`, which only checked `end >= start`. The report then materializes one bucket per day across the whole span.

Measured against the shipped code:

```
0001-01-01 .. 9999-12-30
  accepted span days: 3,652,058
  6.47s and ~215MB of date strings — before a single row is read
```

There is a second failure at the very top of the range. The endpoint catches `ValueError` only:

```python
try:
    return query_report(start=start, end=end)
except ValueError as exc:
    return JSONResponse({"error": str(exc)}, status_code=400)
```

but `end_day + timedelta(days=1)` raises **OverflowError** on `date.max`, which is not a `ValueError`. So `?end=9999-12-31` escaped the handler entirely and surfaced as a 500 traceback instead of a 400.

## Fix

Cap the span at 366 days before that arithmetic runs, which resolves both: the oversized span is refused, and `date.max` never reaches the `timedelta` add.

The bound matches what `dashboard-api/routers/usage.py` already applies to its own `/report` (`MAX_REPORT_RANGE_DAYS = 366`, merged in #1859). dashboard-api is the proxy; token-spy is the backend that does the work and is reachable directly, so the bound belongs on both sides.

Applied to `db.py` **and** `db_postgres.py` — both back the same endpoint from identical code.

## Verification

5 new tests in `TestReportRangeCap`, **3 red before the fix**:
- an oversized span raises `ValueError` (→ 400)
- `9999-12-31` raises `ValueError`, not the `OverflowError` that escaped as a 500
- both backends carry the same cap
- a normal year (`2026-01-01..2026-12-31`) is still accepted — passes before and after, so the cap does not narrow real usage
- a reversed range keeps its existing error

token-spy suite: 10 passed.

The Postgres assertion checks the source rather than importing the module: `db_postgres.py` imports `psycopg2` at module scope, which is absent wherever Postgres is not the configured backend.

### DIFF
--- a/ods/extensions/services/token-spy/db.py
+++ b/ods/extensions/services/token-spy/db.py
@@ -148,11 +148,25 @@ def query_usage(agent: str | None = None, hours: int = 24, limit: int = 200) -> 
     return [dict(r) for r in rows]
 
 
+# Inclusive-day span cap for the report endpoint. Mirrors
+# dashboard-api/routers/usage.py:MAX_REPORT_RANGE_DAYS — both build one dict
+# per day in the requested span, so the bound belongs on each side.
+MAX_REPORT_RANGE_DAYS = 366
+
+
+
 def _parse_report_dates(start: str, end: str) -> tuple[date, date, date]:
     start_day = date.fromisoformat(start)
     end_day = date.fromisoformat(end)
     if end_day < start_day:
         raise ValueError("end must be on or after start")
+    # The report materializes one bucket per day in the span, so an unbounded
+    # range turns a single request into millions of them. Cap it before the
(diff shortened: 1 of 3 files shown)


Lightheartdevs (approved):
Audit + fix: 366-day inclusive cap mirrors dashboard-api exactly, applied to both sqlite and postgres parsers, out-of-range raises a clean 400. Fixed the leftover OverflowError→500 on short spans ending at date.max (reject date.max explicitly in both parsers); added coverage. Local: 11 passed.

PR #1927, opened 2026-07-20
test(render-configs): pin the write loop to the atomic replace

tang-vu:
﻿## Scope change after rebase

This PR originally added an `atomic_write` helper to `ods/scripts/render-runtime-configs.py`. `main` has since landed `atomic_write_text()`, which is a strict superset — same temp-file-and-`os.replace` shape, plus mode preservation and a `PermissionError` retry — and the `--write` loop already calls it.

The implementation half of this PR is therefore fully superseded and has been dropped. Rebased onto current `main`; what remains is the one piece of coverage `main` does not have.

## The gap

`test_atomic_write_failure_preserves_known_good_config` exercises `atomic_write_text()` in isolation. Nothing asserts that the `--write` loop in `render()` actually routes through it.

That matters because a regression back to `target.write_text(...)` is invisible to every existing test:

- final bytes: identical
- file mode: `0644` either way on a freshly created file
- temp-file leftovers: none either way

`config/litellm/<mode>.yaml` and `config/model-router/endpoints.json` are bind-mounted into running containers, so a truncate-then-write re-render leaves a window in which the container reads an empty or partial config.

## The test

`test_write_path_routes_through_the_atomic_replace` drives `render()` in-process with `--write` against a pre-existing target and injects a `PermissionError` into `os.replace`. That is what separates the two paths: the atomic path leaves the previous file intact, an in-place write has already destroyed it.

Verified red on a regressed write loop:

```
AssertionError: the write loop did not go through the atomic replace
```

Green on `main` as it stands. No production code touched — test-only.

## Why this matters

Runtime config rendering is invoked during installation and mode/model changes. The write-loop regression proves an interrupted atomic replace leaves the last valid router configuration intact.

## Overlap check

Triage refresh (2026-08-08): searched open and closed Osmantic/ODS PRs by semantic keywords (render runtime atomic write), inspected the changed production files, and checked patch equivalence against current upstream/main. No strict duplicate or merged equivalent was found.

## Test plan

- [x] `python -m pytest -q ods/tests/test-render-runtime-configs.py`


Lightheartdevs:
Thanks for tightening the runtime config replacement path; this is a useful failure mode to cover, but current `main` has since picked up a more complete helper.

Quickest path to merge:
- Rebase onto current `main` and resolve the conflicts in `ods/scripts/render-runtime-configs.py` and `ods/tests/test-render-runtime-configs.py`.
- Reuse or extend `atomic_write_text(...)` instead of adding a parallel `atomic_write` helper; the current helper already preserves existing mode, retries `PermissionError`, and cleans up temp files on failure.
- If there is still unique value here, adapt the tests to assert behavior through `atomic_write_text` rather than replacing the current implementation.

What I checked:
- `git merge-tree` against `origin/main`, which shows conflicts in the renderer and tests.
- `git diff --check origin/main...origin/pr/1927`.

Thanks again for working on the config durability issue.

Lightheartdevs:
Thanks for narrowing this to the write-loop coverage that current `main` was still missing.

What looks good:
- The added test exercises `render(... --write)` itself, not just `atomic_write_text()`, by injecting an `os.replace` failure and verifying the existing `config/model-router/endpoints.json` content survives unchanged.
- The test is wired into the file's standalone `main()` harness as well as pytest discovery, matching the existing test style in `ods/tests/test-render-runtime-configs.py`.
- The PR is test-only and keeps production renderer behavior unchanged while pinning the atomic replacement contract for bind-mounted runtime configs.

What I checked:
- Reviewed the current one-file diff and the prior discussion about `atomic_write_text()` superseding the original implementation change.
- Ran `git diff --check origin/main...origin/pr/1927`; it passed.
- Ran `py -3.13 -m pytest ods/tests/test-render-runtime-configs.py::test_write_path_routes_through_the_atomic_replace -q`; it passed.
- Checked GitHub status for this head: Python/shell lint, type check, dashboard jobs, Linux integration, distro matrix, macOS smoke, PowerShell lint, env validation, and secret scan are green.

Thanks again for preserving the durability regression as focused coverage.

PR #1928, opened 2026-07-20
fix(talk): keep retrying a transient offline status

tang-vu:
## Problem

The offline retry is a one-shot `setTimeout` re-armed by an effect dependency:

```jsx
useEffect(() => {
  if (!retryStatusRefresh || status !== 'offline') return undefined
  const timer = setTimeout(refreshStatus, TALK_STATUS_RETRY_MS)
  return () => clearTimeout(timer)
}, [refreshStatus, retryStatusRefresh, status])
```

When the retry fails, `refreshStatus` sets `status` to `'offline'` and `retryStatusRefresh` to `true` — the values they already hold. React bails out of a re-render on identical state, so **no dependency changes and the effect never re-runs**. `refreshStatus` is a `useCallback` keyed on `liveMicSupported`, so it is stable too.

Net effect: exactly one retry, 1s after going offline, then the page sits on "ODS Talk is offline" until a manual reload. A backend container taking longer than one interval to become ready is the normal case, so the transient-offline handling this effect exists for did not cover it.

Every other poller in the dashboard (`Dashboard`, `ServiceMap`, `Extensions`, `useDownloadProgress`) uses `setInterval` and reschedules on its own — this one was the outlier.

The existing test passes because it makes the backend ready on the **second** call, which is exactly the one retry that does happen.

## Fix

Advance a `statusAttempt` counter after each retry, so a dependency always changes and the poll re-arms. The cleanup cancels the pending bump, so an unmount or a status change does not schedule another round.

A settled incompatibility still switches retrying off (`setRetryStatusRefresh(!compatibilityReason)`), so the page does not poll forever on a permanent answer.

## Verification

2 new tests. Against `main` (with `--no-cache` — vitest's transform cache will otherwise serve the patched module):

```
× keeps retrying when readiness takes more than one attempt 5026ms
Tests  1 failed | 14 skipped (15)
```

- readiness on the 4th call recovers (red before the fix, times out on `offline`)
- a permanent incompatibility issues no further polls — guards against the fix turning into an unbounded poll

`ODSTalk.test.jsx`: 15 passed. ESLint clean.

### DIFF
--- a/ods/extensions/services/dashboard/src/pages/ODSTalk.jsx
+++ b/ods/extensions/services/dashboard/src/pages/ODSTalk.jsx
@@ -55,6 +55,9 @@ export default function ODSTalk() {
   const [status, setStatus] = useState('loading')
   const [statusText, setStatusText] = useState('Connecting to ODS Talk...')
   const [retryStatusRefresh, setRetryStatusRefresh] = useState(false)
+  // Advances after every offline retry so the polling effect re-arms; a
+  // repeated failure otherwise leaves every dependency unchanged.
+  const [statusAttempt, setStatusAttempt] = useState(0)
   const [sending, setSending] = useState(false)
   const [recording, setRecording] = useState(false)
   const [spokenReplies, setSpokenReplies] = useState(() => {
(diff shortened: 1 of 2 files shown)


Lightheartdevs (approved):
Audit: removes a scheduler-race dependence in ODSTalk status re-arm (traced both commit orderings — no dead-end state, cleanup prevents post-unmount setState). /api/talk/status reports text_chat:false on swap rather than 503, so transient offline is treated correctly; net-positive for the fleet Talk step. Noted: unbounded 1 Hz poll (deliberate, matches other dashboard pollers) with badge flicker while offline.

PR #1933, opened 2026-07-21
fix(model-router): bound SSE rewriter holdback buffer

Lightheartdevs:
## Fix-forward carry-over from #1914/#1931 review

`_SSERewriter.buffer` appended every streamed chunk but only trimmed it when an SSE delimiter matched. An upstream that streams **without a delimiter** grew the buffer without bound — `MAX_BODY_BYTES` only guards the inbound request body, not the rewriter's held-back partial event. (Flagged in the #1914 Claude-lane review; #1920 had bounded its holdback at 1MB.)

## Change
- New `MAX_SSE_HOLDBACK` (default 1MB, `ODS_ROUTER_MAX_SSE_HOLDBACK_BYTES`).
- `_SSERewriter.feed()` flushes the partial buffer via `_rewrite_sse_event` (same best-effort rewrite `finish()` does) once it exceeds the cap, then resets. Memory bounded; framing continues correctly afterward.
- Read-path only; no observe/legacy parity impact.

## Test
- `test_sse_holdback_is_bounded_without_delimiter`: a delimiter-less oversized chunk is flushed (not held), buffer stays ≤ cap, subsequent well-framed event still frames.
- Full router suite: **39 passed**, ruff clean.

### DIFF
--- a/ods/extensions/services/model-router/app/main.py
+++ b/ods/extensions/services/model-router/app/main.py
@@ -59,6 +59,9 @@ PROBE_KEY_PATH: Path | None = (
 INSTANCE_ID = str(uuid.uuid4())
 
 MAX_BODY_BYTES = int(os.environ.get("ODS_ROUTER_MAX_BODY_BYTES", str(2 * 1024 * 1024)))
+# Bound the SSE rewriter's held-back partial event so a delimiter-less upstream
+# cannot grow it without limit (MAX_BODY_BYTES only guards the request body).
+MAX_SSE_HOLDBACK = int(os.environ.get("ODS_ROUTER_MAX_SSE_HOLDBACK_BYTES", str(1024 * 1024)))
 MAX_QUEUE_DEPTH = int(os.environ.get("ODS_ROUTER_MAX_QUEUE_DEPTH", "64"))
 QUEUE_WAIT_SECONDS = int(os.environ.get("ODS_ROUTER_QUEUE_WAIT_SECONDS", "60"))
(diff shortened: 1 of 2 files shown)


Lightheartdevs:
Thanks for taking on the model-router streaming edge case; bounding the SSE rewriter holdback is a useful hardening step for malformed or delimiter-less upstream streams.

What looks good:
- `ods/extensions/services/model-router/app/main.py` now caps `_SSERewriter.buffer` with `ODS_ROUTER_MAX_SSE_HOLDBACK_BYTES`, so an upstream that never sends an SSE delimiter cannot grow memory without bound.
- The overflow path uses the same `_rewrite_sse_event()` best-effort rewrite behavior as `finish()`, then resets the buffer, which keeps the existing alias-redaction path coherent instead of bypassing it.
- `ods/extensions/services/model-router/tests/test_router.py` adds focused regression coverage for the delimiter-less stream and verifies framing still works after a forced flush.

What I checked:
- Reviewed the PR title, description, changed files, and patch for #1933.
- Ran `python -m py_compile extensions/services/model-router/app/main.py` locally on this head.
- Ran `python -m pytest extensions/services/model-router/tests/test_router.py -q` locally: 42 passed, with only existing FastAPI `on_event` deprecation warnings.
- Ran `git diff --check origin/main...HEAD` locally.
- Confirmed GitHub checks are green, including Python lint, type check, dashboard API/frontend, shell/PowerShell lint, integration smoke, matrix smoke, macOS smoke, secret scan, and env validation.

PR #1946, opened 2026-07-21
fix(stability): replace Hermes config and session writes atomically

vaibhavsrv:
## Problem

In `ods/scripts/patch-hermes-config.py` and `ods/scripts/session-cleanup.sh`, config and session file updates previously used in-place writes or temporary files created with default `mkstemp` umask permissions (0600), which discarded target file mode permissions (`chmod`) and ownership (`chown`) during atomic replacement.

Atomic replacement via `os.replace(tmp_path, target)` creates a new inode with the permissions of `tmp_path`, so target permissions and ownership metadata were lost if not read and reapplied to `tmp_path` prior to replacement.

## Fix

1. Update `patch-hermes-config.py` and `session-cleanup.sh` to read existing file stat metadata (`st = os.stat(path)`) prior to replacement and preserve mode (`chmod`) and owner/group (`chown`) on `tmp_path`.
2. Clean up temporary files on exceptions in both write paths.

## Verification

Added regression tests (Test 11) to `ods/tests/test-session-cleanup.sh` verifying mode preservation (0640) across atomic replacement. Ran `bash ods/tests/test-session-cleanup.sh` locally: 16/16 passed. Tested `patch-hermes-config.py`: syntax and execution verified.

### DIFF
--- a/ods/scripts/patch-hermes-config.py
+++ b/ods/scripts/patch-hermes-config.py
@@ -10,7 +10,9 @@ the rest of the user's config.
 from __future__ import annotations
 
 import argparse
+import os
 import re
+import tempfile
 from pathlib import Path
 
 
@@ -275,7 +277,19 @@ def patch_config(
         updated += "\n"
     if updated == original:
         return False
-    path.write_text(updated, encoding="utf-8")
+    fd, tmp_path = tempfile.mkstemp(dir=str(path.parent), suffix=".tmp")
+    try:
+        with os.fdopen(fd, "w", encoding="utf-8") as f:
+            f.write(updated)
+            f.flush()
+            os.fsync(f.fileno())
+        os.replace(tmp_path, str(path))
+    except BaseException:
+        try:
+            os.unlink(tmp_path)
+        except OSError:
+            pass
+        raise
     return True
 
 
--- a/ods/scripts/patch-hermes-config.py
+++ b/ods/scripts/patch-hermes-config.py
@@ -283,6 +283,17 @@ def patch_config(
             f.write(updated)
             f.flush()
             os.fsync(f.fileno())
+        if path.exists():
+            try:
+                st = os.stat(path)
+                os.chmod(tmp_path, st.st_mode)
+                if hasattr(os, "chown"):
+                    try:
+                        os.chown(tmp_path, st.st_uid, st.st_gid)
+                    except OSError:
+                        pass
+            except OSError:
+                pass
         os.replace(tmp_path, str(path))
     except BaseException:
         try:
--- a/ods/scripts/session-cleanup.sh
+++ b/ods/scripts/session-cleanup.sh
@@ -152,17 +152,28 @@ if [ -n "$WIPE_IDS" ]; then
 
     for ID in $WIPE_IDS; do
         "$PYTHON_CMD" -c "
-import json, sys
+import json, sys, os, tempfile
 sessions_file = sys.argv[1]
 target_id = sys.argv[2]
-with open(sessions_file, 'r') as f:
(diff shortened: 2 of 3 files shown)


Lightheartdevs:
Thanks for tightening the Hermes/session cleanup writes; these are exactly the kinds of files where partial writes are painful to debug.

Quickest path to merge:
- `ods/scripts/patch-hermes-config.py`: before `os.replace`, preserve the existing config file mode and owner/group where possible. The current `mkstemp` file will typically land as a new 0600 file, unlike the previous in-place write that preserved inode metadata; current `main`'s `render-runtime-configs.py:atomic_write_text` is a good local pattern for mode preservation.
- `ods/scripts/session-cleanup.sh`: apply the same metadata-preservation rule to the embedded Python rewrite of `sessions.json`, or add a clear reason/tests proving 0600/new-owner replacement is intentional for that file.
- Add focused tests for interrupted replace cleanup plus metadata preservation.

What I checked:
- Static diff review of the two touched write paths.
- Compared local atomic-write conventions in `ods/scripts/render-runtime-configs.py` and `ods/installers/macos/install-macos.sh`.
- `git diff --check origin/main...origin/pr/1946`.

Thanks again for moving these writes toward safer replacement semantics.

vaibhavsrv:
Thanks @Lightheartdevs for the sharp review on mode and ownership preservation!

I've updated `ods/scripts/patch-hermes-config.py`, `ods/scripts/session-cleanup.sh`, and `ods/tests/test-session-cleanup.sh`:
1. **Preserved File Mode & Ownership**: Both Python write paths now inspect `os.stat(path)` before replacement and reapply existing permissions (`os.chmod(tmp_path, st.st_mode)`) and owner/group (`os.chown(tmp_path, st.st_uid, st.st_gid)` where permitted) to `tmp_path` prior to calling `os.replace`.
2. **Added Mode Preservation Regression Test**: Added Test 11 to `ods/tests/test-session-cleanup.sh` verifying that custom mode permissions (e.g. `0640`) on `sessions.json` are preserved across atomic replacement.
3. **Updated PR Description**: Updated the PR body to reflect metadata preservation and error cleanup.

Local verification:
- `bash ods/tests/test-session-cleanup.sh` passed 100% green (16/16 passed).

Appreciate the review!

Lightheartdevs:
Thanks for strengthening the Hermes config and session cleanup replacement paths; preserving metadata before `os.replace` is the right follow-up.

Human call needed:
- The code now preserves mode and owner/group for both `ods/scripts/patch-hermes-config.py` and the embedded `sessions.json` rewrite in `ods/scripts/session-cleanup.sh` before replacement, and the new regression checks the `0640` mode case.
- I could not independently validate that focused mode test from this Windows runner: the default `bash` command resolves to a broken WSL relay, and Git Bash reports the new Test 11 as `expected 640, got 644` because POSIX `chmod`/`stat` mode bits are not reliable on this NTFS/Git Bash path. Since `ods/tests/test-session-cleanup.sh` is not currently part of the GitHub `test-linux.yml` checks, a maintainer should either accept the contributor's POSIX local `16/16` result or rerun `bash ods/tests/test-session-cleanup.sh` on Linux/macOS before merging.

What I checked:
- Reviewed the PR metadata, prior advisory thread, current diff, changed write paths, and the new mode-preservation regression.
- `gh pr checks 1946` is green across Test Linux, lint, Dashboard, secret scan, env validation, and matrix smoke.
- In a detached worktree at `ea26dd2c`, `git diff --check origin/main...HEAD` and `python -m py_compile ods/scripts/patch-hermes-config.py` passed.
- Attempted the focused shell test locally with Git Bash; I am not treating that Windows filesystem mode result as product evidence.

Thanks again for moving these writes toward safer replacement semantics.

PR #1947, opened 2026-07-21
fix(installer): implement atomic writing for registry/summaries and resolve macOS test hang

vaibhavsrv:
## Summary

Prevent in-place truncation when updating installer state files and fix a macOS-specific test hang.

Previously, several installer and upgrade paths updated JSON state files directly using shell redirection or in-place Python writes. These write paths truncate the destination before writing new content, which can leave the file incomplete if the process is interrupted during the update.

This change updates those write paths to use atomic file replacement:

- `background-tasks.sh`
  - Writes the background task registry to a temporary file.
  - Flushes the file with `fsync()`.

- `ods-update.sh`
  - Writes version status updates to a temporary file created with `mktemp`.
  - Atomically replaces the destination using `mv -f`.

- `13-summary.sh`
  - Writes the installation summary to a temporary file.
  - Flushes the file with `fsync()`.

As a result, these state files are no longer exposed to partially written or truncated content during updates.

In addition, this change fixes a macOS-specific issue in `test-bootstrap-upgrade-resume-status.sh`. The test previously derived its upgrade lock key from unresolved temporary paths, while `bootstrap-upgrade.sh` resolved the corresponding paths using their physical location. On macOS, where `/var` is a symlink to `/private/var`, this produced different lock keys and prevented the test from locating the existing lock directory, resulting in repeated retry attempts. The test now resolves temporary directories using `pwd -P`, matching the daemon's behavior and eliminating the mismatch.

## Verification

Verified using the existing test suites:

- `test-bootstrap-upgrade-resume-status.sh`
  - Completes successfully on macOS without hanging (approximately 4 seconds).

- `make bats`
  - Passed, including the background task registry regression tests.

### DIFF
--- a/ods/installers/lib/background-tasks.sh
+++ b/ods/installers/lib/background-tasks.sh
@@ -31,7 +31,9 @@ bg_task_start() {
     # Add task to registry
     python3 - "$BG_TASK_REGISTRY" "$task_id" "$pid" "$description" "$log_file" <<'PY'
 import json
+import os
 import sys
+import tempfile
 from pathlib import Path
 
 registry_path = Path(sys.argv[1])
@@ -48,7 +50,19 @@ tasks.append({
     "log_file": log_file,
     "status": "running"
 })
-registry_path.write_text(json.dumps(tasks, indent=2))
+fd, tmp_path = tempfile.mkstemp(dir=str(registry_path.parent), suffix=".tmp")
+try:
+    with os.fdopen(fd, "w") as f:
+        json.dump(tasks, f, indent=2)
+        f.flush()
+        os.fsync(f.fileno())
+    os.replace(tmp_path, str(registry_path))
+except BaseException:
+    try:
+        os.unlink(tmp_path)
+    except OSError:
+        pass
+    raise
 PY
 }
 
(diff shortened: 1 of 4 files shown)


Lightheartdevs:
Thanks for hardening these installer state writes; the background task registry, summary JSON, and update version file all sit on paths where a partial write can make recovery harder.

What looks good:
- `ods/installers/lib/background-tasks.sh` and `ods/installers/phases/13-summary.sh` now write through peer temp files, flush and `fsync` the temp contents, clean up on exceptions, and finish with `os.replace()` on the destination.
- `ods/ods-update.sh` now routes both `.version` update sites through a temp file followed by `mv -f`, so the common successful path no longer truncates the live version file before the new JSON is available.
- The macOS fixture change in `ods/tests/test-bootstrap-upgrade-resume-status.sh` uses `pwd -P`, matching the bootstrap upgrade lock-key behavior around `/var` versus `/private/var` without changing production installer logic.

What I checked:
- Reviewed the PR description, full patch, changed files, and current discussion for #1947.
- Checked the atomic-write sections in `background-tasks.sh`, `13-summary.sh`, and `ods-update.sh` against the installer architecture's generated-state/recovery expectations.
- Ran `git diff --check origin/main...pr-1947` locally: passed.
- GitHub checks are green for dashboard API/frontend, Python lint, ShellCheck, PowerShell lint, Linux integration smoke, distro matrix, macOS smoke, env validation, secret scan, and Claude review. I did not rerun the Bash suites locally because this Windows automation host does not provide a usable `/bin/bash` runtime.

Lightheartdevs (approved):
Maintainer audit complete: atomic peer-file replacement correctly prevents partial installer state, the production paths remain compatible, current CI is green, and the macOS path-normalization fixture matches runtime behavior.

PR #1948, opened 2026-07-21
fix(security): write generated API key files atomically

vaibhavsrv:
## Summary

Prevent in-place truncation when generating fallback API keys for Dashboard API and Token Spy.

Previously, both services generated a fallback API key when the corresponding environment variable was not set and wrote it directly to disk using `Path.write_text()`. Since this truncates the destination file before writing the new content, an interrupted write could leave the key file empty or partially written, preventing subsequent startups from using a valid persisted key.

This change updates both key generation paths to use atomic file replacement:

- `dashboard-api`
  - Writes the generated `dashboard-api-key.txt` to a temporary file in the target directory.

- `token-spy`
  - Writes the generated `token-spy-api-key.txt` to a temporary file in the target directory.

As a result, the persisted API key is always either the complete previous value or the complete newly generated value, avoiding partially written or truncated key files.

## Verification

Verified using the existing Python test suites:

- `python3 -m pytest -v ods/extensions/services/dashboard-api/tests/test_security.py`
  - **6 passed**

- `python3 -m pytest -v ods/extensions/services/token-spy/tests/test_usage_report.py`
  - **11 passed**

- `make lint`
  - Passed.

### DIFF
--- a/ods/extensions/services/dashboard-api/security.py
+++ b/ods/extensions/services/dashboard-api/security.py
@@ -15,7 +15,9 @@ if not DASHBOARD_API_KEY:
     DASHBOARD_API_KEY = secrets.token_urlsafe(32)
     key_file = Path("/data/dashboard-api-key.txt")
     key_file.parent.mkdir(parents=True, exist_ok=True)
-    key_file.write_text(DASHBOARD_API_KEY)
+    tmp_key_file = key_file.with_name(f".{key_file.name}.tmp")
+    tmp_key_file.write_text(DASHBOARD_API_KEY)
+    os.replace(str(tmp_key_file), str(key_file))
     key_file.chmod(0o600)
     logger.warning(
         "DASHBOARD_API_KEY not set. Generated temporary key and wrote to %s (mode 0600). "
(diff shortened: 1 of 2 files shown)


vaibhavsrv:
Thanks for reviewing!

I've updated the PR to ensure mode `0600` permissions are set directly on the temporary key file *before* performing the atomic `os.replace()`. This guarantees the generated security key is owner-only from the instant it lands on disk.

Local security tests (`pytest ods/extensions/services/dashboard-api/tests/test_security.py ods/extensions/services/token-spy/tests/`) passed cleanly (17 passed).

Lightheartdevs:
Thanks for tightening the fallback API key writes for Dashboard API and Token Spy; avoiding truncated or partially written key files is the right fix for this startup path.

What looks good:
- `ods/extensions/services/dashboard-api/security.py` and `ods/extensions/services/token-spy/main.py` now write generated keys to a sibling temp file and land them with `os.replace`, so an interrupted write should leave either the prior complete key or the new complete key.
- The latest head sets `0600` on the temp file before the atomic replace, which closes the short permission window that would exist if chmod only happened after the destination was replaced.
- The change is tightly scoped to the two generated-key fallback paths and does not alter the configured-env-key path or broader auth behavior.

What I checked:
- Reviewed the PR body, contributor follow-up comment, and current diff against the repo's security/runtime assumptions for generated local secrets.
- Checked GitHub CI for this head: Python lint/type, shell lint, PowerShell lint, dashboard API/frontend, integration smoke, distro matrix, env validation tiers, and secret scan are passing.
- I did not rerun the service pytest suites locally in this Windows scheduled run; the PR reports the focused security tests passing, and CI is green.

Thanks again for closing the permission detail before merge.

PR #1961, opened 2026-07-21
fix/remote-session-cleanup-atomic-writes 

programmerrocket:
## Summary

Prevent in-place truncation when updating `sessions.json` on remote Token Spy agents.

Previously, the inline Python script executed by `_kill_remote_session` updated the remote `sessions.json` file using `open(..., "w")`. Since this truncates the destination before writing the new contents, an interrupted update (for example, due to an SSH disconnect or process termination) could leave the session database incomplete or empty.

This change updates the remote write path to use atomic file replacement:

- Writes the updated session data to a temporary file in the destination directory using `tempfile.mkstemp()`.

As a result, remote agents always observe either the complete previous session database or the complete updated version, avoiding partially written or truncated state during remote session cleanup.

## Verification

Verified using the existing Token Spy test suite:

- `python3 -m pytest -v ods/extensions/services/token-spy/tests/test_usage_report.py`
  - **11 passed**

- `make lint`
  - Passed.

### DIFF
--- a/ods/extensions/services/token-spy/main.py
+++ b/ods/extensions/services/token-spy/main.py
@@ -1332,7 +1332,7 @@ def _kill_remote_session(agent: str, reason: str = "dashboard") -> dict:
     sessions_dir = remote["sessions_dir"]
 
     script = (
-        "import os, glob, json, sys\n"
+        "import os, glob, json, sys, tempfile\n"
         "sdir = sys.argv[1]\n"
         "files = sorted(glob.glob(os.path.join(sdir, '*.jsonl')), key=os.path.getsize, reverse=True)\n"
         "if not files:\n"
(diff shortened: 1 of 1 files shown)


Lightheartdevs:
Thanks for tightening the remote Token Spy session cleanup path; avoiding a truncated remote `sessions.json` is exactly the kind of durability fix that fits this stack.

What looks good:
- `ods/extensions/services/token-spy/main.py` now writes the remote session database through a same-directory temporary file and swaps it into place with `os.replace()`, so an interrupted SSH-side cleanup no longer leaves readers with an empty or partial file.
- The cleanup path removes the temporary file on write/replace failure and preserves the existing fail-soft warning behavior around `sessions.json` update errors.
- The change stays inside `_kill_remote_session()` and does not alter session selection, kill signaling, API shape, or local Token Spy persistence.

What I checked:
- Reviewed the PR description, changed file, patch, and surrounding `_kill_remote_session()` remote script construction.
- Ran `python -m pytest ods/extensions/services/token-spy/tests/test_usage_report.py -q` locally on this head: 11 passed.
- Confirmed GitHub checks for this head are green, including Python lint, dashboard API/frontend, shell lint, secret scan, integration smoke, distro matrix, macOS smoke, and env validation.

PR #1974, opened 2026-07-22
fix/support-bundle-atomic-redact

westwood18:
## Summary

Prevent in-place truncation when redacting files for support bundle generation.

Previously, `redact_file()` in `ods/scripts/ods-support-bundle.sh` rewrote the redacted file directly using `Path.write_text()`. Since this truncates the destination before writing the new contents, an interrupted write could leave the redacted copy incomplete or empty.

This change updates the redaction path to use atomic file replacement:

- Writes the redacted contents to a temporary file in the destination directory.
- Atomically replaces the original file using `os.replace()` after the write completes successfully.

As a result, the support bundle always contains either the complete original redacted file or the complete updated version, avoiding partially written or truncated files if redaction is interrupted.

## Verification

Verified using the project's existing checks:

- `make lint`
  - Passed.

### DIFF
--- a/ods/scripts/ods-support-bundle.sh
+++ b/ods/scripts/ods-support-bundle.sh
@@ -126,8 +126,10 @@ redact_file() {
     [[ -f "$file" ]] || return 0
 
     "$PYTHON_CMD" - "$file" <<'PY'
+import os
 import re
 import sys
+import tempfile
 from pathlib import Path
 
 path = Path(sys.argv[1])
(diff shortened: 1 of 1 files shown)


Lightheartdevs:
Thanks for hardening the support-bundle redaction path; avoiding truncated diagnostics is useful since support bundles are often created while someone is already debugging an unstable install.

What looks good:
- `ods/scripts/ods-support-bundle.sh` now redacts into a peer temp file and uses `os.replace()` so callers observe either the previous complete file or the newly redacted file.
- The cleanup path removes the temporary file on exceptions without masking the original error.
- The change stays inside `redact_file()` and does not alter the redaction patterns or bundle manifest behavior.

What I checked:
- Reviewed the patch against the current support-bundle collection flow and Python usage in the script.
- Checked the repo's Python assumptions across installer/test docs; the support-bundle helper already requires a runnable `python3`/`python`.
- `gh pr checks` reports no checks for this branch; I did not run a full support-bundle generation locally.

Issue #1964, opened 2026-07-21
bug(hermes): stock INSECURE public bind breaks Dream Talk on post-hardening Hermes images


jpvelasco:
## Description

Stock ODS Hermes packaging still assumes an unauthenticated non-loopback dashboard bind via `HERMES_DASHBOARD_INSECURE=1`. Current Hermes Agent images refuse that bind unless a password or OAuth provider is configured.

That breaks the documented Hermes / hermes-proxy / Dream Talk path even when containers report healthy. Dream Talk depends on scraping `window.__HERMES_SESSION_TOKEN__` from the Hermes HTML, which only works when the Hermes dashboard is in ungated (loopback) mode.

Related but different: #1544 (pruned ephemeral `sha-` image tags on release 2.5.2 / 2.5.3).

## Steps to Reproduce

1. Install from latest release tag `v2.5.3`, or from current `main`, with Hermes and hermes-proxy enabled (default recommended path).
2. Note stock compose still sets:
   - `HERMES_DASHBOARD_HOST=0.0.0.0`
   - `HERMES_DASHBOARD_INSECURE=1`
   - proxy / Dream Talk targeting Docker DNS `*:9119` on the bridge network
3. Use a Hermes image that includes non-loopback auth-gate hardening (current Hub tags / recent pins). On `v2.5.3`, first hit may be image pull 404 for `sha-dd0923bb...` (#1544); substitute a retained modern tag if needed to reach this failure mode.
4. Start the stack and inspect Hermes dashboard logs / bind state.
5. Optionally force basic auth so Hermes binds on `0.0.0.0`, then open Dream Talk and start a chat session.
6. Optionally bind Hermes to loopback only (to keep token inject and avoid Hermes password UI) without packaging updates for Dream Talk reachability.

## Expected Behavior

After a normal install with Hermes enabled:

1. Hermes dashboard is reachable on the Docker network as designed.
2. Owner-card gated advanced UI works on the hermes-proxy port.
3. Dream Talk can create a session via the server-side bridge without a Hermes password UI.
4. Dashboard health / feature launch links match real reachability.
5. Shipped Hermes image pin is pullable (retained date tag or digest).

## Actual Behavior

1. **Public bind without auth provider:** Hermes logs that `INSECURE` no longer disables the auth gate and refuses to bind `0.0.0.0`. Dashboard never listens for other containers.
[...]

PR #1976, opened 2026-07-22
fix(hermes): support basic auth in hermes_bridge and compose for post-hardening auth gate (#1964)

westwood18:
## Summary

Restore Dashboard API connectivity when Hermes is configured with Basic Authentication.

Recent Hermes releases require an authentication provider before allowing non-loopback dashboard binds. When operators configured Basic Authentication for the Hermes dashboard, the Dashboard API bridge continued making unauthenticated requests, causing authentication failures when retrieving session tokens or establishing the JSON-RPC WebSocket connection.

This change adds Basic Authentication support to the Hermes bridge:

- Adds `_get_hermes_auth_header()` to construct an `Authorization: Basic` header from:
  - credentials embedded in `HERMES_INTERNAL_URL`
  - `HERMES_DASHBOARD_BASIC_AUTH_USERNAME` / `HERMES_DASHBOARD_BASIC_AUTH_PASSWORD`
  - `HERMES_BASIC_AUTH_USER` / `HERMES_BASIC_AUTH_PASS`
- Applies the generated authentication header to:
  - HTTP session token requests
  - JSON-RPC WebSocket connections
- Updates `compose.yaml` to:
  - pass through Basic Auth environment variables
  - allow `HERMES_DASHBOARD_HOST` to be overridden via `${HERMES_DASHBOARD_HOST:-0.0.0.0}`

As a result, Dashboard API can successfully communicate with Hermes deployments protected by Basic Authentication while preserving existing behavior for deployments that do not require authentication.

## Verification

Added regression coverage in `ods/extensions/services/dashboard-api/tests/test_talk.py`:

- `test_hermes_bridge_get_auth_from_url_and_env`
  - Verifies Basic Authentication header generation from:
    - credentials embedded in `HERMES_INTERNAL_URL`
    - Basic Auth environment variables

### Test Results

- `python3 -m pytest -v ods/extensions/services/dashboard-api/tests/test_talk.py`
  - **38 passed**

- `make lint`
  - Passed.

### DIFF
--- a/ods/extensions/services/dashboard-api/hermes_bridge.py
+++ b/ods/extensions/services/dashboard-api/hermes_bridge.py
@@ -27,6 +27,7 @@ visitors doesn't pin Hermes resources forever.
 from __future__ import annotations
 
 import asyncio
+import base64
 import hashlib
 import json
 import logging
@@ -35,6 +36,7 @@ import re
 import time
 from dataclasses import dataclass, field
 from typing import Any, AsyncIterator
+from urllib.parse import urlparse
 
 import aiohttp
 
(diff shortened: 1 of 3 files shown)


Lightheartdevs:
Thanks for tightening the Hermes auth path here; this is the right place to restore Dashboard API connectivity after the dashboard auth hardening.

What looks good:
- `ods/extensions/services/dashboard-api/hermes_bridge.py` applies the Basic auth header to both the dashboard token fetch and the JSON-RPC WebSocket connection, which matches the bridge's two-step session flow.
- The no-auth path remains unchanged because `_get_hermes_auth_header()` returns an empty header map when no username is configured.
- `ods/extensions/services/hermes/compose.yaml` passes through the dashboard Basic auth env vars without forcing values for installs that do not use them.

What I checked:
- Reviewed the patch plus surrounding Hermes bridge connection-pool/session-token code on current repo context.
- Ran `python -m pytest -q tests/test_talk.py` from `ods/extensions/services/dashboard-api`: 38 passed.
- `gh pr checks` reports no checks for this branch, so the local targeted test is the main automated evidence I found.

PR #1985, opened 2026-07-22
fix(macos): honor compose platform pins during image pre-pull

### LINKED ISSUE #1958: Could not install ODS on MacStudio M3 Ultra 96Gb
### Description

Hi, I saw your video today on HuggingFace and wanted to give ODS a try, when I install I get the following message:
> Pulling Compose image (4/4): ghcr.io/huggingface/text-embeddings-inference:cpu-1.9.1
  [XX] Failed to pull Compose image after retries: ghcr.io/huggingface/text-embeddings-inference:cpu-1.9.

And then the installer ends with this message:
INFO] Compose image already cached: ghcr.io/speaches-ai/speaches:0.9.0-rc.3-cpu
  [XX] 1 Compose image(s) could not be pulled before launch
  > macOS installer will not allow Docker Compose to pull images implicitly.
  > Fix Docker registry/network/disk access, then re-run ./installers/macos/install-macos.sh.
  [!!] Compose failure report saved: /Users/mstudio-2/ods/install-report-2026-07-21-112940.txt

Thanks,

### Steps to Reproduce

git clone https://github.com/Osmantic/ODS.git
cd ODS/ods
./install.sh

Select option 1 - Full Stack

### Expected Behavior

Install ODS

### Actual Behavior

Installer ends

### Operating System

Tahoe 26.3.1

### GPU

Apple

### Docker Version

4.83.0

### VRAM

96gb

### Logs

```shell

```

### Screenshots

_No response_

### LINKED ISSUE #1987: bug(macos): pre-pull strips per-service platform pins, causing arm64 pull failures for amd64-only images
### Description

The macOS installer's pre-pull step builds its image list using `docker compose config --images`, which returns image names only and drops all per-service `platform:` pins defined in the Compose files. As a result, when Docker pulls these images on Apple Silicon without a platform flag, it defaults to the host platform (`linux/arm64/v8`) and fails for amd64-only images, even though the Compose file correctly specifies `platform: linux/amd64`.

Currently, the only affected service is `extensions/services/embeddings/compose.yaml` (the TEI image, which is amd64-only), but this failure mode applies to any future extension that introduces a platform-pinned image. Because the installer treats a failed pre-pull as a fatal error, it aborts the installation instead of allowing Docker Compose to perform the pull later with the correct platform pin applied.

Root cause confirmed by @bokiko in #1958: `docker compose config --images` strips platform metadata, leaving `_macos_pull_image_with_retry` with no way to determine or pass the required `--platform` flag when iterating the image list.

### Steps to Reproduce

1. Use any Apple Silicon Mac (M1/M2/M3/M4).
2. Clone the repository:

   ```bash
   git clone https://github.com/Osmantic/ODS.git
   ```
3. Change into the installer directory:

   ```bash
   cd ODS/ods
   ```
4. Run the installer and select **Full Stack**:

   ```bash
   ./install.sh
   ```
5. Observe the installation failing during the pre-pull of:

   ```text
   ghcr.io/huggingface/text-embeddings-inference:cpu-1.9.1
   ```

### Independent Verification

Run the following commands:

```bash
docker pull ghcr.io/huggingface/text-embeddings-inference:cpu-1.9.1
```

Expected result:

```text
Error: no matching manifest for linux/arm64/v8
```

Then run:

```bash
docker pull --platform linux/amd64 ghcr.io/huggingface/text-embeddings-inference:cpu-1.9.1
```

Expected result:

```text
Pull succeeds (the image is published for amd64 only).
```

### Environment

Confirmed reproducible on:

* **M3 Ultra** — macOS Tahoe 26.3.1, Docker Desktop 4.83.0
* **M4 Pro** — macOS 26.5.1, Colima (aarch64 engine)

See #1958 for the original investigation and confirmation.

### Expected Behavior

The macOS pre-pull step should honor the `platform:` pin defined for each service in its Compose file. When a service specifies a platform override, the installer should pull the image with the corresponding `--platform` flag, ensuring Docker selects the correct image manifest regardless of the host architecture.

With the correct platform applied, amd64-only images (such as the TEI image) are pulled successfully, and the installation completes normally on Apple Silicon systems.

### Actual Behavior

The macOS pre-pull step builds its image list using `docker compose config --images`, which drops all `platform:` metadata. As a result, the installer performs a plain:

```bash
docker pull <image>
```

without a corresponding `--platform` flag.

On Apple Silicon, Docker defaults to `linux/arm64/v8`. For the amd64-only TEI image, no arm64 manifest exists, so the pull fails. Because the failure is caused by an unavailable manifest rather than a transient network issue, retries cannot succeed.

After exhausting its retries, the installer aborts the installation instead of allowing Docker Compose to perform the pull later with the correct platform pin:

```text
[XX] Failed to pull Compose image after retries:
     ghcr.io/huggingface/text-embeddings-inference:cpu-1.9.1
[XX] 1 Compose image(s) could not be pulled before launch
```

As a result, the macOS installer terminates before Docker Compose has an opportunity to pull the image using the `platform: linux/amd64` setting defined in the Compose file.

### Operating System

macOS (Apple Silicon) — confirmed on Tahoe 26.3.1 and macOS 26.5.1

### GPU

Apple Silicon (M1/M2/M3/M4 — any Apple unified memory chip)

### Docker Version

Docker Desktop 4.83.0 (confirmed); also reproducible on Colima aarch64 engine

### VRAM

Not applicable — failure is a registry manifest issue, not GPU/memory dependent. Reproduces on any Apple Silicon regardless of unified memory size.

### Logs

```shell

```

### Screenshots

_No response_

bokiko:
Fixes #1958.
Fixes #1987.

## Problem

The macOS installer pre-pulls compose images with a plain `docker pull "$image"`. The image list comes from the compose config with per-service `platform:` pins dropped. `ghcr.io/huggingface/text-embeddings-inference:cpu-1.9.1` publishes `linux/amd64` only, so on Apple Silicon the unpinned pull resolves the host platform and fails deterministically:

```
Error response from daemon: no matching manifest for linux/arm64/v8 in the manifest list entries: no match for platform in manifest: not found
```

Retries cannot help, and the install aborts even though `extensions/services/embeddings/compose.yaml` already pins `platform: linux/amd64` so this service can run under emulation.

A second edge case existed when the same tag was already cached for the host architecture: the generic `docker image inspect` shortcut skipped the pinned pull, then Compose started with `--pull never` and could reuse the incompatible variant.

## Fix

- `_macos_compose_external_images` now emits `image<TAB>platform` from rendered Compose JSON instead of discarding each service platform pin.
- `_macos_pull_image_with_retry` accepts the optional platform and uses `docker pull --platform ...` for pinned services.
- Platform-pinned images always resolve their exact manifest; only unpinned images may use the generic cached-tag shortcut.
- The `config --images` fallback remains compatible with older behavior when platform metadata is unavailable.

## Scope and invariants

- macOS installer only; Linux and Windows paths are unchanged.
- Built local images remain excluded from registry pulls.
- Unpinned external images preserve the existing plain-pull and cache behavior.
- Retry count, backoff, failure reporting, and `compose up --pull never` semantics are unchanged.
- Fresh install, rerun, partial image cache, and wrong-architecture cached-tag cases all converge on the platform declared by the resolved Compose service.

### Real Apple Silicon receipt

Reproduced on an M4 Pro (macOS 26.5.1, Docker 29.2.1): the unpinned pull fails with the manifest error above, while `--platform linux/amd64` succeeds (digest `b7772cdd...`).

### Current-head validation

- `bash -n ods/installers/macos/install-macos.sh ods/tests/contracts/test-installer-hardening.sh`
- `bash ods/tests/contracts/test-installer-hardening.sh`
- `bash ods/tests/contracts/test-installer-contracts.sh`
- `bash ods/tests/test-docker-image-pull-retry.sh` (13/13)
- `bash ods/tests/smoke/macos-dispatch.sh`
- Real Compose JSON render of `extensions/services/embeddings/compose.yaml` confirms `platform: linux/amd64` is preserved.
- `git diff --check osmantic/main...HEAD`

The hardening contract now also prevents a future regression where a platform-pinned image is skipped merely because an incompatible variant of the same tag is cached.


bokiko:
Quick linkage note: #1987 now tracks the exact platform-pin pre-pull bug this PR fixes, and the PR body has been updated so it will close #1987 on merge.

No urgency — just keeping the review context tied together.

Lightheartdevs:
﻿
Thanks for taking on the macOS Compose pre-pull path here; preserving per-service platform pins is exactly the right place to fix the Apple Silicon TEI failure.

Quickest path to merge:
- `ods/installers/macos/install-macos.sh`: the patch looks directionally sound, but there are now two open PRs aimed at the same macOS platform-pin pre-pull issue (`#1985` and `#1989`). A maintainer should choose which branch is the canonical fix so we do not merge overlapping installer changes twice.
- GitHub checks: this head still shows `distro: ubuntu-22.04` failing while the other lanes are green. If that failure is unrelated/transient, a rerun or maintainer note would make the status clear; if it reproduces, resolve it before merge even though the touched path is macOS-specific.
- `ods/tests/contracts/test-installer-hardening.sh`: the added contract assertions cover the important strings, but this is a high-risk installer path. Keep the PR body's M4 Pro validation in the merge record, or rerun the macOS smoke/contract lane after picking the canonical PR.

What I checked:
- Diff/static review of `ods/installers/macos/install-macos.sh` and `ods/tests/contracts/test-installer-hardening.sh`.
- PR body validation notes for the Apple Silicon failure and pinned `docker pull --platform` behavior.
- GitHub checks for this head: most lanes pass, but `distro: ubuntu-22.04` is failing.
- Local shell contract was not runnable in this Windows automation environment because `/bin/bash` is unavailable here.

Thanks again for chasing the field failure down to the installer pre-pull behavior.

bokiko:
Thanks, that matches my read.

If maintainers want the broader helper-level fix, #1989 looks like the better canonical PR: it carries platform metadata through the shared compose image helpers as well as the macOS path.

I'm happy for #1985 to be closed as superseded if #1989 is the direction maintainers prefer. If you prefer the narrower macOS-only patch instead, this PR is still available as the minimal fix. I won't push further changes here unless asked.

gabsprogrammer:
Updated the branch onto current `main` and closed one additional platform invariant:

- platform-pinned services no longer trust a generic cached tag, because that tag may contain the host-architecture variant;
- pinned images always run `docker pull --platform <compose-platform> ...` before the existing `compose up --pull never` path;
- unpinned services retain the existing cache shortcut;
- added a hardening regression for the wrong-architecture cached-tag case.

Validation on the updated head:

- Bash parser checks passed.
- Installer hardening contracts passed.
- Full installer contracts passed.
- Docker image pull retry suite passed (13/13).
- macOS dispatch smoke passed.
- Real Compose JSON rendering preserves `embeddings.platform = linux/amd64`.
- `git diff --check` passed.

Lightheartdevs:
Thanks for tightening the macOS image pre-pull path after the branch refresh; this now addresses the Apple Silicon TEI failure without changing the shared Linux helper contract.

What looks good:
- `ods/installers/macos/install-macos.sh` now carries each Compose service's `platform` through the JSON path and passes it to `docker pull --platform`, while the older `config --images` fallback still behaves like the current main path.
- Platform-pinned images no longer trust a generic cached tag, which closes the wrong-architecture cache case before the installer reaches `compose up --pull never`.
- The added contract assertions in `ods/tests/contracts/test-installer-hardening.sh` cover the platform extraction, pinned pull, and cache-skip guard that matter for this regression.

What I checked:
- Diff/static review of `ods/installers/macos/install-macos.sh` and `ods/tests/contracts/test-installer-hardening.sh` against current `main` and the overlapping #1989 discussion.
- GitHub checks for this head are green across dashboard API/frontend, Python lint/type, shell and PowerShell lint, secret scan, Linux integration smoke, matrix smoke including macOS smoke, and env validation.
- Local Bash validation was not runnable in this Windows automation image because `/bin/bash` is unavailable here.

Thanks again for chasing this through the exact installer pre-pull behavior.

Lightheartdevs:
Audit repair pushed in 43da9bb1. The pre-pull now inspects cached image architecture, reuses a matching amd64 cache offline, pulls only when absent/mismatched/unverifiable, and remains fail-closed if a wrong-architecture replacement cannot be obtained. The new seven-case executable contract, syntax/diff checks, and all GitHub CI are green. Holding merge for physical Apple Silicon receipts: missing cache, correct amd64 cache offline, wrong arm64 cache, registry unavailable, TEI health, restart, and an actual RAG query.

Lightheartdevs:
Thanks for tightening the macOS pre-pull fix again; the latest head now handles the important cache distinction between a usable pinned-platform image and a same-tag image for the wrong architecture.

Quickest path to merge:
- Attach the physical Apple Silicon receipts called out in the latest maintainer note for `43da9bb1`: missing cache, matching `linux/amd64` cache offline, wrong `arm64` cache replacement, registry unavailable/fail-closed behavior, TEI health, restart persistence, and an actual RAG query. This changes `ods/installers/macos/install-macos.sh`, so the high-risk installer map points to platform-backed validation before merge.
- Keep the overlap with prior platform-pin discussion explicit in the merge record. This PR is now the narrower macOS installer fix with cache architecture inspection; that is a coherent path as long as maintainers are intentionally choosing it over broader helper-level changes.

What I checked:
- Reviewed the current PR body, discussion, and diff for `ods/installers/macos/install-macos.sh`, `ods/tests/contracts/test-installer-contracts.sh`, `ods/tests/contracts/test-installer-hardening.sh`, and `ods/tests/test-macos-compose-image-cache.sh`.
- Ran `C:\Program Files\Git\bin\bash.exe ods/tests/test-macos-compose-image-cache.sh`: passed all seven cache/platform cases.
- Ran `git diff --check main...HEAD`: passed.
- Checked GitHub CI for the current head: lint, dashboard API/frontend, Linux integration smoke, distro matrix including macOS smoke, PowerShell lint, env validation, and secret scan are green.
- Attempted `ods/tests/contracts/test-installer-hardening.sh` locally; it reached this Windows/Git Bash host's unsupported-OS bootstrap branch, so I would treat the GitHub Linux lane and requested Apple Silicon receipts as the authoritative signals for that script.

Thanks again for chasing this through the real cache behavior.

Lightheartdevs:
Thanks for keeping the macOS pre-pull branch current with `main`; the cache/platform handling remains the right area to fix for the Apple Silicon TEI install failure.

Quickest path to merge:
- Add the physical Apple Silicon receipts already called out for `ods/installers/macos/install-macos.sh`: missing cache, matching `linux/amd64` cache offline, wrong `arm64` cache replacement, registry unavailable/fail-closed behavior, TEI health, restart persistence, and an actual RAG query. This is a high-risk installer/runtime path, so the hardware-backed evidence matters even with green CI.
- Keep the merge record explicit that this PR is the narrower macOS pre-pull fix with cached-image architecture inspection, chosen intentionally over the broader overlapping helper-level approach discussed earlier.
- No code issue jumped out from the current diff, but I would hold the final merge decision until the Apple Silicon receipts are attached or a maintainer explicitly accepts the narrower validation.

What I checked:
- Reviewed the current PR body, latest discussion, current head SHA, and diff for `ods/installers/macos/install-macos.sh`, `ods/tests/contracts/test-installer-contracts.sh`, `ods/tests/contracts/test-installer-hardening.sh`, and `ods/tests/test-macos-compose-image-cache.sh`.
- Compared the latest head to the prior advisory SHA; the new head is a merge-from-main refresh, while the macOS pre-pull behavior remains the previously audited patch.
- Checked GitHub CI for the current head: shell lint, dashboard API/frontend, Python lint, Linux integration smoke, distro matrix including macOS smoke, PowerShell lint, env validation, and secret scan are passing.

Thanks again for chasing this through the real installer cache behavior.

gabsprogrammer:
Re-audited and refreshed on current `main` at `bdfd8574`.

### Canonical scope
This remains the intentionally narrow macOS installer fix. It is preferred over the broader `#1989` helper-level approach because `#1989` changes the shared image-entry contract and has known cross-consumer regressions in Phase 08 and `ods-cli`; this PR confines the behavior to the macOS pre-pull path.

### Current-head evidence
- Executable cache/platform contract passed all seven cases: missing cache, matching amd64 cache reuse, wrong-architecture replacement, normalized arm64 cache reuse, and fail-closed behavior when replacement pulls fail.
- macOS dispatch smoke passed.
- Docker image pull retry suite passed `13/13`.
- Bash parser checks and `git diff --check` passed.
- Real Compose JSON rendering preserves `embeddings.image = ghcr.io/huggingface/text-embeddings-inference:cpu-1.9.1` with `platform = linux/amd64`.
- All published CI checks are green on this exact head, including ShellCheck, integration smoke, macOS smoke, distro matrix, dashboard API/frontend, PowerShell lint, env validation, and secret scan.

The independent review found no remaining reproducible code defect in scope, so no speculative behavior changes were added. The outstanding merge gate is still the requested end-to-end receipt from physical Apple Silicon: TEI health, restart persistence, and a real RAG query in addition to the cache scenarios. That cannot be truthfully produced from this Windows host; it requires a physical Apple Silicon run or explicit maintainer acceptance of the existing automated coverage plus the M4 Pro pull receipt already recorded in the PR body.

gabsprogrammer (approved):

PR #1986, opened 2026-07-22
Custom installer via curl breaks with multi-GPUs, resolved if each interactive prompt has `< /dev/tty`

keennay:
This is a simple fix, installation via `curl -fsSL https://install.osmantic.com/ods.sh | bash` breaks for certain multiple GPU setups.

This is resolved if each interactive prompt has `< /dev/tty` on its read command, provided in this PR.

Video capture of failure on my end:

https://github.com/user-attachments/assets/8aa75ad3-0358-436a-a083-99cfcdbd9ee3

### DIFF
--- a/ods/get-ods.sh
+++ b/ods/get-ods.sh
@@ -382,7 +382,7 @@ if [[ -d "$INSTALL_DIR" ]]; then
             exit 1
         else
             echo -n "  Remove and reinstall? [y/N] "
-            read -r response
+            read -r response < /dev/tty
             if [[ "$response" =~ ^[Yy]$ ]]; then
                 remove_install_dir "$INSTALL_DIR" || error "Failed to remove incomplete install at $INSTALL_DIR. Try: sudo rm -rf \"$INSTALL_DIR\""
             else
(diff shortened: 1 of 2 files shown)


Lightheartdevs:
Thanks for tracking down the curl-pipe stdin problem in the custom multi-GPU installer path; that failure mode is very plausible when the bootstrap script is read from the pipeline.

Quickest path to merge:
- `ods/installers/phases/03-features.sh`: add a small contract test or harness that runs the custom multi-GPU prompt path with stdin occupied by a pipe and confirms the selection/GPU prompts read from `/dev/tty`. The touched lines look consistent with the earlier feature prompts, but this installer path is high-impact enough that we should not merge it without a reproducible check.
- `ods/get-ods.sh`: include the existing-install reinstall prompt in that same verification, since this PR also changes its `read` source and that script is the public curl bootstrap path.
- Trigger the normal shell validation checks after the test lands; GitHub currently shows no check runs for this PR, so there is no CI signal for syntax, ShellCheck, or installer contracts yet.

What I checked:
- Reviewed the diff for `ods/get-ods.sh` and `ods/installers/phases/03-features.sh` against the existing `/dev/tty` pattern already used earlier in phase 03.
- Confirmed `git diff --check origin/main...HEAD` is clean.
- Tried to run local `bash -n` on the touched shell files, but this Windows automation environment does not have a usable Bash runtime available.

gabsprogrammer:
It has already been fixed and merged into another PR with a more comprehensive solution, but I appreciate the interest in contributing to ODS.

PR #1989, opened 2026-07-22
fix(macos): preserve per-service platform pins in Compose pre-pull (#1987)

### LINKED ISSUE #1987: bug(macos): pre-pull strips per-service platform pins, causing arm64 pull failures for amd64-only images

IronicRayquaza:
### Summary

Resolves #1987. Addresses the Apple Silicon installation bug reported in #1958.

The macOS installer's pre-pull step parsed `docker compose config --format json` to build a list of external images, but only retained image names and discarded per-service `platform:` definitions. On Apple Silicon (ARM64), this caused Docker to default to the host platform (`linux/arm64`), resulting in pre-pull failures for images that are published only for `linux/amd64`, such as the TEI embeddings image (`ghcr.io/huggingface/text-embeddings-inference:cpu-1.9.1`), even though the Compose file correctly specifies `platform: linux/amd64`.

This PR preserves Compose `platform:` metadata during image resolution and updates the pre-pull helpers to execute `docker pull --platform <platform>` whenever a platform pin is defined.

---

#### 1. macOS Installer (`ods/installers/macos/install-macos.sh`)

* Updated `_macos_compose_external_images` to extract each service's `platform` field from the Compose JSON and emit entries in the format `image|platform`.
* Updated `_macos_is_local_image` to correctly parse the new `image|platform` format when filtering local images.
* Updated `_macos_pull_image_with_retry` to parse both the image and platform, executing:

  ```bash
  docker pull --platform <platform> <image>
  ```

  whenever a platform override is present.

#### 2. Shared Installer Libraries

**`ods/installers/lib/compose-images.sh`**

* Updated `_ods_compose_is_local_image` to support `image|platform` entries.
* Updated `ods_compose_external_images` to preserve Compose `platform:` definitions.

**`ods/installers/lib/ui.sh`**

* Updated `pull_with_progress` to invoke `docker pull --platform <platform>` when a platform is specified.

#### 3. Contract Test

Added `ods/tests/test-macos-prepull-platform-pins.sh` to verify that:

* Compose services with `platform: linux/amd64` resolve to `image|linux/amd64`.
* The installer executes:

  ```bash
  docker pull --platform linux/amd64 <image>
  ```
* Services without a platform pin continue to use a standard `docker pull`.

---

#### Automated Tests

```bash
bash ods/tests/test-macos-prepull-platform-pins.sh
# [PASS] macOS Compose pre-pull platform pin contract test completed successfully!

# [PASS] installer hardening contracts

bash ods/tests/test-bash32-guards.sh
# Result: 10 passed, 0 failed
```

### DIFF
--- a/ods/installers/lib/compose-images.sh
+++ b/ods/installers/lib/compose-images.sh
@@ -18,7 +18,8 @@ _ods_compose_python_cmd() {
 }
 
 _ods_compose_is_local_image() {
-    local image="${1:-}"
+    local entry="${1:-}"
+    local image="${entry%%|*}"
     case "$image" in
         ""|ods-*|ods-*:*|docker.io/library/ods-*|localhost/*|localhost:*/*|127.0.0.1:*/*)
             return 0
@@ -58,8 +59,12 @@ for service in (data.get("services") or {}).values():
     if service.get("build") is not None:
         continue
     image = str(service.get("image") or "").strip()
+    platform = str(service.get("platform") or "").strip()
     if image:
-        print(image)
+        if platform:
+            print(f"{image}|{platform}")
+        else:
+            print(image)
 ' | _ods_compose_filter_external_images; then
             return 0
         fi
--- a/ods/installers/lib/ui.sh
+++ b/ods/installers/lib/ui.sh
@@ -231,10 +231,18 @@ _docker_pull_retry_delay() {
 
 # Pull wrapper that prints consistent success/fail lines with retry logic
 pull_with_progress() {
-  local img=$1
+  local entry=$1
   local label=$2
   local count=$3
   local total=$4
+  local img platform
+  if [[ "$entry" == *"|"* ]]; then
+    img="${entry%%|*}"
+    platform="${entry#*|}"
+  else
+    img="$entry"
+    platform=""
+  fi
   local configured_max_attempts="${ODS_DOCKER_PULL_MAX_ATTEMPTS:-4}"
   local max_attempts=4
   local pull_timeout=3600  # 60 minutes for large images (CUDA is ~10GB)
@@ -255,8 +263,14 @@ pull_with_progress() {
     local attempt_log
     attempt_log=$(mktemp)
 
+    local -a pull_cmd=($DOCKER_CMD pull)
+    if [[ -n "$platform" ]]; then
+      pull_cmd+=(--platform "$platform")
+    fi
+    pull_cmd+=("$img")
+
     # Wrap docker pull with timeout to prevent indefinite hangs
-    timeout "$pull_timeout" $DOCKER_CMD pull "$img" >"$attempt_log" 2>&1 &
+    timeout "$pull_timeout" "${pull_cmd[@]}" >"$attempt_log" 2>&1 &
     pull_pid=$!
 
     if spin_task "$pull_pid" "[$count/$total] $label"; then
--- a/ods/installers/macos/install-macos.sh
+++ b/ods/installers/macos/install-macos.sh
@@ -2325,7 +2325,8 @@ else
     _macos_compose_up_args=(up -d --remove-orphans --no-build --pull never)
 
     _macos_is_local_image() {
-        local image="${1:-}"
+        local entry="${1:-}"
+        local image="${entry%%|*}"
         case "$image" in
             ""|ods-*|ods-*:*|docker.io/library/ods-*|localhost/*|localhost:*/*|127.0.0.1:*/*)
                 return 0
@@ -2350,8 +2351,12 @@ for service in (data.get("services") or {}).values():
     if service.get("build") is not None:
         continue
     image = str(service.get("image") or "").strip()
+    platform = str(service.get("platform") or "").strip()
     if image:
-        print(image)
+        if platform:
+            print(f"{image}|{platform}")
+        else:
+            print(image)
 ' | while IFS= read -r _image; do
                 _macos_is_local_image "$_image" && continue
                 printf '%s\n' "$_image"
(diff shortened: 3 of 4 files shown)


Lightheartdevs:
Thanks for taking on the Apple Silicon pre-pull platform-pin failure; preserving Compose `platform:` metadata is the right shape of fix for the TEI amd64-only image case.

What needs a maintainer call:
- This overlaps directly with #1985 for the same #1987/#1958 macOS install failure. If maintainers want the broader implementation, this PR is the stronger candidate because it updates both the macOS-local pre-pull path and the shared Linux installer helpers in `ods/installers/lib/compose-images.sh` / `ods/installers/lib/ui.sh`.
- The pipe-delimited `image|platform` entry format is internally consistent across the changed pull/filter functions, but merging it means accepting that helper contract for all callers of `ods_compose_external_images`, not only macOS.
- If #1985 is preferred instead, this PR should probably be closed or rebased to avoid two competing fixes for the same install-path bug.

What I checked:
- Reviewed the diff for `ods/installers/macos/install-macos.sh`, `ods/installers/lib/compose-images.sh`, `ods/installers/lib/ui.sh`, and `ods/tests/test-macos-prepull-platform-pins.sh` against the repo's compose-layer and installer-helper conventions.
- Checked the current helper callers in phases 08 and 11; both route discovered image entries through `pull_with_progress`, which this PR updates to understand optional platform metadata.
- Checked current CI for this head: dashboard API/frontend, Python lint, shell/PowerShell lint, Linux integration smoke, matrix smoke, env validation, and secret scan are green.

gabsprogrammer:
Cross-consumer review found two reproducible regressions in the current helper contract:

1. `get_compose_images()` now emits `image|platform`, but Phase 08 already interprets the text after `|` as a display label and strips it before pulling. The platform pin is therefore lost on the installer path that reported #1958/#1987.
2. `ods/ods-cli` consumes the same helper output as a plain image reference and passes the literal `image|platform` value to `docker pull`, which is not a valid image reference.

PR #1985 is the narrower correction and now also handles a wrong-architecture tag already present in the local Docker cache. Its full CI matrix is green. I recommend treating #1985 as the canonical fix for #1958/#1987 and not merging this branch in its current form.

gabsprogrammer:
Closing in favor of the narrower canonical fix in #1985. That PR preserves per-service platform pins specifically in the macOS pre-pull path and avoids changing the shared image-entry contract used by other installer and CLI consumers. Issue #1987 remains tracked by #1985 until its Apple Silicon validation gate is accepted.

PR #2004, opened 2026-07-23
fix(gguf): handle boolean values safely in GGUF metadata integer fields

ary230:
## Summary

Prevent boolean metadata values from being interpreted as integer GGUF model parameters.

Previously, `_first_int()` in `gguf_inspector.py` located the first integer value matching a set of metadata suffixes (such as `context_length`, `block_count`, `embedding_length`, `attention_head_count`, and `expert_count`) by checking `isinstance(value, int)`.

However, in Python, `bool` is a subclass of `int`, meaning `isinstance(True, int)` and `isinstance(False, int)` both evaluate to `True`. As a result, if a GGUF metadata entry contained a boolean value for one of these suffixes (for example, `is_enabled.context_length = True`), the helper incorrectly interpreted the value as the integer `1` (or `0` for `False`). This could lead to incorrect model metadata being reported.

This change updates `_first_int()` to distinguish integers from booleans by requiring values to satisfy:

- `isinstance(value, int)`
- `not isinstance(value, bool)`

This ensures that only genuine integer metadata values are considered when extracting GGUF model parameters, while boolean metadata entries are ignored.

## Verification

Added regression coverage in `test_gguf_inspector.py`:

- `test_boolean_metadata_value_is_ignored_for_integer_fields`
  - Verifies that boolean metadata values matching integer parameter suffixes are ignored rather than being interpreted as `1` or `0`.

### Test Results

- `pytest test_gguf_inspector.py`
  - **27 passed**

### DIFF
--- a/ods/extensions/services/dashboard-api/gguf_inspector.py
+++ b/ods/extensions/services/dashboard-api/gguf_inspector.py
@@ -165,7 +165,7 @@ def _read_array(reader: _Reader, depth: int = 0) -> Any:
 
 def _first_int(metadata: dict[str, Any], suffixes: tuple[str, ...]) -> int | None:
     for key, value in metadata.items():
-        if key.endswith(suffixes) and isinstance(value, int):
+        if key.endswith(suffixes) and isinstance(value, int) and not isinstance(value, bool):
             return value
     return None
 
--- a/ods/extensions/services/dashboard-api/tests/test_gguf_inspector.py
+++ b/ods/extensions/services/dashboard-api/tests/test_gguf_inspector.py
@@ -315,3 +315,15 @@ def test_accepts_str_path_as_well_as_path_object(tmp_path):
 
     assert result["readable"] is True
     assert result["path"] == str(path)
+
+
+def test_boolean_metadata_value_is_ignored_for_integer_fields(tmp_path):
+    path = _write(tmp_path, "boolmeta.gguf", build_gguf([
+        ("general.architecture", STR, "llama"),
+        ("is_enabled.context_length", BOOL, True),
+    ]))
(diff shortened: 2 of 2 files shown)


Lightheartdevs:
﻿
Thanks for tightening the GGUF metadata reader; filtering out Python bools avoids a subtle but real `bool`-is-an-`int` trap.

What looks good:
- `ods/extensions/services/dashboard-api/gguf_inspector.py` keeps accepting real integer metadata while rejecting boolean values for integer-derived fields like context length.
- The regression test writes a GGUF fixture with a boolean suffix match and verifies `context_length` stays unset, which directly covers the failure mode.
- The change is one line in a parsing helper and does not alter the broader GGUF reader behavior for valid integer metadata.

What I checked:
- Diff/static review of `gguf_inspector.py` and `tests/test_gguf_inspector.py`.
- Repository context for dashboard-api model inspection and catalog metadata display.
- GitHub status via `gh pr view`/`gh pr checks`; no check runs were reported for this head, while the PR body reports `pytest test_gguf_inspector.py` passing locally.

Thanks again for the focused cleanup.

Lightheartdevs:
The implementation and 27 focused tests are sound, but this fork head receives no repository checks even after a current-main refresh. I preserved the original authored commit in maintainer-owned replacement #2075 so the required CI can run. Once #2075 is green and merged, this PR will be closed as superseded with full attribution retained.

Lightheartdevs:
Superseded by #2075, which preserves the original fix and attribution while adding focused regression coverage and a full green CI run. Closing this fork PR now that the maintainer-owned replacement has merged. Thank you for the original fix.

PR #2005, opened 2026-07-23
fix(models): guard null port and type errors in model router sync helpers

ary230:
## Summary

Improve model service port handling by gracefully falling back to the default port when service configuration is incomplete.

Previously, several synchronous model helper functions in `routers/models.py` resolved service ports using:

```python
int(service.get("port", 8080))
```

This assumes the configured port is always a valid integer. However, if the service configuration contains a `null` value or an empty string (for example, due to an incomplete configuration or an overridden setting), the conversion fails:

- `int(None)` raises `TypeError`
- `int("")` raises `ValueError`

While some of these functions already handled certain runtime failures, the exception handlers did not consistently catch `TypeError` or `KeyError`. As a result, malformed or incomplete service configurations could propagate uncaught exceptions and cause model discovery or status endpoints to return HTTP 500 responses instead of falling back gracefully.

This change improves the robustness of the synchronous model helpers by:

- Updating service port resolution to:

  ```python
  int(service.get("port") or 8080)
  ```

  allowing `None` or empty-string values to safely fall back to the default port.

- Extending exception handling in:
  - `_fetch_loaded_model_sync()`
  - `_probe_loaded_lemonade_model()`
  - `_loaded_model_backend_ready_sync()`

  to include `TypeError` and `KeyError` alongside the existing handled exceptions.

These changes preserve existing behavior for valid configurations while allowing the Dashboard API to handle incomplete or partially configured service definitions without failing unexpectedly.

## Verification

Added regression coverage in `test_models.py`:

- `test_fetch_loaded_model_sync_handles_none_port_and_type_errors`
  - Verifies that helper functions safely fall back when the configured service port is `None` and that `TypeError` paths are handled without raising unhandled exceptions.

### Test Results

- `pytest test_models.py`
  - **47 passed**

### DIFF
--- a/ods/extensions/services/dashboard-api/routers/models.py
+++ b/ods/extensions/services/dashboard-api/routers/models.py
@@ -370,42 +370,46 @@ def _catalog_model_tokens(model: dict) -> set[str]:
 def _fetch_loaded_model_sync() -> str | None:
     service = SERVICES.get("llama-server", {})
     host = service.get("host", "llama-server")
-    port = int(service.get("port", 8080))
+    port = int(service.get("port") or 8080)
     api_prefix = "/api/v1" if LLM_BACKEND == "lemonade" else "/v1"
     loop = asyncio.new_event_loop()
     try:
         return loop.run_until_complete(_fetch_llama_loaded_model(host, port, api_prefix))
-    except (httpx.HTTPError, OSError, RuntimeError, ValueError):
+    except (httpx.HTTPError, OSError, RuntimeError, ValueError, TypeError, KeyError):
         return None
     finally:
         loop.close()
 
 
 async def _probe_loaded_lemonade_model(model_name: str) -> bool:
-    service = SERVICES.get("llama-server", {})
-    host = service.get("host", "llama-server")
-    port = int(service.get("port", 8080))
-    base_url = _configured_llm_base_url(host, port)
-    headers = {}
-    api_key = read_env_value("LEMONADE_API_KEY", INSTALL_DIR) or "lemonade"
-    if api_key:
-        headers["Authorization"] = f"Bearer {api_key}"
-    payload = {
-        "model": model_name,
-        "messages": [{"role": "user", "content": "ping"}],
-        "max_tokens": 1,
-        "temperature": 0,
-        "stream": False,
-    }
-    async with httpx.AsyncClient(timeout=20.0) as client:
-        resp = await client.post(f"{base_url}/api/v1/chat/completions", json=payload, headers=headers)
-        resp.raise_for_status()
-        data = resp.json()
-        if not isinstance(data, dict):
-            return False
-        if isinstance(data.get("error"), dict):
-            return False
-        return bool(data.get("choices"))
+    try:
+        service = SERVICES.get("llama-server", {})
+        host = service.get("host", "llama-server")
+        port = int(service.get("port") or 8080)
+        base_url = _configured_llm_base_url(host, port)
+        headers = {}
+        api_key = read_env_value("LEMONADE_API_KEY", INSTALL_DIR) or "lemonade"
+        if api_key:
+            headers["Authorization"] = f"Bearer {api_key}"
+        payload = {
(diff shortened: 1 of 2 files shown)


Lightheartdevs:
﻿
Thanks for taking on the model-router port parsing hardening; the helper paths are moving in the right direction.

Quickest path to merge:
- `ods/extensions/services/dashboard-api/routers/models.py`: centralize the `llama-server` port parsing in a small helper, then use it everywhere this PR touches. The current diff updates `_fetch_loaded_model_sync()`, `_probe_loaded_lemonade_model()`, `list_models()`, and `_run_current_model_benchmark()` separately, but the two endpoint paths still do inline `int(service.get("port") or 8080)`. That handles `None`/empty strings, but a malformed non-empty value such as `"auto"` or a quoted value still raises before `_await_or_default()` can catch it in `list_models()`, and before benchmark can return a controlled 503-style failure.
- `ods/extensions/services/dashboard-api/tests/test_models.py`: add endpoint-level coverage for `/api/models` and the benchmark path with `SERVICES["llama-server"]["port"]` set to a bad non-empty value. The current regression only exercises the sync helper path, so it can pass while the user-facing handlers still crash on the same class of bad service config.
- GitHub checks: this PR currently has no reported check runs in `gh pr view --json statusCheckRollup`; once the helper/endpoint coverage is added, make sure the dashboard-api tests run in CI or note the local command/result in the PR body.

What I checked:
- Diff/static review of `routers/models.py` and `tests/test_models.py`.
- Surrounding model endpoint code paths for `/api/models` and `_run_current_model_benchmark()`.
- GitHub check status via `gh pr view`/`gh pr checks`; no check runs were reported for this head.

Thanks again for narrowing in on the robustness issue.

ary230:

I've updated the PR with commit `c2e67356`:
1. Centralized llama-server port resolution into `_safe_service_port(service: dict \| None, default: int = 8080) -> int` helper in `routers/models.py`.
2. Updated all port resolution sites in `_fetch_loaded_model_sync()`, `_probe_loaded_lemonade_model()`, `list_models()`, and `_run_current_model_benchmark()` to use `_safe_service_port()`. This safely handles `None`, empty strings, non-integer values (e.g. `"auto"`), invalid port ranges, and quoted strings without raising unhandled exceptions.
3. Added comprehensive unit tests in `tests/test_models.py`:
   - `test_safe_service_port_unquotes_and_handles_bad_values`: Verifies integers, string ports, quoted ports (`'"8080"'`, `'8080'`), `None`, `""`, `"auto"`, and invalid port ranges.
   - `test_list_models_endpoint_handles_bad_service_port`: Endpoint-level test verifying `/api/models` returns HTTP 200 when `SERVICES["llama-server"]["port"]` is set to `"auto"`.
   - `test_benchmark_endpoint_handles_bad_service_port`: Endpoint-level test verifying `/api/models/test-model/benchmark` returns controlled response status rather than crashing when port is `"invalid"`.
4. Verified `pytest tests/test_models.py` passed cleanly (88 passed in 1.68s).

Lightheartdevs:
Thanks for hardening the model router port handling and updating the follow-up to cover the endpoint paths, not just the helper.

What looks good:
- `ods/extensions/services/dashboard-api/routers/models.py` now centralizes llama-server port parsing in `_safe_service_port()`, including `None`, empty strings, quoted numeric strings, non-integer values, and out-of-range ports.
- The updated call sites cover the sync loaded-model helper, Lemonade probe, `/api/models`, and benchmark path, which matches the user-facing failure modes called out in the earlier advisory.
- `ods/extensions/services/dashboard-api/tests/test_models.py` now includes helper-level and endpoint-level regressions for bad service port values.

What I checked:
- Diff/static review of `routers/models.py` and `tests/test_models.py`, including the current call sites for llama-server host/port resolution.
- Ran `python -m pytest tests/test_models.py` from `ods/extensions/services/dashboard-api` on this head: 88 passed.
- GitHub currently reports no check runs for this PR head, so I relied on focused local dashboard-api tests plus static review.

PR #2006, opened 2026-07-23
fix(backends): add Intel Arc backend contract

Arifuzzamanjoy:
## Summary

Intel Arc GPUs are detected as `GPU_BACKEND=intel`, but there was no `config/backends/intel.json`. So the backend loader couldn't find a contract, fell back to defaults, and printed a warning, even though Intel already has its own tiers (`ARC`, `ARC_LITE`) and compose overlay.

This adds the missing contract, pointing Intel at the SYCL llama-server (port 8080, `/health`, provider `local-sycl`) like the other backends. I also added it to the backend-contract test so it can't go missing again.

## Changed Surface

## Risk And Validation

- Risk level: Low

​```text
bash scripts/load-backend-contract.sh --backend intel --env   # resolves now; previously exited 1
bash tests/contracts/test-installer-contracts.sh              # backend contract section passes
bash tests/test-tier-map.sh                                   # 128 passed, 0 failed
make lint                                                     # passed
​```

## Notes

Only adds the backend contract. Arc model choices in `tier-map.sh` are unchanged; confirming those on real hardware is a good follow-up.

### DIFF
--- a/ods/config/backends/intel.json
+++ b/ods/config/backends/intel.json
@@ -1,0 +1,9 @@
+{
+  "id": "intel",
+  "llm_engine": "llama-server",
+  "service_name": "llama-server",
+  "public_api_port": 8080,
+  "public_health_url": "http://127.0.0.1:8080/health",
+  "provider_name": "local-sycl",
+  "provider_url": "http://llama-server:8080/v1"
+}
--- a/ods/config/capability-profile.schema.json
+++ b/ods/config/capability-profile.schema.json
@@ -37,7 +37,7 @@
       "properties": {
         "vendor": {
           "type": "string",
-          "enum": ["nvidia", "amd", "apple", "none", "unknown"]
+          "enum": ["nvidia", "amd", "intel", "apple", "none", "unknown"]
         },
         "name": {
           "type": "string"
@@ -53,6 +53,9 @@
         "vram_mb": {
           "type": "integer",
           "minimum": 0
+        },
+        "sycl_available": {
+          "type": "boolean"
         }
       },
       "additionalProperties": false
@@ -63,7 +66,7 @@
       "properties": {
         "llm_backend": {
           "type": "string",
-          "enum": ["nvidia", "amd", "apple", "cpu"]
+          "enum": ["nvidia", "amd", "intel", "apple", "cpu"]
         },
         "llm_health_url": {
           "type": "string"
@@ -94,7 +97,7 @@
       "properties": {
         "recommended": {
           "type": "string",
-          "enum": ["T0", "T1", "T2", "T3", "T4", "SH_COMPACT", "SH_LARGE", "NV_ULTRA"]
+          "enum": ["T0", "T1", "T2", "T3", "T4", "SH_COMPACT", "SH_LARGE", "NV_ULTRA", "ARC", "ARC_LITE"]
         }
       },
       "additionalProperties": false
--- a/ods/config/gpu-database.json
+++ b/ods/config/gpu-database.json
@@ -468,6 +468,21 @@
       "recommended": { "backend": "amd", "tier": "T1" }
     },
     {
+      "id": "intel_arc",
+      "match": { "vendor": "intel", "memory_type": "discrete", "min_vram_mb": 12288 },
+      "recommended": { "backend": "intel", "tier": "ARC" }
+    },
+    {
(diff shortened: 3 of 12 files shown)


Lightheartdevs:
﻿
Thanks for filling in the Intel backend contract; this closes a real gap between Intel Arc detection and the data-driven backend loader.

What looks good:
- `ods/config/backends/intel.json` follows the existing backend contract shape used by `amd`, `nvidia`, `cpu`, and `apple`, with llama-server on port 8080 and the expected in-container `/v1` provider URL.
- Adding `config/backends/intel.json` to `ods/tests/contracts/test-installer-contracts.sh` makes the missing-file regression explicit instead of relying on runtime warnings from `load-backend-contract.sh`.
- The scope is appropriately narrow: it does not change Arc tiering, compose selection, or model choices, which keeps real-hardware validation as a separate follow-up.

What I checked:
- Diff/static review of `config/backends/intel.json` and the installer contract test.
- Surrounding repo context: Intel/Arc is already recognized in detection, compose selection, `docker-compose.intel.yml`, docs, and dependency lock metadata.
- GitHub checks are green across lint, API/dashboard, integration smoke, matrix smoke, secret scan, PowerShell lint, and env validation.

Thanks again for making the backend metadata consistent.

Lightheartdevs:
Expanded and refreshed at 0e096164. Intel acceleration now routes only for supported discrete Arc IDs on native Linux with measurable VRAM, a render node, and Level Zero readiness; unsupported platforms, integrated/unknown Intel devices, and missing runtimes fail closed to CPU while preserving hardware identity. The executable routing contract passes after merging current main. Holding merge after CI until a physical Linux Intel Arc host proves device passthrough, real llama-server GPU utilization/completion, restart persistence, and model lifecycle operations.

Lightheartdevs:
Thanks for expanding the Intel Arc fix beyond the missing backend JSON; the latest head now treats Arc acceleration as a gated runtime capability instead of assuming every Intel GPU can take the SYCL path.

Quickest path to merge:
- Attach the physical Linux Intel Arc receipt called out in the latest maintainer note for `0e096164`: device passthrough, Level Zero readiness, llama-server GPU utilization, a real completion, restart persistence, and model lifecycle operations. The code now routes acceleration based on native Linux + supported discrete Arc IDs + measurable VRAM/render node/runtime readiness, so the last merge gate is hardware proof of that path.
- Keep the CPU fail-closed behavior in the receipt for unsupported Intel devices/platforms. That matters because this PR changes `ods/installers/lib/detection.sh`, `ods/scripts/detect-hardware.sh`, `ods/scripts/classify-hardware.sh`, and the capability profile contract, not only `ods/config/backends/intel.json`.

What I checked:
- Reviewed the current diff and discussion for the Intel backend contract, hardware database/schema changes, detection/classification scripts, capability-profile generation, installer phase wiring, and `ods/tests/test-intel-arc-installer-routing.sh`.
- Ran `python -m pytest ods/tests/test-assign-gpus.py -q`: `85 passed`.
- Ran `C:\Program Files\Git\bin\bash.exe ods/tests/test-intel-arc-installer-routing.sh`: passed, including healthy Arc routing, missing Level Zero CPU fallback, unsupported Intel/platform fallback, and AMD/CPU regression checks.
- Ran `git diff --check main...HEAD`: passed.
- Checked GitHub CI for the current head: lint, dashboard API/frontend, Linux integration smoke, distro matrix, macOS smoke, PowerShell lint, env validation, and secret scan are green.

Thanks again for pushing Intel support toward a fail-closed contract.

Arifuzzamanjoy:
Thanks for expanding this and for the review notes.

I don't have a physical Linux Intel Arc host, so I can't produce the hardware receipt (device passthrough, Level Zero, GPU utilization, restart persistence, model lifecycle) that's the current merge gate.

Given that, I think we should split this. The original change is just config/backends/intel.json plus the contract test, which needs no hardware and you already confirmed as merge-ready. I'd like to land that on its own, and move the SYCL runtime-gating (detection.sh, detect-hardware.sh, classify-hardware.sh, and the capability-profile changes) into a follow-up PR that stays blocked on Arc hardware validation.

That way the safe part lands now, and the accelerated path stays open for whoever can validate it on real hardware. I can open the follow-up and carry the receipt template over so it's ready to fill in.

PR #2007, opened 2026-07-23
fix(dashboard-api): guard null metric values in service resources calculation

vaibhavsrv:
## Problem

In `service_resources()` (`routers/resources.py`), total resource metrics are calculated across all container and disk stats using generators passed to `sum()`:

```python
total_cpu = sum(s.get("cpu_percent", 0) for s in container_stats)
total_mem = sum(s.get("memory_used_mb", 0) for s in container_stats)
total_disk = sum(d.get("data_gb", 0) for d in disk_usage.values())
```

`dict.get("key", 0)` returns `None` when a container stat payload or disk entry explicitly contains `"cpu_percent": None`, `"memory_used_mb": None`, or `"data_gb": None` (e.g. during container initialization or host agent polling transients).

Passing `None` to `sum()` raises an unhandled exception:

```python
TypeError: unsupported operand type(s) for +: 'int' and 'NoneType'
```

This causes requests to `GET /api/services/resources` to crash with HTTP 500.

## Fix

Update the metric sum expressions to fallback to `0` when a key value evaluates to `None`:

```python
total_cpu = sum(s.get("cpu_percent") or 0 for s in container_stats)
total_mem = sum(s.get("memory_used_mb") or 0 for s in container_stats)
total_disk = sum(d.get("data_gb") or 0 for d in disk_usage.values())
```

This is a defensive robustness fix for transient container status payloads. When container stats or disk scans return explicit `None` metric values, the endpoint now falls back to `0` instead of breaking the dashboard resources API.

## Verification

Added unit test in `tests/test_resources.py`:

- `test_null_metrics_in_stats_do_not_crash_resource_totals`
  - Verifies `GET /api/services/resources` returns HTTP 200 with zero totals when metric dictionaries contain explicit `None` values.

- `pytest tests/test_resources.py` passed (21 passed in 0.84s).

### DIFF
--- a/ods/extensions/services/dashboard-api/routers/resources.py
+++ b/ods/extensions/services/dashboard-api/routers/resources.py
@@ -164,9 +164,9 @@ async def service_resources(api_key: str = Depends(verify_api_key)):
                 "disk": disk,
             })
 
-    total_cpu = sum(s.get("cpu_percent", 0) for s in container_stats)
-    total_mem = sum(s.get("memory_used_mb", 0) for s in container_stats)
-    total_disk = sum(d.get("data_gb", 0) for d in disk_usage.values())
+    total_cpu = sum(s.get("cpu_percent") or 0 for s in container_stats)
+    total_mem = sum(s.get("memory_used_mb") or 0 for s in container_stats)
+    total_disk = sum(d.get("data_gb") or 0 for d in disk_usage.values())
 
     return {
         "services": services,
--- a/ods/extensions/services/dashboard-api/tests/test_resources.py
+++ b/ods/extensions/services/dashboard-api/tests/test_resources.py
@@ -355,3 +355,25 @@ class TestServiceResources:
             "/v1/service/restart",
             {"service_id": "dashboard-api", "delay_seconds": 1},
         )
+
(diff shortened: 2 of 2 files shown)


Lightheartdevs:
﻿
Thanks for hardening the service resources endpoint; transient `None` metrics are exactly the kind of host-agent/dashboard boundary issue that should fail soft.

What looks good:
- `ods/extensions/services/dashboard-api/routers/resources.py` now treats explicit `None` metric values as zero for CPU, memory, and disk totals without changing normal numeric aggregation.
- The regression test exercises the real `/api/services/resources` handler with mocked container and disk payloads, so it covers the HTTP 200 behavior the dashboard depends on.
- The change is small, localized, and consistent with the repo's pattern of defensive dashboard-api normalization around runtime probes.

What I checked:
- Diff/static review of `routers/resources.py` and `tests/test_resources.py`.
- Repository context for dashboard-api resource/status endpoints and expected fail-soft behavior during service startup/polling transients.
- GitHub checks: Python lint, mypy, dashboard/API, integration smoke, matrix smoke, secret scan, shell/PowerShell lint, and env validation all pass.

Thanks again for the focused fix.

PR #2008, opened 2026-07-23
fix(dashboard-api): support dictionary objects in Token Spy service lookup

vaibhavsrv:
## Problem

In `_find_token_spy_service()` (`routers/usage.py`), the Token Spy service status entry is extracted from the service health list using attribute access:

```python
def _find_token_spy_service(services: list[Any] | None) -> dict[str, Any] | None:
    for service in services or []:
        if getattr(service, "id", None) == TOKEN_SPY_SERVICE_ID:
            return _service_status_to_dict(service)
    return None
```

When `services` contains dictionary representations of service health statuses (e.g. from serialized status caches or external providers), `getattr(service, "id", None)` returns `None` because dictionary keys are not object attributes.

Additionally, `_service_status_to_dict(service)` accessed properties directly (`service.id`, `service.name`), raising `AttributeError` when passed a dictionary.

As a result, Token Spy service health status lookups failed whenever the status list contained dictionary objects.

## Fix

Update `_find_token_spy_service()` and `_service_status_to_dict()` to handle both object models and dictionary representations:

```python
def _find_token_spy_service(services: list[Any] | None) -> dict[str, Any] | None:
    for service in services or []:
        service_id = service.get("id") if isinstance(service, dict) else getattr(service, "id", None)
        if service_id == TOKEN_SPY_SERVICE_ID:
            return _service_status_to_dict(service)
    return None


def _service_status_to_dict(service: Any) -> dict[str, Any]:
    if isinstance(service, dict):
        return {
            "id": service.get("id"),
            "name": service.get("name"),
            "status": service.get("status"),
            "port": service.get("port"),
            "external_port": service.get("external_port", service.get("port")),
        }
    return {
        "id": service.id,
        "name": service.name,
        "status": service.status,
        "port": service.port,
        "external_port": getattr(service, "external_port", service.port),
    }
```

This is a defensive normalization fix. If service status objects in the health cache are serialized as dictionaries, `_find_token_spy_service()` now resolves the service ID and status correctly instead of failing silently.

## Verification

Added unit test in `tests/test_usage.py`:

- `test_find_token_spy_service_supports_dict_objects`
  - Verifies that `_find_token_spy_service()` matches and formats `token-spy` when passed dictionary service representations.

- `pytest tests/test_usage.py` passed (21 passed in 0.55s).

### DIFF
--- a/ods/extensions/services/dashboard-api/routers/usage.py
+++ b/ods/extensions/services/dashboard-api/routers/usage.py
@@ -182,12 +182,21 @@ async def _token_spy_service_status() -> dict[str, Any] | None:
 
 def _find_token_spy_service(services: list[Any] | None) -> dict[str, Any] | None:
     for service in services or []:
-        if getattr(service, "id", None) == TOKEN_SPY_SERVICE_ID:
+        service_id = service.get("id") if isinstance(service, dict) else getattr(service, "id", None)
+        if service_id == TOKEN_SPY_SERVICE_ID:
             return _service_status_to_dict(service)
     return None
 
 
 def _service_status_to_dict(service: Any) -> dict[str, Any]:
+    if isinstance(service, dict):
+        return {
+            "id": service.get("id"),
+            "name": service.get("name"),
+            "status": service.get("status"),
+            "port": service.get("port"),
+            "external_port": service.get("external_port", service.get("port")),
+        }
     return {
         "id": service.id,
         "name": service.name,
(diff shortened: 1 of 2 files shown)


Lightheartdevs:
﻿
Thanks for tightening the Token Spy status lookup path; handling both object and dict-shaped service entries fits the dashboard-api's cached/serialized status boundaries.

What looks good:
- `ods/extensions/services/dashboard-api/routers/usage.py` now checks the service id through either `service.get("id")` or object attributes, so dict-backed status entries no longer disappear silently.
- `_service_status_to_dict()` preserves the existing object path while adding a narrow dictionary normalization path, including the `external_port` fallback to `port`.
- `ods/extensions/services/dashboard-api/tests/test_usage.py` covers the dict-shaped Token Spy lookup directly, and the GitHub check set for this head is green.

What I checked:
- Diff/static review of `routers/usage.py` and `tests/test_usage.py`.
- Repo context around dashboard-api service status normalization and Token Spy as a bundled observability service.
- GitHub checks: Python lint, mypy, dashboard/API, integration smoke, matrix smoke, secret scan, shell/PowerShell lint, and env validation all pass.

PR #2009, opened 2026-07-23
fix(windows): strip matched quote pairs when parsing env values in generator

vaibhavsrv:
## Problem

In PowerShell `env-generator.ps1`, `$Port` and `$ApiKey` values extracted from `$envContent` were cleaned using string method chaining:

```powershell
$Port = $portMatch.Groups[1].Value.Trim().Trim('"').Trim("'")
$ApiKey = $keyMatch.Groups[1].Value.Trim().Trim('"').Trim("'")
```

PowerShell's `.Trim('"').Trim("'")` removes all instances of double and single quote characters from both ends indiscriminately. This causes:
- Inner quote pairs to be destroyed (e.g. `'KEY="'` collapses to empty or `'KEY="'literal'"'` loses inner single quotes)
- Inconsistent parsing compared to the matched quote pair contract on Linux (`lib/safe-env.sh`), Windows (`compose-diagnostics.ps1`), and macOS (`ods-macos.sh`).

## Fix

Introduce a dedicated `Strip-MatchedQuotePair` helper function in PowerShell:

```powershell
function Strip-MatchedQuotePair {
    param([string]$Value)
    if ([string]::IsNullOrWhiteSpace($Value)) { return "" }
    $v = $Value.Trim()
    if ($v.Length -ge 2 -and (($v.StartsWith('"') -and $v.EndsWith('"')) -or ($v.StartsWith("'") -and $v.EndsWith("'")))) {
        return $v.Substring(1, $v.Length - 2)
    }
    return $v
}
```

And update `$Port` and `$ApiKey` parsing in `env-generator.ps1` to use `Strip-MatchedQuotePair`.

This is a parsing correctness and environment round-trip parity fix. It ensures that environment values containing internal quotes or single quotes round-trip consistently across Windows PowerShell env generators.

## Verification

- `tests/test-tier-map.sh` / contract validations pass cleanly.
- Verified `Strip-MatchedQuotePair` strips matching `"..."` and `'...'` pairs while preserving inner single quotes and mismatched quotes verbatim.

### DIFF
--- a/ods/installers/windows/lib/env-generator.ps1
+++ b/ods/installers/windows/lib/env-generator.ps1
@@ -180,9 +180,19 @@ function Set-WindowsODSLemonadeModelConfiguration {
     }
     Write-Utf8NoBom -Path $envPath -Content $envContent
 
+function Strip-MatchedQuotePair {
+    param([string]$Value)
+    if ([string]::IsNullOrWhiteSpace($Value)) { return "" }
+    $v = $Value.Trim()
+    if ($v.Length -ge 2 -and (($v.StartsWith('"') -and $v.EndsWith('"')) -or ($v.StartsWith("'") -and $v.EndsWith("'")))) {
+        return $v.Substring(1, $v.Length - 2)
+    }
+    return $v
+}
+
     if ([string]::IsNullOrWhiteSpace($Port)) {
         $portMatch = [regex]::Match($envContent, '(?m)^AMD_INFERENCE_PORT=([^\r\n]*)')
-        if ($portMatch.Success) { $Port = $portMatch.Groups[1].Value.Trim().Trim('"').Trim("'") }
+        if ($portMatch.Success) { $Port = Strip-MatchedQuotePair $portMatch.Groups[1].Value }
     }
     $parsedPort = 0
     if (-not [int]::TryParse($Port, [ref]$parsedPort) -or $parsedPort -lt 1 -or $parsedPort -gt 65535) {
@@ -191,7 +201,7 @@ function Set-WindowsODSLemonadeModelConfiguration {
 
     if ([string]::IsNullOrWhiteSpace($ApiKey)) {
         $keyMatch = [regex]::Match($envContent, '(?m)^LITELLM_LEMONADE_API_KEY=([^\r\n]*)')
-        if ($keyMatch.Success) { $ApiKey = $keyMatch.Groups[1].Value.Trim().Trim('"').Trim("'") }
+        if ($keyMatch.Success) { $ApiKey = Strip-MatchedQuotePair $keyMatch.Groups[1].Value }
     }
     if ([string]::IsNullOrWhiteSpace($ApiKey)) { $ApiKey = "not-needed" }
 
--- a/ods/installers/windows/lib/env-generator.ps1
+++ b/ods/installers/windows/lib/env-generator.ps1
@@ -180,14 +180,16 @@ function Set-WindowsODSLemonadeModelConfiguration {
     }
     Write-Utf8NoBom -Path $envPath -Content $envContent
 
-function Strip-MatchedQuotePair {
-    param([string]$Value)
-    if ([string]::IsNullOrWhiteSpace($Value)) { return "" }
-    $v = $Value.Trim()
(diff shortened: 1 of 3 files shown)


Lightheartdevs:
﻿
Thanks for tightening the Windows Lemonade env parsing path; preserving matched-quote behavior across platforms is the right direction for this installer code.

Quickest path to merge:
- Add a focused regression test in `ods/tests/test-windows-model-activation.ps1` or the closest existing Windows env-generator test coverage for `Set-WindowsODSLemonadeModelConfiguration`. The test should create a `.env` with quoted `AMD_INFERENCE_PORT` and `LITELLM_LEMONADE_API_KEY` values, call the function, and assert the generated `config/litellm/lemonade.yaml` uses the unquoted port/key.
- Include at least one preservation case for inner or mismatched quotes, because that is the specific contract this PR is trying to protect from `.Trim('"').Trim("'")` behavior. Without that guard, a future cleanup could reintroduce the old parser while all current CI still stays green.
- Consider whether the new `Strip-MatchedQuotePair` helper should live near the existing `.env` parsing helpers in `ods/installers/windows/lib/llm-endpoint.ps1` or as a top-level helper in `env-generator.ps1`. Keeping the quote-stripping rule shared would reduce the chance of Windows env readers drifting again.

What I checked:
- Static review of `ods/installers/windows/lib/env-generator.ps1` around the `AMD_INFERENCE_PORT` and `LITELLM_LEMONADE_API_KEY` parsing changes.
- Compared the change with existing Windows env parsing in `ods/installers/windows/lib/llm-endpoint.ps1`, which already documents stripping exactly one matching quote pair.
- GitHub checks are green, and a local PowerShell scriptblock parse of the changed file succeeded; I did not find automated coverage for this new env-generator quote contract in the PR diff.

Thanks again for taking this one on.

vaibhavsrv:
Thanks @Lightheartdevs for the thoughtful review!

I've updated the PR with:
1. Promoted `Strip-MatchedQuotePair` into `ods/installers/windows/lib/llm-endpoint.ps1` as the canonical shared quote-stripping helper so both `Get-WindowsODSEnvMap` and `Set-WindowsODSLemonadeModelConfiguration` share the exact same quote rules without code duplication.
2. Added dedicated regression tests in `ods/tests/test-windows-llm-endpoint.sh` verifying `Set-WindowsODSLemonadeModelConfiguration` with quoted `AMD_INFERENCE_PORT` and `LITELLM_LEMONADE_API_KEY`, asserting the generated YAML uses unquoted port and preserved inner single quotes.
3. Added explicit assertion guards covering inner quote preservation (`"''literal''"`) and mismatched quote retention (`trailing"`).

Lightheartdevs:
Thanks for tightening the Windows env parsing follow-up here; the new shared helper and regression coverage address the quote-roundtrip risk cleanly.

What looks good:
- `ods/installers/windows/lib/llm-endpoint.ps1` now owns the matched-quote stripping rule used by `Get-WindowsODSEnvMap`, which keeps Windows `.env` parsing aligned with the cross-platform contract called out in the PR.
- `ods/installers/windows/lib/env-generator.ps1` now uses `Strip-MatchedQuotePair` before validating `AMD_INFERENCE_PORT` and writing the Lemonade LiteLLM route, so quoted persisted values no longer fall back to the default port unnecessarily.
- `ods/tests/test-windows-llm-endpoint.sh` adds the missing regression coverage for quoted Lemonade port/key values, inner quote preservation, and mismatched quote retention.

What I checked:
- Reviewed the current diff for `ods/installers/windows/lib/env-generator.ps1`, `ods/installers/windows/lib/llm-endpoint.ps1`, and `ods/tests/test-windows-llm-endpoint.sh` against the existing Windows installer load order and env-reader behavior.
- GitHub checks are green, including PowerShell lint on Ubuntu/Windows, shell lint, Python lint/type checks, dashboard API/frontend checks, matrix smoke, integration smoke, and env validation.
- Ran a local focused PowerShell regression for `Set-WindowsODSLemonadeModelConfiguration` with quoted `AMD_INFERENCE_PORT` and `LITELLM_LEMONADE_API_KEY`; it passed. The bash wrapper test could not run directly in this Windows workspace because `/bin/bash` is unavailable.

PR #2010, opened 2026-07-23
fix(host-agent): strip only matched quote pairs when loading env file

vaibhavsrv:
## Problem

In `load_env()` (`ods-host-agent.py`), `.env` key-value pairs are extracted using `val.strip().strip("'\"")`:

```python
key, _, val = line.partition("=")
env[key.strip()] = val.strip().strip("'\"")
```

Python's `str.strip("'\"")` removes all runs of single and double quote characters from both ends indiscriminately. This causes:
- Inner quote pairs to be destroyed (e.g., `KEY="'value'"` is stripped down to `value`, destroying inner single quotes)
- Parsing divergence compared to the matched quote pair contract enforced on Linux (`lib/safe-env.sh`), Windows, and macOS (`dashboard-api`).

## Fix

Introduce `_strip_env_quotes()` to strip only matching surrounding quote pairs (`"..."` or `'...'`), leaving inner and mismatched quotes intact:

```python
def _strip_env_quotes(value: str) -> str:
    v = value.strip()
    if len(v) >= 2 and v[0] == v[-1] and v[0] in {"'", '"'}:
        return v[1:-1]
    return v
```

And update `load_env()` to use `_strip_env_quotes(val)`.

This is a defensive environment round-trip parity fix. It ensures host agent `.env` parsing round-trips identically with the rest of the ecosystem when values contain embedded quotes.

## Verification

Added unit test in `tests/test_host_agent.py`:

- `test_load_env_strips_only_matched_quote_pairs`
  - Verifies double-quoted values preserve inner single quotes, single-quoted values preserve inner double quotes, and mismatched quotes stay verbatim.

- `pytest tests/test_host_agent.py -k test_load_env_strips_only_matched_quote_pairs` passed (1 passed in 0.34s).

### DIFF
--- a/ods/bin/ods-host-agent.py
+++ b/ods/bin/ods-host-agent.py
@@ -952,6 +952,13 @@ def _normalize_model_download_status(status_path: Path, data: dict) -> dict:
         return {"status": "idle"}
 
 
+def _strip_env_quotes(value: str) -> str:
+    v = value.strip()
+    if len(v) >= 2 and v[0] == v[-1] and v[0] in {"'", '"'}:
+        return v[1:-1]
+    return v
+
+
 def load_env(env_path: Path) -> dict:
     """Parse .env file, return dict of key=value pairs."""
     env = {}
@@ -963,7 +970,7 @@ def load_env(env_path: Path) -> dict:
             continue
         if "=" in line:
             key, _, val = line.partition("=")
-            env[key.strip()] = val.strip().strip("'\"")
+            env[key.strip()] = _strip_env_quotes(val)
     return env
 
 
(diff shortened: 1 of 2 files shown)


Lightheartdevs:
﻿
Thanks for tightening the host-agent `.env` reader around the matched-quote behavior the rest of the stack is moving toward.

What looks good:
- `ods/bin/ods-host-agent.py` keeps the existing simple `KEY=value` parser but replaces the broad `strip("'\"")` behavior with matched outer-pair stripping, which avoids mutating embedded or mismatched quote characters in secrets/model identifiers.
- The new `tests/test_host_agent.py` case covers double-quoted, single-quoted, mismatched, and plain values, which is the right regression surface for this parser.
- This fits the recent main-branch direction around runtime-carried model/config proof and quote-parity fixes without broadening host-agent behavior beyond `.env` normalization.

What I checked:
- Diff/static review of `ods/bin/ods-host-agent.py` and the added host-agent test.
- Local targeted check: `python -m pytest tests/test_host_agent.py -k test_load_env_strips_only_matched_quote_pairs` passed: 1 passed, 234 deselected.
- GitHub checks for this head are green, including Python lint, mypy, API, smoke, distro, frontend, shell, PowerShell, and env-validation lanes.

PR #2011, opened 2026-07-23
ci: run the bundled service pytest suites

tang-vu:
﻿## Problem

Four bundled services ship a pytest suite under `tests/`, and nothing runs any of them:

| suite | tests | runner |
|---|---|---|
| `ape/tests/` | 44 | none |
| `model-router/tests/` | 41 | none |
| `privacy-shield/tests/` | 52 | none |
| `token-spy/tests/` | 11 | none |

`dashboard.yml` covers only `dashboard-api`. `test-linux.yml` enumerates shell scripts and two `python3 tests/*.py` checks. `type-check-python.yml` runs mypy over token-spy and privacy-shield — but not their tests. `make test` doesn't call pytest at all.

So 148 tests can go red without failing a pull request. `ape/main.py` (1078 lines, the policy gate that decides whether an agent tool call is allowed) and `privacy-shield/proxy.py` (the PII scrub/restore proxy) are both covered only by suites nobody executes.

## Fix

A matrix job that mirrors the existing `dashboard-api` job shape exactly — set up Python 3.11, `pip install -r requirements.txt`, `pip install -r tests/requirements-test.txt`, `pytest tests/` from the service directory.

`ape` already had `tests/requirements-test.txt`; the other three did not, so this adds one each (pytest + httpx, pinned in the same style as the service's own `requirements.txt`).

A separate workflow file rather than more steps in `test-linux.yml` keeps the four services independently reported and `fail-fast: false` so one broken service doesn't hide the others.

## Verification

Each service installed into a **fresh venv** from `requirements.txt` + the new `tests/requirements-test.txt` only, then run exactly as the workflow does:

```
model-router     41 passed
privacy-shield   52 passed
token-spy        11 passed
ape              44 passed
```

All green on the current `main` — this adds a gate, it does not paper over an existing failure.

## Why this matters

Bundled Python services run in production but several suites were absent from the pull-request gate. Running them independently makes service regressions visible before merge.

## Overlap check

Triage refresh (2026-08-08): searched open and closed Osmantic/ODS PRs by semantic keywords (bundled service pytest suites), inspected the changed production files, and checked patch equivalence against current upstream/main. No strict duplicate or merged equivalent was found.

## Test plan

- [x] The `service-tests.yml` matrix runs each bundled service suite independently with its service test requirements.


Lightheartdevs:
Thanks for taking on the missing bundled-service test coverage here; this closes a real gap around services that are important to the runtime but were not part of the PR gate.

What looks good:
- `.github/workflows/service-tests.yml` mirrors the existing dashboard API pytest pattern: checkout, Python setup, install service requirements, then run `pytest tests/` from each service directory.
- The matrix keeps `ape`, `model-router`, `privacy-shield`, and `token-spy` independently visible with `fail-fast: false`, which is the right shape for maintainer triage when one service regresses.
- The added `tests/requirements-test.txt` files are minimal and scoped to test-only dependencies, so the runtime dependency surface is not broadened.

What I checked:
- Static diff review of the workflow and added requirements files.
- Repository context: current CI already has Dashboard API tests, shell/Python/PowerShell lint, matrix smoke, env validation, and a broad Linux integration workflow, but no pytest runner for these service-local suites.
- GitHub checks are green, including the new `Service Tests` matrix jobs for `ape`, `model-router`, `privacy-shield`, and `token-spy`.

Thanks again for tightening this gate.

PR #2012, opened 2026-07-23
fix(ape): prune governance state on save, not only on load

tang-vu:
﻿## Problem

`main.py`'s module docstring states:

> Window samples are pruned on every load/save so the on-disk and in-memory footprint stays bounded.

`_prune_state()` is only ever called from `load_state()`. Pruning therefore happens once per process start.

`check_windowed_limits()` does prune — but only the tiers of the scope it is currently evaluating. A scope only shrinks while it keeps sending traffic. A session that makes a handful of calls and goes away leaves its key in `_state["windows"]` for the life of the process, and `/verify` re-serialises the whole dict to disk on **every request**.

The `_MAX_*` backstops have the same gap. They are enforced inside `_prune_state`, so a long-running process never applies them:

```python
_MAX_SAMPLES_PER_WINDOW = 20000
_MAX_BREAKER_SAMPLES = 5000
_MAX_PENDING_APPROVALS = 1000     # "memory-bounding policy #2786"
_MAX_PENDING_GRANTS = 1000
```

Measured against the shipped module (500 single-call sessions dated a week back, then one live request):

```
window keys held in memory after 500 expired sessions: 501
window keys persisted to state.json                  : 501
keys whose every sample is older than the 5min window: 500
state.json size                                      : 25042 bytes

pending approvals persisted (cap is 1000)            : 1500
```

Two consequences: `state.json` grows monotonically with the number of distinct sessions that ever called `/verify`, and every `/verify` pays `json.dumps` over that growing dict inside `_STATE_LOCK` — a lock every request has to take.

`require_approval` escalations make the approvals path reachable without a cooperating client: each one mints a token that is stored until someone calls `/approve`. An agent that keeps tripping the escalation and never gets approved accumulates them past the documented cap.

## Fix

`save_state()` prunes under the lock before serialising — which is what the docstring already describes. `_prune_state` was already correct; it just had one caller.

Cost is bounded by the same walk `json.dumps` already does, and pruning is what keeps that walk small.

## Verification

7 new cases in `tests/test_state_pruning.py`, **4 red before the fix**:

- `test_expired_scopes_are_dropped_on_save` — 200 expired scopes, `_state["windows"]` must be `{}` after a save (and on disk)
- `test_state_file_stops_growing_with_dead_sessions` — file size stable across 100 → 500 expired scopes
- `test_pending_approval_cap_applies_without_a_restart` — cap enforced, oldest dropped first, newest kept
- `test_pending_grant_cap_applies_without_a_restart`

3 guards against over-pruning (green both before and after):

- `test_live_samples_are_kept` — in-window samples still count
- `test_open_circuit_breaker_survives_a_save` — a breaker in cooldown is not cleared
- `test_window_counter_still_survives_restart` — re-asserts the persistence contract from `test_windowed_governance_contract.py`

Full APE suite: **51 passed**.

> Note: these tests only gate once something runs the APE suite — no workflow does today. #2011 adds that.

## Why this matters

Long-running APE instances continually save governance state. Pruning only at startup lets expired scopes and counters grow indefinitely between restarts.

## Overlap check

Triage refresh (2026-08-08): searched open and closed Osmantic/ODS PRs by semantic keywords (APE governance state pruning), inspected the changed production files, and checked patch equivalence against current upstream/main. No strict duplicate or merged equivalent was found.

## Test plan

- [x] `pytest ods/extensions/services/ape/tests/test_state_pruning.py`


Lightheartdevs:
Thanks for tightening the APE governance-state lifecycle; this is a good fix for a slow-growth runtime problem that would otherwise stay invisible in short CI runs.

What looks good:
- `ods/extensions/services/ape/main.py` now calls `_prune_state(time.time())` inside `save_state()` while holding `_STATE_LOCK`, matching the module's documented load/save pruning contract and keeping serialization under the same synchronization boundary.
- The new `ods/extensions/services/ape/tests/test_state_pruning.py` coverage exercises the important failure modes: expired scope eviction, stable state-file size, pending approval/grant caps, and guards against pruning live samples or active breaker cooldowns.
- The change is narrow: it reuses the existing pruning function instead of adding a separate cleanup path, which keeps the persisted-state rules centralized.

What I checked:
- Static review of the PR diff and surrounding APE state-management code.
- Repository context: APE is a bundled policy gate under `ods/extensions/services/ape`, and recent maintainer direction favors bounded operational state plus explicit contract tests for runtime surfaces.
- GitHub checks are green for the existing workflows on this head SHA; the PR description also reports the full APE suite passing locally.

PR #2013, opened 2026-07-23
fix(amd-topo): report PCIe generation and lane count in the shared schema

tang-vu:
﻿## Problem

`amd-topo.sh`'s header says it

> Mirrors nvidia-topo.sh output schema for vendor-agnostic consumption by assign_gpus.py and the dashboard API.

Two fields of that shared schema mean different things depending on which vendor filled them in.

**`pcie_gen`** — `nvidia-topo.sh` reads `nvidia-smi --query-gpu=pcie.link.gen.current`, a PCIe *generation* (1-6). sysfs exposes the transfer rate instead, and the AMD path passed its leading digits straight through:

```bash
pcie_gen=$(cat "$card_dir/current_link_speed" 2>/dev/null | grep -oP '^\d+' || ...)
```

`current_link_speed` is `"16.0 GT/s PCIe"`, so a Gen4 card reported `16`. Gen5 reported `32`, Gen3 `8` — and Gen1 (`"2.5 GT/s"`) reported `2`, which collides with Gen2's own value.

**`pcie_width`** — nvidia reports a bare lane count (`"16"`). The AMD path prefixed an `x` while building the TSV row:

```bash
gpus_tsv+="...${pcie_gen}	x${pcie_width}	..."
```

so the same field holds `"x16"` on AMD.

`tests/fixtures/topology_json/*.json` pin the intended shape — `"pcie_gen": "4"`, `"pcie_width": "16"`.

**Honest severity:** no consumer thresholds `pcie_gen` today (`assign_gpus.py` and `gpu.py` read `memory_gb` and the link ranks), so nothing is currently broken at runtime. The value is written into `GPU_TOPOLOGY_JSON`, persisted, and surfaced from there — this is a correctness/parity fix so the next consumer that compares generations across vendors doesn't read AMD cards as four generations ahead of where they are.

## Fix

- `amd_pcie_gen_from_speed()` — GT/s → generation, `unknown` for anything unrecognised
- `_amd_pcie_gen()` / `_amd_pcie_width()` — read `current_*`, fall back to `max_*` when the current attribute is absent, `"Unknown"`, or `0` (a card parked in a low-power link state reports exactly that)
- drop the `x` prefix

## Verification

13 new assertions in `tests/test-amd-topo.sh`, **all red before the fix**:

```
✗ gpu 0 pcie_gen is a generation, not GT/s: expected '4', got '16'
✗ gpu 0 pcie_width has no 'x' prefix:       expected '16', got 'x16'
✗ gpu 1 falls back to max_link_speed:       expected '4', got '16'
✗ gpu 1 falls back to max_link_width:       expected '16', got 'x16'
```

plus the full GT/s table (2.5/5/8/16/32/64, no-decimal `"8 GT/s"`, `"Unknown"`, empty). The end-to-end case drives `detect_amd_topo` over a fake sysfs tree through the existing `ODS_DRM_SYS` seam — one healthy card and one in a low-power link state.

Suite after the fix: **36/36**. `bash -n` clean, no new shellcheck findings.

## Test wiring

`tests/test-amd-topo.sh` and `tests/test-nvidia-topo.sh` had no runner at all — neither `make test` nor any workflow invoked them, so the existing 23 + 7 assertions never gated anything either. Both are now wired into `make test` and `test-linux.yml` (`bash -n`/YAML validated; `test-nvidia-topo.sh` passes 7/7 unchanged).

## Why this matters

Consumers use one topology schema across NVIDIA and AMD. Reporting PCIe transfer-rate digits or prefixed lane widths as different semantics leads dashboards and scheduling logic to compare unlike values.

## Overlap check

Triage refresh (2026-08-08): searched open and closed Osmantic/ODS PRs by semantic keywords (AMD topology PCIe generation lanes), inspected the changed production files, and checked patch equivalence against current upstream/main. No strict duplicate or merged equivalent was found.

## Test plan

- [x] `bash ods/tests/test-amd-topo.sh`


Lightheartdevs:
Thanks for tightening the AMD topology schema parity; keeping `pcie_gen` and `pcie_width` vendor-neutral prevents subtle dashboard/API mistakes later.

What looks good:
- `ods/installers/lib/amd-topo.sh` now converts sysfs GT/s link speeds into PCIe generation numbers, matching the `nvidia-topo.sh` meaning for `pcie_gen` instead of reporting transfer-rate digits like `16` or `32`.
- AMD `pcie_width` now emits the bare lane count without the `x` prefix, matching the topology JSON fixtures and the NVIDIA schema shape.
- `ods/tests/test-amd-topo.sh` covers the GT/s conversion table, low-power current-link fallback to max-link fields, and end-to-end fake sysfs detection; the topology tests are now wired into both `ods/Makefile` and `.github/workflows/test-linux.yml`.

What I checked:
- Reviewed the PR description, full patch, and surrounding topology fixture/schema expectations on `origin/main`.
- Checked the architecture fit: topology data is persisted into `GPU_TOPOLOGY_JSON` and surfaced vendor-agnostically through installer/dashboard paths, even if no current consumer thresholds `pcie_gen` yet.

PR #2014, opened 2026-07-23
fix(installer): answer for the current run's background tasks

tang-vu:
﻿## Problem

`BG_TASK_REGISTRY` defaults to `/tmp/ods-bg-tasks.json` — a fixed path shared by every run on the host. Nothing clears it.

`bg_task_start` appended unconditionally:

```python
tasks = json.loads(registry_path.read_text())
tasks.append({"id": task_id, "pid": pid, ...})
```

`bg_task_status` took the **first** matching record:

```python
task = next((t for t in tasks if t["id"] == task_id), None)
```

Re-running the installer registers the same well-known ids again — `"sdxl-download"` (phase 11) and `"full-model-download"` (bootstrap upgrade) — so every lookup answers with the *previous* run's record.

Reproduced against the shipped lib (register under a dead pid, then register the same id under a live one):

```
records in registry : 2
live pid            : 694   (dead pid was 684)
bg_task_status rc   : 1   (0=running 1=completed 2=failed 3=not found)
```

The task is running; the registry says it finished.

Both directions land on a user:

- **`13-summary.sh`** prints `SDXL Lightning model download completed` for a download still in flight, or reports a failure by grepping `ERROR` out of the *previous* run's log.
- **`12-health.sh`** treats `bg_task_status ... == 0` as "a model download is active" and **skips the blocking LLM health wait** entirely. A pid from an earlier run that the kernel has since recycled onto an unrelated process keeps `os.kill(pid, 0)` succeeding forever, so the installer finishes and declares success without ever probing `llama-server`.

## Fix

- `bg_task_start` replaces any record carrying the same id instead of appending a duplicate.
- The lookup prefers the **newest** match, so a registry file left by an earlier release (already holding duplicates) still resolves to the right run.

Also: a junk file at the shared `/tmp` path aborted the caller. `/tmp` is world-writable and the registry is pure scratch state, but `json.loads` raising made `bg_task_start` exit non-zero — and phase 11 calls it as the last command of an `if` body under `set -euo pipefail`, so the install dies. Unparseable content now restarts the file, and `bg_task_status` answers `3` (not found), which every call site already handles.

Narrow `(OSError, ValueError)` at an I/O boundary, mapping to a distinct meaningful state — nothing about a real task is swallowed.

## Verification

New `tests/test-background-tasks.sh`, 11 assertions, **6 red before the fix**:

```
[FAIL] one record per task id: expected '1', got '2'
[FAIL] status reports the live task as running: expected '0', got '1'
[FAIL] newest record wins: expected '0', got '1'
[FAIL] bg_task_start succeeds over a corrupt registry: expected '0', got '1'
[FAIL] registry rebuilt with just this task: expected '1', got ''
[FAIL] status returns 3 (not found): expected '3', got '1'
```

The remaining 5 pin the documented status codes (`0` running / `1` completed / `2` failed-per-log / `3` not found) and pass both before and after — the contract `12-health.sh` and `13-summary.sh` switch on is unchanged.

Wired into `make test` and `test-linux.yml`.

## Why this matters

Installer reruns reuse fixed background-task IDs. Status and wait calls must answer for the current process rather than a stale record from an earlier download.

## Overlap check

Triage refresh (2026-08-08): searched open and closed Osmantic/ODS PRs by semantic keywords (background task current run stale), inspected the changed production files, and checked patch equivalence against current upstream/main. No strict duplicate or merged equivalent was found.

## Test plan

- [x] `bash ods/tests/test-background-tasks.sh`

### DIFF
--- a/.github/workflows/test-linux.yml
+++ b/.github/workflows/test-linux.yml
@@ -151,6 +151,9 @@ jobs:
       - name: Windows Missing Service Hint Tests
         run: bash tests/test-windows-missing-service-hints.sh
 
+      - name: Background Task Registry Tests
+        run: bash tests/test-background-tasks.sh
+
       - name: Install Readiness Summary Tests
         run: bash tests/test-install-readiness-summary.sh
 
--- a/ods/Makefile
+++ b/ods/Makefile
@@ -83,6 +83,9 @@ test: ## Run unit and contract tests
 	@echo "=== brave-search searxng compat contract ==="
 	@bash tests/test-brave-search-searxng-compat.sh
 	@echo ""
+	@echo "=== installer background task registry ==="
+	@bash tests/test-background-tasks.sh
+	@echo ""
 	@echo "=== ods-cli pipefail tolerance ==="
 	@bash tests/test-ods-cli-pipefail-tolerance.sh
 	@echo ""
--- a/ods/installers/lib/background-tasks.sh
+++ b/ods/installers/lib/background-tasks.sh
@@ -22,12 +22,12 @@ bg_task_start() {
     local pid="$2"
     local description="$3"
     local log_file="$4"
-    
+
     # Create registry if it doesn't exist
     if [[ ! -f "$BG_TASK_REGISTRY" ]]; then
         echo "[]" > "$BG_TASK_REGISTRY"
     fi
-    
+
     # Add task to registry
     python3 - "$BG_TASK_REGISTRY" "$task_id" "$pid" "$description" "$log_file" <<'PY'
 import json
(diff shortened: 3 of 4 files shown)


Lightheartdevs:
Thanks for taking on the background task registry behavior; stale installer state at a fixed `/tmp` path can create exactly the kind of misleading recovery signal that is hard for users to diagnose.

What looks good:
- `ods/installers/lib/background-tasks.sh` now replaces records by task id on start and reads the newest matching record for legacy duplicate registries, so reruns answer for the current download rather than an earlier `sdxl-download` or `full-model-download` entry.
- Corrupt or non-list scratch registry content now maps to a recoverable empty/not-found state instead of aborting installer phases under `set -euo pipefail`.
- `ods/tests/test-background-tasks.sh` covers re-registration, legacy duplicates, corrupt registries, and the documented status codes, and the test is wired into both `ods/Makefile` and `.github/workflows/test-linux.yml`.

What I checked:
- Reviewed the full patch and the installer risk against the repo's current priority on idempotent reruns and clear recovery behavior.
- Checked the call-site contract described by the PR for health/summary phases: status codes `0/1/2/3` remain unchanged.

PR #2015, opened 2026-07-23
fix(mdns): strip only matched quote pairs when reading env file

vaibhavsrv:
## Problem

In `_read_env()` (`ods-mdns.py`), `.env` key-value pairs are extracted using `value.strip().strip('"').strip("'")`:

```python
key, _, value = stripped.partition("=")
env[key.strip()] = value.strip().strip('"').strip("'")
```

Python's `str.strip()` chaining removes all runs of single and double quote characters from both ends indiscriminately. This causes:
- Inner quote pairs to be destroyed (e.g. `ODS_DEVICE_NAME="'device'"` is stripped down to `device`, destroying inner single quotes)
- Inconsistent environment parsing compared to the matched quote pair contract enforced on Linux (`lib/safe-env.sh`), Windows (`compose-diagnostics.ps1`), and macOS (`ods-macos.sh`).

## Fix

Introduce `_strip_env_quotes()` to strip only matching surrounding quote pairs (`"..."` or `'...'`), leaving inner and mismatched quotes intact:

```python
def _strip_env_quotes(value: str) -> str:
    v = value.strip()
    if len(v) >= 2 and v[0] == v[-1] and v[0] in {"'", '"'}:
        return v[1:-1]
    return v
```

And update `_read_env()` to use `_strip_env_quotes(value)`.

This is a defensive environment parsing parity fix. It ensures that `.env` settings read by mDNS publisher service round-trip identically across platforms without stripping internal quote characters.

## Verification

- Verified `_strip_env_quotes()` strips matching `"..."` and `'...'` pairs while leaving inner quotes and mismatched quotes verbatim.
- Verified mDNS service script initializes cleanly.

### DIFF
--- a/ods/bin/ods-mdns.py
+++ b/ods/bin/ods-mdns.py
@@ -96,6 +96,13 @@ POLL_INTERVAL = _safe_positive_int_env("ODS_MDNS_POLL_INTERVAL", 30)
 _HOSTNAME_RE = re.compile(r"^[a-zA-Z0-9]([a-zA-Z0-9-]{0,30}[a-zA-Z0-9])?$")
 
 
+def _strip_env_quotes(value: str) -> str:
+    v = value.strip()
+    if len(v) >= 2 and v[0] == v[-1] and v[0] in {"'", '"'}:
+        return v[1:-1]
+    return v
+
+
 def _read_env() -> dict[str, str]:
     """Return current .env values, ignoring comments and blank lines."""
     if not ENV_FILE.is_file():
@@ -106,7 +113,7 @@ def _read_env() -> dict[str, str]:
         if not stripped or stripped.startswith("#") or "=" not in stripped:
             continue
         key, _, value = stripped.partition("=")
-        env[key.strip()] = value.strip().strip('"').strip("'")
+        env[key.strip()] = _strip_env_quotes(value)
     return env
 
 
--- a/ods/tests/test_mdns_subdomains.py
+++ b/ods/tests/test_mdns_subdomains.py
@@ -20,6 +20,8 @@ from __future__ import annotations
 import ast
 from pathlib import Path
 
+import pytest
+
 
 REPO_ROOT = Path(__file__).resolve().parent.parent
 MDNS_SCRIPT = REPO_ROOT / "bin" / "ods-mdns.py"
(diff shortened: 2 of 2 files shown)


Lightheartdevs:
Thanks for tightening the mDNS `.env` parsing to preserve inner and mismatched quote characters consistently with the rest of the config readers.

Quickest path to merge:
- Add a focused regression for `ods/bin/ods-mdns.py`, either as a small Python test or a shell test that imports/runs the helper with a temporary `.env`. It should cover `ODS_DEVICE_NAME='device'` preserving the inner single quotes after the outer pair is stripped, plus a mismatched quote value staying intact.
- Include that test in the appropriate repo test path if there is already a lightweight mDNS/bin test target, or at least name the exact local command in the PR so maintainers can reproduce it. Right now the patch changes parser behavior but only modifies `ods/bin/ods-mdns.py`.

What I checked:
- Reviewed the PR description and full patch for #2015.
- Checked current mDNS usage: `_read_env()` feeds `ODS_DEVICE_NAME` and service ports for local `.local` publishing, so parser parity matters for real runtime behavior.
- Checked current GitHub checks for this head: Python lint/type check, dashboard API/frontend, shell/PowerShell lint, Linux integration smoke, matrix smoke, env validation, and secret scan are green.

vaibhavsrv:

I've updated the PR with:
1. Added focused unit tests in `ods/tests/test_mdns_subdomains.py`:
   - `test_strip_env_quotes_preserves_inner_and_mismatched_quotes`: Asserts `"hello"` -> `hello`, `'hello'` -> `hello`, `"'device'"` -> `'device'` (inner single quotes preserved), `trailing"` -> `trailing"`, and `"` -> `"`.
   - `test_read_env_strips_matched_quotes`: Creates a temporary `.env` file, calls `_read_env()`, and asserts `ODS_DEVICE_NAME` unquotes outer double quotes while preserving inner single quotes, and mismatched quotes remain intact.
2. Verified all tests pass via `pytest tests/test_mdns_subdomains.py` (4 passed in 0.04s).

Lightheartdevs:
Thanks for cleaning up the mDNS env reader without widening its parsing rules.

What looks good:
- ods/bin/ods-mdns.py now strips only matched outer quote pairs, so mismatched or literal quote characters are preserved instead of being silently rewritten.
- The added tests in ods/tests/test_mdns_subdomains.py cover both the helper behavior and .env file loading, while keeping the existing static mDNS subdomain contract intact.
- The change is small and consistent with the repo's recent focus on quoted .env handling across installer/runtime surfaces.

What I checked:
- Static review of the mDNS env parser diff and existing subdomain contract tests.
- GitHub checks are green, including Python lint, mypy, shell lint, integration-smoke, distro smoke matrix, and tier env validation.

PR #2016, opened 2026-07-23
fix(dashboard): stop the usage sparkline losing the month you came from

tang-vu:
## Problem

`appendUsageHistory()` narrowed the stored samples to the period it had just fetched and wrote **that** back as the whole store:

```js
const samples = readUsageHistory()
  .filter(item => item && item.ts >= cutoff && item.period === sample.period)
...
writeUsageHistory(next)
```

`localStorage['ods-usage-summary-history-v1']` is a single array covering every period, and the page reads it back per period — `useUsageReport` seeds its state with

```js
useState(() => readUsageHistory().filter(item => item.period === `${range.start}:${range.end}`))
```

So the reader expects several periods to coexist while the writer keeps exactly one.

**What the user sees:** click ◀ on the month selector and the month you were just looking at is deleted from storage. The Cost Estimate / Tokens / Requests sparklines lose their live signal and `seriesFromHistory` silently falls back to the report's daily series, so the tile stops showing how the number actually moved during the session. Navigate back and forth and you never accumulate history for either month.

The page polls both the current and previous month every 10s but only appends the current one, so this fires on every month switch, not just an unusual one.

## Fix

Age everything out by the 24h cutoff first, then split: samples for the fetched period get the new sample appended, samples for other periods are carried through untouched.

`appendUsageHistory` still returns only the fetched period's samples, so `setHistory` and the sparklines are unchanged.

Per-period depth stays at 240; an overall cap keeps the entry bounded now that several periods can share it.

**Also**: the store is parsed straight out of `localStorage`, and a malformed entry made `useUsageReport`'s initial `item.period` lookup throw *during render* — the Usage page went blank rather than starting from an empty history. `readUsageHistory` now drops non-array / falsy entries.

## Verification

`src/pages/Usage.history.test.jsx` drives the real page through the month selector (no test-only exports — it asserts against `localStorage` directly). 4 cases, **3 red before the fix**:

```
× keeps the current month when the selector moves to the previous one
  expected [ '2026-04-01:2026-04-30' ] to deeply equal
           [ '2026-04-01:2026-04-30', '2026-05-01:2026-05-31' ]
× still has the original month after navigating back to it
  expected [ '2026-05-01:2026-05-31' ] to deeply equal
           [ '2026-04-01:2026-04-30', '2026-05-01:2026-05-31' ]
× renders when storage holds a malformed entry
```

The fourth case (24h ageing applied across every period) passes before and after — it guards the fix against over-keeping.

Dashboard suite: **222 passed (24 files)**. `npm run lint` clean, production compile clean.

### DIFF
--- a/ods/extensions/services/dashboard/src/pages/Usage.history.test.jsx
+++ b/ods/extensions/services/dashboard/src/pages/Usage.history.test.jsx
@@ -1,0 +1,120 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
+import { fireEvent, screen, waitFor } from '@testing-library/react'
+import { render } from '../test/test-utils'
+import Usage from './Usage' // eslint-disable-line no-unused-vars
+
+// The sparkline history store is keyed by period and read back per period, so
+// navigating between months must not drop the month you came from.
+const STORAGE_KEY = 'ods-usage-summary-history-v1'
+const MAY = '2026-05-01:2026-05-31'
+const APRIL = '2026-04-01:2026-04-30'
+
+function reportFor(start, end, spend) {
+  return {
+    period: { start, end },
+    source: { name: 'token-spy', status: 'ok', detail: null },
+    summary: {
+      spend_usd: spend,
+      requests: spend * 10,
+      input_tokens: 0,
+      output_tokens: 0,
+      cache_read_tokens: 0,
+      cache_write_tokens: 0,
+      total_tokens: spend * 100,
+      tracked_providers: 1,
+      billing_providers: 1,
+      local_providers: 0,
+      untracked_providers: 0,
+      paid_cost_usd: spend,
+      local_cost_usd: 0,
+    },
+    daily: [],
+    models: [],
+    services: [],
+    sources: [],
+  }
+}
+
+function installFetchMock() {
+  let call = 0
+  vi.stubGlobal('fetch', vi.fn(async (url) => {
+    const text = String(url)
+    if (text.includes('/api/usage/readiness')) {
+      return { ok: true, json: async () => ({ status: 'ready', actions: {} }) }
+    }
+    const params = new URLSearchParams(text.split('?')[1] || '')
+    call += 1
+    return {
+      ok: true,
+      json: async () => reportFor(params.get('start'), params.get('end'), call),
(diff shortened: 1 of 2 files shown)


Lightheartdevs:
Thanks for fixing the usage sparkline retention across month navigation; preserving that local history makes the dashboard's live usage cards much less surprising.

What looks good:
- `ods/extensions/services/dashboard/src/pages/Usage.jsx` now ages the shared localStorage history first, then updates only the fetched period while carrying other periods forward.
- The per-period cap remains intact, and the new overall cap keeps the shared store bounded now that multiple periods can coexist.
- `ods/extensions/services/dashboard/src/pages/Usage.history.test.jsx` drives the real Usage page through month navigation and malformed storage cases instead of relying on test-only exports.

What I checked:
- Reviewed the full patch and current `useUsageReport` initialization path against the dashboard's period-keyed history model.
- Checked that the reader now rejects non-array storage and filters falsy entries before render-time `item.period` access.
- Checked current GitHub checks for this head: dashboard frontend/API, shell/Python lint, Linux integration smoke, distro matrix, env validation, PowerShell lint, and secret scan are green.

Lightheartdevs (approved):

tang-vu:
Hi Mike — thanks again for reviewing and merging this change.

I built a small prototype from this contributor workflow called ProofPort: it turns one merged public PR into a portable work receipt, with the maintainer signing the exact claim and a Solana Attestation Service devnet record providing independently checkable signer/schema/lifecycle state.

The current receipt for this PR is deliberately self-issued and proves only that the technical flow works. Since you actually approved and merged #2016, would you be open to a 10-minute independent maintainer test?

You would sign in with GitHub, review the frozen public PR facts, connect your own devnet wallet, sign a readable payload authorization, and sign the SAS devnet transaction. No real funds, private repository access, seed phrase, private key, payment, or endorsement is requested; you can reject the claim, and the result will preserve only the assurance checks the product actually establishes.

Prototype: https://proofport.tangvu.dev
Current self-issued proof: https://proofport.tangvu.dev/proof/8c0cf4d8-b0f7-42f8-97cc-bf4ce57a2f8f

If you are open to it, a reply here is enough and I will send the short setup steps. Thank you.

PR #2018, opened 2026-07-23
fix(dashboard-api): guard null nodes payload in cluster status refresh

vaibhavsrv:
## Problem

In `ClusterStatus.refresh()` (`agent_monitor.py`), node data is extracted from the cluster proxy response using `data.get("nodes", [])`:

```python
if proc.returncode == 0:
    data = json.loads(stdout.decode())
    self.nodes = data.get("nodes", [])
    self.total_gpus = len(self.nodes)
    self.active_gpus = sum(1 for n in self.nodes if n.get("healthy", False))
```

If the status endpoint returns JSON containing `"nodes": null` (e.g., during proxy startup or cluster node re-initialization), `data.get("nodes", [])` returns `None` instead of `[]`.

Subsequent list operations (`len(self.nodes)`) raise `TypeError: object of type 'NoneType' has no len()`, causing background cluster health checks to crash.

## Fix

Update `ClusterStatus.refresh()` to verify `data` is a dictionary, fall back to `[]` when `data.get("nodes")` is `None` or non-list, and ensure element checks verify `isinstance(n, dict)`:

```python
if proc.returncode == 0:
    data = json.loads(stdout.decode())
    if isinstance(data, dict):
        nodes_data = data.get("nodes")
        self.nodes = nodes_data if isinstance(nodes_data, list) else []
        self.total_gpus = len(self.nodes)
        self.active_gpus = sum(
            1 for n in self.nodes if isinstance(n, dict) and n.get("healthy", False)
        )
        self.failover_ready = self.active_gpus > 1
```

This is a defensive robustness fix for cluster proxy telemetry parsing. When the cluster proxy returns null or malformed node fields, background monitoring handles the state gracefully without crashing.

## Verification

Added unit test in `tests/test_agent_monitor.py`:

- `test_refresh_handles_null_nodes_payload`
  - Verifies `ClusterStatus.refresh()` handles `{"nodes": null}` without raising `TypeError`.

- `pytest tests/test_agent_monitor.py` passed (20 passed in 0.13s).


Lightheartdevs:
Thanks for hardening the cluster status refresh around malformed proxy payloads; dashboard background metrics should degrade cleanly when optional telemetry is incomplete.

What looks good:
- `ods/extensions/services/dashboard-api/agent_monitor.py` now treats missing or `null` `nodes` as an empty list and ignores non-dict node entries when counting healthy GPUs, avoiding the previous `len(None)` / `n.get(...)` failures.
- The change stays inside `ClusterStatus.refresh()` and preserves the existing default empty cluster state and `failover_ready` calculation.
- `ods/extensions/services/dashboard-api/tests/test_agent_monitor.py` adds a focused async regression for `{nodes: null}`.

What I checked:
- Reviewed the full patch and surrounding `ClusterStatus` / `get_full_agent_metrics` behavior on `origin/main`.
- Checked current GitHub checks for this head: dashboard API/frontend, Python lint/type check, shell/PowerShell lint, Linux integration smoke, matrix smoke, env validation, and secret scan are green.
- Confirmed this does not change the cluster proxy contract or installer/runtime wiring.

PR #2019, opened 2026-07-23
fix(qrcode): read the LAN IP from the src field, not column 7

tang-vu:
## Problem

`_ods_lan_ip` parsed `ip route get` output by column:

```bash
lan_ip=$(ip route get 1 2>/dev/null | awk '{print $7; exit}')
```

`ip route get` prints `via <gateway>` **only when the route has one**. On an on-link default route those two fields are absent and everything after shifts left by two, so column 7 lands on the uid:

```
route: 1.0.0.0 via 192.168.1.1 dev eth0 src 192.168.1.50 uid 1000
  positional $7 -> 192.168.1.50
  src field     -> 192.168.1.50

route: 1.0.0.0 dev eth0 src 172.28.112.1 uid 1000
  positional $7 -> 1000          ← the uid
  src field     -> 172.28.112.1
```

On-link default routes are normal on WSL2, point-to-point and VPN interfaces (`tun0`/`wg0`), and several cloud VM images.

The value flows straight into user-facing output:

- `print_success_card` prints `Remote Access (LAN): Dashboard: http://1000:3001`
- `print_dashboard_qr` encodes that same string into the QR code the install summary tells the user to scan from their phone

Nothing validates it, so the failure is silent — the user just gets a QR that goes nowhere.

## Fix

Read the address from the `src` field by name, which is what `installers/phases/06-directories.sh` already does for `HOST_LAN_IP`:

```bash
ip -4 route get 1.1.1.1 | awk '{for (i = 1; i <= NF; i++) if ($i == "src") { print $(i + 1); exit }}'
```

`-4` is added for the same reason the installer uses it: an IPv6 answer can never fit the `http://<ip>:3001` shape these callers build.

The `ifconfig` fallback is untouched.

## Verification

`tests/test-qrcode-helper.sh` gains three cases — gateway route, on-link route, and the dashboard URL that comes out of the on-link case. Red before the fix:

```
[FAIL] _ods_lan_ip misread an on-link (gateway-less) route: got '1000'
```

The existing macOS-`ifconfig` and explicit-URL assertions still pass.

**Test wiring:** `tests/test-qrcode-helper.sh` had no runner — neither `make test` nor any workflow invoked it — so its existing assertions never gated anything either. Now wired into both.

### DIFF
--- a/.github/workflows/test-linux.yml
+++ b/.github/workflows/test-linux.yml
@@ -68,6 +68,9 @@ jobs:
       - name: Health Check Tests
         run: bash tests/test-health-check.sh
 
+      - name: QR Code Helper Tests
+        run: bash tests/test-qrcode-helper.sh
+
       - name: n8n Cookie Policy Tests
         run: bash tests/test-n8n-cookie-policy.sh
 
--- a/ods/Makefile
+++ b/ods/Makefile
@@ -55,6 +55,9 @@ test: ## Run unit and contract tests
 	@echo "=== safe .env loading (quote/CRLF handling) ==="
 	@bash tests/test-safe-env.sh
 	@echo ""
+	@echo "=== QR code helper LAN detection ==="
+	@bash tests/test-qrcode-helper.sh
+	@echo ""
 	@echo "=== Bind address sweep ==="
 	@bash tests/test-bind-address-sweep.sh
 	@bash tests/test-n8n-cookie-policy.sh
--- a/ods/lib/qrcode.sh
+++ b/ods/lib/qrcode.sh
@@ -7,10 +7,17 @@
 # ═══════════════════════════════════════════════════════════════
 
 # Best-effort LAN IP detection for remote access URLs.
+#
+# The address is read from the `src` field by name, not by column. `ip route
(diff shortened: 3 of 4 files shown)


Lightheartdevs:
Thanks for fixing the QR-code LAN address detection; this is a practical installer UX bug for WSL2, VPN, and on-link route setups.

What looks good:
- `ods/lib/qrcode.sh` now reads the IPv4 route `src` field by name instead of depending on column 7, so gateway-less `ip route get` output no longer turns `uid` into the advertised host.
- The helper now matches the existing `HOST_LAN_IP` parsing pattern in `ods/installers/phases/06-directories.sh`, keeping installer and success-card network assumptions aligned.
- `ods/tests/test-qrcode-helper.sh` adds gateway, on-link, and rendered dashboard URL coverage, and the test is wired into both `ods/Makefile` and `.github/workflows/test-linux.yml`.

What I checked:
- Reviewed the PR description, full patch, and current QR success-card flow on `origin/main`.
- Checked the route parsing against the existing installer phase's `ip -4 route get 1.1.1.1` source-field logic.

Lightheartdevs (approved):

PR #2020, opened 2026-07-23
fix(security): create credential files owner-only, not after the fact

tang-vu:
﻿## Problem

Two credential files are written at the process umask and tightened afterwards, so each spends a window at `0644` on a default Linux host (`umask 022`):

**`privacy-shield/key_management.persist_key`**

```python
with open(path, "w", encoding="utf-8") as f:   # created 0644
    f.write(key)
try:
    os.chmod(path, 0o600)
except Exception:
    pass                                        # ← silent
```

**`dashboard-api/routers/magic_link._write_store`**

```python
tmp.write_text(json.dumps(store, indent=2))     # tmp created 0644
tmp.replace(store_path)                         # published 0644
try:
    store_path.chmod(0o600)                     # ← after the fact
except OSError:
    pass
```

`/data` is a bind mount shared with the host, so anything readable there is readable by other local users on the box.

- The shield API key authenticates **every** request through the PII proxy — `_token_valid` compares against exactly this value.
- The magic-link store holds the SHA-256 hashes that back redemption, plus the audit trail. Its `.json.tmp` sits in the same directory at `0644` for the whole write, not just the gap between the two calls.

privacy-shield is worse than a window: the chmod was under `except Exception: pass`, so a filesystem that rejects chmod (the comment names Docker volumes) leaves the key world-readable **permanently** and logs nothing.

## Fix

`oauth_passthrough._atomic_write_0600` already documents the right ordering — tighten the temp file, *then* rename, so the destination is never observable at the loose mode. This applies that shape to both sites:

- `key_management.persist_key` creates via `os.open(..., 0o600)`, so the mode comes from the create rather than a follow-up call. It still chmods afterwards because `O_CREAT` leaves an existing file's mode alone — and a chmod failure now logs a warning naming the exposure instead of passing silently.
- `magic_link._write_store` chmods the tmp **before** the replace, and downgrades the swallow to a debug log, matching the oauth helper.

**Deliberately not touched:** `dashboard-api/security.py` and `token-spy/main.py` have the same write-then-chmod ordering, but #1948 is already reworking both files. The same point applies there and is better made in that PR than in a conflicting one.

## Verification

File modes are POSIX-only, so both tests skip on Windows; these results are from a Linux run.

**privacy-shield** — 4 new cases, **2 red before the fix**:

```
FAIL: test_mode_comes_from_the_create_not_a_later_chmod
AssertionError: 420 != 384 : key file was created at the umask default and
only tightened afterwards — readable by other users until the chmod lands
FAIL: test_generated_key_lands_owner_only
AssertionError: 420 != 384
```

(420 = `0o644`, 384 = `0o600`.) The proof neutralises `os.chmod` so the assertion pins the *ordering*, not the end state — a write-then-chmod cannot pass it. The other two cover the normal result and an existing loose file being tightened.

**magic_link** — 1 new case, **red before the fix**. It spies on the rename and asserts the file being published already carries `0600`:

```
>       assert all(mode == 0o600 for mode in published_modes)
E       assert False
```

Suites: privacy-shield `7/7`, dashboard-api `test_magic_link` **57 passed**, full dashboard-api suite green.

## Why this matters

API tokens and generated encryption keys live on host-mounted storage. Creating them owner-only from the first write avoids even a short permissive exposure window.

## Overlap check

Triage refresh (2026-08-08): searched open and closed Osmantic/ODS PRs by semantic keywords (credential file owner permissions), inspected the changed production files, and checked patch equivalence against current upstream/main. No strict duplicate or merged equivalent was found.

## Test plan

- [x] `pytest ods/extensions/services/dashboard-api/tests/test_magic_link.py ods/extensions/services/privacy-shield/tests/test_key_management.py`


Lightheartdevs:
Thanks for tightening the credential-file write paths; avoiding even short world-readable windows matters on host-mounted `/data` state.

What looks good:
- `ods/extensions/services/privacy-shield/key_management.py` now creates generated shield keys with `os.open(..., 0o600)`, so the mode comes from file creation rather than a later best-effort chmod.
- `ods/extensions/services/dashboard-api/routers/magic_link.py` chmods the temporary token store before publishing it with `replace`, which matches the repo's existing atomic secret-write direction.
- The new tests cover both the final mode and the ordering: privacy-shield tests neutralize `chmod`, and the magic-link test observes the temp file mode at rename time.

What I checked:
- Reviewed the full patch and surrounding auth/privacy paths: `SHIELD_API_KEY` gates the privacy proxy, and the magic-link store contains token hashes used for redemption.
- Checked the PR's explicit scope note against related write-then-chmod sites in `dashboard-api/security.py` and `token-spy/main.py`; keeping those for the already-overlapping follow-up avoids broadening this PR.
- Checked current GitHub checks for this head: Python lint/type check, dashboard API/frontend, shell/PowerShell lint, Linux integration smoke, matrix smoke, env validation, and secret scan are green.

PR #2023, opened 2026-07-23
fix(token-spy): keep tool-call chains intact when trimming history

tang-vu:
﻿## Problem

`_filter_history` groups messages into atomic units first, and says why:

```python
# Step 2: Group conversation messages into atomic units
# An atomic unit is: [user msg, assistant reply, tool_call/tool result chain]
# We must not split these or the API contract breaks.
```

The `max_total_chars` pass then flattened them and popped **one message at a time**:

```python
while len(filtered_conv) > always_keep_last_n:
    total = sum(len(json.dumps(m, separators=(",", ":"))) for m in filtered_conv)
    if total <= max_total:
        break
    # Remove the oldest non-system unit
    filtered_conv.pop(0)          # ← pops a message, not a unit
    result.messages_removed += 1
```

The comment says unit; the code pops a message.

Driving the shipped filter with a conversation whose first exchange is a tool call (`max_total_chars=620`, `always_keep_last_n=2`):

```
resulting roles: ['system', 'tool', 'assistant', 'user', 'assistant', 'user']
orphaned tool messages (no preceding assistant tool_calls): [1]
```

The loop stopped right after removing the assistant `tool_calls` turn but before its result, so the request forwarded to llama-server opens with a tool result answering nothing. Every OpenAI-compatible server rejects that shape. token-spy is a transparent proxy in front of the model, so the user's chat request just fails — and the filter that broke it is the one meant to keep requests small.

### `always_keep_last_n` was never enforced on the `max_pairs` path

It is documented as a floor for the whole filter, but only the `max_total_chars` loop consulted it. `max_pairs` sliced units unconditionally, and the block that was supposed to backstop it ended in a bare `pass`:

```python
# Step 7: Apply always_keep_last_n safety — ensure the last N raw messages ...
    if not tail_ids.issubset(existing_ids):
        pass  # Units-based approach already preserves recent messages
```

Measured: `always_keep_last_n=6` with `max_pairs=1` kept **1** conversation message.

## Fix

Both trims go through a single `_drop_leading_units(units, max_to_drop, keep_last_n)` helper that removes **whole units** and stops before the retained conversation falls below the floor. The flatten moves after the size trim, so there is no message-level list left to pop from. The dead Step 7 block goes away — its job is now done where the trimming happens.

Also: `drop_old_tool_calls` strips `tool_calls` off an older assistant turn, but a pure tool-call turn carries `content: null` — what was left was a message with neither content nor tool_calls. Those are dropped along with the tool results they belonged to.

Same defaults, same knobs, same `FilterResult` fields.

## Verification

`filters.py` had no test file. `tests/test_filters.py` adds 16 cases, **5 red before the fix**:

```
FAILED TestHistoryKeepsToolChainsIntact::test_max_total_chars_never_orphans_a_tool_result
FAILED TestHistoryKeepsToolChainsIntact::test_no_orphans_at_any_size_boundary[620]
FAILED TestHistoryKeepsToolChainsIntact::test_no_orphans_at_any_size_boundary[700]
FAILED TestAlwaysKeepLastN::test_max_pairs_does_not_cut_below_the_floor
FAILED TestDropOldToolCalls::test_stripped_assistant_turn_is_not_left_empty
```

The other 11 pass before and after: trimming still shrinks the body, system messages survive, tool results still truncate, the tool blocklist still filters, and `enabled: false` is still a no-op.

After the fix: **16/16**, and the reproduction now yields `['system', 'user', 'assistant', 'user']` with no orphans, and `always_keep_last_n=6` keeps 7 messages instead of 1.

> Note: no workflow runs the token-spy suite today, so these tests only gate once something does — #2011 adds that.

## Why this matters

OpenAI-compatible requests require assistant tool calls and their tool results to remain a valid chain. Character or pair trimming must not split that atomic exchange.

## Overlap check

Triage refresh (2026-08-08): searched open and closed Osmantic/ODS PRs by semantic keywords (Token Spy tool-call chain trimming), inspected the changed production files, and checked patch equivalence against current upstream/main. No strict duplicate or merged equivalent was found.

## Test plan

- [x] `pytest ods/extensions/services/token-spy/tests/test_filters.py`


Lightheartdevs:
Thanks for taking on the Token Spy history trimming path; keeping tool-call exchanges atomic is important for preserving the OpenAI-compatible request contract.

What looks good:
- `ods/extensions/services/token-spy/filters.py` moves both `max_pairs` and `max_total_chars` trimming through whole conversation units, so an assistant `tool_calls` turn and its `tool` results are no longer split by size pressure.
- The `always_keep_last_n` floor is enforced where units are dropped, replacing the old dead tail-check block with behavior that applies to both trim paths.
- `ods/extensions/services/token-spy/tests/test_filters.py` adds the missing coverage for orphaned tool results, floor enforcement, stripped empty assistant turns, disabled filters, truncation, and tool blocklisting.

What I checked:
- Reviewed the PR description, full patch, and surrounding Token Spy filter implementation on `origin/main`.
- Compared the change against the repo architecture: Token Spy is a transparent proxy in the observability layer, so forwarding invalid message shapes is a user-visible runtime failure rather than only telemetry loss.
- Checked current GitHub checks for this head: Python lint/type check, dashboard API/frontend, shell/PowerShell lint, Linux integration smoke, matrix smoke, env validation, and secret scan are green. The PR correctly notes the Token Spy-specific suite is not yet wired into CI.

PR #2024, opened 2026-07-23
fix(session-signer): validate integer expiry before computing hmac signature

vikkiydv:
## Problem

In `verify()` (`session_signer.py`), signed cookie validation checked the HMAC-SHA256 signature *before* validating the integer timestamp format of `expiry_str`:

```python
payload = f"{random_id}.{expiry_str}"
expected_sig = _sign(payload)

if not hmac.compare_digest(expected_sig.encode("utf-8"), claimed_sig.encode("utf-8")):
    return False, "bad-signature"

try:
    expiry = int(expiry_str)
    return False, "malformed"
```

If `cookie_value` contains a non-integer or malformed expiry string (e.g. `random_id.abc.claimed_sig`), `verify()` computed `_sign(payload)` and performed timing-safe digest comparison before returning `(False, "malformed")`.

## Fix

Reorder validation steps so `expiry_str` integer parsing occurs *before* computing `_sign(payload)`:

```python
random_id, expiry_str, claimed_sig = parts
if not random_id or not expiry_str or not claimed_sig:
    return False, "malformed"

try:
    expiry = int(expiry_str)
    return False, "malformed"

payload = f"{random_id}.{expiry_str}"
expected_sig = _sign(payload)

if not hmac.compare_digest(expected_sig.encode("utf-8"), claimed_sig.encode("utf-8")):
    return False, "bad-signature"
```

This is a defensive input validation ordering fix. Non-integer timestamps are rejected immediately as `"malformed"` without wasting CPU cycles computing HMAC-SHA256 signatures on unparseable expiry strings.

## Verification

- `pytest tests/test_session_signer.py` passed (21 passed in 1.15s).

### DIFF
--- a/ods/extensions/services/dashboard-api/session_signer.py
+++ b/ods/extensions/services/dashboard-api/session_signer.py
@@ -147,6 +147,11 @@ def verify(cookie_value: str) -> Tuple[bool, str]:
     if not random_id or not expiry_str or not claimed_sig:
         return False, "malformed"
 
+    try:
+        expiry = int(expiry_str)
+    except (ValueError, TypeError):
+        return False, "malformed"
+
     payload = f"{random_id}.{expiry_str}"
     expected_sig = _sign(payload)
     # Constant-time compare to defeat signature timing oracles. Encoded to
@@ -155,12 +160,6 @@ def verify(cookie_value: str) -> Tuple[bool, str]:
     # verify() must return a reason, never raise.
     if not hmac.compare_digest(expected_sig.encode("utf-8"), claimed_sig.encode("utf-8")):
         return False, "bad-signature"
-
-    # Signature is good — check expiry.
-    try:
-        expiry = int(expiry_str)
-    except (ValueError, TypeError):
-        return False, "malformed"
 
     # `<=` because the expiry timestamp is the moment of invalidation, not
     # the last valid second. A cookie issued with ttl=60 stops being valid
--- a/ods/extensions/services/dashboard-api/tests/test_session_signer.py
+++ b/ods/extensions/services/dashboard-api/tests/test_session_signer.py
@@ -172,15 +172,16 @@ class TestVerify:
         assert ok is False
         assert reason == "no-secret"
 
-    def test_non_integer_expiry(self):
-        """A malformed expiry field that survives split should be caught."""
-        # Sign a payload with a non-integer expiry. Signature will be valid
-        # but the int() parse will fail.
-        random_id = "abc123"
-        bad_expiry = "not-a-number"
-        payload = f"{random_id}.{bad_expiry}"
-        sig = session_signer._sign(payload)
-        cookie = f"{payload}.{sig}"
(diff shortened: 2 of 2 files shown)


Lightheartdevs:
Thanks for tightening the session cookie verifier around malformed expiry values before doing HMAC work.

Quickest path to merge:
- `ods/extensions/services/dashboard-api/tests/test_session_signer.py`: add a focused regression for a three-part cookie with a non-integer expiry, and assert it returns `(False, malformed)`. Ideally monkeypatch `_sign` to fail the test if it is called, since the behavior this PR changes is validation order rather than only the final reason string.
- Re-trigger CI for this head once the test is in place. GitHub currently reports no checks for `fix/session-signer-int-expiry-validation-order`, so maintainers do not have the usual dashboard-api, Python lint/type, smoke, env validation, and secret-scan evidence for this security-adjacent path.

What I checked:
- Reviewed the PR description and full patch for `ods/extensions/services/dashboard-api/session_signer.py` against the existing verifier contract and constant-time signature check comments.
- Checked the existing `tests/test_session_signer.py` coverage on `origin/main`; it covers malformed cookie shape, bad signatures, non-ASCII signatures, expiry, and secret rotation, but not this specific non-integer-expiry-before-signing order.
- Ran `gh pr checks 2024`; GitHub returned no checks reported for this branch.

vikkiydv:

I've updated the PR with commit `85e82925`:
1. Updated `test_non_integer_expiry` in `ods/extensions/services/dashboard-api/tests/test_session_signer.py` to monkeypatch `session_signer._sign` with a function that raises `AssertionError`.
2. Verifies that passing a 3-part cookie with a non-integer expiry (`"random_id.not-a-number.claimed_sig"`) returns `(False, "malformed")` BEFORE computing the HMAC signature, guaranteeing `_sign()` is never called.
3. Verified `pytest tests/test_session_signer.py` passed cleanly (21 passed in 1.16s).

Lightheartdevs:
Thanks for tightening the session cookie verifier and then adding the regression that proves malformed expiries are rejected before any HMAC work.

What looks good:
- `ods/extensions/services/dashboard-api/session_signer.py` now parses `expiry_str` before building/signing the payload, which keeps malformed three-part cookies on the `malformed` path without spending signature work.
- `ods/extensions/services/dashboard-api/tests/test_session_signer.py` now monkeypatches `_sign` to fail if it is reached for a non-integer expiry, so the test covers the changed ordering rather than only the final response tuple.
- The change stays inside the existing verifier contract: valid signatures still use `hmac.compare_digest`, expiry is still checked after signature validation, and malformed inputs remain rejected.

What I checked:
- Diff/static review of `session_signer.py` and `tests/test_session_signer.py` against the existing dashboard-api auth/session boundary.
- Ran `python -m pytest tests/test_session_signer.py` from `ods/extensions/services/dashboard-api` on this head: 21 passed.
- GitHub currently reports no check runs for this PR head, so I relied on the focused local test plus static review.

Thanks again for following through on the review feedback.

PR #2025, opened 2026-07-23
fix(token-spy): stop the SSE token stream re-sending every event

tang-vu:
﻿## Problem

`/token_events` streams usage rows over SSE. `main.py` polls `query_recent_events(after_id=last_id)` every 2s and advanced the cursor inside the emit loop:

```python
for event in events:
    yield f"data: {json.dumps(event_data)}\n\n"
    last_id = event.get("id")
```

`query_recent_events` ordered the `after_id` branch by **`timestamp DESC`**, so the last row iterated was the *oldest* of the batch, and `last_id` ended up on the smallest id. The next poll's `WHERE id > last_id` then re-selected every newer row it had just sent.

Driven against the shipped sqlite backend — 5 rows, three consecutive idle polls, no new data:

```
poll 0: cursor_after=1  emitted ids=[5, 4, 3, 2, 1]
poll 1: cursor_after=2  emitted ids=[5, 4, 3, 2]
poll 2: cursor_after=3  emitted ids=[5, 4, 3]
total emitted: 12  unique: 5  duplicates: 7
```

Every connected dashboard re-renders the same events on a 2-second loop, and any client-side counter fed from the stream drifts upward without bound.

## Fix

- **`db.py`** — the `after_id` branch now orders by **`id ASC`**. `id` is the `AUTOINCREMENT` key, i.e. insertion order, so this is a genuine forward cursor: a burst larger than one page drains across polls instead of returning the same newest page every time.
- **`main.py`** — advance the cursor to the **max** id in the batch, not the last one iterated. The initial `after_id=None` query still returns the newest page in `timestamp DESC` order (the backlog dump), so seeding from its last row would replay the whole backlog on poll two.

After the fix the same scenario emits each id exactly once and idle polls return nothing; a forward burst of 120 rows drains as `4..123` in order with no gaps or repeats.

## Scope — Postgres left for a separate change (flagged, not half-fixed)

This fixes the **sqlite** backend (the default local deployment) and the shared SSE loop. The Postgres backend has a deeper, independent defect: `requests.id` is a random `uuid4()`, so `WHERE r.id > %s` is a **lexicographic UUID comparison unrelated to insertion time** — it was never a valid forward cursor, with or without this bug. Fixing it correctly needs a timestamp-based composite cursor and a running Postgres to verify, so I've deliberately left it out rather than ship an unverifiable half-fix. Calling it out here so it isn't lost.

## Verification

`tests/test_recent_events_cursor.py`, 4 cases, **2 red before the fix**:

```
FAILED test_new_events_after_idle_polls_are_delivered_in_order
FAILED test_forward_burst_larger_than_one_page_drains_without_skips
```

- `test_each_event_is_streamed_once` — drain to empty, assert no duplicates
- `test_cursor_does_not_walk_backwards_on_the_initial_batch` — newest-first backlog, cursor must jump to max, next poll empty
- the two above — ordered delivery after idle polls, and a >1-page forward burst with no skips

`main.py` compiles; existing `test_usage_report.py` still green.

> Note: no workflow runs the token-spy suite today, so these gate once something does — #2011 adds that.

## Why this matters

The Token Spy SSE endpoint is a live production stream. Re-sending the same SQLite rows on every poll wastes bandwidth and causes dashboards to duplicate events.

## Overlap check

Triage refresh (2026-08-08): searched open and closed Osmantic/ODS PRs by semantic keywords (Token Spy SSE cursor events), inspected the changed production files, and checked patch equivalence against current upstream/main. No strict duplicate or merged equivalent was found. The PostgreSQL UUID-cursor problem remains intentionally separate in #2173 because it requires a different ordering design.

## Test plan

- [x] `pytest ods/extensions/services/token-spy/tests/test_recent_events_cursor.py`

### DIFF
--- a/ods/extensions/services/token-spy/db.py
+++ b/ods/extensions/services/token-spy/db.py
@@ -483,6 +483,14 @@ def query_recent_events(limit: int = 100, after_id: str = None):
     conn.row_factory = sqlite3.Row
 
     if after_id:
+        # Forward pagination for the SSE tail: return the OLDEST unseen rows
+        # first and let the caller advance its cursor to the last (largest)
+        # id. Ordering by timestamp DESC here made the caller's cursor walk
+        # backwards — it set last_id to the oldest row of each batch, so the
+        # next `id > last_id` re-selected everything newer than that, resending
+        # the whole batch every poll. `id` is the AUTOINCREMENT key, so ASC is
+        # insertion order and a burst larger than one page is drained across
+        # polls without skips.
         rows = conn.execute(
             """
             SELECT
@@ -492,7 +500,7 @@ def query_recent_events(limit: int = 100, after_id: str = None):
                 estimated_cost_usd as cost_usd, timestamp
             FROM usage
             WHERE id > ?
-            ORDER BY timestamp DESC
+            ORDER BY id ASC
             LIMIT ?
             """,
             (after_id, limit)
--- a/ods/extensions/services/token-spy/main.py
+++ b/ods/extensions/services/token-spy/main.py
@@ -2429,6 +2429,19 @@ async def token_events(request: Request):
                 # Query recent events
                 events = query_recent_events(limit=50, after_id=last_cursor)
 
+                # Advance to the highest cursor in the complete batch. SQLite's
+                # initial window is newest-first, while PostgreSQL emits a
+                # stable (timestamp, UUID) cursor; taking the maximum works for
+                # both without walking the SQLite cursor backwards.
+                batch_max_cursor = max(
+                    (
+                        event.get("_cursor", event.get("id"))
+                        for event in events
+                        if event.get("_cursor", event.get("id")) is not None
+                    ),
+                    default=None,
+                )
+
                 for event in events:
                     # Format event as SSE
                     event_data = {
@@ -2445,9 +2458,9 @@ async def token_events(request: Request):
                     }
 
                     yield f"data: {json.dumps(event_data)}\n\n"
-                    # PostgreSQL supplies a stable (timestamp, UUID) cursor.
-                    # SQLite and older backends continue to use the row ID.
-                    last_cursor = event.get("_cursor", event.get("id"))
+
+                if batch_max_cursor is not None:
+                    last_cursor = batch_max_cursor
 
                 # Heartbeat to keep connection alive
                 yield ":heartbeat\n\n"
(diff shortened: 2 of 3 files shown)


Lightheartdevs:
Thanks for digging into the Token Spy SSE cursor behavior and backing it with a concrete duplicate-stream regression.

Decision point:
- The sqlite fix in `ods/extensions/services/token-spy/db.py` is directionally right: the live `after_id` cursor should page forward by insertion id, not keep returning newest-first batches.
- The shared `event_stream()` change in `ods/extensions/services/token-spy/main.py` also fixes the initial backlog case by advancing to the batch max id instead of the oldest row in a timestamp-desc batch.
- The PR deliberately leaves `DB_BACKEND=postgres` unresolved, but `ods/extensions/services/token-spy/README.md` documents PostgreSQL/TimescaleDB as the production backend. That may be acceptable as a narrow sqlite/default-deploy fix, but I would want a maintainer to decide whether this should merge as scoped or wait for a Postgres cursor follow-up/issue before landing.

What I checked:
- Reviewed the PR title/description, full patch, and new `ods/extensions/services/token-spy/tests/test_recent_events_cursor.py` coverage.
- Checked current Token Spy docs and service config on `origin/main`; sqlite is the zero-config/default path, while Postgres is still documented as a supported production backend.
- GitHub checks are green, including Python lint, mypy, API/frontend jobs, smoke, distro, env validation, PowerShell lint, and secret scan. The PR notes the Token Spy-specific suite is not yet wired into CI.

tang-vu:
On the open decision point — the Postgres backend: keeping this PR scoped to SQLite is deliberate, and I have split the Postgres path out as #2173 rather than leave it implicit.

Short version of why it is not fixable here:

- `db_postgres.py:query_recent_events()` has the same `WHERE r.id > %s ORDER BY r.timestamp DESC` shape, but `r.id` is a **UUID**, not a monotonic key. So it is not the same bug — bytewise UUID comparison carries no insertion order, which means new rows sorting below the cursor are dropped from `/events` permanently, not merely re-sent.
- `event_stream()` is shared by both backends and advances the cursor with `max(...)`. Python's `UUID` is comparable, so the Postgres path fails silently instead of erroring.
- The `ORDER BY id ASC` fix does not transfer. A correct fix needs a monotonic cursor column, and `init_db()` only ensures the tenant row exists — `tenants` / `requests` / `agents` are provisioned externally and there is no migration or `CREATE TABLE` for them in the repo. Choosing between adding a `BIGSERIAL`, a `(timestamp, id)` keyset cursor, or UUIDv7 is a schema decision, and guessing at it blind would be worse than scoping it out.

So the SQLite fix here stands on its own for the default deployment, and #2173 tracks the Postgres cursor with the analysis and the three candidate fixes.

PR #2027, opened 2026-07-23
fix(lemonade): strip matched quote pairs when normalizing base url

vaibhavsrv:
## Problem

In `normalize_base_url()` (`lemonade_client.py`), base URLs are sanitized using `(base_url or DEFAULT_BASE_URL).strip().rstrip("/")`:

```python
base = (base_url or DEFAULT_BASE_URL).strip().rstrip("/")
```

When `base_url` is passed with surrounding quotes (e.g. read directly from environment variables or configuration mappings), `.rstrip("/")` leaves trailing quote characters intact (e.g., `"http://localhost:13305"` becomes `http://localhost:13305"`).

This causes suffix checks (`base.lower().endswith(suffix.lower())`) to fail and results in invalid host URL formatting when initializing HTTP clients.

## Fix

Update `normalize_base_url()` to strip matching surrounding quote pairs (`"..."` or `'...'`) before trimming trailing slashes:

```python
def normalize_base_url(base_url: str, api_base_path: str = DEFAULT_API_BASE_PATH) -> str:
    """Return a Lemonade host base URL without an API suffix."""
    base = (base_url or DEFAULT_BASE_URL).strip()
    if len(base) >= 2 and base[0] == base[-1] and base[0] in {"'", '"'}:
        base = base[1:-1].strip()
    base = (base or DEFAULT_BASE_URL).rstrip("/")
    api_path = _clean_path(api_base_path).rstrip("/")
    for suffix in (api_path, "/api/v1", "/v1"):
        if suffix and base.lower().endswith(suffix.lower()):
            return base[: -len(suffix)].rstrip("/") or DEFAULT_BASE_URL
    return base
```

This is a URL normalization robustness fix. It ensures that quoted host base URLs read from environment configuration round-trip cleanly without breaking HTTP client endpoint construction.

## Verification

Added unit test in `tests/test_lemonade_client.py`:

- `test_normalize_base_url_strips_api_suffixes`
  - Verified `normalize_base_url('"http://localhost:13305/api/v1"')` strips quotes and returns `"http://localhost:13305"`.

- `pytest tests/test_lemonade_client.py` passed (9 passed in 0.14s).

### DIFF
--- a/ods/extensions/services/dashboard-api/lemonade_client.py
+++ b/ods/extensions/services/dashboard-api/lemonade_client.py
@@ -22,7 +22,10 @@ def _clean_path(path: str) -> str:
 
 def normalize_base_url(base_url: str, api_base_path: str = DEFAULT_API_BASE_PATH) -> str:
     """Return a Lemonade host base URL without an API suffix."""
-    base = (base_url or DEFAULT_BASE_URL).strip().rstrip("/")
+    base = (base_url or DEFAULT_BASE_URL).strip()
+    if len(base) >= 2 and base[0] == base[-1] and base[0] in {"'", '"'}:
+        base = base[1:-1].strip()
+    base = (base or DEFAULT_BASE_URL).rstrip("/")
     api_path = _clean_path(api_base_path).rstrip("/")
     for suffix in (api_path, "/api/v1", "/v1"):
         if suffix and base.lower().endswith(suffix.lower()):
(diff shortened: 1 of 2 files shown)


Lightheartdevs:
Thanks for cleaning up the Lemonade URL normalization path for quoted config values.

What looks good:
- `ods/extensions/services/dashboard-api/lemonade_client.py` strips only matched surrounding quotes before trailing-slash and API-suffix normalization, so existing unquoted URLs keep the same behavior.
- The fallback to `DEFAULT_BASE_URL` still runs after quote stripping, which keeps empty quoted values from becoming an empty base URL.
- `ods/extensions/services/dashboard-api/tests/test_lemonade_client.py` extends the existing suffix test to cover the quoted `/api/v1` case that would previously leave an invalid trailing quote.

What I checked:
- Reviewed the PR title/description, changed files, and full patch for #2027.
- Compared the change against the repo's Lemonade/Windows configuration direction on recent `origin/main` commits.
- GitHub checks are green, including Python lint, mypy, API tests, smoke, distro, env validation, PowerShell lint, and secret scan jobs.

PR #2029, opened 2026-07-23
fix(dashboard): make the GPU topology legend match the matrix colours

tang-vu:
﻿## Problem

`TopologyView` colours every interconnect cell in the matrix with `linkStyle(link.rank)`. The legend underneath hard-coded its own rank per label — and those numbers disagreed with the ranks `installers/lib/nvidia-topo.sh` / `amd-topo.sh` actually emit:

```
link type   real rank   matrix colour     legend rank   legend swatch
NVLink        100        green              100          green   ✓
PIX            50        yellow              60          indigo  ✗
PXB            40        yellow              40          yellow  ✓
PHB            30        orange              20          orange  (right colour, wrong rank)
SYS            10        red                  5          red     (right colour, wrong rank)
```

So for a PIX link the legend swatch was **indigo** while the grid above it drew the cell **yellow** — the key contradicted the thing it explains. An operator reading the legend to interpret the matrix gets PIX wrong.

(`nvidia-topo.sh link_rank`: `PIX=50 PXB=40 PHB=30 SYS=10`; `PIX=60` was never a value the backend produces.)

## Fix

The legend now derives each swatch from the real emitted rank through the **same** `linkStyle()` the matrix uses, via a shared `TOPOLOGY_LEGEND` table, so the two can't drift apart again.

Colours and thresholds are untouched — this only corrects the legend's rank constants to the values the backend emits. (PIX and PXB still share the yellow band; that's a pre-existing property of the 5-colour scale, not changed here.)

## Verification

`TopologyView` had no test. `tests/TopologyView.legend.test.jsx` adds 3 cases, **all red before the fix**:

```
× colours each legend entry with its real emitted rank
× does not claim PIX is a high-bandwidth (indigo) link
× renders the legend swatch in the same colour the matrix uses for that link
```

The third renders the real component with a PIX interconnect and asserts the legend PIX swatch dot carries the same colour class as the matrix PIX cell.

Dashboard suite green (221), lint clean, production build clean.

## Why this matters

Operators read the topology legend to interpret the GPU matrix. The legend must use the same rank-to-colour function as the rendered cells or it explains the visualization incorrectly.

## Overlap check

Triage refresh (2026-08-08): searched open and closed Osmantic/ODS PRs by semantic keywords (topology legend colours), inspected the changed production files, and checked patch equivalence against current upstream/main. No strict duplicate or merged equivalent was found.

## Test plan

- [x] Dashboard component test: `ods/extensions/services/dashboard/src/components/__tests__/TopologyView.legend.test.jsx`


Lightheartdevs:
Thanks for tightening the topology legend so it matches the backend ranks operators actually see in the matrix.

What looks good:
- `ods/extensions/services/dashboard/src/components/TopologyView.jsx` now drives the legend through the same exported `linkStyle()` path as matrix cells, which removes the PIX/PXB/PHB/SYS drift without changing the existing colour thresholds.
- The new `TOPOLOGY_LEGEND` ranks line up with the emitted topology ranks described in the PR, and keeping that table next to `linkStyle()` makes future visual drift easier to catch.
- `ods/extensions/services/dashboard/src/components/__tests__/TopologyView.legend.test.jsx` covers both the constants and the rendered PIX case, including the exact mismatch this fixes.

What I checked:
- Reviewed the PR title/description, changed files, and full patch for #2029.
- Checked repo context around the dashboard/frontend surface and recent validation priorities on `origin/main`.
- GitHub checks are green across dashboard frontend, API, lint, secret scan, smoke, and distro/env validation jobs.

PR #2032, opened 2026-07-23
fix(helpers): validate finite float values in prometheus metrics parser

vaibhavsrv:
## Problem

In `get_llama_metrics()` (`helpers.py`), Prometheus metrics lines are parsed by splitting lines and converting values using `float(parts[-1])`:

```python
if metric_name.endswith("tokens_predicted_total"):
    try:
        metrics["tokens_predicted_total"] = float(parts[-1])
    except ValueError:
        pass
```

When Prometheus endpoints emit non-finite values such as `NaN` or `Inf` (e.g., when counters reset or model metrics uninitialize), `float("NaN")` and `float("Inf")` parse without raising `ValueError`. Non-finite floats then propagate into token throughput calculations and arithmetic comparison operations.

## Fix

Validate `math.isfinite(val)` prior to accepting parsed Prometheus float values:

```python
if metric_name.endswith("tokens_predicted_total"):
    try:
        val = float(parts[-1])
        if math.isfinite(val):
            metrics["tokens_predicted_total"] = val
    except ValueError:
        pass
if metric_name.endswith("tokens_predicted_seconds_total"):
    try:
        val = float(parts[-1])
        if math.isfinite(val):
            metrics["tokens_predicted_seconds_total"] = val
    except ValueError:
        pass
```

This is a telemetry parser validation fix. It guarantees that invalid non-finite metric values (`NaN`/`Inf`) emitted by upstream inference services do not pollute dashboard metrics.

## Verification

Added unit test in `tests/test_helpers.py`:

- `test_get_llama_metrics_ignores_nan_and_inf_prometheus_values`
  - Verified `NaN` and `Inf` metric lines are safely skipped and cause fallback handling.

- `pytest tests/test_helpers.py` passed (102 passed in 0.35s).

### DIFF
--- a/ods/extensions/services/dashboard-api/helpers.py
+++ b/ods/extensions/services/dashboard-api/helpers.py
@@ -458,12 +458,16 @@ async def get_llama_metrics(model_hint: Optional[str] = None) -> dict:
             metric_name = parts[0].split("{", 1)[0]
             if metric_name.endswith("tokens_predicted_total"):
                 try:
-                    metrics["tokens_predicted_total"] = float(parts[-1])
+                    val = float(parts[-1])
+                    if math.isfinite(val):
+                        metrics["tokens_predicted_total"] = val
                 except ValueError:
                     pass
             if metric_name.endswith("tokens_predicted_seconds_total"):
                 try:
-                    metrics["tokens_predicted_seconds_total"] = float(parts[-1])
+                    val = float(parts[-1])
+                    if math.isfinite(val):
+                        metrics["tokens_predicted_seconds_total"] = val
                 except ValueError:
                     pass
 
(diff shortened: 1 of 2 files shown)


Lightheartdevs:
Thanks for hardening the Prometheus metrics parser; this is right in the dashboard's runtime-health risk area.

What looks good:
- `ods/extensions/services/dashboard-api/helpers.py` now rejects `NaN` and `Inf` before storing Prometheus token counters, matching the module's existing `math.isfinite()` pattern in `_non_negative_number()` and TPS validation.
- The change keeps incomplete or invalid metrics on the existing fallback path, so bogus upstream telemetry cannot pollute dashboard throughput or lifetime-token calculations.
- `ods/extensions/services/dashboard-api/tests/test_helpers.py` covers both non-finite metric fields and verifies the cumulative fallback behavior.

What I checked:
- Static review of the PR diff and surrounding `get_llama_metrics()` parsing/fallback logic.
- Local test: `python -m pytest tests/test_helpers.py -q` from `ods/extensions/services/dashboard-api` passed with 100 passed, 2 skipped.

PR #2035, opened 2026-07-23
fix(usage): capture routed model telemetry

gabsprogrammer:
## Summary

The Settings Account card can report `0` tokens, requests, and models even while Open WebUI, Hermes, or another ODS client is actively generating. Token Spy only observes traffic sent through its proxy endpoints, while the supported ODS data plane normally routes through LiteLLM and, when the switchboard is enabled, Model Router.

This PR adds an authenticated, metadata-only usage path for both routing modes:

- `observe`: LiteLLM reports successful completion metadata to Token Spy.
- `enabled`: Model Router reports the completion and LiteLLM skips it, preserving exactly-once accounting.
- direct Model Router clients are covered.
- non-streaming, SSE streaming, Chat Completions, Completions, and Responses usage shapes are normalized.

## Data-flow contract

```text
observe: client -> LiteLLM -> native/cloud runtime
                     | success metadata
                     v
                  Token Spy

enabled: client -> LiteLLM -> Model Router -> active runtime
                                   | success metadata
                                   v
                                Token Spy
```

### Invariants

- Inference never waits for Token Spy.
- A successful routed completion produces at most one usage event.
- Failed and incomplete streams are not counted as completed requests.
- Prompts, messages, generated content, tool definitions/names, and credentials are never transmitted or stored.
- The ingest endpoint accepts only an allowlisted, bounded metadata schema and requires `TOKEN_SPY_API_KEY`.
- Local runtimes (`llama-server`, Lemonade, HipFire, MLX, Ollama, vLLM) remain zero-cost local usage.
- Lemonade's optional `extra.` namespace is normalized so one physical model is not counted twice.

## Implementation

- adds `POST /api/ingest/routed` to Token Spy with bearer authentication, a 16 KiB request limit, strict field rejection, type/range bounds, and existing database-schema reuse;
- adds bounded asynchronous telemetry queues to LiteLLM and Model Router;
- preserves existing LiteLLM callbacks and Langfuse configuration while generating `/tmp/config.yaml`;
- requires an explicit terminal signal (`[DONE]`, non-null `finish_reason`, or `response.completed`) before streaming usage and route evidence are committed;
- rate-limits delivery warnings and drops telemetry rather than delaying or failing model responses;
- documents the supported routed telemetry contract.

## Compatibility and failure behavior

- no database migration and no public host port;
- no installer-specific filesystem path; Linux, Windows, and macOS use the same Compose service contract;
- existing installs receive the path on normal in-place recreate;
- missing key disables emission;
- disabled, unavailable, slow, or rejecting Token Spy does not affect inference;
- cloud usage without authoritative billing metadata is recorded as `untracked`, not assigned an invented cost.

### Automated

- `python -m pytest ods/extensions/services/model-router/tests ods/extensions/services/token-spy/tests ods/extensions/services/litellm/tests ods/extensions/services/dashboard-api/tests/test_usage.py -q` -> **92 passed**
- Ruff on all touched Python and tests -> passed
- `python -m py_compile` on touched runtime modules -> passed
- `git diff --check osmantic/main...HEAD` -> passed
- real `docker compose config --quiet` renders -> passed for Linux/CPU, Windows AMD local, and macOS overlays

### Real Windows AMD runtime receipt

Using the local Lemonade-backed ODS stack:

- one non-stream LiteLLM completion recorded exactly **1 request / 22 tokens**;
- one streaming completion recorded exactly **1 request / 15 tokens**;
- one direct Model Router completion recorded exactly once;
- switchboard `enabled` recorded one event, not both LiteLLM and Router events;
- stopping Token Spy still returned a successful model completion in about 2.2 seconds;
- unauthenticated ingest returned `401`;
- a payload containing a forbidden `prompt` field returned `422`;
- the Settings Account card updated from zero to live token, request, model, and activity values.

## Residual limitations

- Delivery is intentionally best-effort and in-memory; a Token Spy outage can lose telemetry, but cannot interrupt inference.
- The SQLite runtime was exercised end-to-end. PostgreSQL uses the existing `log_usage` contract and is covered structurally, but was not exercised against a live PostgreSQL service in this Windows receipt.
- Physical Linux and macOS inference were not run locally; their Compose renders and shared Python paths are covered.

## Independent final review

The branch was updated to the latest `main` and reviewed again without assuming the original implementation was correct. Minimal break attempts covered truncated EOF, `[DONE]`-only streams, non-null `finish_reason`, `response.completed`, `response.failed`, malformed and oversized ingest payloads, missing/invalid authentication, disabled telemetry, queue delivery failure, switchboard observe/enabled deduplication, and local-model identity normalization.

One reproducible defect was found and fixed: a 2xx SSE connection that ended normally without any terminal model event was previously counted as a completed request. Model Router now requires a protocol-level terminal signal before committing telemetry or activation evidence. Regression tests cover truncated Chat Completions and terminal Responses API streams. No additional reproducible defect remains inside this PR's scope; no speculative subsystem changes were added.

### DIFF
--- a/ods/docker-compose.base.yml
+++ b/ods/docker-compose.base.yml
@@ -316,6 +316,8 @@ services:
       - ODS_ROUTER_INTERNAL_KEY=${ODS_ROUTER_INTERNAL_KEY:-}
       - DASHBOARD_API_KEY=${DASHBOARD_API_KEY:-}
       - ODS_FLEET_PROBE_KEY=${ODS_FLEET_PROBE_KEY:-}
+      - TOKEN_SPY_URL=${TOKEN_SPY_URL:-http://token-spy:8080}
+      - TOKEN_SPY_API_KEY=${TOKEN_SPY_API_KEY:-}
       # File-based probe key inside the already-mounted state volume: the
       # fleet harness can create/rotate/remove it without recreating the
       # router (a recreate destroys in-memory route evidence). Absent file
--- a/ods/extensions/services/litellm/README.md
+++ b/ods/extensions/services/litellm/README.md
@@ -18,6 +18,8 @@ ODS does not expose LiteLLM as a normal dashboard quicklink. The upstream LiteLL
 - **Multi-provider routing**: Anthropic, OpenAI, Together AI, MiniMax, and local llama-server
 - **Master key auth**: Secure all requests with `LITELLM_KEY`
 - **Drop params**: Unsupported parameters silently ignored across backends
+- **Usage telemetry**: Reports metadata-only completion usage to Token Spy
+  asynchronously when that recommended service is available
 
 ## Operating Modes
 
@@ -72,6 +74,8 @@ Environment variables (set in `.env`):
 | `LITELLM_KEY` | *(required)* | Master API key — generate with `echo "sk-ods-$(openssl rand -hex 16)"` |
 | `LITELLM_PORT` | `4000` | External + internal port |
 | `ODS_MODE` | `local` | Operating mode: `local`, `cloud`, or `hybrid` |
+| `TOKEN_SPY_URL` | `http://token-spy:8080` | Optional internal usage ingest target |
+| `TOKEN_SPY_API_KEY` | *(generated)* | Shared key used for authenticated usage ingest |
 | `ANTHROPIC_API_KEY` | *(empty)* | Required for `cloud` and `hybrid` modes |
 | `OPENAI_API_KEY` | *(empty)* | Required for OpenAI models in `cloud` mode |
 | `TOGETHER_API_KEY` | *(empty)* | Required for Together AI models in `cloud` mode |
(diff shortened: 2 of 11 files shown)


Lightheartdevs:
Thanks for taking on the routed usage telemetry path here; this closes a real observability gap without moving prompts or response content into Token Spy.

What looks good:
- The telemetry boundary fits the repo's architecture: LiteLLM reports in `observe`, Model Router reports in `enabled`, and both send to Token Spy through the existing internal service/auth pattern rather than adding a public port or a new persistence surface.
- The failure behavior is appropriately defensive for the inference path. Model Router and LiteLLM enqueue telemetry best-effort, drop or warn on delivery trouble, and the tests cover that telemetry failures do not change model responses.
- The ingest contract is narrow and privacy-conscious: `routed_telemetry.py` rejects unknown fields such as prompt/content/authorization, bounds counters and strings, normalizes local backends as zero-cost, and reuses the existing Token Spy usage schema.

What I checked:
- Reviewed the final diff against current `main`, including `ods/docker-compose.base.yml`, `ods/extensions/services/litellm/compose.yaml`, the LiteLLM callback, Model Router stream/non-stream telemetry, Token Spy ingest validation, and the new tests/docs.
- Ran focused local tests: `python -m pytest ods/extensions/services/model-router/tests/test_router.py ods/extensions/services/token-spy/tests/test_routed_telemetry.py ods/extensions/services/litellm/tests/test_token_spy_callback.py -q` -> 61 passed.
- Checked GitHub CI for this head: frontend, dashboard API, Ruff, mypy, shellcheck, compose validation, Linux integration-smoke, matrix smoke, secret scan, and env validation are green.

PR #2037, opened 2026-07-24
fix(agents): guard null last_update timestamp in agent metrics html endpoint

vaibhavsrv:
## Problem

In `get_agent_metrics_html()` (`routers/agents.py`), agent timestamp formatting evaluated `last_update = agent.get("last_update", "")` and split on `"T"`:

```python
last_update = agent.get("last_update", "")
last_update_time = last_update.split("T")[1][:8] if "T" in last_update else "N/A"
```

If `agent_metrics.last_update` was `None` or non-datetime, `AgentMetrics.to_dict()` raised `AttributeError: 'NoneType' object has no attribute 'isoformat'`. Furthermore, if `agent.get("last_update")` returned `None`, `"T" in last_update` raised `TypeError: argument of type 'NoneType' is not iterable`, causing 500 Internal Server Error when fetching `/api/agents/metrics.html`.

## Fix

Update `AgentMetrics.to_dict()` and `get_agent_metrics_html()` to safely handle `None` or non-string timestamp values:

```python
# agent_monitor.py
"last_update": self.last_update.isoformat() if hasattr(self.last_update, "isoformat") else str(self.last_update or "")

# routers/agents.py
last_update = agent.get("last_update") or ""
if not isinstance(last_update, str):
    last_update = str(last_update)
last_update_time = (
    last_update.split("T")[1][:8]
    if "T" in last_update and len(last_update.split("T")) > 1
    else "N/A"
)
```

This is a robust HTML fragment formatting fix. It guarantees that uninitialized or missing agent timestamps render `"Updated: N/A"` safely without raising 500 server errors.

## Verification

Added unit test in `tests/test_agents.py`:

- `test_handles_null_last_update_timestamp`
  - Verified `/api/agents/metrics.html` renders `"Updated: N/A"` cleanly when `last_update` is `None`.

- `pytest tests/test_agents.py` passed (7 passed in 0.64s).

### DIFF
--- a/ods/extensions/services/dashboard-api/agent_monitor.py
+++ b/ods/extensions/services/dashboard-api/agent_monitor.py
@@ -34,7 +34,7 @@ class AgentMetrics:
             "tokens_per_second": round(self.tokens_per_second, 2),
             "error_rate_1h": round(self.error_rate_1h, 2),
             "queue_depth": self.queue_depth,
-            "last_update": self.last_update.isoformat()
+            "last_update": self.last_update.isoformat() if hasattr(self.last_update, "isoformat") else str(self.last_update or "")
         }
 
 
--- a/ods/extensions/services/dashboard-api/routers/agents.py
+++ b/ods/extensions/services/dashboard-api/routers/agents.py
@@ -27,8 +27,14 @@ async def get_agent_metrics_html(api_key: str = Depends(verify_api_key)):
 
     cluster_class = "status-ok" if cluster.get("failover_ready") else "status-warn"
     failover_text = "Ready \u2705" if cluster.get("failover_ready") else "Single GPU \u26a0\ufe0f"
-    last_update = agent.get("last_update", "")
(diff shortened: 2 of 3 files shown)


Lightheartdevs:
Thanks for making the agent metrics HTML path tolerate a missing timestamp; that keeps the dashboard fragment from failing when monitor state is only partially initialized.

What looks good:
- `ods/extensions/services/dashboard-api/agent_monitor.py` now serializes absent or non-datetime `last_update` values without raising before the router sees them.
- `ods/extensions/services/dashboard-api/routers/agents.py` normalizes non-string timestamps before formatting, preserving the existing `Updated: N/A` behavior for missing data.
- `ods/extensions/services/dashboard-api/tests/test_agents.py` covers the null timestamp path through the authenticated HTML endpoint.

What I checked:
- Reviewed the PR diff and nearby agent metrics HTML escaping/formatting code.

PR #2038, opened 2026-07-24
fix(updates): strip matched quote pairs and allow leading whitespace in version reader

vaibhavsrv:
## Problem

In `_read_current_version()` (`routers/updates.py`), `.env` version parsing checked line equality using `line.startswith("ODS_VERSION=")` and stripped quotes via `.strip().strip("\"'")`:

```python
for line in _read_utf8(env_file).splitlines():
    if line.startswith("ODS_VERSION="):
        return line.split("=", 1)[1].strip().strip("\"'")
```

This pattern suffered from two bugs:
1. Lines with leading whitespace (e.g., `  ODS_VERSION=2.0.0`) failed `.startswith()` checks and were skipped.
2. `.strip().strip("\"'")` indiscriminately stripped internal quotes from version values (e.g. `ODS_VERSION="'2.1.0'"` lost its inner quotes).

## Fix

Update `_read_current_version()` to ignore comment lines, strip leading whitespace before checking prefix assignment, strip matching surrounding quote pairs (`"..."` or `'...'`), and catch `UnicodeError`:

```python
def _read_current_version() -> str:
    """Read installed version from .env (preferred) or .version file."""
    env_file = Path(INSTALL_DIR) / ".env"
    if env_file.is_file():
        try:
            for line in _read_utf8(env_file).splitlines():
                stripped = line.strip()
                    continue
                if stripped.startswith("ODS_VERSION="):
                        return val[1:-1]
                    return val
            pass
```

This is a version reading robustness fix. It guarantees that version strings read from `.env` or `.version` files handle leading whitespace and preserve internal quotes accurately.

## Verification

Added unit test in `tests/test_updates.py`:

- `test_read_current_version_handles_leading_whitespace_and_matched_quotes`
  - Verified values with leading whitespace and inner single quotes preserve inner quotes correctly.

- `pytest tests/test_updates.py` passed (28 passed in 1.13s).

### DIFF
--- a/ods/extensions/services/dashboard-api/routers/updates.py
+++ b/ods/extensions/services/dashboard-api/routers/updates.py
@@ -44,24 +44,32 @@ def _read_utf8(path: Path) -> str:
 def _read_current_version() -> str:
     """Read installed version from .env (preferred) or .version file."""
     env_file = Path(INSTALL_DIR) / ".env"
-    if env_file.exists():
+    if env_file.is_file():
         try:
             for line in _read_utf8(env_file).splitlines():
-                if line.startswith("ODS_VERSION="):
-                    return line.split("=", 1)[1].strip().strip("\"'")
-        except OSError:
+                stripped = line.strip()
+                if not stripped or stripped.startswith("#"):
+                    continue
+                if stripped.startswith("ODS_VERSION="):
+                    val = stripped.split("=", 1)[1].strip()
+                    if len(val) >= 2 and val[0] == val[-1] and val[0] in {"'", '"'}:
+                        return val[1:-1]
+                    return val
+        except (OSError, UnicodeError):
             pass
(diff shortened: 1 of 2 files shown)


Lightheartdevs:
Thanks for hardening the installed-version reader; update checks are a user-facing path where forgiving config parsing is helpful.

What looks good:
- `ods/extensions/services/dashboard-api/routers/updates.py` now handles leading whitespace, comments, file-vs-directory checks, matched outer quotes, and Unicode decode failures for `.env` and `.version` reads.
- The fallback order remains the same: `.env`, then `.version`, then manifest handling.
- `ods/extensions/services/dashboard-api/tests/test_updates.py` adds focused coverage for leading whitespace and matched quote handling.

What I checked:
- Reviewed the PR diff and surrounding update router version-resolution code.

PR #2039, opened 2026-07-24
fix(usage): read Token Spy API keys safely without changing them

vaibhavsrv:
## Problem

In `_token_spy_api_key()` (`routers/usage.py`), Token Spy API key parsing checked `if TOKEN_SPY_API_KEY:` and read `TOKEN_SPY_KEY_FILE`:

```python
def _token_spy_api_key() -> str:
    if TOKEN_SPY_API_KEY:
        return TOKEN_SPY_API_KEY
    try:
        return TOKEN_SPY_KEY_FILE.read_text(encoding="utf-8").strip()
    except OSError:
        return ""
```

This pattern suffered from three edge-case bugs:
1. `if TOKEN_SPY_API_KEY:` evaluated to `True` for whitespace-only strings (e.g. `"   "`), returning whitespace instead of reading the key file fallback.
2. Quoted API keys (e.g. `TOKEN_SPY_API_KEY="my-secret-key"`) retained surrounding quote characters in HTTP headers.
3. If `TOKEN_SPY_KEY_FILE` raised `UnicodeError` when reading non-UTF8 bytes, it escaped `except OSError:` and raised an uncaught 500 error.

## Fix

Update `_token_spy_api_key()` to unquote environment keys and key file contents, verify `TOKEN_SPY_KEY_FILE.is_file()`, and handle `UnicodeError`:

```python
def _token_spy_api_key() -> str:
    raw_key = (TOKEN_SPY_API_KEY or "").strip()
    if len(raw_key) >= 2 and raw_key[0] == raw_key[-1] and raw_key[0] in {"'", '"'}:
        raw_key = raw_key[1:-1].strip()
    if raw_key:
        return raw_key
    try:
        if TOKEN_SPY_KEY_FILE.is_file():
            val = TOKEN_SPY_KEY_FILE.read_text(encoding="utf-8").strip()
                return val[1:-1].strip()
            return val
        pass
    return ""
```

This is a telemetry authentication key reader fix. It ensures that Token Spy API keys read from `.env` or shared key files are clean, unquoted, and handle unreadable/binary key files fail-soft.

## Verification

Added unit test in `tests/test_usage.py`:

- `test_token_spy_api_key_unquotes_and_handles_unicode_error`
  - Verified quoted keys are unquoted properly and non-UTF8 key files return empty string fallback.

- `pytest tests/test_usage.py` passed (21 passed in 0.70s).

### DIFF
--- a/ods/extensions/services/dashboard-api/routers/usage.py
+++ b/ods/extensions/services/dashboard-api/routers/usage.py
@@ -330,12 +330,20 @@ async def usage_readiness(api_key: str = Depends(verify_api_key)):
 
 
 def _token_spy_api_key() -> str:
-    if TOKEN_SPY_API_KEY:
-        return TOKEN_SPY_API_KEY
+    raw_key = (TOKEN_SPY_API_KEY or "").strip()
+    if len(raw_key) >= 2 and raw_key[0] == raw_key[-1] and raw_key[0] in {"'", '"'}:
+        raw_key = raw_key[1:-1].strip()
+    if raw_key:
+        return raw_key
     try:
-        return TOKEN_SPY_KEY_FILE.read_text(encoding="utf-8").strip()
-    except OSError:
-        return ""
+        if TOKEN_SPY_KEY_FILE.is_file():
+            val = TOKEN_SPY_KEY_FILE.read_text(encoding="utf-8").strip()
+            if len(val) >= 2 and val[0] == val[-1] and val[0] in {"'", '"'}:
+                return val[1:-1].strip()
+            return val
+    except (OSError, UnicodeError):
+        pass
+    return ""
 
 
 def _configured_local_runtime_metrics_urls() -> list[str]:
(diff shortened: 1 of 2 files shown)


Lightheartdevs:
Thanks for cleaning up Token Spy key normalization in the usage router; quoted env/file values are an easy way for auth between bundled services to fail mysteriously.

What looks good:
- `ods/extensions/services/dashboard-api/routers/usage.py` now trims whitespace, removes only matched outer quotes, falls back past blank env values, and treats unreadable or non-UTF8 key files as absent.
- The change keeps the logic local to `_token_spy_api_key()` and does not alter the Token Spy routing or metrics paths.
- `ods/extensions/services/dashboard-api/tests/test_usage.py` covers quoted env values and Unicode decode failure on the key file fallback.

What I checked:
- Reviewed the PR diff and existing usage router key-file fallback test context.

Lightheartdevs:
Maintainer audit found that stripping matched quote characters only in dashboard-api could make its bearer token differ from Token Spy's exact secret comparison. I pushed a follow-up that keeps API keys opaque, trims only surrounding whitespace, retains the regular-file and Unicode-error hardening, and adds regression coverage for exact env/file secret preservation. Focused result: 21 passed. Waiting for the refreshed full CI before merge.

Lightheartdevs (approved):
Maintainer repair verified: Token Spy secrets remain opaque and identical between client and server, unreadable/non-file key paths fail closed, focused tests pass, and refreshed CI is fully green.

PR #2040, opened 2026-07-24
fix(templates): cleanup visiting set on dependency ordering exception

vaibhavsrv:
## Problem

In `_runtime_dependency_order()` (`routers/templates.py`), depth-first search dependency traversal adds visiting service IDs to `_visiting` before recursing into dependencies:

```python
_visiting.add(service_id)
for dep in read_direct_deps(service_id):
    _runtime_dependency_order(...)
_visiting.remove(service_id)
```

If `read_direct_deps()` or a nested dependency lookup raises an exception (such as `HTTPException` or `RuntimeError`), `_visiting.remove(service_id)` is bypassed. If the `_visiting` set is re-used or referenced by upper-level handlers, unremoved service IDs lead to false-positive `HTTPException(400, "Circular dependency detected")` errors on subsequent dependency resolutions.

## Fix

Wrap recursive dependency traversal in a `try...finally:` block to ensure `_visiting.remove(service_id)` is always called upon scope exit:

```python
_visiting.add(service_id)
try:
    for dep in read_direct_deps(service_id):
        _runtime_dependency_order(
            dep,
            read_direct_deps,
            _visiting=_visiting,
            _visited=_visited,
            _order=_order,
        )
        if dep not in _order:
            _order.append(dep)
finally:
    _visiting.remove(service_id)
_visited.add(service_id)
return _order
```

This is a dependency graph traversal safety fix. It guarantees state isolation for cycle detection sets during service template dependency resolution.

## Verification

Added unit test in `tests/test_templates.py`:

- `test_runtime_dependency_order_cleans_up_visiting_set_on_exception`
  - Verified that exceptions during dependency reading remove visited node IDs from the `_visiting` set.

- `pytest tests/test_templates.py` passed (32 passed in 0.54s).

### DIFF
--- a/ods/extensions/services/dashboard-api/routers/templates.py
+++ b/ods/extensions/services/dashboard-api/routers/templates.py
@@ -41,17 +41,19 @@ def _runtime_dependency_order(
         return _order
 
     _visiting.add(service_id)
-    for dep in read_direct_deps(service_id):
-        _runtime_dependency_order(
-            dep,
-            read_direct_deps,
-            _visiting=_visiting,
-            _visited=_visited,
-            _order=_order,
-        )
-        if dep not in _order:
-            _order.append(dep)
-    _visiting.remove(service_id)
+    try:
+        for dep in read_direct_deps(service_id):
+            _runtime_dependency_order(
+                dep,
+                read_direct_deps,
+                _visiting=_visiting,
+                _visited=_visited,
+                _order=_order,
+            )
+            if dep not in _order:
+                _order.append(dep)
+    finally:
+        _visiting.remove(service_id)
     _visited.add(service_id)
     return _order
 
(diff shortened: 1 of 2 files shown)


Lightheartdevs:
Thanks for tightening the template dependency traversal state cleanup; this is a good small fix for a subtle follow-on failure mode.

What looks good:
- `ods/extensions/services/dashboard-api/routers/templates.py` now removes the active node from `_visiting` in a `finally`, so exceptions during dependency reads do not poison later cycle checks.
- The fix preserves the existing `_visited` / `_order` behavior on successful traversal and only changes exceptional cleanup.
- `ods/extensions/services/dashboard-api/tests/test_templates.py` adds direct regression coverage for the leaked `_visiting` set case.

What I checked:
- Reviewed the dependency-order diff and surrounding template router tests.

PR #2043, opened 2026-07-24
fix(tailscale): fall back to offline payload when host agent is unreachable

vaibhavsrv:
## Problem

In `tailscale_status()` (`routers/tailscale.py`), Tailscale status endpoint resolution forwarded requests to `_proxy_agent("/v1/tailscale/status")`:

```python
@router.get("/api/tailscale/status", dependencies=[Depends(verify_api_key)])
async def tailscale_status() -> dict:
    return await asyncio.to_thread(_proxy_agent, "/v1/tailscale/status", 15)
```

If the host agent returned HTTP 503 (`AgentUnavailable`) or HTTP 504 (`AgentTimeout`) when the host daemon was down or running in standalone development mode, `_proxy_agent` raised `HTTPException(503)` or `HTTPException(504)`.

This violated the endpoint's API contract (which promises to always return HTTP 200 with `running: false` status rather than throwing HTTP errors for unconfigured or unreachable daemons).

## Fix

Catch `HTTPException` with status codes 503 or 504 in `tailscale_status()` and return an offline payload (`{"running": false, "authenticated": false, "detail": ...}`):

```python
@router.get("/api/tailscale/status", dependencies=[Depends(verify_api_key)])
async def tailscale_status() -> dict:
    try:
        return await asyncio.to_thread(_proxy_agent, "/v1/tailscale/status", 15)
    except HTTPException as exc:
        if exc.status_code in {503, 504}:
            return {"running": False, "authenticated": False, "detail": exc.detail}
        raise
```

This is an API contract compliance fix. It guarantees that the remote access settings section of the dashboard UI displays an offline state gracefully instead of throwing HTTP 503 errors when host agents are unavailable.

## Verification

Updated unit tests in `tests/test_tailscale.py`:

- `test_status_returns_200_offline_when_agent_unreachable`
- `test_status_returns_200_offline_when_agent_request_times_out`
  - Verified `HTTP 200` with `running=False` is returned when host agent calls time out or fail.

- `pytest tests/test_tailscale.py` passed (8 passed in 0.50s).

### DIFF
--- a/ods/extensions/services/dashboard-api/routers/tailscale.py
+++ b/ods/extensions/services/dashboard-api/routers/tailscale.py
@@ -68,4 +68,9 @@ async def tailscale_status() -> dict:
         dns_name, ips, online}, "magic_dns_suffix": "...", "tailnet_name": "..."}`
         — fully on the tailnet
     """
-    return await asyncio.to_thread(_proxy_agent, "/v1/tailscale/status", 15)
+    try:
+        return await asyncio.to_thread(_proxy_agent, "/v1/tailscale/status", 15)
+    except HTTPException as exc:
+        if exc.status_code in {503, 504}:
+            return {"running": False, "authenticated": False, "detail": exc.detail}
+        raise
--- a/ods/extensions/services/dashboard-api/tests/test_tailscale.py
+++ b/ods/extensions/services/dashboard-api/tests/test_tailscale.py
@@ -99,30 +99,34 @@ def test_status_when_fully_joined(test_client):
 # ---------------------------------------------------------------------------
 
 
-def test_status_returns_503_when_agent_unreachable(test_client):
+def test_status_returns_200_offline_when_agent_unreachable(test_client):
     with patch(
         "routers.tailscale.request_agent_json",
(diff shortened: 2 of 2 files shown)


Lightheartdevs:
Thanks for tightening the Tailscale status path so the dashboard can show an offline state instead of surfacing host-agent transport failures.

What looks good:
- `ods/extensions/services/dashboard-api/routers/tailscale.py` now matches the route docstring's "always 200" contract for unreachable or timed-out host-agent status calls while still re-raising other HTTP exceptions such as 500s.
- `ods/extensions/services/dashboard-api/tests/test_tailscale.py` updates the failure-mode expectations for `AgentUnavailable`, `AgentTimeout`, and a host-agent 504 payload, so the dashboard-facing contract is covered at the API boundary.
- The change stays inside the dashboard-api proxy layer and does not alter host-agent behavior, compose wiring, auth, or installer/runtime state.

What I checked:
- Reviewed the PR diff and surrounding `routers/tailscale.py` / `tests/test_tailscale.py` context against the dashboard-api boundary conventions.

PR #2045, opened 2026-07-24
fix(privacy): handle invalid non-integer port values safely in privacy shield endpoints

vaibhavsrv:
## Problem

In `get_privacy_shield_status()` and `get_privacy_shield_stats()` (`routers/privacy.py`), `SHIELD_PORT` configuration was read using `int(os.environ.get("SHIELD_PORT", ...))`:

```python
shield_port = int(os.environ.get("SHIELD_PORT", str(_ps.get("port", 0))))
```

If `SHIELD_PORT` was set to a non-integer string (e.g. `"auto"`) or contained surrounding quotes (e.g. `"\"8080\""`), `int(...)` raised an unhandled `ValueError`, crashing the Privacy Shield endpoints with HTTP 500.

## Fix

Introduce `_safe_port()` helper to unquote matched quote pairs and safely parse port integers with a fallback default:

```python
def _safe_port(val: any, default: int = 0) -> int:
    try:
        raw = str(val or "").strip()
            raw = raw[1:-1].strip()
        return int(raw)
        return default
```

This is a port configuration parsing safety fix. It guarantees that Privacy Shield status and statistics endpoints handle quoted or invalid port strings without raising unhandled internal server errors.

## Verification

Added unit test in `tests/test_privacy.py`:

- `test_privacy_shield_status_handles_invalid_and_quoted_port`
  - Verified quoted port strings (e.g. `"\"8080\""`) are parsed cleanly and invalid strings fall back to `0`.

- `pytest tests/test_privacy.py` passed (18 passed in 0.55s).

### DIFF
--- a/ods/extensions/services/dashboard-api/routers/privacy.py
+++ b/ods/extensions/services/dashboard-api/routers/privacy.py
@@ -23,11 +23,21 @@ logger = logging.getLogger(__name__)
 router = APIRouter(tags=["privacy"])
 
 
+def _safe_port(val: any, default: int = 0) -> int:
+    try:
+        raw = str(val or "").strip()
+        if len(raw) >= 2 and raw[0] == raw[-1] and raw[0] in {"'", '"'}:
+            raw = raw[1:-1].strip()
+        return int(raw)
+    except (ValueError, TypeError):
+        return default
+
+
 @router.get("/api/privacy-shield/status", response_model=PrivacyShieldStatus)
 async def get_privacy_shield_status(api_key: str = Depends(verify_api_key)):
     """Get Privacy Shield status and configuration."""
     _ps = SERVICES.get("privacy-shield", {})
-    shield_port = int(os.environ.get("SHIELD_PORT", str(_ps.get("port", 0))))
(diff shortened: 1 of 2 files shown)


Lightheartdevs:
Thanks for tightening the Privacy Shield port parsing path; quoted or malformed env values are exactly the kind of config edge that can make dashboard status routes fail harder than they should.

What looks good:
- `ods/extensions/services/dashboard-api/routers/privacy.py` now centralizes the `SHIELD_PORT` parsing for both status and stats, so invalid env values fall back instead of raising before the endpoint can return a dashboard-friendly response.
- The matched-quote handling is scoped to the outer quote pair and keeps the existing manifest/default port behavior when `SHIELD_PORT` is unset.
- `ods/extensions/services/dashboard-api/tests/test_privacy.py` adds focused coverage for quoted, invalid, and missing port values.

What I checked:
- Reviewed the PR diff and surrounding Privacy Shield status/stats routes against the dashboard-api service manifest conventions.
- Checked the current service manifest loader path: service ports are normalized to integers before entering `SERVICES`, so the helper fallback stays compatible with current registry data.

PR #2049, opened 2026-07-24
fix(model-state): handle unicode decoding errors safely when reading state schema and file

vaibhavsrv:
## Problem

In `get_model_state()` and `_validate_document()` (`routers/model_state.py`), model state reading loaded `model-state.json` and its JSON schema using `.read_text(encoding="utf-8")`:

```python
# _validate_document
except (OSError, ValueError, SchemaError) as exc:

# get_model_state
except OSError as exc:
    logger.warning("model-state read failed: %s", exc)
    return _invalid_response([f"read failed: {exc}"])
```

If `model-state.json` contained corrupted non-UTF8 binary data, `path.read_text()` raised `UnicodeDecodeError`. Because `UnicodeDecodeError` is not a subclass of `OSError`, it bypassed `except OSError:` in `get_model_state()`, raising an unhandled HTTP 500 error instead of returning a fail-soft diagnostic error payload.

## Fix

Add `UnicodeError` to the caught exception tuples in both `_validate_document()` and `get_model_state()`:

```python
# _validate_document
except (OSError, UnicodeError, ValueError, SchemaError) as exc:

# get_model_state
except (OSError, UnicodeError) as exc:
    logger.warning("model-state read failed: %s", exc)
    return _invalid_response([f"read failed: {exc}"])
```

This is a model state reader robustness fix. It guarantees that corrupted or non-UTF8 binary state files return sanitized fail-soft diagnostic responses rather than 500 internal server errors.

## Verification

Added unit test in `tests/test_model_state.py`:

- `test_get_model_state_handles_unicode_decode_error`
  - Verified non-UTF8 binary data files return `valid=False` with `"read failed"` diagnostics without crashing.

- `pytest tests/test_model_state.py` passed (30 passed in 0.91s).

### DIFF
--- a/ods/extensions/services/dashboard-api/routers/model_state.py
+++ b/ods/extensions/services/dashboard-api/routers/model_state.py
@@ -63,7 +63,7 @@ def _validate_document(doc: Any) -> list[str]:
         schema = json.loads(_schema_path().read_text(encoding="utf-8"))
         Draft202012Validator.check_schema(schema)
         validator = Draft202012Validator(schema)
-    except (OSError, ValueError, SchemaError) as exc:
+    except (OSError, UnicodeError, ValueError, SchemaError) as exc:
         return [f"state schema unavailable or invalid: {exc}"]
 
     errors = []
@@ -109,7 +109,7 @@ async def get_model_state(api_key: str = Depends(verify_api_key)):
         raw = path.read_text(encoding="utf-8")
     except FileNotFoundError:
         return _invalid_response([], exists=False)
-    except OSError as exc:
+    except (OSError, UnicodeError) as exc:
         logger.warning("model-state read failed: %s", exc)
         return _invalid_response([f"read failed: {exc}"])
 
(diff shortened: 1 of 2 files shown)


Lightheartdevs:
Thanks for tightening the model-state reader around corrupted local state files.

What looks good:
- `ods/extensions/services/dashboard-api/routers/model_state.py` now treats `UnicodeError` like the existing file/schema boundary failures, so non-UTF8 state or schema content returns a diagnostic payload instead of a dashboard-api 500.
- The route still preserves the existing missing-file and invalid-JSON behavior, and it keeps the endpoint read-only/sanitizing rather than attempting repair.
- `ods/extensions/services/dashboard-api/tests/test_model_state.py` adds focused coverage for a binary `model-state.json` failure.

What I checked:
- Reviewed the PR title/body, full patch, and the surrounding model-state endpoint contract from the PR branch.
- Checked current GitHub status for this head: Python lint/type check, dashboard API/frontend, shell/PowerShell lint, Linux integration smoke, matrix smoke, env validation, and secret scan are green.
- Confirmed there was no existing advisory marker for this head SHA.

PR #2050, opened 2026-07-24
fix(auth): strip matched quote pairs when reading cookie domain environment variable

vaibhavsrv:
## Problem

In `_cookie_domain()` (`routers/auth.py`), cookie domain resolution extracted `ODS_COOKIE_DOMAIN` using `(os.environ.get("ODS_COOKIE_DOMAIN") or "").strip()`:

```python
def _cookie_domain() -> Optional[str]:
    raw = (os.environ.get("ODS_COOKIE_DOMAIN") or "").strip()
    return raw or None
```

If `ODS_COOKIE_DOMAIN` was configured in `.env` with surrounding quotes (e.g. `ODS_COOKIE_DOMAIN="'.example.local'"` or `ODS_COOKIE_DOMAIN='".example.local"'`), the `Domain` attribute in the `Set-Cookie` header retained the outer quote characters, causing browsers to reject the `ods-session` cookie for invalid domain syntax.

## Fix

Update `_cookie_domain()` in `routers/auth.py` to strip matching surrounding quote pairs (`"..."` or `'...'`):

```python
def _cookie_domain() -> Optional[str]:
    """Cookie ``Domain`` attribute. Empty/None = host-only cookie."""
    raw = (os.environ.get("ODS_COOKIE_DOMAIN") or "").strip()
        raw = raw[1:-1].strip()
    return raw or None
```

This is an authentication cookie domain parsing fix. It guarantees that quotes around `ODS_COOKIE_DOMAIN` in configuration files do not corrupt `Set-Cookie` Domain headers or break cross-subdomain SSO.

## Verification

Updated unit tests in `tests/test_auth.py`:

- `test_cookie_domain_strips_matched_quotes`
  - Verified quoted domain strings (e.g. `"'example.local'"`) are unquoted cleanly.

- `pytest tests/test_auth.py` passed.

### DIFF
--- a/ods/extensions/services/dashboard-api/routers/auth.py
+++ b/ods/extensions/services/dashboard-api/routers/auth.py
@@ -48,17 +48,10 @@ SESSION_TTL_SECONDS = 12 * 3600
 
 
 def _cookie_domain() -> Optional[str]:
-    """Cookie ``Domain`` attribute. Empty/None = host-only cookie.
-
-    ODS_COOKIE_DOMAIN is set by the installer to ``<ODS_DEVICE_NAME>.local``
-    when ods-proxy + mDNS are wired so a single redemption authenticates
-    across chat.<name>.local, hermes.<name>.local, and the rest. With it
-    empty, the cookie is host-only — functional for single-host setups
-    but breaks subdomain SSO. Same logic as routers/magic_link.py;
-    duplicated here intentionally so the two cookie-issuing paths stay
-    coupled without one depending on the other.
-    """
+    """Cookie ``Domain`` attribute. Empty/None = host-only cookie."""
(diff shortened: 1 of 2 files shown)


Lightheartdevs:
Thanks for taking on the cookie-domain parsing edge case.

What looks good:
- `ods/extensions/services/dashboard-api/routers/auth.py` keeps host-only cookies as the empty/default behavior and only unquotes matched `ODS_COOKIE_DOMAIN` values before setting the cookie Domain attribute.
- `ods/extensions/services/dashboard-api/tests/test_auth.py` adds focused coverage for the unauthenticated verify-session path and quoted domain parsing.
- The change is compatible with the existing magic-link/admin-session SSO model for proxy subdomains.

What I checked:
- Static review of the auth router diff and surrounding session cookie mint/verify logic.

PR #2059, opened 2026-07-24
fix(installer): make multi-GPU setup pipeline-safe

gabsprogrammer:
## Summary

Makes the Linux multi-GPU setup safe when ODS is launched through the documented hosted bootstrap:

```bash
curl -fsSL https://install.osmantic.com/ods.sh | bash
```

- reads every interactive multi-GPU prompt from the controlling terminal
- guarantees a non-empty default llama-server assignment when auxiliary services consume every GPU
- converts custom-assignment cancellation into a valid automatic assignment instead of continuing with empty state
- adds behavioral regression coverage with inherited stdin forced to EOF

## Root cause

`get-ods.sh` is consumed through a pipe and eventually `exec`s the checked-out installer without restoring stdin. The initial install menu already reads from `/dev/tty`, but four later prompts in phase `03-features` inherited the exhausted curl pipeline:

1. automatic/custom mode selection
2. per-service GPU selection
3. llama-server GPU list
4. custom assignment confirmation

The first `read` returned non-zero and `set -e` aborted phase 03 before the user could respond. Detection, driver validation, and topology discovery had already succeeded.

## Confirmed edge cases fixed

The adversarial review reproduced two adjacent invalid-state paths in the same custom setup flow:

- assigning all GPUs to auxiliary services left the default llama-server list empty; pressing Enter produced a zero-GPU llama assignment
- declining the custom confirmation returned with an empty `GPU_ASSIGNMENT_JSON`, so downstream extraction failed instead of recovering

The first case now defaults llama-server to the detected GPU list. The second falls back to topology-aware automatic assignment.

## Invariants preserved

- interactive prompts use the controlling terminal; data transforms continue to use stdin/process substitution normally
- Enter on the mode prompt still selects automatic assignment
- every successful path produces at least one llama-server GPU
- cancellation cannot reach persistence with empty assignment state
- non-interactive assignment and valid persisted-assignment reuse are unchanged
- topology detection, tier selection, assignment policy, `.env` encoding, Compose overlays, and runtime consumers are unchanged

## Regression coverage

`tests/test-phase03-multigpu-tty.sh` uses util-linux `script` to provide a controlling PTY while launching the phase harness with `</dev/null>`, reproducing the hosted bootstrap contract. It covers a heterogeneous three-GPU topology and verifies:

- automatic selection with closed stdin
- empty mode selection retaining the automatic default
- complete custom assignment with closed stdin
- empty llama-server default when all auxiliary services have GPU selections
- custom cancellation recovering through automatic assignment

A portable static contract also requires all four interactive multi-GPU reads to use `/dev/tty`. The behavioral cases assert the selected path and exact service GPU UUIDs, use a bounded timeout to prevent a prompt regression from hanging CI, and avoid Bash features newer than the macOS system Bash.

### GitHub CI

All reported checks pass, including:

- Linux integration smoke and installer contracts
- ShellCheck
- Ubuntu 22.04/24.04, Debian 12, Mint 21.3, Fedora 41, Rocky 9, Arch, Manjaro, CachyOS, and openSUSE matrix
- macOS smoke
- PowerShell lint on Ubuntu and Windows
- dashboard API/frontend
- env-schema tiers, Python lint, and secret scan

The Linux installer-contract log confirms all five PTY behavioral cases pass.

### Focused local validation

- Bash parser on Linux installer phases, macOS installer scripts, and the new test
- PowerShell parser across the Windows installer tree
- JSON/YAML parsers for env schema, workflow, and Compose files
- Python compile for `assign_gpus.py`
- `python -m pytest tests/test-assign-gpus.py extensions/services/dashboard-api/tests/test_gpu_detailed.py -q` -> `116 passed`
- Bash 3.2 guard contracts -> `10 passed`
- hosted bootstrap verifier -> passed
- real `docker compose config --quiet` renders for NVIDIA multi-GPU, AMD multi-GPU, macOS, and CPU overlays
- `git diff --check`

## Independent adversarial review

A separate clean-room review found no reproducible production defect after the fixes. It did reproduce two defects in the first test revision: custom cases could pass if the selector incorrectly called automatic assignment, and `mapfile` made the portable contract incompatible with Bash 3.2. Both were corrected. A deliberate selector mutation now fails on the missing custom-path receipt and exact GPU mapping.
## Compatibility and scope

| Area | Result |
|---|---|
| Linux clean hosted install | Fixed and behaviorally tested |
| Linux direct installer | Same prompts, now explicitly terminal-bound |
| Update / second interactive run | Uses the same fixed terminal path |
| Non-interactive rerun / persisted assignment | Unchanged; official contracts pass |
| Disabled optional services | Compose gating unchanged |
| Partial custom cancellation | Recovers to valid automatic assignment |
| Windows | Separate PowerShell installer untouched; parser/lint pass |
| macOS | Separate installer untouched; parser/smoke/Compose render pass |
| Dashboard API/frontend | Consumers unchanged; tests pass |
| Documentation | Documented bootstrap command now matches implementation; no text change required |

No persistent format, public API, Compose schema, secret handling, or runtime contract changes are introduced.

### DIFF
--- a/.github/workflows/test-linux.yml
+++ b/.github/workflows/test-linux.yml
@@ -137,6 +137,7 @@ jobs:
       - name: Installer Contract Checks
         run: |
           bash tests/contracts/test-installer-contracts.sh
+          bash tests/test-phase03-multigpu-tty.sh
           bash tests/contracts/test-preflight-fixtures.sh
           bash tests/test-preflight-resolver-bug.sh
           python3 tests/contracts/test-network-exposure-contracts.py
--- a/ods/Makefile
+++ b/ods/Makefile
@@ -28,6 +28,7 @@ test: ## Run unit and contract tests
 	@echo ""
 	@echo "=== Installer contracts ==="
 	@bash tests/contracts/test-installer-contracts.sh
+	@bash tests/test-phase03-multigpu-tty.sh
 	@bash tests/contracts/test-preflight-fixtures.sh
 	@bash tests/test-preflight-resolver-bug.sh
 	@bash tests/test-linux-cloud-mode.sh
--- a/ods/installers/phases/03-features.sh
+++ b/ods/installers/phases/03-features.sh
@@ -360,7 +360,7 @@ run_custom() {
   for svc in whisper comfyui embeddings; do
     local valid=false
     while ! $valid; do
-      read -rp "  GPU for ${WHT}${svc}${NC} (0-$((GPU_COUNT-1))): " chosen
+      read -rp "  GPU for ${WHT}${svc}${NC} (0-$((GPU_COUNT-1))): " chosen < /dev/tty
       if [[ "$chosen" =~ ^[0-9]+$ ]] && [[ $chosen -ge 0 ]] && [[ $chosen -lt $GPU_COUNT ]]; then
         CUSTOM_ASSIGNMENT[$svc]=$chosen; valid=true
       else
@@ -378,8 +378,11 @@ run_custom() {
     $found || default_llama+="${idx},"
   done
   default_llama="${default_llama%,}"
+  if [[ -z "$default_llama" ]]; then
+    default_llama=$(IFS=,; echo "${GPU_INDICES[*]}")
+  fi
 
-  read -rp "  GPUs for ${WHT}llama_server${NC} [${default_llama}]: " llama_input
+  read -rp "  GPUs for ${WHT}llama_server${NC} [${default_llama}]: " llama_input < /dev/tty
   llama_input="${llama_input:-$default_llama}"
   IFS=',' read -ra LLAMA_GPUS_CUSTOM <<< "$llama_input"
   for g in "${LLAMA_GPUS_CUSTOM[@]}"; do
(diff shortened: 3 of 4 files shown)


Lightheartdevs:
Thanks for taking on the hosted-bootstrap multi-GPU prompt path here; this is exactly the kind of installer edge case that can make an otherwise-good hardware detection run fail at the last mile.

What looks good:
- `ods/installers/phases/03-features.sh` keeps the `/dev/tty` reads scoped to the interactive multi-GPU TUI, while non-interactive and dry-run installs still reuse persisted assignment or run automatic assignment.
- The custom path now closes the two bad state exits: an empty llama default falls back to all detected GPU indices, and a declined custom confirmation falls back to automatic assignment instead of leaving `GPU_ASSIGNMENT_JSON` empty.
- The new `ods/tests/test-phase03-multigpu-tty.sh` exercises the hosted `curl | bash` shape with stdin closed, and wires that regression into both `ods/Makefile` and `.github/workflows/test-linux.yml`.

What I checked:
- Reviewed the PR diff, surrounding phase-03 multi-GPU control flow, repo installer architecture docs, high-risk change map, and current `main` test expectations.
- `gh pr checks` is green across Linux integration smoke, ShellCheck, distro matrix, macOS smoke, PowerShell lint, dashboard/API, env validation, Python lint, and secret scan.
- `git diff --check origin/main...HEAD` passed locally. I could not run the new Bash regression locally in this Windows environment because `bash` resolves through WSL and `/bin/bash` is unavailable here.

PR #2062, opened 2026-07-24
fix(oracle): tolerate an explicit null aliases in a catalog entry

w3-surfer:
## Problem

`_model_aliases` expands a catalog entry's aliases with:

```python
aliases.update(str(alias or "") for alias in model.get("aliases", []))   # line 171
for part in model.get("gguf_parts", []) or []:                           # line 172
```

`.get("aliases", [])` only substitutes the default when the key is **absent**. An entry carrying an explicit JSON null (`"aliases": null`) yields `None`, so iterating raises:

```
TypeError: 'NoneType' object is not iterable
```

That propagates out of `normalize_catalog_entry` → `load_model_catalog`, which backs `routers/models.py` and `routers/talk.py`.

The **very next line** already guards the same shape with `or []`, so the two lines are simply inconsistent.

## Fix

One word, matching the adjacent line:

```diff
-    aliases.update(str(alias or "") for alias in model.get("aliases", []))
+    aliases.update(str(alias or "") for alias in model.get("aliases", []) or [])
```

Aliases are still derived from `id`/`name`/`gguf` when the field is null, and listed aliases keep working.

## Honest note on reachability

This is a robustness/consistency fix, not a report of a break in a shipped path. I checked before opening it:

- The shipped `config/model-library.json` (52 entries) does not use the `aliases` field at all.
- The HuggingFace import path (`routers/models.py`) does not write `aliases` either.

So reaching it today requires a hand-edited or fork-customized catalog that writes `"aliases": null` — which `docs/VALIDATION_REPRODUCIBILITY.md` explicitly contemplates ("changed model catalogs" for forks). Feel free to close if you'd rather not carry the guard.

## Verification

Ran `normalize_catalog_entry` directly against the current file and against the pre-fix line:

```
[fixed]  null_aliases=PASS          listed_aliases=PASS
[buggy]  null_aliases=FAIL(TypeError)  listed_aliases=PASS
```

Added two tests to `test_performance_oracle.py`:
- `test_normalize_catalog_entry_tolerates_explicit_null_aliases` — fails before this change, passes after.
- `test_normalize_catalog_entry_keeps_listed_aliases` — guards against the fix silently dropping real aliases.

### DIFF
--- a/ods/extensions/services/dashboard-api/performance_oracle.py
+++ b/ods/extensions/services/dashboard-api/performance_oracle.py
@@ -168,7 +168,7 @@ def _model_aliases(model: dict[str, Any]) -> set[str]:
         str(model.get("gguf_file") or ""),
         str(model.get("llm_model_name") or ""),
     }
-    aliases.update(str(alias or "") for alias in model.get("aliases", []))
+    aliases.update(str(alias or "") for alias in model.get("aliases", []) or [])
     for part in model.get("gguf_parts", []) or []:
         if isinstance(part, dict):
             aliases.add(str(part.get("file") or ""))
--- a/ods/extensions/services/dashboard-api/tests/test_performance_oracle.py
+++ b/ods/extensions/services/dashboard-api/tests/test_performance_oracle.py
@@ -11,6 +11,7 @@ from performance_oracle import (
     model_compatibility_runtime_context,
     model_app_compatibility,
     model_publisher,
+    normalize_catalog_entry,
     rank_pre_download_models,
 )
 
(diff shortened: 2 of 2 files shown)


Lightheartdevs:
Thanks for tightening the catalog normalization path here; this is a small fix, but it fits the project’s fork/custom-catalog posture well.

What looks good:
- `ods/extensions/services/dashboard-api/performance_oracle.py` now treats explicit `aliases: null` the same way the adjacent `gguf_parts` handling already treats null input, so `normalize_catalog_entry()` keeps deriving aliases from `id` and `gguf` instead of raising during catalog load.
- The added tests in `ods/extensions/services/dashboard-api/tests/test_performance_oracle.py` cover both the regression case and the non-regression case for listed aliases, which is the right scope for this one-line guard.
- This stays inside the performance oracle/catalog normalization boundary and does not change shipped catalog data, routing behavior, installer state, or compose/runtime contracts.

What I checked:
- Reviewed the PR title, description, changed files, patch, and existing discussion for #2062.
- Checked surrounding code in `_model_aliases()`, `normalize_catalog_entry()`, and the model/talk router catalog consumers.
- Ran `python -m pytest ods/extensions/services/dashboard-api/tests/test_performance_oracle.py -q` locally on `7f9c055b7592882dbf119d8665c6425187ebe763`: 40 passed.
- Confirmed GitHub checks are passing, including dashboard API tests, Python lint, type check, integration smoke, distro matrix, shell lint, secret scan, and env validation.

PR #2069, opened 2026-07-25
fix(hermes): read slash-worker pids from a known ps layout

tang-vu:
## Problem

`scripts/prune-hermes-slash-workers.sh` asks the container shell to choose its own process-table command:

```sh
ps -eo pid=,etimes=,args= 2>/dev/null || ps -ef 2>/dev/null
```

and then parses whatever comes back as `pid=$1, age=$2`. The two layouts do not line up — `ps -ef` leads with **UID**, so on the fallback branch column 1 is the UID and column 2 is the pid.

Two failure modes, depending on what the Hermes image's `ps` prints:

| image prints | result |
|---|---|
| UID **name** (`root`) | column 1 fails the numeric-pid check, every worker row is dropped → `[PASS] no Hermes slash workers found` while the leak this tool exists to clear keeps growing |
| UID **number** (uid with no passwd entry) | the uid is read as the pid and the real pid is read as an age → `--force` signals an unrelated process and the workers stay up |

Both are silent: `ods repair hermes-workers` reports success either way.

A second, smaller defect: the "container is not running" notice was written to **stdout**, which is the worker stream. It was fed straight into the parser and discarded, so a stopped container also printed `no Hermes slash workers found`.

## Fix

- Request each layout in its own `docker exec` so the caller knows which columns came back, and normalise to `pid / age / command` in one place.
- The POSIX fallback is now `ps -eo pid=,args=` — pid stays in column 1, and the age is recorded as unknown (`-1`) instead of being invented from whatever sits in column 2.
- Unknown ages never satisfy the max-age policy on their own. When a count overage still has to pick victims and no ages are available, the tie breaks on the lowest pid (longest-lived worker) rather than on sort order.
- The stopped-container notice goes to stderr.

Behaviour on procps images — the common case — is unchanged; tests 3 below pin that.

## Tests

`tests/test-hermes-slash-worker-prune.sh` (new) drives the real script through a `docker` shim that models both image families and records which pids get killed. 12 assertions.

Verified red on the pre-fix script:

```
[FAIL] posix container: selects real worker pids, never the uid column
       expected [101,102] got [1000]
[FAIL] posix container: kills real worker pids
       expected [101,102] got [1000]
[FAIL] posix container: uid column must not be read as a pid
       output selected pid=1000
[FAIL] posix container: unknown age is not treated as older than max-age
       expected [] got [1000]
[FAIL] stopped container: reports that the container is not running
       [PASS] no Hermes slash workers found

Passed: 7  Failed: 5
```

Green after the fix (12/12), on both Git Bash and Linux.

`tests/contracts/test-hermes-worker-guardrails.py` already existed but was an orphan — no Makefile target, no CI job, not in `release-gate.sh` — so nothing guarded this script at all. Both tests are now wired into `make test` and the Linux CI job.

`shellcheck -S warning` clean on both scripts.

### DIFF
--- a/.github/workflows/test-linux.yml
+++ b/.github/workflows/test-linux.yml
@@ -134,6 +134,11 @@ jobs:
       - name: CPU-Only Path Tests
         run: bash tests/test-cpu-only-path.sh
 
+      - name: Hermes Slash Worker Prune Tests
+        run: |
+          bash tests/test-hermes-slash-worker-prune.sh
+          python3 tests/contracts/test-hermes-worker-guardrails.py
+
       - name: Installer Contract Checks
         run: |
           bash tests/contracts/test-installer-contracts.sh
--- a/ods/Makefile
+++ b/ods/Makefile
@@ -26,6 +26,10 @@ test: ## Run unit and contract tests
 	@bash tests/test-installer-context-parity.sh
 	@bash tests/test-hermes-context-floor.sh
 	@echo ""
+	@echo "=== Hermes slash worker prune ==="
+	@bash tests/test-hermes-slash-worker-prune.sh
+	@python3 tests/contracts/test-hermes-worker-guardrails.py
+	@echo ""
 	@echo "=== Installer contracts ==="
 	@bash tests/contracts/test-installer-contracts.sh
 	@bash tests/test-phase03-multigpu-tty.sh
--- a/ods/scripts/prune-hermes-slash-workers.sh
+++ b/ods/scripts/prune-hermes-slash-workers.sh
@@ -67,9 +67,37 @@ if [[ -z "$CONTAINER" ]]; then
     exit 1
 fi
 
+# Turn one process-table layout into the normalized row this script works on:
+#
+#     pid <TAB> age_seconds <TAB> original ps line
+#
+# `has_age=1` means the caller asked ps for an elapsed-time column, so field 2
+# holds seconds. `has_age=0` means the layout only has `pid` then the command,
+# and age is recorded as -1 ("unknown") rather than being read out of whichever
+# column happens to sit in position 2.
+normalize_workers() {
+    local has_age="$1"
+    awk -v has_age="$has_age" '
+        $0 !~ /tui_gateway[.]slash_worker/ {
+            next
+        }
+        {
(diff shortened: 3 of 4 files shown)


Lightheartdevs:
Thanks for taking on the Hermes slash-worker pruning path; this fixes a quiet but risky parser mismatch where the repair command could report success while leaving workers running or targeting the wrong pid.

What looks good:
- `ods/scripts/prune-hermes-slash-workers.sh` now probes `ps` layouts separately and normalizes them before policy decisions, so the fallback path no longer treats a `ps -ef` UID column as either `pid` or `age`.
- Unknown ages are handled conservatively: they do not satisfy the max-age policy, and count-based pruning falls back to the lowest pid when elapsed time is unavailable.
- The stopped-container message now goes to stderr, keeping stdout as the normalized worker stream, and the new shim-based test exercises procps, POSIX/no-age, force-kill, and stopped-container behavior.

What I checked:
- Reviewed the PR body, patch, and changed files: `.github/workflows/test-linux.yml`, `ods/Makefile`, `ods/scripts/prune-hermes-slash-workers.sh`, and `ods/tests/test-hermes-slash-worker-prune.sh`.
- Ran `C:\Program Files\Git\bin\bash.exe tests/test-hermes-slash-worker-prune.sh`: 12 passed, 0 failed.
- Ran `python3 tests/contracts/test-hermes-worker-guardrails.py`: passed.
- GitHub checks are green, including Linux integration smoke, shell lint, distro matrix, macOS smoke, dashboard, env validation, Python lint, PowerShell lint, and secret scan.

Lightheartdevs:
Maintainer audit found one remaining false-green: when both in-container process-table layouts were unreadable, the cleanup warned but returned success and then claimed no workers existed. I pushed a fail-closed correction plus a negative fixture. Focused result is now 15/15, including nonzero exit, explicit inspection failure, and no false 'no workers' message. Waiting for refreshed CI before merge.

Lightheartdevs (approved):
Maintainer repair verified: worker rows are parsed from known layouts, unreadable process tables now fail closed without a false zero-worker claim, 15 focused assertions pass, and the rerun GitHub suite is green.

PR #2070, opened 2026-07-25
fix(memory-shepherd): stop a blocked run deleting the live lock

tang-vu:
﻿## Problem

`memory-shepherd.sh` arms its lock-cleanup trap *before* it checks whether another run already holds the lock:

```sh
cleanup_lock() { rm -f "$LOCKFILE"; }
trap cleanup_lock EXIT

if [ -f "$LOCKFILE" ]; then
    ...
    log "Another reset running (lock age: ${lock_age}s) — exiting"
    exit 0      # <-- trap fires and removes the OTHER run's lock
fi
echo $$ > "$LOCKFILE"
```

The invocation that correctly backs off takes the running invocation's lockfile with it. The next invocation then finds no lock and starts alongside a run that is still going — the exact thing the lock exists to prevent.

### Why this is reachable

Two shipped units drive the same script:

- `memory-shepherd-workspace.timer` — `OnUnitActiveSec=60s`, and its service has **two** `ExecStart` lines
- `memory-shepherd-memory.timer` — every 3h

A remote agent reset does two `scp` round-trips, so a run outliving the 60s gap is ordinary rather than exotic. Once two runs overlap they share a fixed archive filename (`${TIMESTAMP}.md`, minute resolution) and a fixed `/tmp/memory-shepherd-<agent>-current.md` scratch path, so the losing run can archive a half-written memory file, or overwrite the archive the other run just wrote — losing the scratch notes this tool exists to preserve before it resets MEMORY.md.

## Fix

Arm the trap only once the lockfile has been claimed.

## Also in this PR

`memory-shepherd.sh` uses an associative array for the parsed config but never got the Bash 4+ guard the repo requires. `tests/test-bash32-guards.sh` already enforces that guard for `lib/progress.sh`, `installers/phases/03-features.sh`, `scripts/pre-download.sh` and `scripts/ods-test-functional.sh`; this script was simply missed. It carries Darwin branches in `_stat_mtime`/`_stat_size` and is covered by `macos-bsd-compat.bats`, so macOS — where `/bin/bash` is 3.2 — is a path it is expected to reach. Today it dies there with `declare: -A: invalid option` instead of the actionable "install a modern bash" message its siblings print. Added the guard and registered the script in `test-bash32-guards.sh`.

`LOCKFILE` now honours `MEMORY_SHEPHERD_LOCKFILE`, which is what makes the locking protocol testable without touching the real lockfile.

## Tests

`tests/test-memory-shepherd-lock.sh` (new) covers busy / stale / uncontended locks — 13 assertions.

Red-on-old was verified by keeping the new env seam and moving **only** the `trap` back to its old position, so the failures isolate the defect rather than the missing seam:

```
[FAIL] busy lock: the other run's lockfile still exists
       lockfile was deleted
[FAIL] busy lock: the other run's pid is untouched
       expected [424242] got [MISSING]

Passed: 11  Failed: 2
```

Green after the fix (13/13). `tests/test-bash32-guards.sh` goes 10 → 12 assertions and passes.

Both are wired into `make test`; `test-bash32-guards.sh` had no CI job either, so the new Linux CI step runs both.

## Why this matters

A second memory-shepherd run is expected to exit when the lock is busy; it must not delete the lock owned by the live process and admit a third concurrent run.

## Overlap check

Triage refresh (2026-08-08): searched open and closed Osmantic/ODS PRs by semantic keywords (memory shepherd lock ownership), inspected the changed production files, and checked patch equivalence against current upstream/main. No strict duplicate or merged equivalent was found.


Lightheartdevs:
Thanks for tightening up the memory-shepherd lock path here; the failure mode is concrete, and moving lock ownership ahead of cleanup registration is the right small fix.

What looks good:
- `ods/memory-shepherd/memory-shepherd.sh` now only arms `cleanup_lock` after this process has actually written `LOCKFILE`, so the busy-lock exit path no longer deletes another run's live lock.
- The `MEMORY_SHEPHERD_LOCKFILE` override keeps the production default unchanged while making the lock protocol testable without touching `/tmp/memory-shepherd.lock`.
- Adding the Bash 4 guard fits the repo's existing convention for scripts that use associative arrays, and wiring both new coverage points into `ods/Makefile` and `.github/workflows/test-linux.yml` closes the prior CI gap.

What I checked:
- Reviewed the PR body, patch, and changed files: `.github/workflows/test-linux.yml`, `ods/Makefile`, `ods/memory-shepherd/memory-shepherd.sh`, `ods/tests/test-bash32-guards.sh`, and `ods/tests/test-memory-shepherd-lock.sh`.
- Ran `C:\Program Files\Git\bin\bash.exe tests/test-memory-shepherd-lock.sh`: 13 passed, 0 failed.
- Ran `C:\Program Files\Git\bin\bash.exe tests/test-bash32-guards.sh`: 12 passed, 0 failed.
- GitHub checks are green, including Linux integration smoke, shell lint, distro matrix, macOS smoke, dashboard, env validation, Python lint, PowerShell lint, and secret scan.

PR #2072, opened 2026-07-25
fix(mode-switch): stop --status aborting before it prints anything

tang-vu:
## Problem

`show_status` read the current mode with a bare pipeline:

```sh
current=$(grep "^ODS_MODE=" "$ENV_FILE" 2>/dev/null | cut -d= -f2)
echo "Current mode: ${current:-local}"
```

The script runs under `set -euo pipefail`. When `ODS_MODE` is absent — or `.env` does not exist at all — `grep` exits 1, `pipefail` propagates it, and `set -e` kills the script **on that assignment**. Nothing is printed: no current mode, no list of available modes, no error. The command just exits 2.

The `${current:-local}` fallback on the very next line shows the absent key was meant to be handled. It was simply unreachable.

Reproduced from a clean checkout, where `.env` has not been generated yet:

```
$ bash scripts/mode-switch.sh --status
$ echo $?
2
```

`scripts/README.md` lists this script for operators and `docs/MODE-SWITCH.md` documents it as the mode-switching backend, so `--status` — the read-only entry point people reach for first — is exactly the invocation that fails.

## Fix

- Guard the read with a `.env` existence check and tolerate a missing key, so the documented default is actually reached.
- Say when the reported value is a default rather than something configured, instead of silently implying `local` is set.
- Read with `cut -d= -f2-` and strip quotes/CR the way the rest of the repo does, so `ODS_MODE="hybrid"` or a CRLF `.env` reports the mode rather than a mangled string.

Switching behaviour is untouched.

## Tests

`tests/test-mode-switch-status.sh` (new, 17 assertions) covers status with and without `.env`, a `.env` that omits `ODS_MODE`, quoted and CRLF values, the bare no-argument invocation, switching in both directions, and rejection of bad input. Nothing tested this script before.

Red on the pre-fix script:

```
[FAIL] no .env: exits 0
[FAIL] no .env: falls back to the default mode
[FAIL] no .env: still lists the available modes
[FAIL] no .env: says why the value is a default
[FAIL] no ODS_MODE key: exits 0
[FAIL] no ODS_MODE key: falls back to the default mode
[FAIL] quoted mode is reported unquoted

Passed: 10  Failed: 7
```

Green after the fix: 17/17 on Git Bash and Linux. `shellcheck -S warning` clean.

Wired into `make test` and the Linux CI job.

### DIFF
--- a/.github/workflows/test-linux.yml
+++ b/.github/workflows/test-linux.yml
@@ -110,6 +110,9 @@ jobs:
       - name: Safe Env Loading Tests
         run: bash tests/test-safe-env.sh
 
+      - name: Mode Switch Tests
+        run: bash tests/test-mode-switch-status.sh
+
       - name: Validate Simulation Summary Tests
         run: bash tests/test-validate-sim-summary.sh
 
--- a/ods/Makefile
+++ b/ods/Makefile
@@ -56,6 +56,9 @@ test: ## Run unit and contract tests
 	@echo "=== safe .env loading (quote/CRLF handling) ==="
 	@bash tests/test-safe-env.sh
 	@echo ""
+	@echo "=== mode switch status and switching ==="
+	@bash tests/test-mode-switch-status.sh
+	@echo ""
 	@echo "=== Bind address sweep ==="
 	@bash tests/test-bind-address-sweep.sh
 	@bash tests/test-n8n-cookie-policy.sh
--- a/ods/scripts/mode-switch.sh
+++ b/ods/scripts/mode-switch.sh
@@ -39,8 +39,15 @@ env_set() {
 }
 
 show_status() {
-    local current
-    current=$(grep "^ODS_MODE=" "$ENV_FILE" 2>/dev/null | cut -d= -f2)
+    local current=""
+    if [[ -f "$ENV_FILE" ]]; then
+        # `grep` exits 1 when the key is absent and the script runs under
+        # `set -euo pipefail`, so a bare pipeline here aborted the whole script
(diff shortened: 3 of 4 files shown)


Lightheartdevs:
Thanks for taking on the `mode-switch --status` path; making the read-only operator command reliable before `.env` exists is a good fit for the current installer-hardening work.

What looks good:
- `ods/scripts/mode-switch.sh` now reaches the intended `local` fallback when `.env` is missing or `ODS_MODE` is absent instead of dying under `set -euo pipefail` on the `grep | cut` assignment.
- The status output distinguishes the missing-`.env` default case, and the value reader handles quoted and CRLF values without changing the actual switching behavior.
- `ods/tests/test-mode-switch-status.sh` covers clean checkout status, missing keys, quoted/CRLF mode values, both switch directions, and bad input; the new test is also wired into `ods/Makefile` and Linux CI.

What I checked:
- Reviewed the PR title, description, changed files, and patch for #2072 against the documented scripts/operator workflow.
- Checked the changed `show_status()` behavior for the repo's usual `.env` quote and CRLF handling expectations.
- Confirmed GitHub checks are passing, including Test Linux, ShellCheck, Matrix Smoke, dashboard frontend/API, Python lint, PowerShell lint, secret scan, and env validation. I could not rerun the bash test locally because this Windows runner only exposes a broken WSL `bash` launcher.

Lightheartdevs (approved):
Maintainer audit complete: missing/quoted/CRLF mode values are handled safely, both QR and mode-switch registrations survive the current-main refresh, focused tests pass, and refreshed CI is green.

PR #2073, opened 2026-07-25
fix(deps): build the core-service list from the services block only

tang-vu:
﻿## Problem

`validate_service_dependencies()` decides whether every declared `depends_on` target is enabled. Extension services come from the registry; core services are recovered from `docker-compose.base.yml` with:

```sh
sed -n 's/^  \([a-z][a-z0-9_-]*\):.*/\1/p' "$_base_compose"
```

That matches **any** two-space-indented key anywhere in the file, not just service names. Run against the current base compose it returns:

```
driver options llama-server open-webui dashboard-api model-router dashboard default
```

`driver` and `options` are keys of the `x-logging: &default-logging` anchor. `default` is the network. Three of the eight names are not services, so a `depends_on` naming one of them is accepted as satisfied.

Being straight about the impact: no shipped manifest depends on `driver`, `options` or `default`, so this hardens the gate rather than unblocking a failing install. But the parse is wrong *now*, and it gets wronger on its own — every anchor, network or volume key added to the file later silently joins the core-service list.

`installers/phases/11-services.sh:1003` runs this immediately before bringing the stack up and aborts the install when it fails, so the list needs to be exactly the services that exist.

## Fix

Read keys only while inside the top-level `services:` block. Verified against PyYAML on the real file — the extraction now returns exactly the five services and nothing else:

```
$ awk '...' docker-compose.base.yml | sort | tr '\n' ' '
dashboard dashboard-api llama-server model-router open-webui

$ python3 -c "import yaml; print(sorted(yaml.safe_load(open('docker-compose.base.yml'))['services']))"
['dashboard', 'dashboard-api', 'llama-server', 'model-router', 'open-webui']
```

Also default `SCRIPT_DIR` when resolving the base compose path, so a caller that sets neither `INSTALL_DIR` nor `SCRIPT_DIR` gets a missed file rather than an unbound-variable abort under `set -u`.

## Tests

`tests/test-validate-dependencies.sh` (new, 8 assertions) drives the real function against a base compose shaped like the shipped one — anchor keys at service indentation, a services block, a networks block — and asserts the anchor and network names are rejected while real services are accepted. It also pins the extraction against the repo's actual `docker-compose.base.yml` using PyYAML as the source of truth, so the parse cannot drift as services are added. Nothing tested this file before.

Red on the pre-fix library:

```
[FAIL] 'driver' is not treated as an enabled service
[FAIL] 'options' is not treated as an enabled service
[FAIL] 'default' is not treated as an enabled service

Passed: 5  Failed: 3
```

Green after the fix: 8/8 on Git Bash and Linux. `shellcheck -S warning` clean.

Wired into `make test` and the Linux CI job.

## Why this matters

Installer dependency checks must derive service IDs only from the Compose `services` block. Top-level networks or extension keys are not running services and cannot satisfy dependencies.

## Overlap check

Triage refresh (2026-08-08): searched open and closed Osmantic/ODS PRs by semantic keywords (dependency core services block), inspected the changed production files, and checked patch equivalence against current upstream/main. No strict duplicate or merged equivalent was found.


Lightheartdevs:
Thanks for tightening the service dependency gate here; this is exactly the kind of small parser fix that prevents installer surprises later.

What looks good:
- `ods/lib/validate-dependencies.sh` now scopes core-service discovery to the top-level `services:` block, so `x-logging` keys and the `default` network no longer masquerade as satisfied dependencies.
- The `SCRIPT_DIR` fallback avoids a separate `set -u` abort path when callers do not provide `INSTALL_DIR`, which matches the repo's defensive shell style.
- `ods/tests/test-validate-dependencies.sh` covers the real failure mode with anchor/network keys plus a PyYAML comparison against the shipped compose file, and it is wired into both `ods/Makefile` and Linux CI.

What I checked:
- Reviewed the PR title, description, changed files, and patch for #2073 against the repo's compose/extension dependency model.
- Checked the surrounding `validate_service_dependencies()` flow and the installer phase context that treats this as a launch gate.
- Confirmed GitHub checks are passing, including Test Linux, ShellCheck, Matrix Smoke, dashboard frontend/API, Python lint, PowerShell lint, secret scan, and env validation. I could not rerun the bash test locally because this Windows runner only exposes a broken WSL `bash` launcher.

PR #2078, opened 2026-07-25
fix(resources): handle permission and os errors safely during data directory scanning

vaibhavsrv:
### Problem
`_scan_service_disk()` in `ods/extensions/services/dashboard-api/routers/resources.py` called `data_path.iterdir()` directly without exception guards. If the data directory or any child directory encountered permission restrictions (e.g. `PermissionError`) or filesystem IO errors (`OSError`), the unhandled exception crashed `_scan_service_disk()` and caused the `/api/services/resources` endpoint to return an HTTP 500 error.

### Fix
Wrap `data_path.iterdir()` and `child.is_dir()` in `try ... except (PermissionError, OSError)` blocks in `_scan_service_disk()`. If iterating `/data` or checking a child directory encounters filesystem errors, a warning is logged and unreadable directories are skipped safely while returning all readable service disk usage entries.

### Threat Model
- **Fail-soft scoping:** Unreadable data subdirectories or broken permissions will no longer cause systemic failure of the service resource endpoint.
- **Side effects:** None; valid data directories continue to be mapped and measured accurately.

### Verification
Ran the unit test suite (`pytest ods/extensions/services/dashboard-api/tests/test_resources.py -q`):
```
.....................                                                    [100%]
21 passed, 1 warning in 0.70s
```

### DIFF
--- a/ods/extensions/services/dashboard-api/routers/resources.py
+++ b/ods/extensions/services/dashboard-api/routers/resources.py
@@ -53,8 +53,16 @@ def _scan_service_disk() -> dict[str, dict]:
     results = {}
     if not data_path.is_dir():
         return results
-    for child in data_path.iterdir():
-        if not child.is_dir():
+    try:
+        children = list(data_path.iterdir())
+    except (PermissionError, OSError) as exc:
+        logger.warning("Could not iterate data directory %s: %s", data_path, exc)
+        return results
+    for child in children:
+        try:
+            if not child.is_dir():
+                continue
+        except (PermissionError, OSError):
             continue
         service_id = _DATA_DIR_MAP.get(child.name, child.name)
         size_gb = dir_size_gb(child)
--- a/ods/extensions/services/dashboard-api/tests/test_resources.py
+++ b/ods/extensions/services/dashboard-api/tests/test_resources.py
@@ -1,5 +1,6 @@
 """Tests for routers/resources.py — per-service resource metrics."""
 
+from pathlib import Path
 from unittest.mock import patch
 
 from host_agent_client import AgentHTTPError, AgentUnavailable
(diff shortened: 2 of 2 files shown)


Lightheartdevs:
Thanks for taking on the resource scanner failure path here; making `/api/services/resources` fail soft on unreadable data directories fits the dashboard API's recent hardening work.

Quickest path to merge:
- Rebase this branch onto current `main` and keep the PR scoped to `ods/extensions/services/dashboard-api/routers/resources.py` plus `ods/extensions/services/dashboard-api/tests/test_resources.py`. The current diff also changes `ods/extensions/services/dashboard-api/routers/workflows.py` and `ods/extensions/services/dashboard-api/tests/test_workflows.py`, which appears to belong to the workflow catalog reader fix rather than this resource-scanning PR.
- After the scope cleanup, rerun `pytest ods/extensions/services/dashboard-api/tests/test_resources.py -q` or rely on the dashboard API CI job if it stays green. The resource change itself looks reasonable: `iterdir()` and `child.is_dir()` now fail closed without taking down the endpoint.

What I checked:
- Reviewed the PR title/body, changed files, and full diff against current `main`.
- Checked existing PR comments and found no advisory marker for this head SHA.
- Checked CI: dashboard `api`, Python lint, mypy, integration smoke, matrix smoke, env validation, secret scan, ShellCheck, and dashboard frontend are passing.

vaibhavsrv:
Thanks @Lightheartdevs for the review and advisory guidance!

I've rebased the branch onto current `main` and cleaned up the PR scope so the diff strictly touches only:
- `ods/extensions/services/dashboard-api/routers/resources.py`
- `ods/extensions/services/dashboard-api/tests/test_resources.py`

(The unrelated workflow reader changes have been dropped from this branch.)

I also verified the local test suite execution:
- `pytest ods/extensions/services/dashboard-api/tests/test_resources.py -q` passed cleanly (21/21 passed).

Lightheartdevs:
Thanks for cleaning up the resource scanner scope after the earlier pass; this now matches the dashboard API fail-soft hardening work cleanly.

What looks good:
- `ods/extensions/services/dashboard-api/routers/resources.py` now handles `DATA_DIR.iterdir()` failures without turning `/api/services/resources` into a 500, which is the right behavior for unreadable or transient data directories.
- The `child.is_dir()` guard also avoids one bad child path poisoning the whole response, while preserving readable service disk entries.
- The diff is now scoped to the resource router and its focused test coverage.

What I checked:
- Reviewed the PR description, current head diff, changed files, and follow-up discussion.

PR #2079, opened 2026-07-25
fix(features): handle missing or non-dictionary requirements safely in feature calculator

vaibhavsrv:
### Problem
`calculate_feature_status()` in `ods/extensions/services/dashboard-api/routers/features.py` extracted `feature["requirements"]` directly without dict type validation. If a feature definition (e.g. dynamically loaded feature plugins or third-party feature manifests) lacked a `"requirements"` key or had `requirements=None`, calling `req.get(...)` raised an unhandled `KeyError` or `AttributeError`.

### Fix
Use `feature.get("requirements")` with an `isinstance(req, dict)` fallback check to ensure `req` is always a valid dictionary before accessing VRAM or service requirement fields.

### Threat Model
- **Fail-soft scoping:** Features with omitted or null requirements dictionaries default safely without crashing feature discovery endpoints (`/api/features`).
- **Side effects:** None; standard features with valid requirements dicts function identically.

### Verification
Ran the unit test suite (`pytest ods/extensions/services/dashboard-api/tests/test_features.py -q`):
```
....................                                                     [100%]
20 passed, 1 warning in 0.69s
```

### DIFF
--- a/ods/extensions/services/dashboard-api/routers/features.py
+++ b/ods/extensions/services/dashboard-api/routers/features.py
@@ -31,7 +31,9 @@ def calculate_feature_status(feature: dict, services: list, gpu_info: Optional[G
             pass
         gpu_vram_free_gb = gpu_vram_gb  # assumes zero current usage; Docker can't measure host memory pressure
 
-    req = feature["requirements"]
+    req = feature.get("requirements")
+    if not isinstance(req, dict):
+        req = {}
     vram_ok = gpu_vram_gb >= req.get("vram_gb", 0)
     vram_fits = gpu_vram_free_gb >= req.get("vram_gb", 0)
 
--- a/ods/extensions/services/dashboard-api/tests/test_features.py
+++ b/ods/extensions/services/dashboard-api/tests/test_features.py
@@ -26,6 +26,29 @@ class TestCalculateFeatureStatusDefaults:
         assert result["setupTime"] == "Unknown"
         assert result["priority"] == 99
 
+    def test_missing_and_non_dict_requirements_use_defaults(self):
+        """Features with missing requirements, requirements=None, or non-dict requirements default safely."""
+        # 1. Missing requirements key
+        f_missing = {"id": "f_missing", "name": "Missing Req"}
+        res1 = calculate_feature_status(f_missing, [], None)
+        assert res1["id"] == "f_missing"
+        assert res1["status"] == "enabled"
+        assert res1["requirements"]["vramOk"] is True
+
+        # 2. requirements = None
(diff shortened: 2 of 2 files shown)


Lightheartdevs:
Thanks for tightening the feature calculator around missing or malformed `requirements`; that is a good fit for the dashboard API's fail-soft behavior.

Quickest path to merge:
- Rebase this branch onto current `main` and narrow the diff to `ods/extensions/services/dashboard-api/routers/features.py` plus `ods/extensions/services/dashboard-api/tests/test_features.py`. Right now the PR also carries `resources.py`/`test_resources.py` and `workflows.py`/`test_workflows.py` changes from adjacent open fixes, so merging this as-is would bundle unrelated dashboard API behavior under the features PR.
- Add or adjust the test to cover both missing `requirements` and explicitly non-dictionary `requirements` values, since the implementation intentionally handles both. The current test name/docstring mentions `requirements=None`, but the fixture only exercises the missing-key case.

What I checked:
- Reviewed the PR title/body, changed files, and full diff against current `main`.
- Checked existing PR comments and found no advisory marker for this head SHA.
- Checked CI: dashboard `api`, Python lint, mypy, integration smoke, matrix smoke, env validation, secret scan, ShellCheck, and dashboard frontend are passing.

vaibhavsrv:
Thanks @Lightheartdevs for the detailed review and guidance!

I've rebased the branch onto current `main` and narrowed the PR scope so the diff strictly touches only:
- `ods/extensions/services/dashboard-api/routers/features.py`
- `ods/extensions/services/dashboard-api/tests/test_features.py`

(All unrelated workflow and resources files have been removed from this PR.)

Additionally, I updated `test_missing_and_non_dict_requirements_use_defaults` in `test_features.py` to explicitly exercise all three cases:
1. Missing `"requirements"` key
2. `requirements = None`
3. Non-dictionary `requirements` string/value

Local test verification:
- `pytest ods/extensions/services/dashboard-api/tests/test_features.py -q` passed cleanly (20/20 passed).

Lightheartdevs:
Thanks for tightening the feature status calculator and following through on the narrower test coverage.

What looks good:
- `ods/extensions/services/dashboard-api/routers/features.py` now treats missing, `None`, and non-dict `requirements` as empty requirements instead of letting malformed feature metadata crash status calculation.
- `ods/extensions/services/dashboard-api/tests/test_features.py` covers all three malformed requirement shapes, which matches the implementation intent.
- The updated diff is scoped to the feature calculator and tests, with the unrelated router changes removed.

What I checked:
- Reviewed the PR description, current diff, changed files, and previous advisory/follow-up discussion.

PR #2081, opened 2026-07-25
fix(user-extensions): handle non-string and non-dict manifest values safely during service scan

vaibhavsrv:
## Problem

When scanning `manifest.yaml` files for user extensions in `scan_user_extension_services`, non-string `name` entries (e.g. dictionary or list values) or raw integer string / `health_port` properties could result in invalid type propagations or unhandled casting exceptions during dictionary assembly.

Manifest field validation checked string type for `health` paths but allowed non-scalar `name` values to propagate directly to downstream service dictionaries.

## Fix

Coerce `name` to a string safely when a scalar (string, int, float) is present, falling back to `service_id` when missing or non-scalar. Safely parse `health_port` and `port` attributes without raising unhandled conversion errors.

## Verification

Added `test_scan_non_string_name_and_health_port_coercion` to `ods/extensions/services/dashboard-api/tests/test_user_extensions.py`. Ran `pytest` locally: 15/15 passed.

### DIFF
--- a/ods/extensions/services/dashboard-api/tests/test_user_extensions.py
+++ b/ods/extensions/services/dashboard-api/tests/test_user_extensions.py
@@ -228,3 +228,25 @@ class TestCaching:
 
         r2 = get_user_services_cached(user_dir, ttl=300.0)
         assert r2 == {}
+
+    def test_scan_non_string_name_and_health_port_coercion(self, tmp_path):
+        """Non-string names fall back safely and health_port is handled cleanly."""
+        user_dir = tmp_path / "user"
+        ext_dir = user_dir / "custom-ext"
+        manifest = {
+            "schema_version": "ods.services.v1",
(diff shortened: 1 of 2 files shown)


Lightheartdevs:
Thanks for hardening the user extension manifest scanner around malformed service metadata.

What looks good:
- `ods/extensions/services/dashboard-api/user_extensions.py` now avoids propagating dict/list values as service names and falls back to the service id for non-scalar names.
- The port parsing stays inside the existing fail-soft skip path, while valid scalar `port`, `external_port_default`, and `health_port` values are normalized before building the service record.
- `ods/extensions/services/dashboard-api/tests/test_user_extensions.py` adds coverage for a malformed name plus explicit `health_port` handling.

What I checked:
- Reviewed the PR description, full diff, and the existing extension scanner contract.

PR #2082, opened 2026-07-25
fix(tailscale): validate host agent response shape in proxy helper

vaibhavsrv:
## Problem

When the host-agent returns a non-dictionary JSON payload (e.g., boolean or primitive string response), `_proxy_agent` in `routers/tailscale.py` returned raw values directly. Downstream routes or caller code expecting a dictionary response could encounter unexpected type errors or invalid status parsing.

The proxy transport translated HTTP errors to `HTTPException` but assumed successful JSON responses from the host-agent were always dictionary instances.

## Fix

Validate that host-agent responses are dictionary instances in `_proxy_agent`. If a non-dictionary value is returned, log a warning and return `{"running": False}` safely.

## Verification

Added `test_status_handles_non_dict_agent_response` to `ods/extensions/services/dashboard-api/tests/test_tailscale.py`. Ran `pytest` locally: 9/9 passed.

### DIFF
--- a/ods/extensions/services/dashboard-api/routers/tailscale.py
+++ b/ods/extensions/services/dashboard-api/routers/tailscale.py
@@ -41,7 +41,11 @@ router = APIRouter(tags=["tailscale"])
 def _proxy_agent(path: str, timeout: int = 15) -> dict:
     """Forward a GET to the host-agent. Translates HTTPError → HTTPException."""
     try:
-        return request_agent_json("GET", path, timeout=timeout)
+        res = request_agent_json("GET", path, timeout=timeout)
+        if not isinstance(res, dict):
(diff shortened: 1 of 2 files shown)


Lightheartdevs:
Thanks for tightening the Tailscale host-agent response handling.

What looks good:
- `ods/extensions/services/dashboard-api/routers/tailscale.py` now validates the successful host-agent response shape before returning it from the dashboard API proxy.
- Falling back to `{running: false}` keeps `/api/tailscale/status` aligned with its existing not-configured/not-running shape instead of leaking an unexpected primitive response to the UI.
- `ods/extensions/services/dashboard-api/tests/test_tailscale.py` adds a focused non-dict response regression.

What I checked:
- Reviewed the PR description, full diff, and the surrounding Tailscale status endpoint contract.

PR #2083, opened 2026-07-25
fix(privacy-shield): safely parse SHIELD_PORT environment variable with quote stripping and default fallback

vaibhavsrv:
## Problem

In `routers/privacy.py`, `SHIELD_PORT` was read directly with `int(os.environ.get("SHIELD_PORT", ...))`. If `SHIELD_PORT` contains quotes or invalid non-numeric text, `int(...)` raises an unhandled `ValueError`, leading to HTTP 500 error tracebacks on `/api/privacy-shield/status` and `/api/privacy-shield/stats`.

Port lookup assumed environment values passed to `int()` were always valid unquoted integer strings.

## Fix

Added a `_get_shield_port()` helper function in `privacy.py` that strips quote pairs, validates numeric format, and falls back to the configured service default port on invalid inputs.

## Verification

Added `test_get_shield_port_invalid_env_fallback` to `ods/extensions/services/dashboard-api/tests/test_privacy.py`. Ran `pytest` locally: 18/18 passed.

### DIFF
--- a/ods/extensions/services/dashboard-api/routers/privacy.py
+++ b/ods/extensions/services/dashboard-api/routers/privacy.py
@@ -23,11 +23,25 @@ logger = logging.getLogger(__name__)
 router = APIRouter(tags=["privacy"])
 
 
+def _get_shield_port() -> int:
+    """Safely parse SHIELD_PORT from environment with fallback to default service port."""
+    default_port = int(SERVICES.get("privacy-shield", {}).get("port", 0))
+    raw = os.environ.get("SHIELD_PORT", "").strip("'\"").strip()
+    if not raw:
+        return default_port
+    try:
+        val = int(raw)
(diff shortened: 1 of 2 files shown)


Lightheartdevs:
Thanks for hardening the Privacy Shield port parsing path here.

What looks good:
- `ods/extensions/services/dashboard-api/routers/privacy.py` centralizes `SHIELD_PORT` parsing through `_get_shield_port()`, so status and stats endpoints share the same fallback behavior.
- Invalid, empty, quoted, and non-positive values now fall back instead of raising `ValueError` during request handling.
- `ods/extensions/services/dashboard-api/tests/test_privacy.py` adds a direct regression for invalid and quoted environment values.

What I checked:
- Reviewed the PR description, full diff, and existing Privacy Shield status/stats call sites.

PR #2084, opened 2026-07-25
fix(talk): ensure active model compatibility is a dictionary before item assignment

vaibhavsrv:
## Problem

In `routers/talk.py`, `_active_model_app_compatibility()` invokes `model_app_compatibility()` and immediately assigns `compatibility["activeModel"] = ...`. Under edge cases where `model_app_compatibility()` returns `None` or a non-dictionary value, this causes an unhandled `TypeError: 'NoneType' object does not support item assignment` when computing Talk status or model viability.

Assumed `model_app_compatibility()` always returns a dictionary instance without defensive type verification before key assignment.

## Fix

Ensure `compatibility` is initialized to `{}` if `model_app_compatibility()` returns non-dict, and check `isinstance(entry, dict)` safely when reading model catalog ID.

## Verification

Added `test_active_model_app_compatibility_handles_non_dict_return` to `ods/extensions/services/dashboard-api/tests/test_talk.py`. Ran `pytest` locally: 38/38 passed.

### DIFF
--- a/ods/extensions/services/dashboard-api/routers/talk.py
+++ b/ods/extensions/services/dashboard-api/routers/talk.py
@@ -89,8 +89,10 @@ def _active_model_app_compatibility() -> dict[str, Any]:
         entry or {},
         runtime_context=model_compatibility_runtime_context(INSTALL_DIR),
     )
+    if not isinstance(compatibility, dict):
+        compatibility = {}
     compatibility["activeModel"] = {
-        "id": entry.get("id") if entry else None,
+        "id": entry.get("id") if isinstance(entry, dict) else None,
         "model": model_name or None,
         "gguf": gguf or None,
     }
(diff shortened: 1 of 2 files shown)


Lightheartdevs:
Thanks for taking on the Talk compatibility edge case around active model metadata.

What looks good:
- `ods/extensions/services/dashboard-api/routers/talk.py` now verifies `model_app_compatibility()` returned a dict before assigning `activeModel`, so an unexpected helper return no longer breaks the Talk compatibility path.
- The `entry` access is now type-checked too, which keeps the fallback response coherent if catalog lookup returns an unexpected shape.
- `ods/extensions/services/dashboard-api/tests/test_talk.py` adds a focused regression for the non-dict compatibility return.

What I checked:
- Reviewed the PR description, full diff, and surrounding Talk compatibility code.

PR #2085, opened 2026-07-25
fix(privacy-shield): scrub compressed IPv6 addresses whole

tang-vu:
﻿## Problem

`PIIDetector.PATTERNS['ip_address']` lists the bare "ends in `::`" branch *before* the branch that also consumes the tail after the compression marker:

```python
r'(?:[0-9a-fA-F]{1,4}:){1,7}:'                    # Trailing ::   <- tried first
r'|'
r'::(?:[0-9a-fA-F]{1,4}:){0,6}[0-9a-fA-F]{1,4}'   # Leading ::
r'|'
r'(?:[0-9a-fA-F]{1,4}:){1,6}:[0-9a-fA-F]{1,4}'    # Middle ::
```

Python alternation is first-match-wins, not longest-match, so a compressed IPv6 address matched only its head and the rest was emitted in clear:

```
'Peer fe80::1ff:fe23:4567:890a connected'
  -> 'Peer <PII_ip_address_2198aff566b4>1ff:fe23:4567:890a connected'

'host 2001:db8::dead end'
  -> 'host <PII_ip_address_96ed29589156>dead end'
```

For a link-local address the leaked tail is the whole interface identifier — on SLAAC hosts that is the MAC address. Every RFC 5952 form except leading-`::` and the fully expanded one leaked its tail to the upstream provider. Round-tripping still looked correct, which is why this was not visible from the response side.

## Fix

Move the head-and-tail branch ahead of the trailing-`::` branch and widen it to consume all remaining hextets:

```python
r'(?:[0-9a-fA-F]{1,4}:){1,7}(?::[0-9a-fA-F]{1,4}){1,7}'  # head::tail
```

The tail branch still requires a doubled colon, so `12:34:56` and MAC addresses are unchanged (they were never matched before either).

## Tests

`tests/test_pii_scrubber.py` gains a parametrised case over six compressed forms asserting nothing but the token survives, a round-trip case, and a clock-time negative case. The three compressed-with-tail cases fail on the previous regex and pass on this one; the existing 34 tests are unaffected.

```
$ python -m pytest tests/test_pii_scrubber.py -q
40 passed
```

## Why this matters

Privacy Shield promises to remove IP addresses before forwarding text. Partial matching of compressed IPv6 can leave the host identifier visible even though the input appears scrubbed.

## Overlap check

Triage refresh (2026-08-08): searched open and closed Osmantic/ODS PRs by semantic keywords (Privacy Shield compressed IPv6), inspected the changed production files, and checked patch equivalence against current upstream/main. No strict duplicate or merged equivalent was found.


Lightheartdevs:
Thanks for tightening the Privacy Shield IPv6 scrubber around compressed addresses.

What looks good:
- `ods/extensions/services/privacy-shield/pii_scrubber.py` now orders the IPv6 alternation so compressed `head::tail` forms are consumed whole instead of leaving the interface identifier in clear text.
- The tests cover several RFC 5952-style compressed forms, round-trip restoration, and a clock-time negative case, which addresses both leakage and false-positive risk.
- The change is narrowly scoped to the IP regex branch and preserves the existing token/restore behavior.

What I checked:
- Reviewed the PR description, patch, and the IPv4/full IPv6/compressed IPv6 alternation order.
- Static review of `ods/extensions/services/privacy-shield/tests/test_pii_scrubber.py` for the new compressed IPv6 and negative assertions.

PR #2086, opened 2026-07-25
fix(litellm): map async call types to the right telemetry endpoint

tang-vu:
﻿## Problem

`_endpoint_path` in `ods_token_spy_callback.py` classifies a request by LiteLLM's `call_type`:

```python
if call_type in {"text_completion", "completion"}:
    return "/v1/completions"
return "/v1/chat/completions"
```

`CallTypes` names the *calling function*, not the wire endpoint. `litellm.completion()` / `acompletion()` is the **chat** API; `litellm.text_completion()` / `atext_completion()` is the legacy one. The set above has `completion` on the legacy side, and it matches only the bare stems, so the two async spellings — the only ones the proxy ever produces — both land wrong:

| `call_type` | wire endpoint | reported before | reported now |
|---|---|---|---|
| `acompletion` | `/v1/chat/completions` | `/v1/chat/completions` | same |
| `atext_completion` | `/v1/completions` | ❌ `/v1/chat/completions` | ✅ `/v1/completions` |
| `completion` | `/v1/chat/completions` | ❌ `/v1/completions` | ✅ `/v1/chat/completions` |
| `text_completion` | `/v1/completions` | `/v1/completions` | same |

In practice every `/v1/completions` request through the proxy was filed under `/v1/chat/completions` in Token Spy, so the `path` breakdown never showed legacy-completions traffic at all. The `responses` branch above already used a substring test, which is why `aresponses` was handled correctly.

## Fix

Match the `text_completion` family explicitly and let everything else fall through to chat, so the sync and async spellings agree:

```python
if call_type in {"text_completion", "atext_completion"}:
    return "/v1/completions"
return "/v1/chat/completions"
```

Both values remain inside `routed_telemetry._ALLOWED_PATHS`, so ingest validation is unchanged.

## Tests

`test_callback_maps_call_types_to_wire_endpoints` pins all six call types plus the missing-`call_type` default. It fails on the previous mapping and passes here.

```
$ python -m pytest extensions/services/litellm/tests/test_token_spy_callback.py -q
5 passed
```

## Why this matters

LiteLLM callback types represent different OpenAI wire endpoints. Correct endpoint attribution is required for Token Spy usage records and per-endpoint analysis.

## Overlap check

Triage refresh (2026-08-08): searched open and closed Osmantic/ODS PRs by semantic keywords (LiteLLM callback endpoint async), inspected the changed production files, and checked patch equivalence against current upstream/main. No strict duplicate or merged equivalent was found.


Lightheartdevs:
Thanks for tightening the LiteLLM-to-Token Spy endpoint mapping.

What looks good:
- `ods/extensions/services/litellm/ods_token_spy_callback.py` now maps the `text_completion` family to `/v1/completions` while leaving `completion`/`acompletion` on the chat endpoint, matching LiteLLM�s function-vs-wire-endpoint distinction.
- The regression test pins sync, async, responses, and missing `call_type` behavior in one compact table of assertions.
- The emitted paths remain inside the existing telemetry allowlist, so this corrects attribution without changing ingest validation shape.

What I checked:
- Reviewed the PR description, patch, callback path logic, and `ods/extensions/services/litellm/tests/test_token_spy_callback.py` coverage.
- Static review of the default fallback to `/v1/chat/completions` for unknown or missing call types.

PR #2087, opened 2026-07-25
fix(token-spy): price on the longest matching model prefix

tang-vu:
## Problem

`COST_PER_MILLION` is documented as *"Cost per million tokens by model prefix (longer prefixes matched first)"*, but the fallback lookup in `estimate_cost` walks the dict in declaration order and breaks on the first substring hit:

```python
for prefix, r in COST_PER_MILLION.items():
    if prefix in model_lower:
        rates = r
        break
```

The table declares `gpt-4o` before `gpt-4o-mini`, and `gpt-4o` is a substring of `gpt-4o-mini`, so mini traffic never reaches its own row:

| model | correct input rate | charged before |
|---|---|---|
| `gpt-4o-mini` | $0.15 / 1M | ❌ $2.50 / 1M |
| `gpt-4o` | $2.50 / 1M | $2.50 / 1M |

Output tokens were off by the same factor ($0.60 → $10.00). That is a ~16x overstatement of spend on the one shadowed pair in the table; the Anthropic and Kimi blocks happen to be written longest-first and were unaffected.

The provider plugins already get this right — `OpenAIProvider.get_model_pricing` iterates `sorted(self.COST_TABLE.keys(), key=len, reverse=True)`. Only this backwards-compatibility fallback disagreed, so the bug surfaced whenever `ProviderRegistry.get_or_none()` returned nothing for the request's provider.

## Fix

Keep the longest match rather than the first, matching the plugins and making the table order-independent for future entries:

```python
for prefix, r in COST_PER_MILLION.items():
    if prefix in model_lower and len(prefix) > len(matched):
        matched, rates = prefix, r
```

No table rows are reordered or re-priced — only the selection rule changes.

## Tests

New `tests/test_cost_estimation.py`:
- the `gpt-4o` / `gpt-4o-mini` pair resolves to its own rate,
- **every** prefix in the table is reachable at its declared input rate, so a future entry that shadows an existing one fails here rather than silently mispricing,
- unknown and empty model names stay free.

The first two fail on the previous lookup. The module is guarded with `pytest.importorskip("fastapi")` so it skips where the service deps are absent.

```
$ python -m pytest extensions/services/token-spy/tests/test_cost_estimation.py -q
3 passed
```

### DIFF
--- a/ods/extensions/services/token-spy/main.py
+++ b/ods/extensions/services/token-spy/main.py
@@ -620,13 +620,16 @@ def estimate_cost(model: str, input_tokens: int, output_tokens: int,
     if provider:
         return provider.calculate_cost(usage, model)
 
-    # Fallback to hardcoded rates for backwards compatibility
+    # Fallback to hardcoded rates for backwards compatibility.
+    # Longest match wins, as the provider plugins already do: a shorter family
+    # prefix must not shadow a more specific one just because it is declared
+    # first (gpt-4o would otherwise price every gpt-4o-mini call).
     rates = None
+    matched = ""
     model_lower = (model or "").lower()
     for prefix, r in COST_PER_MILLION.items():
-        if prefix in model_lower:
-            rates = r
-            break
+        if prefix in model_lower and len(prefix) > len(matched):
+            matched, rates = prefix, r
     if not rates:
         return 0.0
 
--- a/ods/extensions/services/token-spy/tests/test_cost_estimation.py
+++ b/ods/extensions/services/token-spy/tests/test_cost_estimation.py
@@ -1,0 +1,60 @@
+"""Fallback pricing lookup contracts for estimate_cost."""
+
+from __future__ import annotations
+
+import importlib.util
+import sys
+from pathlib import Path
+from uuid import uuid4
+
+import pytest
+
+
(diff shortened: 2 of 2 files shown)


Lightheartdevs:
Thanks for fixing the Token Spy fallback pricing path without changing the rate table itself.

What looks good:
- `ods/extensions/services/token-spy/main.py` now uses the longest matching prefix for the fallback table, matching the provider plugin behavior and preventing `gpt-4o` from shadowing `gpt-4o-mini`.
- `ods/extensions/services/token-spy/tests/test_cost_estimation.py` covers the known shadowed pair, unknown/empty models, and every declared fallback prefix so future table additions cannot silently become unreachable.
- The change is limited to fallback selection logic; no pricing constants or provider-specific implementations are being churned.

What I checked:
- Diff/static review of `estimate_cost()` and the new fallback pricing tests.
- Local PR-worktree test: `python -m pytest ods/extensions/services/token-spy/tests/test_cost_estimation.py -q` passed with `3 passed` and existing FastAPI deprecation warnings.
- GitHub checks are passing, including Dashboard api/frontend, Python lint/type checks, Test Linux integration-smoke, Matrix Smoke, ShellCheck, PowerShell lint, env validation, and secret scan.

PR #2090, opened 2026-07-25
fix(gpu): stop hybrid splits stranding a GPU on odd counts

tang-vu:
## Problem

`largest_pow2_divisor` documents two properties — *"p divides n evenly"* and *"Minimum return value is 2"* — that cannot both hold, because no power of two divides an odd number. The floor wins:

```python
return max(2, p)
```

So for every odd GPU count, `tensor_parallel_size * pipeline_parallel_size` comes out one short and a GPU that was assigned to llama-server is never placed:

| llama GPUs | tp | pp | placed |
|---|---|---|---|
| 4 | 2 | 2 | 4 ✅ |
| **5** | **2** | **2** | **4** ❌ |
| 6 | 2 | 3 | 6 ✅ |
| **7** | **2** | **3** | **6** ❌ |

Five GPUs is not a corner case — it is what the shipped 8-GPU NV12 fixture produces, because `assign_services` pushes `remaining[3:]` back to llama specifically so "no GPU sits idle":

```
$ python scripts/assign_gpus.py --topology tests/fixtures/topology_json/nvidia_smi_topo_matrix_8gpus_nv12_full_mesh.json --model-size 100000
llama gpus 5 [0, 1, 5, 6, 7]
parallelism {'mode': 'hybrid', 'tensor_parallel_size': 2, 'pipeline_parallel_size': 2, ...}
```

An A100 is reserved by llama-server and left out of the topology — the exact outcome the push-back logic exists to prevent.

`installers/phases/03-features.sh` carries a hand-maintained mirror of this rule (marked `NOTE: keep in sync with assign_gpus.py select_parallelism()`) that hardcodes `tp=2; pp=$((n/2))`, so the manual multi-GPU selection path stranded a GPU the same way.

## Fix

- `largest_pow2_divisor` returns the honest divisor, `1` when none exists, and its docstring now says so.
- `_hybrid_or_pipeline` falls back to pipeline when no even tensor grouping exists; pipeline spans every GPU. The two hybrid call sites and the four duplicated pipeline literals collapse into it.
- The shell mirror gets the same odd-count guard.

Even counts are unaffected — 4, 6, 8, 16 still get their hybrid split at the same `gpu_memory_utilization`.

Not touched: the shell mirror still hardcodes `tp=2` for even counts where Python picks the largest divisor ≤ √n (they diverge at n=16: `tp=2 pp=8` vs `tp=4 pp=4`). That is a pre-existing difference in a separate code path and changing it would alter working 16-GPU configurations, so it is out of scope here.

## Tests

Three existing cases pinned `tp=2 pp=2` for a five-GPU assignment — they were asserting the stranded plan, and are updated to the pipeline outcome with the reason inline. Those five-GPU fixtures were also the *only* hybrid output the suite exercised, and the assertions never multiplied `tp * pp` out, which is how this survived.

New `TestParallelismCoversAllGpus` checks the missing invariant directly against `select_parallelism`: `tp * pp == GPU count` for counts 1–16 across every rank band, odd NVLink counts fall back to pipeline, even ones stay hybrid, and `largest_pow2_divisor` divides its input for 1–64.

```
$ python -m pytest tests/test-assign-gpus.py -q
89 passed
```

Against the previous algorithm, 7 fail. `bash -n` passes on the installer phase and `tests/test-phase03-multigpu-tty.sh` still passes.

### DIFF
--- a/ods/installers/phases/03-features.sh
+++ b/ods/installers/phases/03-features.sh
@@ -418,13 +418,18 @@ run_custom() {
 
   # NOTE: keep in sync with assign_gpus.py select_parallelism()
   local mode tp pp mem_util
+  # An odd GPU count has no even tensor grouping: tp*pp would come out one
+  # short and the leftover GPU would be listed but never placed. Pipeline
+  # covers every GPU, so fall back instead of splitting.
   if   [[ $n -eq 1 ]];         then mode="none";     tp=1;  pp=1;        mem_util=0.95
   elif [[ $min_rank -ge 80 ]]; then
-    if   [[ $n -le 3 ]];       then mode="tensor";   tp=$n; pp=1;        mem_util=0.92
-    else                            mode="hybrid";   tp=2;  pp=$((n/2)); mem_util=0.93; fi
+    if   [[ $n -le 3 ]];        then mode="tensor";   tp=$n; pp=1;        mem_util=0.92
+    elif (( n % 2 != 0 ));      then mode="pipeline"; tp=1;  pp=$n;       mem_util=0.95
+    else                             mode="hybrid";   tp=2;  pp=$((n/2)); mem_util=0.93; fi
   elif [[ $min_rank -le 10 ]]; then mode="pipeline"; tp=1;  pp=$n;       mem_util=0.95
   elif [[ $n -le 3 ]];         then mode="pipeline"; tp=1;  pp=$n;       mem_util=0.95
-  elif [[ $min_rank -ge 40 ]]; then mode="hybrid";   tp=2;  pp=$((n/2)); mem_util=0.93
+  elif [[ $min_rank -ge 40 && $((n % 2)) -eq 0 ]]; then
+                                    mode="hybrid";   tp=2;  pp=$((n/2)); mem_util=0.93
   else                              mode="pipeline"; tp=1;  pp=$n;       mem_util=0.95
   fi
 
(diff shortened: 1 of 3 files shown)


Lightheartdevs:
Thanks for taking on the odd-count GPU planning hole here; this is exactly the kind of invariant that should be pinned in tests.

What looks good:
- `ods/scripts/assign_gpus.py` now keeps the key placement invariant explicit: `tensor_parallel_size * pipeline_parallel_size` covers every assigned llama GPU, and odd high-bandwidth counts fall back to pipeline instead of silently dropping one GPU from the plan.
- `ods/installers/phases/03-features.sh` gets the matching odd-count guard in the manual multi-GPU path, which matters because the repo already calls out that shell logic as a mirror of `select_parallelism()`.
- `ods/tests/test-assign-gpus.py` adds direct coverage for GPU counts 1-16 across rank bands, odd NVLink fallback, even NVLink hybrid behavior, and the divisor helper itself.

What I checked:
- Diff/static review of the Python planner, shell mirror, and updated fixture expectations.
- Local PR-worktree test: `python -m pytest ods/tests/test-assign-gpus.py -q` passed with `89 passed`.
- I could not run `bash -n ods/installers/phases/03-features.sh` locally because this Windows environment has no `bash` on PATH, but GitHub ShellCheck, Matrix Smoke, Test Linux integration-smoke, Python lint/type checks, Dashboard checks, env validation, PowerShell lint, and secret scan are all passing.

PR #2095, opened 2026-07-25
fix(updates): order pre-releases below the release they precede

tang-vu:
## Problem

`_build_version_result` compared versions by keeping only the fully-numeric dot components:

```python
current_parts = [int(x) for x in current.split(".") if x.isdigit()][:3]
```

The filter **drops** non-numeric components instead of stopping at them, so a digit that belongs to a pre-release identifier gets promoted into the release triple:

```
"2.6.0-rc.2".split(".")  ->  ["2", "6", "0-rc", "2"]
        isdigit filter   ->  ["2", "6",          "2"]  ->  [2, 6, 2]
```

An rc or beta build therefore reported itself as **newer** than every release it precedes:

```
current=2.0.0-beta.1  latest=2.0.0  update_available=False
current=2.0.0-beta.1  latest=2.0.1  update_available=False
current=2.6.0-rc.2    latest=2.6.0  update_available=False
current=2.6.0-rc.2    latest=2.7.0  update_available=False   # 2.7.0 vs parsed 2.6.2 -> only this one recovers
```

A machine left on a pre-release never saw the upgrade off it. `/releases/latest` excludes pre-releases on GitHub's side, so this only ever affected `current` — but the changelog shows ODS has shipped non-plain versions (`2.0.0-strix-halo`, `0.3.0-dev`), and any `-rc.N` / `-beta.N` build hits the promotion.

## Fix

- `_release_parts` parses the release triple up to the first non-numeric component and pads to three, so `2.6.0-rc.2` and `2.6.0+build.7` both give `[2, 6, 0]`.
- A pre-release sorts below the same release number, so the stable build of that version registers as an update.

Stable-to-stable comparison is unchanged, including the `2.9.0 -> 2.10.0` ordering that a string compare would get wrong.

## Tests

`tests/test_updates.py` gains a parametrised case over five pre-release and five stable pairs, plus a direct `_release_parts` check. Six fail on the previous parser.

```
$ python -m pytest extensions/services/dashboard-api/tests/test_updates.py -q
38 passed
```

### DIFF
--- a/ods/extensions/services/dashboard-api/routers/updates.py
+++ b/ods/extensions/services/dashboard-api/routers/updates.py
@@ -142,6 +142,27 @@ def _normalize_version(value: Optional[str]) -> str:
     return (value or "").strip().lstrip("v")
 
 
+def _release_parts(value: str) -> list[int]:
+    """Numeric release identifiers, stopping at the pre-release suffix.
+
+    A pre-release's identifiers are not part of the release number, so parsing
+    must stop at the first non-numeric component rather than skipping it.
+    Skipping promotes the digit in "2.6.0-rc.2" into the patch slot, making the
+    installed build read as 2.6.2 — newer than every release it precedes.
+    """
+    release = re.split(r"[-+]", value, maxsplit=1)[0]
+    parts: list[int] = []
+    for component in release.split("."):
+        if not component.isdigit():
+            break
+        parts.append(int(component))
+    return (parts + [0, 0, 0])[:3]
+
+
+def _is_prerelease(value: str) -> bool:
+    return "-" in value
+
+
 def _build_version_result(current: str, payload: Optional[dict]) -> dict:
     current = _normalize_version(current)
     result = {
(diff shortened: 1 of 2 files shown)


Lightheartdevs:
Thanks for fixing the prerelease ordering path in the dashboard update check.

What looks good:
- `ods/extensions/services/dashboard-api/routers/updates.py` now parses the release triple before `-`/`+` suffixes and treats a prerelease as lower than the matching stable release, which matches the upgrade behavior users need.
- The tests cover the exact promoted-digit bug (`2.6.0-rc.2` no longer parses as `2.6.2`) plus stable comparisons like `2.9.0` vs `2.10.0`.

Human decision point:
- CI has one red matrix job, `distro: cachyos`, but the failed log is a Docker Hub pull timeout for `cachyos/cachyos:latest` before the repo tests run. A maintainer should decide whether to rerun that job or accept it as infrastructure noise under the repo�s branch-protection policy.

What I checked:
- Reviewed the PR description, patch, `_release_parts`, `_is_prerelease`, and the added `ods/extensions/services/dashboard-api/tests/test_updates.py` cases.
- Checked CI: Python lint, mypy, dashboard API/frontend, shell lint, PowerShell lint, Linux integration smoke, the other distro matrix jobs, secret scan, and env validation are passing.
- Inspected the failed CachyOS job log; it failed while pulling the container image with Docker registry timeout errors, before project commands executed.

Lightheartdevs:
Thanks for tightening the dashboard update check around prerelease version ordering.

What looks good:
- `ods/extensions/services/dashboard-api/routers/updates.py` now stops parsing release parts at `-` or `+`, so `2.6.0-rc.2` no longer compares as `2.6.2`.
- The same-version prerelease-to-stable path now reports an update, while stable comparisons such as `2.9.0` to `2.10.0` remain covered by tests.

What I checked:
- Diff/static review of `routers/updates.py` and `tests/test_updates.py`.
- `gh pr checks 2095`: current head is green across dashboard `api`, integration smoke, Python lint/type, matrix smoke, and related gates.
- `git diff --check origin/main...refs/remotes/pr/2095`.

Thanks again for cleaning up this edge case.

PR #2096, opened 2026-07-25
fix(token-spy): stop the skills block inflating the file above it

tang-vu:
## Problem

`analyze_system_prompt` measures each workspace file from its `## FILE.md` heading up to the next known marker. The marker list is the workspace filenames plus three trailing sections:

```python
end_markers = ["Silent Replies", "Heartbeats", "Runtime"]
```

The skill-injection heading is missing from it, even though the same function measures that block separately a few lines below. When the skills block sits between the workspace files and `## Heartbeats` — where the agent emits it — the last workspace file's span runs straight through it, and those characters land in **two** buckets.

`base_prompt_chars` is computed as the total minus everything accounted, so the double count is subtracted twice. On a small sample prompt:

| field | before | after |
|---|---|---|
| `workspace_soul_chars` | 62 | 14 |
| `skill_injection_chars` | 48 | 48 |
| `base_prompt_chars` | 65 | 113 |

SOUL.md's actual body is 12 characters. The dashboard panel exists to show *where* the context window is going, and it was attributing the entire skill injection to whichever workspace file happened to precede it while under-reporting the base prompt by exactly that amount.

## Fix

- Hoist the heading to `SKILLS_SECTION_HEADING` and use it both as an end marker and in the block's own matcher, so the two can't drift apart again.
- The end-marker pattern used `\b`, which never matches after a heading ending in `)`. Replaced with a `(?=\W|$)` lookahead — same behaviour for the existing word-ending markers, and it works for `Skills (mandatory)`.

Prompts with no skills section are unaffected.

## Tests

New `tests/test_system_prompt_analysis.py`:
- the skills block is not folded into the preceding file,
- a file's measured size does not change when a skills block is added after it,
- the breakdown reconciles with the total, with and without a skills section,
- the empty-input contract.

Two fail on the previous marker list.

```
$ python -m pytest extensions/services/token-spy/tests/ -q
24 passed
```

Guarded with `pytest.importorskip("fastapi")` so it skips where the service deps are absent.

### DIFF
--- a/ods/extensions/services/token-spy/main.py
+++ b/ods/extensions/services/token-spy/main.py
@@ -481,6 +481,10 @@ async def on_shutdown():
 # ── Analysis ─────────────────────────────────────────────────────────────────
 
 # Map of known workspace filenames to their DB column names
+# Heading that opens the skill-injection block. Shared so the block is both
+# measured on its own and treated as a boundary for the workspace files above it.
+SKILLS_SECTION_HEADING = "Skills (mandatory)"
+
 WORKSPACE_FILE_MAP = {
     "AGENTS.md": "workspace_agents_chars",
     "SOUL.md": "workspace_soul_chars",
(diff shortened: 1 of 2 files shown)


Lightheartdevs:
Thanks for tightening Token Spy�s system-prompt section accounting here.

What looks good:
- `ods/extensions/services/token-spy/main.py` now treats `## Skills (mandatory)` as both its own measured section and a boundary for preceding workspace files, which fixes the double-counting path without changing prompts that have no skills section.
- Replacing the `\b` end-marker guard with a punctuation-safe lookahead is a good fit for headings like `Skills (mandatory)` while preserving the existing word-ending marker behavior.
- The new tests check the regression directly and also verify that the prompt breakdown reconciles back to `system_prompt_total_chars`.

What I checked:
- Reviewed the PR description, patch, and the surrounding `analyze_system_prompt` marker logic.
- Static review of `ods/extensions/services/token-spy/tests/test_system_prompt_analysis.py`, including the no-skills and empty-prompt contracts.

PR #2098, opened 2026-07-25
fix(dashboard): show days in the Settings uptime

tang-vu:
## Problem

Two pages render the same `uptime` value from the same `/api/settings/summary` field, with two different formatters.

`Dashboard.jsx` carries a day unit:

```js
if (d > 0) return `${d}d ${h}h ${m}m`
```

`Settings.jsx` stops at hours:

```js
const formatUptime = (secs = 0) => {
  const hours = Math.floor(secs / 3600)
  const mins = Math.floor((secs % 3600) / 60)
  return hours > 0 ? `${hours}h ${mins}m` : `${mins}m`
}
```

So a server up for three days reads **"3d 11h 20m"** on the Dashboard and **"83h 20m"** on Settings. For an always-on local box — which is what ODS is — the Settings figure keeps climbing into the hundreds of hours and stops being readable.

## Fix

Move the day-aware implementation into `src/utils/formatUptime.js` and import it from both pages, so the two cannot drift again. No change to the Dashboard's output.

One behaviour change on Settings: an absent uptime (`0`, `null`, a field the API did not return) now renders as an em dash instead of `"0m"`, matching what Dashboard already did for services that report none. The Settings tile falls back to `'Unknown'` only on a null-ish value, and an em dash is truthy, so it displays the dash — the value is missing rather than zero.

## Tests

New `src/utils/__tests__/formatUptime.test.js` covers the day, hour, minute, and absent branches, including the three-day case that motivated the fix.

```
$ npx vitest run --no-cache src/utils src/pages/Settings.test.jsx src/pages/Dashboard.test.jsx
Test Files  5 passed (5)
     Tests  43 passed (43)
$ npm run lint    # clean
$ npm run build   # ✓ built
```

### DIFF
--- a/ods/extensions/services/dashboard/src/pages/Dashboard.jsx
+++ b/ods/extensions/services/dashboard/src/pages/Dashboard.jsx
@@ -25,6 +25,7 @@ import {
 import { memo, useEffect, useMemo, useRef, useState } from 'react'
 import { Link } from 'react-router-dom'
 import { serviceUrl } from '../lib/serviceUrls'
+import { formatUptime } from '../utils/formatUptime'
 
 // Compute overall health from services (excludes not_deployed from counts)
 function computeHealth(services) {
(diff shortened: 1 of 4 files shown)


Lightheartdevs:
Thanks for cleaning up the Settings uptime display and removing the duplicate formatter.

What looks good:
- `ods/extensions/services/dashboard/src/utils/formatUptime.js` gives Dashboard and Settings one shared formatter for the same `/api/settings/summary` value, so long uptimes no longer drift between pages.
- The tests cover the important branches: multi-day output, sub-day output, sub-hour output, and absent values.
- The Settings behavior change from `0m` to an em dash for absent uptime is consistent with the Dashboard�s existing missing-value display.

What I checked:
- Reviewed the PR description, patch, and the previous duplicated implementations in `Dashboard.jsx` and `Settings.jsx`.
- Checked CI for this head: dashboard frontend/API, Linux integration smoke, matrix smoke, shell lint, PowerShell lint, Python lint, secret scan, and env validation are passing.
- Static review of the new `formatUptime` utility and `src/utils/__tests__/formatUptime.test.js` coverage.

PR #2099, opened 2026-07-25
fix(token-spy): store the MEMORY.md prompt bucket on SQLite

tang-vu:
## Problem

`usage.workspace_memory_chars` is declared in the SQLite schema (`db.py:58`) and the Postgres backend both writes and reads it (`db_postgres.py:167,196,251`), but the SQLite `log_usage` insert column list skips it (`db.py:122`).

`AnthropicProvider.WORKSPACE_FILE_MAP` emits `workspace_memory_chars` for every request whose system prompt carries a `MEMORY.md` section (`providers/anthropic.py:41`). On the default SQLite backend that value is dropped without an error and reads back as 0 forever, so the same agent shows a smaller system-prompt breakdown on SQLite than on Postgres.

## Fix

Add `workspace_memory_chars` to the insert column list.

## Tests

Two regression tests in `tests/test_usage_report.py`:
- a MEMORY.md round trip through `log_usage` → `SELECT`;
- a schema-driven check that every `workspace_*` column declared on the table is actually written, so the next schema addition cannot go missing the same silent way.

Both fail on `main` (`workspace_memory_chars` reads 0), pass with the fix. Full token-spy suite: 22 passed.

### DIFF
--- a/ods/extensions/services/token-spy/db.py
+++ b/ods/extensions/services/token-spy/db.py
@@ -119,7 +119,7 @@ def log_usage(entry: dict):
         "system_prompt_total_chars",
         "workspace_agents_chars", "workspace_soul_chars", "workspace_tools_chars",
         "workspace_identity_chars", "workspace_user_chars", "workspace_heartbeat_chars",
-        "workspace_bootstrap_chars",
+        "workspace_bootstrap_chars", "workspace_memory_chars",
         "skill_injection_chars", "base_prompt_chars",
         "conversation_history_chars",
         "input_tokens", "output_tokens", "cache_read_tokens", "cache_write_tokens",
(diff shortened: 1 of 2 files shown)


Lightheartdevs:
Thanks for closing the SQLite gap in Token Spy�s workspace prompt accounting.

What looks good:
- `ods/extensions/services/token-spy/db.py` now writes `workspace_memory_chars`, bringing SQLite in line with the declared schema, the Anthropic provider output, and the Postgres backend.
- The new tests cover both the concrete `MEMORY.md` round trip and the broader �every declared workspace column is written� contract, which is a good fit for avoiding another silent default-to-zero regression.
- The change is tightly scoped to the insert column list and usage-report tests, with no reporting API shape change.

What I checked:
- Reviewed the PR description, patch, SQLite schema/insert path, Postgres parity references, and Anthropic workspace-file mapping.
- Reviewed the added `ods/extensions/services/token-spy/tests/test_usage_report.py` assertions for the MEMORY bucket and schema-driven workspace coverage.

PR #2100, opened 2026-07-25
fix(installer): read the LAN IP by its route label, not field position

tang-vu:
## Problem

`_ods_lan_ip` (`ods/lib/qrcode.sh:13`) reads the host LAN address positionally:

```bash
lan_ip=$(ip route get 1 2>/dev/null | awk '{print $7; exit}')
```

Field 7 is the source address **only when the destination is reached through a gateway**:

```
1.1.1.1 via 192.168.1.1 dev eth0 src 192.168.1.42 uid 1000
                                      ^-- $7
```

A directly-connected destination prints no `via <gateway>`, which shifts every later field two positions left:

```
1.1.1.1 dev eth0 src 10.0.0.5 uid 1000
                                  ^-- $7
```

`_ods_lan_ip` then returns `1000`. That value feeds `print_success_card` and `print_dashboard_qr`, so the end of a successful install advertises `http://1000:3001` and encodes it into the QR code the user scans with their phone.

The repo already has the correct idiom — `installers/phases/06-directories.sh:583` keys off the `src` label when detecting `HOST_LAN_IP`. This brings `qrcode.sh` in line.

## Fix

- Parse by the `src` label instead of a fixed field number, and pin the lookup to `ip -4`.
- Reject anything that is not a dotted quad, so unparseable output (`RTNETLINK answers: Network is unreachable`) falls back to `localhost` rather than reaching the URL.

## Tests

New `ods/tests/test-lan-ip-detection.sh` (wired into `make test`) stubs `ip`/`ifconfig` and covers the gateway route, the on-link route, a route with no `src`, unparseable output, and the localhost fallback in the QR URL.

On `main` the on-link case fails with `expected 10.0.0.5, got '1000'`; all five pass with the fix.

### DIFF
--- a/ods/Makefile
+++ b/ods/Makefile
@@ -56,6 +56,9 @@ test: ## Run unit and contract tests
 	@echo "=== safe .env loading (quote/CRLF handling) ==="
 	@bash tests/test-safe-env.sh
 	@echo ""
+	@echo "=== LAN IP detection for the success card / QR ==="
+	@bash tests/test-lan-ip-detection.sh
+	@echo ""
 	@echo "=== Bind address sweep ==="
 	@bash tests/test-bind-address-sweep.sh
 	@bash tests/test-n8n-cookie-policy.sh
--- a/ods/lib/qrcode.sh
+++ b/ods/lib/qrcode.sh
@@ -10,12 +10,23 @@
 _ods_lan_ip() {
     local lan_ip=""
     if command -v ip >/dev/null 2>&1; then
-        lan_ip=$(ip route get 1 2>/dev/null | awk '{print $7; exit}')
+        # Key off the `src` label rather than a fixed field number. Only a
+        # route through a gateway prints `... via GW dev IF src ADDR`; a
+        # directly-connected route omits `via GW`, which shifts every later
+        # field two positions left and turns a positional read into whatever
+        # follows (the `uid` value). Matches the detection in
+        # installers/phases/06-directories.sh.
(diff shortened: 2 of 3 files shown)


Lightheartdevs:
Thanks for fixing the LAN-IP detection at the source of the success-card and QR URL behavior.

Quickest path to merge:
- Rebase or update the branch against current `main`; GitHub reports this PR as `mergeable_state: dirty`, so it cannot merge in its current state.
- Rerun the normal PR checks after the conflict is resolved. `gh pr checks 2100` currently reports no checks for this head, so CI has not validated the new `ods/tests/test-lan-ip-detection.sh` wiring in `ods/Makefile`.
- Preserve the `src`-label parsing in `ods/lib/qrcode.sh` during the update. That matches the repo�s existing route parsing direction and avoids the direct-route field-position bug described in the PR.

What I checked:
- Reviewed the PR description and patch for `ods/lib/qrcode.sh`, `ods/Makefile`, and the new LAN-IP detection test.
- Checked mergeability and checks: GitHub reports the branch dirty, and no checks are currently reported for this head.
- Compared the change against the recent main-branch direction around QR/LAN-IP parsing and installer success output.

Thanks again for tightening this path up.

tang-vu:
Closing as a duplicate — #2019 landed the same fix (same diagnosis: field 7 only holds the source address when the route has a gateway) before this one, and `ods/tests/test-qrcode-helper.sh` already covers it. Sorry for the noise.

PR #2101, opened 2026-07-25
fix(installer): normalize the install dir on hosts without GNU realpath

tang-vu:
## Problem

`normalize_path` (`ods/installers/lib/path-utils.sh`) probes for the *name* `realpath`, and only falls back to `grealpath` when no `realpath` exists at all:

```bash
if command -v realpath &>/dev/null; then
    realpath -m "$path" 2>/dev/null || echo "$path"
elif command -v grealpath &>/dev/null; then     # <- "GNU realpath via Homebrew (macOS)"
    grealpath -m "$path" 2>/dev/null || echo "$path"
```

macOS 12.3+ ships a **BSD** realpath under exactly that name. So on every currently-supported macOS:

1. the first branch always wins and the branch written for macOS is unreachable;
2. BSD realpath has no `-m` (`realpath: illegal option -- m`), so the call fails;
3. the `|| echo "$path"` fallback returns the raw string — no normalization at all.

`-m` is not incidental. The install dir does not exist yet during a first install, and BSD realpath refuses paths whose components are missing, so even a hypothetical `realpath` without the flag could not do this job.

`installers/macos/lib/constants.sh:33` derives `ODS_INSTALL_DIR` from this function, so whatever the operator typed — trailing slash, `.`, `..` — propagates verbatim into every derived path, the generated `.env`, and the launchd plists.

## Fix

- Probe for the working **flag**, not the name: try `grealpath -m`, then `realpath -m`, and accept the first that actually succeeds. Linux is unchanged (its `realpath` is GNU and answers `-m`).
- When no GNU realpath exists at all, normalize lexically (collapse `.`, `..`, duplicate and trailing slashes) instead of returning the string unchanged. Symlinks stay unresolved — the one thing that genuinely needs the filesystem.

Kept Bash 3.2-safe (macOS system bash); `make test`'s `test-bash32-guards.sh` still passes.

## Tests

New `ods/tests/test-path-normalization.sh` (wired into `make test`) stubs a BSD-style realpath and a host with no realpath, and asserts normalization in both, that `grealpath` is not shadowed, and that empty input still returns non-zero.

On `main` the first assertion fails: `bsd: trailing slash not stripped (got '/Users/example/ods/')`.

### DIFF
--- a/ods/Makefile
+++ b/ods/Makefile
@@ -66,6 +66,9 @@ test: ## Run unit and contract tests
 	@echo "=== mode switch status and switching ==="
 	@bash tests/test-mode-switch-status.sh
 	@echo ""
+	@echo "=== install-dir path normalization ==="
+	@bash tests/test-path-normalization.sh
+	@echo ""
 	@echo "=== Bind address sweep ==="
 	@bash tests/test-bind-address-sweep.sh
 	@bash tests/test-n8n-cookie-policy.sh
--- a/ods/installers/lib/path-utils.sh
+++ b/ods/installers/lib/path-utils.sh
@@ -12,36 +12,78 @@
 #   Add platform-specific path handling here.
 # ============================================================================
 
+# Collapse ".", "..", duplicate and trailing slashes without touching the
+# filesystem. Purely textual, so an install dir that does not exist yet
+# normalizes exactly like one that does.
+_normalize_path_lexically() {
+    local path="$1" segment
+    local -a segments=() kept=()
+    local saved_ifs="$IFS"
+    IFS='/' read -r -a segments <<< "$path"
+    IFS="$saved_ifs"
+
+    for segment in ${segments[@]+"${segments[@]}"}; do
+        case "$segment" in
+            "" | ".")
+                ;;
+            "..")
+                # A leading ".." has nothing to pop; drop it, since the caller
+                # already made the path absolute and /.. is /.
+                if [[ ${#kept[@]} -gt 0 ]]; then
+                    kept=(${kept[@]+"${kept[@]:0:${#kept[@]} - 1}"})
+                fi
+                ;;
+            *)
+                kept+=("$segment")
+                ;;
+        esac
+    done
+
+    local joined=""
+    for segment in ${kept[@]+"${kept[@]}"}; do
+        joined="$joined/$segment"
+    done
+    echo "${joined:-/}"
+}
+
 # Normalize a path (resolve symlinks, remove trailing slashes, make absolute)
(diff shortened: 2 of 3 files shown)


Lightheartdevs:
Thanks for taking on the macOS install-dir normalization path here; the `realpath -m` capability probe is the right shape for this installer edge case.

Quickest path to merge:
- Rebase or update the branch against current `main`; GitHub currently reports this PR as `mergeable_state: dirty` / `mergeable: false`, so it cannot merge as-is.
- After resolving the conflict, rerun the normal PR checks. GitHub currently reports no checks for this head, so there is not yet CI evidence for the new `ods/tests/test-path-normalization.sh` entry in `ods/Makefile`.
- Keep an eye on `ods/installers/lib/path-utils.sh` while resolving: the fallback normalization should stay Bash 3.2-compatible because macOS system bash is still part of this repo�s installer contract.

What I checked:
- Reviewed the PR description, patch, and the path normalization helper added under `ods/installers/lib/path-utils.sh`.
- Checked GitHub mergeability through the PR API: `mergeable=false`, `mergeable_state=dirty`, `rebaseable=false`.
- Checked PR checks with `gh pr checks 2101`: no checks are currently reported for this branch.

Thanks again for narrowing this down to the installer utility layer.

Lightheartdevs:
Thanks for tightening the installer path normalization around the macOS/BSD `realpath` case.

What looks good:
- `ods/installers/lib/path-utils.sh` now probes for a working GNU-style `-m` implementation instead of trusting the `realpath` command name, which preserves the intended Homebrew `grealpath` path on macOS while keeping Linux behavior intact.
- The lexical fallback handles missing install directories by collapsing `.`, `..`, duplicate slashes, trailing slashes, `/`, and `/..` without filesystem dependence, which matches first-install requirements.
- `ods/tests/test-path-normalization.sh` covers BSD-style `realpath`, no-realpath fallback, `grealpath` precedence, and empty-input failure, and it is wired into `ods/Makefile`.

What I checked:
- Reviewed the updated patch, prior advisory thread, and current mergeability API result: `mergeable=true`, `rebaseable=true`.
- Checked GitHub status rollup for the current head: Dashboard, Ruff, ShellCheck, Linux integration, env validation, macOS smoke, distro matrix, PowerShell lint, and secret scan are green.
- Tried to run `ods/tests/test-path-normalization.sh` locally, but this Windows runner only exposes the WSL `bash.exe` relay and WSL has no `/bin/bash`, so local shell execution was unavailable here.

Thanks again for preserving the Bash 3.2-compatible installer surface.

PR #2102, opened 2026-07-25
fix(installer): stop the readiness summary flagging healthy containers

tang-vu:
## Problem

`_ods_readiness_container_state` returns whatever this reports:

```
{{if .State.Health}}{{.State.Health.Status}}{{else}}{{.State.Status}}{{end}}
```

So a container **with** a healthcheck yields `starting` / `healthy` / `unhealthy`, and one **without** yields `running`. The classifier only accepted `running`, `starting` and `host`:

```bash
elif [[ "$container_state" == "running" || "$container_state" == "starting" || "$container_state" == "host" ]]; then
    state="starting"
```

`healthy` falls through to the `else`, so **the container Docker actively vouches for ranks below one Docker only knows is up**, and the summary prints the self-contradicting line:

```
[!!] LiteLLM                      needs attention - container healthy, HTTP 404
```

The HTTP probe and the container's own healthcheck are not always the same endpoint — a service answering 404 on `/` while its own check passes is still coming up. The net effect is a false alarm at the end of an otherwise clean install, pointing the operator at a container that is fine. `installers/windows/lib/readiness-summary.ps1` carries the identical list and the identical bug.

## Fix

Group `healthy` with `running` on both platforms. `unhealthy` still needs an operator and is unchanged.

## Tests

- `tests/test-install-readiness-summary.sh`: adds a healthy container behind a 404 probe and an unhealthy one, asserting the first reads `starting` and the second keeps `needs attention`; also asserts the string `container healthy` never appears in the attention list.
- `tests/test-windows-readiness-summary.sh`: same two cases in the pwsh runtime block.

On `main` the Bash suite fails 2 assertions (`healthy container reads as starting`, `healthy container is never reported as needing attention`). Both suites are green with the fix; PowerShell output verified locally under Windows PowerShell 5.1.

### DIFF
--- a/ods/installers/lib/readiness-summary.sh
+++ b/ods/installers/lib/readiness-summary.sh
@@ -74,7 +74,14 @@ ods_readiness_summary() {
         if _ods_readiness_is_ready_code "$http_code"; then
             state="ready"
             detail="HTTP $http_code"
-        elif [[ "$container_state" == "running" || "$container_state" == "starting" || "$container_state" == "host" ]]; then
+        # `healthy` belongs here with `running`: docker inspect reports the
+        # health status when the container declares a healthcheck and the plain
+        # status otherwise, so leaving it out made the container Docker vouches
+        # for rank below one Docker only knows is up. The probe URL and the
+        # healthcheck are not always the same endpoint — a service answering
+        # 404 on / while its own check passes is starting up, not broken.
+        elif [[ "$container_state" == "running" || "$container_state" == "healthy" \
+             || "$container_state" == "starting" || "$container_state" == "host" ]]; then
             state="starting"
             detail="HTTP $http_code"
         elif [[ "$container_state" == "missing" || "$container_state" == "docker-unavailable" ]]; then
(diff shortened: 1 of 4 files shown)


Lightheartdevs:
Thanks for tightening up the install readiness summary on both the Bash and Windows paths.

What looks good:
- `ods/installers/lib/readiness-summary.sh` and `ods/installers/windows/lib/readiness-summary.ps1` now treat Docker `healthy` consistently with the existing `running`/`starting` bucket, while leaving `unhealthy` in the operator-attention path.
- The regression tests cover the exact false-positive case: a healthy container whose HTTP probe returns 404 no longer reports `container healthy` as needing attention.
- The change keeps Linux/Windows installer parity, which matches the installer architecture expectation that generated/readiness behavior stays aligned across platforms.

What I checked:
- Reviewed the PR description, patch, and surrounding readiness-summary classification behavior.
- Checked CI: shell lint, PowerShell lint on Ubuntu and Windows, dashboard API/frontend, Linux integration smoke, matrix smoke, secret scan, and env validation are passing for this head.
- Static review of the added Bash and PowerShell test cases, including the preserved `container unhealthy` assertion.

PR #2103, opened 2026-07-25
fix(ape): let a served circuit-breaker cooldown actually reopen the gate

tang-vu:
## Problem

`circuit_breaker_blocked` clears the trip when the cooldown expires but leaves the decisions that caused it in place:

```python
if tripped_until and now >= tripped_until:
    _state["breaker"]["tripped_until"] = 0.0    # decisions[] untouched
```

`window_seconds` is longer than `cooldown_seconds` by design (defaults: 300 vs 120), so those denials are still inside the scoring window. The first decision after the cooldown calls `record_breaker_decision`, which re-scores the same denials and trips the breaker again — **even when that decision was an allow**.

There is no way out of this loop from inside the cooldown: a blocked `/verify` returns at `main.py:804` *before* `record_breaker_decision`, so no fresh evidence is ever recorded. The breaker can never accumulate the `min_samples` of clean traffic it needs to prove recovery, and it re-trips on the first request after every cooldown until the original denials age out of `window_seconds` on their own.

Net effect: the breaker stays shut for `window_seconds`, not `cooldown_seconds`, and every agent tool call routed through APE is denied for that whole time.

## Fix

Clear the decision window when the cooldown is served. Reopening then rests on fresh evidence, which is what a cooldown is for. A system that is genuinely still failing re-trips as soon as `min_samples` new decisions arrive — covered by the second test.

## Tests

`ape/tests/test_main.py`:
- `test_circuit_breaker_serves_traffic_again_after_cooldown` — trips the breaker, serves the cooldown by rewinding `tripped_until` (no sleeping), then asserts **two** consecutive allowed calls succeed. The second is the regression: on `main` the first call feeds the breaker and slams it shut again.
- `test_circuit_breaker_retrips_on_fresh_denials_after_cooldown` — guards the other direction, so clearing the history cannot disarm the breaker.

On `main` the first test fails; the second passes. Full APE suite: 46 passed.

### DIFF
--- a/ods/extensions/services/ape/main.py
+++ b/ods/extensions/services/ape/main.py
@@ -583,6 +583,14 @@ def circuit_breaker_blocked(policy: dict, now: float) -> tuple[bool, str]:
                           f"{datetime.fromtimestamp(tripped_until, timezone.utc).isoformat()})")
         if tripped_until and now >= tripped_until:
             _state["breaker"]["tripped_until"] = 0.0
+            # Drop the decisions that caused the trip along with the trip.
+            # window_seconds is longer than cooldown_seconds by design, so
+            # keeping them means the first decision after the cooldown is
+            # re-scored against the same denials and re-trips immediately —
+            # the breaker can never accumulate the min_samples it needs to
+            # prove recovery, and stays open for window_seconds instead of
+            # cooldown_seconds. Reopening has to rest on fresh evidence.
+            _state["breaker"]["decisions"] = []
     return (False, "")
 
 
--- a/ods/extensions/services/ape/main.py
+++ b/ods/extensions/services/ape/main.py
@@ -338,6 +338,26 @@ def _coerce_state(raw: Any) -> dict[str, Any]:
     return state
 
 
+def _clear_expired_trip(cb: dict, now: float) -> None:
+    """Clear a served breaker cooldown together with the denials that caused it.
+
+    window_seconds is longer than cooldown_seconds by design, so keeping the
+    deny history past the cooldown means the first decision after it is
+    re-scored against those same denials and re-trips immediately — the breaker
+    can never accumulate the min_samples it needs to prove recovery, and stays
+    open for window_seconds instead of cooldown_seconds. Reopening has to rest
+    on fresh evidence.
+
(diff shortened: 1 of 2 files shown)


Lightheartdevs:
Thanks for taking on the APE circuit-breaker recovery path; the cooldown behavior described here is a real operator-facing failure mode.

Quickest path to merge:
- Update `_prune_state()` in `ods/extensions/services/ape/main.py` to clear `_state["breaker"]["decisions"]` when it clears an expired `tripped_until`. Right now `_prune_state()` can run during `_load_state()` and set `tripped_until` back to `0.0` before `circuit_breaker_blocked()` sees the expired trip, which bypasses the new decision-window reset.
- Add a regression test in `ods/extensions/services/ape/tests/test_main.py` for restart/load recovery: trip the breaker, persist or load state with an expired `tripped_until`, then verify the first allowed decision after load does not re-score the stale denials and re-open the breaker.
- Keep the existing fresh-denials test; it is the right guard that clearing the cooldown history does not permanently disable the breaker.

What I checked:
- Reviewed the PR description, patch, `circuit_breaker_blocked()`, `record_breaker_decision()`, `_prune_state()`, and `_load_state()`.
- Static review of the new cooldown tests; they cover the in-process path but not the load/prune path above.

Thanks again for pushing this forward; this looks close once the persisted-state path follows the same recovery rule.

Lightheartdevs:
Thanks for tightening the APE circuit-breaker cooldown path, including the restart case from the previous review.

What looks good:
- `ods/extensions/services/ape/main.py` now clears an expired trip through `_clear_expired_trip()` from both `circuit_breaker_blocked()` and `_prune_state()`, so a served cooldown does not immediately re-score stale denials after a request or after process restart.
- `ape/tests/test_main.py` covers normal cooldown recovery, retripping on fresh denials, and cooldown expiry across `load_state()`.

What I checked:
- Diff/static review of `ape/main.py` and `ape/tests/test_main.py`.
- `gh pr checks 2103`: current head is green across API tests, integration smoke, Python lint/type, matrix smoke, and related gates.
- `git diff --check origin/main...refs/remotes/pr/2103`.

Thanks again for closing the recovery loop.

PR #2104, opened 2026-07-25
fix(ape): stop the per-minute limiter retaining every session id

tang-vu:
## Problem

`_session_request_times` is keyed by the caller-supplied `session_id` and nothing ever removed an entry:

```python
key = session_id or "_global"
if key not in _session_request_times:
    _session_request_times[key] = deque()
```

Timestamps inside a deque are trimmed to the 60-second window, but the deque itself — and its key — lives for the life of the process. An agent framework that mints a session per conversation, or any caller sending arbitrary ids, grows the table without limit.

This is the one unbounded structure in the module. Every other one is explicitly capped: `_MAX_SAMPLES_PER_WINDOW`, `_MAX_BREAKER_SAMPLES`, `_MAX_PENDING_APPROVALS`, `_MAX_PENDING_GRANTS`, plus the `_prune_state` sweep for all of them.

## Fix

Sweep scopes whose whole minute window has aged out, at most once per window (entries only become collectable after 60 idle seconds, so sweeping more often just rescans live ones).

Only **dead** entries are dropped. A scope still holding timestamps is still enforcing a limit, and evicting it would hand its caller a fresh budget — flooding new session ids would become a way to clear an exhausted one. That rules out LRU/size-cap eviction here; sweeping only aged-out entries bounds the table by concurrent traffic instead of by uptime, with no such bypass.

## Tests

`ape/tests/test_main.py`:
- `test_rate_limit_table_does_not_retain_finished_sessions` — 200 finished sessions, aged past the window, leave exactly one live entry behind.
- `test_rate_limit_sweep_never_clears_a_live_budget` — an exhausted session stays exhausted after 500 fresh ids force a sweep. This is the guard against turning the fix into a limiter bypass.
- `test_sweep_keeps_scopes_with_a_live_window` — a scope inside its window survives a sweep.

On `main` the first and third fail (retention, and the helper does not exist); the bypass guard passes both before and after, as it should. Full APE suite: 47 passed.

### DIFF
--- a/ods/extensions/services/ape/main.py
+++ b/ods/extensions/services/ape/main.py
@@ -432,17 +432,46 @@ def save_state() -> None:
 # Legacy minute-window limiter (kept verbatim for the requests_per_minute
 # policy knob so existing deployments behave identically).
 _session_request_times: dict[str, deque] = {}
+_last_rate_limit_sweep: float = 0.0
+
+
+def sweep_idle_rate_limit_scopes(cutoff: float) -> None:
+    """Drop scopes whose whole minute window has aged out.
+
+    The table is keyed by the caller-supplied session id and nothing ever
+    removed an entry, so it grew once per distinct session for the life of
+    the process. Every other governance structure here is explicitly bounded
+    (_MAX_SAMPLES_PER_WINDOW, _MAX_BREAKER_SAMPLES, _MAX_PENDING_APPROVALS,
+    _MAX_PENDING_GRANTS); this one was not.
+
+    Only fully-aged scopes are dropped. A scope still holding timestamps is
+    still enforcing a limit, and evicting it would hand its caller a fresh
+    budget — flooding new session ids would become a way to clear an
+    exhausted one. Sweeping only dead entries bounds the table by concurrent
+    traffic instead of by uptime, with no such bypass.
+    """
+    for key in [key for key, times in _session_request_times.items()
+                if not times or times[-1] < cutoff]:
+        del _session_request_times[key]
 
 
 def check_rate_limit(policy: dict, session_id: Optional[str]) -> bool:
     """Return True if the request is within the legacy per-minute limit."""
+    global _last_rate_limit_sweep
     limit = policy.get("rate_limit", {}).get("requests_per_minute", RATE_LIMIT)
     key = session_id or "_global"
+    now = time.monotonic()
+    cutoff = now - 60.0
+
+    # One pass per window at most: entries only become collectable after 60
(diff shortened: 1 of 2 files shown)


Lightheartdevs:
Thanks for bounding the APE per-minute limiter�s session table without turning eviction into a rate-limit bypass.

What looks good:
- `ods/extensions/services/ape/main.py` only sweeps scopes whose whole 60-second window has aged out, so live/exhausted session budgets are preserved instead of reset by eviction.
- The tests cover the retention problem, the bypass guard for an exhausted live scope, and the direct helper behavior for scopes still inside their window.
- The sweep cadence is conservative: at most once per window, which keeps the cleanup proportional to when entries can actually become collectable.

Human decision point:
- CI is not fully green at review time: `integration-smoke` is still pending, and `distro: cachyos` failed before repo tests ran because Docker could not pull `cachyos/cachyos:latest` from Docker Hub. A maintainer should wait for the pending check and decide whether to rerun/waive the CachyOS infrastructure failure.

What I checked:
- Reviewed the PR description, patch, `check_rate_limit()`, `sweep_idle_rate_limit_scopes()`, and the existing rate-limit call path.
- Checked CI: Python lint, mypy, dashboard API/frontend, shell lint, PowerShell lint, Linux smoke, most distro matrix jobs, secret scan, and env validation are passing; `integration-smoke` was pending at the last refresh.
- Inspected the failed CachyOS job log; it failed while pulling the job container image, before project commands executed.

tang-vu:
The one red check is a Docker Hub registry timeout while pulling the distro image in `matrix-smoke` (`Error response from daemon: Get "https://registry-1.docker.io/v2/": Client.Timeout exceeded`), before any repo code runs. Unrelated to this diff — needs a rerun, which I don't have rights to trigger.

Lightheartdevs:
Thanks for bounding the APE per-minute limiter's session table without turning cleanup into a rate-limit bypass.

What looks good:
- `ods/extensions/services/ape/main.py` sweeps only fully aged-out session scopes, so live and exhausted minute windows keep enforcing their budgets.
- The new tests cover table retention, the bypass guard for an exhausted live scope, and the helper behavior for scopes still inside the 60-second window.

What I checked:
- Diff/static review of `ape/main.py` and `ape/tests/test_main.py`.
- `gh pr checks 2104`: current head is green across API tests, integration smoke, Python lint/type, matrix smoke, and related gates.
- `git diff --check origin/main...refs/remotes/pr/2104`.

Thanks again for keeping the fix scoped to the real retention problem.

PR #2105, opened 2026-07-25
fix(ape): make path_guard deny a destination it cannot read

tang-vu:
## Problem

`path_guard` reads the destination from three fixed keys and **allows** the call when none of them is present:

```python
path = str(args.get("path", args.get("file", args.get("filename", ""))))
if not path:
    return True, "no path specified"
```

`classify_intent` routes any tool whose name contains a write verb (`write`, `create`, `append`, `save`, `put`) into `WriteFile`, which the shipped `config/ape/policy.yaml` puts under `path_guard`. So a tool that names its destination anything else — `dest`, `target`, `output_path`, `destination` — reaches the guard with nothing to check and is waved through, never compared against `allowed_paths`.

```
tool_name: save_output
args:      {"destination": "/etc/cron.d/backdoor", "content": "..."}
→ intent WriteFile, decision allow, reason "no path specified"
```

The `allowlist` branch immediately above already treats the equivalent case the other way:

```python
command = args.get("command", args.get("cmd", ""))
if not command:
    return False, "empty command denied"
```

Two modes in the same function, same situation, opposite answers — and the guard is the one that fails open.

## Fix

Deny, and name the keys that were looked for so an operator can either fix the tool contract or change the mode.

**Behaviour change:** a tool classified as `WriteFile` that genuinely carries no destination argument now gets denied instead of allowed. That is the intended direction for a guard mode, and the reason string says exactly what to do about it.

## Tests

`ape/tests/test_main.py` gains a `path_guard` block — the mode had no coverage at all:
- write inside the zone is allowed (positive control; skipped on Windows since the `allowed_paths` prefix match is POSIX-separator only and APE ships as a Linux container);
- write outside the zone is denied;
- **a destination under an unrecognised key is denied** — the regression;
- an empty `path` value is denied.

On `main` the last two fail. Full APE suite: 47 passed, 1 skipped.

### DIFF
--- a/ods/extensions/services/ape/main.py
+++ b/ods/extensions/services/ape/main.py
@@ -685,7 +685,16 @@ def evaluate(intent: str, tool_name: str, args: dict, policy: dict) -> tuple[boo
     if mode == "path_guard":
         path = str(args.get("path", args.get("file", args.get("filename", ""))))
         if not path:
-            return True, "no path specified"
+            # Fail closed, like the empty-command case in allowlist above. A
+            # write whose destination this guard cannot see is precisely what
+            # it exists to stop: classify_intent routes any tool named for a
+            # write verb here, so args naming the destination something else
+            # (dest, target, output_path) would otherwise walk straight past
+            # allowed_paths.
+            return False, (
+                "no readable destination in args (expected path, file, or "
+                "filename); path_guard cannot verify the write"
+            )
         real = os.path.realpath(path)
         allowed_paths = intent_policy.get("allowed_paths", [])
         if any(real == p or real.startswith(p.rstrip("/") + "/") for p in allowed_paths):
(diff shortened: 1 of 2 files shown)


Lightheartdevs:
Thanks for tightening the APE `path_guard` behavior; failing closed is the right default for this policy boundary.

What looks good:
- `ods/extensions/services/ape/main.py` now denies write-classified tool calls when the guard cannot find a readable destination key, rather than letting unrecognized destination names bypass `allowed_paths`.
- The new reason string tells operators exactly which argument names the guard inspected, which should make tool-contract fixes straightforward.
- `ods/extensions/services/ape/tests/test_main.py` adds positive and negative coverage for the mode, including the regression where `destination` used to walk past the guard.

What I checked:
- Reviewed the patch against APE's current intent classification and the existing fail-closed behavior for empty allowlist commands.
- CI is green for this head, including Ruff, mypy, API tests, integration smoke, and distro/env checks.

PR #2106, opened 2026-07-25
fix(support-bundle): redact collected files with the same secret set as .env

tang-vu:
## Problem

`scripts/ods-support-bundle.sh` has two redactors with two different secret-word sets.

`write_redacted_env` (`.env` only) was already widened, with this comment:

> `USER|EMAIL|BEARER` cover schema `secret:true` keys the old pattern missed — `N8N_USER`, `LANGFUSE_INIT_USER_EMAIL`, `LANGFUSE_MINIO_ROOT_USER` — which are published in cleartext when this `env.redacted` is shared on a public issue.

`redact_file` — which runs over **every other file in the archive** (`collect_shell` output, `copy_if_exists` copies, `write_file` content) — kept the narrow set:

```python
secret_word = r"(?:KEY|TOKEN|SECRET|PASSWORD|PASS|SALT|AUTH|CREDENTIAL)"
```

That is the path that matters most. `validation/compose-config.txt` is the rendered compose config, which prints every environment value; `docker/info.txt`, `diagnostics/ods-doctor.log` and the container logs capture command output the same way. So the exact keys the `.env` redactor was fixed for shipped in cleartext one directory over, inside the archive users are told to attach to public issues.

Reproduced on `main` with a stubbed docker whose `info` echoes resolved env values:

```
--- docker/info.txt on main ---
N8N_USER=support-bundle-super-secret
LANGFUSE_MINIO_ROOT_USER=support-bundle-super-secret
DASHBOARD_API_KEY=[REDACTED]
```

## Fix

Define the word set once (`ODS_SUPPORT_SECRET_WORDS`) and pass it to both redactors as argv, so they cannot drift again. `REDACTION_VERSION` goes to `2` so a bundle records which set produced it.

Over-masking is the intended direction here — the same trade the `.env` redactor and `ods config show` already make.

## Tests

`tests/test-support-bundle.sh` gains a run with a stubbed `docker` whose output echoes resolved env values, then asserts:
- no schema-secret value survives anywhere in the collected bundle (fails on `main`);
- `DASHBOARD_API_KEY` is still redacted, so the original keyword set was not lost.

### DIFF
--- a/ods/scripts/ods-support-bundle.sh
+++ b/ods/scripts/ods-support-bundle.sh
@@ -2,7 +2,17 @@
 set -euo pipefail
 
 TOOL_VERSION="1"
-REDACTION_VERSION="1"
+REDACTION_VERSION="2"
+
+# Key-name substrings that mark a value as secret, shared by BOTH redactors in
+# this script: redact_file (every collected file) and write_redacted_env (.env).
+# They must not drift — validation/compose-config.txt is `docker compose
+# config` output, which renders every environment value, so anything the file
+# redactor misses ships in cleartext inside the archive users attach to public
+# issues. USER|EMAIL|BEARER are here because .env.schema.json marks N8N_USER,
+# LANGFUSE_INIT_USER_EMAIL and LANGFUSE_MINIO_ROOT_USER secret:true; this
+# mirrors that set and the CLI's config-show masking.
+ODS_SUPPORT_SECRET_WORDS="KEY|TOKEN|SECRET|PASSWORD|PASS|SALT|AUTH|CREDENTIAL|USER|EMAIL|BEARER"
 DEFAULT_LOG_TAIL=200
 MAX_LOG_CONTAINERS=25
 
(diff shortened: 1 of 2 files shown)


Lightheartdevs:
Thanks for taking on the support-bundle redaction parity issue; this closes a practical public-issue leakage path.

What looks good:
- `ods/scripts/ods-support-bundle.sh` now uses one `ODS_SUPPORT_SECRET_WORDS` definition for both `.env` redaction and collected-file redaction, which removes the drift that let schema-secret `USER`/`EMAIL` values survive outside `env.redacted`.
- Bumping `REDACTION_VERSION` gives maintainers useful evidence about which masking rules produced a shared bundle.
- `ods/tests/test-support-bundle.sh` adds a realistic collected-command fixture that echoes resolved env values and verifies no schema-secret value remains anywhere in the bundle while preserving existing key redaction.

What I checked:
- Reviewed the PR description and patch against the repo's support-bundle, config-show, and installer redaction conventions.
- CI is green for this head, including shell lint, secret scan, integration smoke, support-bundle coverage through the Linux test job, and distro/env checks.

PR #2107, opened 2026-07-25
fix(token-spy): count MEMORY.md as its own workspace section

tang-vu:
## Problem

`analyze_system_prompt` (the live Anthropic proxy route, `main.py:673`) measures each workspace file **from its own marker up to the next _known_ marker**:

```python
all_file_names = list(WORKSPACE_FILE_MAP.keys())
...
content_end = file_positions[i + 1][0] if i + 1 < len(file_positions) else len(after_ctx)
```

So `WORKSPACE_FILE_MAP` is a **section-boundary table**, not just a list of metrics. `MEMORY.md` is missing from it, while `providers/anthropic.py:41` and both usage schemas (`db.py:58`, `db_postgres.py:167`) carry it.

The consequence is not a metric that reads zero — it is a metric attributed to the **wrong file**. Without `MEMORY.md` as a boundary, its entire body folds into whichever workspace file precedes it in the prompt.

Measured on `main` with a 400-char SOUL.md followed by a 900-char MEMORY.md:

```
workspace_soul_chars = 1315   (expected 401)
```

The agent's memories are silently billed to SOUL.md, and the two breakdowns in this service disagree about the same prompt depending on which path served the request.

## Fix

Add the marker. One line, restoring parity with the provider map and the usage columns.

## Tests

New `tests/test_system_prompt_analysis.py` — the function had no coverage:
- each marker gets its own bucket, with SOUL.md/MEMORY.md/HEARTBEAT.md adjacent (this is the misattribution case);
- `main.WORKSPACE_FILE_MAP` equals `AnthropicProvider.WORKSPACE_FILE_MAP`, so the two paths cannot drift again;
- `base_prompt_chars` still nets out every workspace bucket.

On `main` the first two fail. Full token-spy suite: 23 passed.

Independent of #2099, which fixes the SQLite insert for the same column.

### DIFF
--- a/ods/extensions/services/token-spy/main.py
+++ b/ods/extensions/services/token-spy/main.py
@@ -489,6 +489,12 @@ WORKSPACE_FILE_MAP = {
     "USER.md": "workspace_user_chars",
     "HEARTBEAT.md": "workspace_heartbeat_chars",
     "BOOTSTRAP.md": "workspace_bootstrap_chars",
+    # Every marker here is also a section boundary: a file's content is measured
+    # up to the next KNOWN marker, so omitting one does not merely lose its
+    # bucket — it folds that file's whole body into the bucket of whichever
+    # workspace file precedes it. Keep in step with the usage table columns and
+    # providers/anthropic.py's WORKSPACE_FILE_MAP.
+    "MEMORY.md": "workspace_memory_chars",
 }
 
 
--- a/ods/extensions/services/token-spy/tests/test_system_prompt_analysis.py
+++ b/ods/extensions/services/token-spy/tests/test_system_prompt_analysis.py
@@ -1,0 +1,86 @@
+"""Token Spy system-prompt breakdown tests.
+
+analyze_system_prompt measures each workspace file from its own marker up to
+the next KNOWN marker, so the marker table is a section-boundary table, not
+just a list of metrics. A file missing from it silently folds its whole body
+into the preceding file's bucket.
+"""
+
+from __future__ import annotations
+
+import importlib.util
+import sys
+from pathlib import Path
+from uuid import uuid4
+
+import pytest
+
+TOKEN_SPY_DIR = Path(__file__).resolve().parent.parent
+
+
+@pytest.fixture(scope="module")
+def token_spy_main():
+    sys.path.insert(0, str(TOKEN_SPY_DIR))
+    spec = importlib.util.spec_from_file_location(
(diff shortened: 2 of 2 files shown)


tang-vu:
The one red check is a Docker Hub registry timeout while pulling the distro image in `matrix-smoke` (`Error response from daemon: Get "https://registry-1.docker.io/v2/": Client.Timeout exceeded`), before any repo code runs. Unrelated to this diff — needs a rerun, which I don't have rights to trigger.

Lightheartdevs:
Thanks for narrowing this Token Spy accounting mismatch down to the missing `MEMORY.md` section boundary.

What needs decision:
- The code change itself looks ready: `ods/extensions/services/token-spy/main.py` brings the proxy route marker table back into parity with `providers/anthropic.py`, and the new `ods/extensions/services/token-spy/tests/test_system_prompt_analysis.py` covers the adjacent `SOUL.md`/`MEMORY.md`/`HEARTBEAT.md` attribution case.
- One CI check is still red, but the failed `distro: rocky-9` job never reached repository code; it failed during GitHub's job-container startup while pulling `rockylinux:9` from Docker Hub with connection timeouts. A maintainer should rerun or decide whether that infrastructure failure can be ignored under the branch protection policy.

What I checked:
- Reviewed the PR description, patch, existing Token Spy workspace maps, and the added tests.
- Checked the failed job log for run `30168899516`, job `89706482638`; the failure is repeated `docker pull rockylinux:9` timeout before checkout or test execution. Other checks on this head passed, including Ruff, mypy, API/frontend, integration smoke, and the remaining distro jobs.

Lightheartdevs:
Thanks for tracking the Token Spy accounting mismatch down to the missing `MEMORY.md` section boundary.

What looks good:
- `ods/extensions/services/token-spy/main.py` brings the proxy route marker table back into parity with the Anthropic provider map, so `MEMORY.md` is measured as its own workspace bucket instead of folding into the prior file.
- `tests/test_system_prompt_analysis.py` covers adjacent `SOUL.md` / `MEMORY.md` / `HEARTBEAT.md` sections, provider-map parity, and base-prompt netting.

What I checked:
- Diff/static review of `token-spy/main.py` and the new Token Spy prompt-analysis tests.
- `gh pr checks 2107`: current head is green across API tests, integration smoke, Python lint/type, matrix smoke, and related gates.
- `git diff --check origin/main...refs/remotes/pr/2107`.

Thanks again for making the accounting contract explicit.

PR #2108, opened 2026-07-25
fix(updates): reject GitHub errors and malformed release payloads

tang-vu:
## Summary

- reject non-2xx GitHub Releases responses before reading their JSON bodies
- validate the latest-release object once for both /api/version and /api/update/dry-run
- validate and normalize every /api/releases/manifest item, including GitHub's legitimate null optional fields
- keep stale-cache and local-manifest fallbacks intact instead of converting upstream failures into false no-update results or HTTP 500s

## Why this matters

All three authenticated dashboard update paths call GitHub's Releases API. GitHub rate-limit, 404, and 5xx bodies are valid JSON. Before this change, /api/version could cache an empty release for five minutes, dry-run could report no update with no error, and a malformed manifest item could escape as AttributeError. Operators could therefore receive a false up-to-date result precisely when GitHub lookup failed.

Reachable production path: dashboard version badge or update preview -> dashboard-api updates router -> GitHub Releases API -> cached/computed update state shown to the operator.

## Root cause and invariant

Root cause: the consumers parsed JSON before enforcing HTTP success, and each endpoint applied a different incomplete shape check.

Invariant: a GitHub response may affect update state only when its HTTP status is successful and its payload matches the expected release schema. Upstream errors must use the existing explicit stale/local/error path.

## Implementation

- Add one latest-release parser shared by the cache refresh and dry-run paths.
- Require a non-empty normalized tag_name and a string-or-null html_url.
- Add a release-list parser that requires object entries, validates field types, accepts null optional strings, and preserves changelog truncation.
- Type the in-memory version cache so the touched update path passes focused mypy checks.
- Continue catching only the existing narrow I/O and decoding exception classes.

## Overlap check

Searched open and closed PRs by semantic release/update terms and by the changed production files routers/updates.py and tests/test_updates.py.

- #2468 is a later same-author PR over the same files. This PR is now the stricter superset: it also enforces HTTP status, requires a usable tag, covers release-shaped error bodies, and retains all useful malformed-field coverage. #2468 was closed after this canonical PR passed all CI because no independent behavior remained.
- #2341 covers GitHub body: null. Its stronger public-boundary regression was ported here; this parser preserves that valid GitHub contract.
- #2095 changes semantic-version ordering after a valid payload is received, so its behavior is independent even though it touches the same files.
- #2822 and #2550 concern local manifest/version-file reads, not remote response validation.

No new PR was opened; this updates the earlier canonical #2108.

## Regression tests

The tests exercise authenticated HTTP endpoints rather than implementation-only helpers:

- /api/version rejects 403, missing-tag, and list payloads without poisoning the cache, while a real release still caches.
- /api/releases/manifest rejects a release-shaped 403 and non-object entries, falls back to the installed version, and accepts null optional GitHub fields.
- /api/update/dry-run rejects a release-shaped 403, missing tag, and list payload while returning an explicit version_check_error.

Red evidence before the endpoint-wide fix: 5 new boundary tests failed while 33 passed. Green result: 39 passed.

## Validation

- python -m pytest tests/test_updates.py -q: 39 passed
- full dashboard-api suite: 2,112 passed, 14 skipped; one unrelated Windows-host failure because the account lacks symlink privilege (WinError 1314 in test_user_extensions.py)
- Ruff CI command over ods/: passed
- focused mypy with imports skipped: no issues in the two changed files
- workflow-style mypy reaches one pre-existing config.py:466 annotation error; no touched-file error remains
- Python compile check: passed
- git diff --check: passed
- GitHub CI at f03ce0bd: all required checks passed, including API, Ruff, mypy, integration-smoke, PowerShell on Windows and Ubuntu, and the 10-distro matrix

## Cross-platform and rollback

The change is pure Python at the HTTP boundary and adds no platform-specific filesystem behavior. Focused tests passed on Windows; GitHub CI covers the Linux runtime. There is no persisted-state or API-schema migration. Reverting the two PR commits restores the previous permissive parsing behavior.

### DIFF
--- a/ods/extensions/services/dashboard-api/routers/updates.py
+++ b/ods/extensions/services/dashboard-api/routers/updates.py
@@ -178,9 +178,23 @@ async def _refresh_release_cache() -> Optional[dict]:
                 f"{_GITHUB_RELEASES_API}/latest",
                 headers=_GITHUB_HEADERS,
             )
+        # An error body is still valid JSON. GitHub answers an exhausted
+        # unauthenticated rate limit with 403 and {"message": ...}, which has
+        # no tag_name — so without these checks the empty result is cached for
+        # the full TTL and also becomes the allow_stale fallback, and the
+        # dashboard reports "up to date" while a release exists. The releases
+        # manifest below already validates its response this way.
+        response.raise_for_status()
         data = response.json()
+        if not isinstance(data, dict):
+            raise httpx.HTTPError(
+                f"unexpected release response: {type(data).__name__}"
+            )
+        tag = str(data.get("tag_name") or "").strip()
+        if not tag:
+            raise httpx.HTTPError("release response carried no tag_name")
         payload = {
-            "latest": data.get("tag_name", "").lstrip("v"),
+            "latest": tag.lstrip("v"),
             "changelog_url": data.get("html_url"),
             "checked_at": datetime.now(timezone.utc).isoformat(),
         }
--- a/ods/extensions/services/dashboard-api/routers/updates.py
+++ b/ods/extensions/services/dashboard-api/routers/updates.py
@@ -7,7 +7,7 @@ import re
 import time
 from datetime import datetime, timezone
 from pathlib import Path
-from typing import Optional
+from typing import Optional, TypedDict
 
 import httpx
 from fastapi import APIRouter, Depends, HTTPException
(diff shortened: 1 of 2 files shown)


Lightheartdevs:
Thanks for cleaning up the dashboard update lookup path; this is exactly the kind of small cache bug that can become confusing operationally.

What looks good:
- `ods/extensions/services/dashboard-api/routers/updates.py` now rejects non-2xx, non-dict, and missing-`tag_name` GitHub responses before writing `_version_cache`, so an API error body no longer becomes an authoritative “no update” result.
- The error handling stays inside the existing `httpx.HTTPError` fallback path, which fits the router's current stale-cache behavior instead of adding a new user-visible failure mode.
- `ods/extensions/services/dashboard-api/tests/test_updates.py` covers the rate-limit body, missing tag, list body crash case, and the successful release response.

What I checked:
- Reviewed the patch against the existing release manifest validation behavior in the same router.
- CI is green for this head, including Ruff, mypy, dashboard API tests, integration smoke, and distro/env checks.

PR #2109, opened 2026-07-25
fix(installer): close the redaction gaps in the compose failure report

tang-vu:
## Problem

`write_compose_failure_report` produces a file whose own header says *"Review before posting publicly"*, and the installer prints its path whenever compose startup fails — it exists to be attached to an issue. Two things reached it unredacted.

**1. The keyword set stops short.**

```awk
secret_re = "(key|token|secret|password|pass|salt|auth|credential)"
```

No `user|email|bearer`, so the schema-secret keys `N8N_USER`, `LANGFUSE_INIT_USER_EMAIL` and `LANGFUSE_MINIO_ROOT_USER` (`.env.schema.json`, `secret: true`) survive the rendered `docker compose config` output. The same set gates the harvested-value pass, so those values are not scrubbed anywhere else in the report either. `scripts/ods-support-bundle.sh` already carries the wider set for exactly these keys.

**2. The installer log tail is piped in raw.**

```bash
tail -n 160 "$log_file"
```

It is installer output — it carries whatever the phases echoed, including env values — and it lands in the same shareable file as the compose config, which *is* filtered.

## Fix

Add `user|email|bearer` to the keyword set, and run the log tail through `_ods_report_redact_stream` like every other captured stream.

## Tests

`tests/test-compose-failure-report.sh` gains `N8N_USER` / `LANGFUSE_INIT_USER_EMAIL` to the `.env` fixture, the compose-config stub, and the installer log, then asserts both keys redact and neither value appears anywhere in the report.

Note the fixture values (`ods-owner-alice`, `alice@example.com`) deliberately contain no secret keyword — with a value like `super-secret-…` the line-level rule fires on the *value* and the key-name coverage goes untested.

All four new assertions fail on `main`. Suite: 17 passed.

### DIFF
--- a/ods/installers/lib/compose-failure-report.sh
+++ b/ods/installers/lib/compose-failure-report.sh
@@ -52,7 +52,11 @@ _ods_report_redact_stream() {
     local env_file="${1:-}"
     awk -v env_file="$env_file" '
         BEGIN {
-            secret_re = "(key|token|secret|password|pass|salt|auth|credential)"
+            # user|email|bearer are here for the same reason scripts/
+            # ods-support-bundle.sh carries them: .env.schema.json marks
+            # N8N_USER, LANGFUSE_INIT_USER_EMAIL and LANGFUSE_MINIO_ROOT_USER
+            # secret:true, and this report is written to be posted on an issue.
+            secret_re = "(key|token|secret|password|pass|salt|auth|credential|user|email|bearer)"
             if (env_file != "") {
                 while ((getline line < env_file) > 0) {
                     sub(/\r$/, "", line)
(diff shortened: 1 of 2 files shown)


Lightheartdevs:
Thanks for tightening the compose failure report redaction path here, especially the installer log tail that gets shared when startup fails.

What looks good:
- `ods/installers/lib/compose-failure-report.sh` now applies the same redaction stream to the installer log tail that it already uses for compose output, closing the obvious shareable-report leak path.
- The expanded key pattern covers the schema-secret `USER`/`EMAIL` cases called out in `.env.schema.json`, which matches the repo's recent direction around support-bundle and config-output redaction.
- `ods/tests/test-compose-failure-report.sh` exercises both key-name masking and raw value absence, so the regression is covered without depending on the value itself containing a secret-looking word.

What I checked:
- Reviewed the PR description, patch, and existing compose failure report flow against the installer/support-bundle redaction conventions on `origin/main`.
- CI is green for this head, including shell lint, secret scan, integration smoke, compose/env validation, and distro smoke jobs.

PR #2110, opened 2026-07-25
fix(dashboard-api): report non-JSON Token Spy responses as content-type warnings

tang-vu:
## Problem

`_fetch_token_spy_metrics()` listed its exception handlers in this order:

```python
except aiohttp.ClientError as e:        # broad
    logger.debug("Token Spy unavailable: %s", e)
except asyncio.TimeoutError:
    ...
except aiohttp.ContentTypeError as e:   # never reached
    logger.warning("Token Spy returned unexpected content type: %s", e)
```

`aiohttp.ContentTypeError` subclasses `ClientResponseError` → `ClientError`, so the content-type handler is dead code. A Token Spy container that is reachable and answers `200` with a non-JSON body (proxy error page, HTML login redirect, wrong `TOKEN_SPY_URL` pointing at another service) is reported at **debug** level as "Token Spy unavailable" instead of the **warning** that names the actual problem — so the operator sees empty agent metrics with nothing in the logs at default verbosity.

## Fix

Order the specific handler ahead of the broad one, with a comment recording why the order matters.

## Test

`test_content_type_error_handled` previously only asserted "does not raise", which passed with the handler dead. It now asserts the WARNING record is emitted.

Verified red on the pre-fix module (`assert []`) and green after.

### DIFF
--- a/ods/extensions/services/dashboard-api/agent_monitor.py
+++ b/ods/extensions/services/dashboard-api/agent_monitor.py
@@ -155,12 +155,16 @@ async def _fetch_token_spy_metrics() -> None:
                                len(data), total_out)
                 else:
                     logger.debug("Token Spy returned status %d", resp.status)
+    # ContentTypeError derives from ClientError, so it must be listed first:
+    # behind the broad handler it can never run, and a Token Spy response that
+    # is reachable but not JSON would be reported as "unavailable" at debug
+    # level instead of as the misconfiguration it is.
+    except aiohttp.ContentTypeError as e:
+        logger.warning("Token Spy returned unexpected content type: %s", e)
     except aiohttp.ClientError as e:
         logger.debug("Token Spy unavailable: %s", e)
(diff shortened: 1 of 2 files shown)


Lightheartdevs:
Thanks for making the Token Spy content-type failure visible at the right log level.

What looks good:
- `ods/extensions/services/dashboard-api/agent_monitor.py` now catches `aiohttp.ContentTypeError` before the broader `ClientError`, so a reachable non-JSON Token Spy response is reported as the misconfiguration it is.
- The behavior remains graceful: the metrics fetch still degrades without raising, but operators get a warning instead of only a debug "unavailable" line.
- `ods/extensions/services/dashboard-api/tests/test_agent_monitor.py` now asserts the warning, so the handler-order regression is actually covered.

What I checked:
- Reviewed the exception hierarchy issue, the agent monitor diff, and the updated test.

PR #2111, opened 2026-07-25
fix(lemonade): keep audio calls inside the classified-error contract

tang-vu:
## Problem

`LemonadeClient` documents itself as failing with `LemonadeClientError`, and that is what callers rely on — `routers/gpu.py` branches on `exc.kind` (`provider_unreachable`/`timeout`) and catches nothing else.

Two methods were outside that contract:

- **`speech()`** had no `try`/`except` at all. `client.post(...)` and `raise_for_status()` raised raw `httpx.ConnectError`, `httpx.ReadTimeout` or `httpx.HTTPStatusError`.
- **`transcribe_wav()`** caught only `httpx.HTTPStatusError`, so connection and timeout failures still escaped raw.

So an unreachable or slow Lemonade backend surfaced as an unhandled 500 with an httpx traceback rather than the reported "provider unreachable" state the rest of the client produces.

## Fix

Extract the existing `_request_json` classification into `_classified_error()` and route every `httpx.HTTPError` (status, timeout, transport) through it from all three request paths. Behaviour of `_request_json` is unchanged — `httpx.HTTPError` is the common base of `HTTPStatusError` and `RequestError`, and the classifier preserves the same order (status → timeout → unreachable).

## Tests

Five cases added in `tests/test_lemonade_client.py` covering `speech()` transport/timeout/status classification, `transcribe_wav()` transport classification, and the success path.

Verified red on the pre-fix module (4 failures) and green after.

### DIFF
--- a/ods/extensions/services/dashboard-api/lemonade_client.py
+++ b/ods/extensions/services/dashboard-api/lemonade_client.py
@@ -98,6 +98,27 @@ def classify_status(status_code: int) -> str:
     return "request_rejected"
 
 
+def _classified_error(exc: httpx.HTTPError) -> LemonadeClientError:
+    """Map an httpx failure onto the client's classified error contract.
+
+    Every public method must fail with LemonadeClientError: callers such as
+    the GPU router branch on ``exc.kind`` and catch nothing else, so a raw
+    httpx transport error escaping a request method becomes an unhandled 500
+    instead of a reported "unreachable" provider.
+    """
+    if isinstance(exc, httpx.HTTPStatusError):
+        payload = _json_payload(exc.response)
+        return LemonadeClientError(
+            classify_status(exc.response.status_code),
+            _payload_message(payload) or exc.response.text or str(exc),
+            status_code=exc.response.status_code,
+            payload=payload,
+        )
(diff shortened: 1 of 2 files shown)


Lightheartdevs:
Thanks for closing the Lemonade audio-client error-contract gap.

What looks good:
- `ods/extensions/services/dashboard-api/lemonade_client.py` now sends `speech()`, `transcribe_wav()`, and `_request_json()` through the same `httpx.HTTPError` classifier, matching the caller contract around `LemonadeClientError.kind`.
- Status, timeout, and transport failures keep distinct classifications, so GPU/router callers can report provider state instead of surfacing raw `httpx` exceptions as 500s.
- `ods/extensions/services/dashboard-api/tests/test_lemonade_client.py` covers transport, timeout, status, and success behavior for the audio path.

What I checked:
- Reviewed the Lemonade client diff, caller expectations described in the PR, and the focused tests.

tang-vu:
Superseded by #2332, which covers the same Lemonade audio error contract while also classifying malformed transcription responses and keeping the exception handlers narrow. Closing this duplicate so review and merge work stay on the newer branch.

PR #2112, opened 2026-07-25
fix(nvidia-topo): rank every NVLink connection above PCIe

tang-vu:
## Problem

`link_rank()` matched an enumerated list of NVLink values:

```bash
NV4 | NV6 | NV8 | NV12 | NV18)  echo 100 ;;   # NVLink gen2/3
NV1 | NV2 | NV3)                echo 80  ;;   # NVLink gen1
```

`nvidia-smi topo -m` reports `NV<n>` where **n is the number of NVLink connections in the pair**, and driver/board generations keep producing counts this list never had — `NV5`, `NV7`, `NV9`, `NV10`, `NV16`, `NV24`. Those fell through to `*) echo 0`.

Rank 0 is **below cross-NUMA PCIe** (`SYS` = 10) and below every other PCIe relationship. The rank flows into `.links[].rank` in the topology JSON, which `installers/phases/03-features.sh` loads into `LINK_RANK` and feeds to `scripts/assign_gpus.py` and `get_rank()`. So on hardware with an unlisted link count, a real NVLink pair is scored worse than a PCIe pair and loses multi-GPU assignment — while `link_label()` (which uses `NV*`) still reports the pair as "NVLink", so the mismatch is invisible in the topology output.

## Fix

Rank on the parsed link count rather than an enumerated list: `NV1`–`NV3` → 80, `NV4`+ → 100. Every previously enumerated value keeps its exact rank; `NV0` and non-numeric `NV` strings still fall through to 0.

## Test

`tests/test-nvidia-topo.sh` gains a `link_rank()`/`link_label()` unit test asserting (a) every `NV1`–`NV24` outranks `SYS` and `PIX`, (b) higher counts never rank below lower ones, (c) all previously enumerated ranks are unchanged.

Red on the pre-fix lib (`NV5/NV7/NV9/NV10/NV16/NV24 rank=0 <= SYS`), green after; the 7 existing fixture tests still pass.

### DIFF
--- a/ods/installers/lib/nvidia-topo.sh
+++ b/ods/installers/lib/nvidia-topo.sh
@@ -18,9 +18,20 @@
 
 link_rank() {
   case "$1" in
-  NV4 | NV6 | NV8 | NV12 | NV18)  echo 100 ;;   # NVLink gen2/3
+  # nvidia-smi reports NV<n>, where n is the number of NVLink connections
+  # between the pair. Rank on n instead of an enumerated list: driver and
+  # board generations keep introducing counts the list never had (NV5, NV9,
+  # NV10, NV16, NV24...), and an unlisted one fell through to rank 0 — below
+  # even cross-NUMA PCIe (SYS, rank 10). A genuine NVLink pair then lost GPU
+  # assignment to a PCIe pair.
+  NV[1-9] | NV[1-9][0-9])
+    if [[ "${1#NV}" -ge 4 ]]; then
+      echo 100                                  # NVLink gen2/3, 4+ links
+    else
+      echo 80                                   # NVLink gen1, 1-3 links
+    fi
+    ;;
   XGMI | XGMI2)                   echo 90  ;;   # AMD Infinity Fabric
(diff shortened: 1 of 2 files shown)


Lightheartdevs:
Thanks for fixing the NVLink ranking path in the NVIDIA topology helper.

What looks good:
- `ods/installers/lib/nvidia-topo.sh` now ranks `NV<n>` from the parsed count instead of a brittle enumerated list, so newer NVLink counts no longer fall below PCIe relationships.
- Previously known ranks stay stable (`NV1`-`NV3` at 80, `NV4+` at 100), so existing topology behavior is preserved while covering newer hardware.
- `ods/tests/test-nvidia-topo.sh` adds direct regression coverage for representative NV counts and guards that NVLink ranks above `SYS` and `PIX`.

What I checked:
- Reviewed the topology helper diff and the GPU-assignment risk against the repo's multi-GPU installer path.
- GitHub checks are green, including shell lint, integration smoke, distro smoke, and macOS smoke.

PR #2055, opened 2026-07-24
fix(workflows): handle directory paths and unicode errors safely in workflow catalog reader

vaibhavsrv:
## Problem

In `load_workflow_catalog()` (`routers/workflows.py`), the catalog path check used `WORKFLOW_CATALOG_FILE.exists()` and opened the catalog file without catching decoding errors:

```python
if not WORKFLOW_CATALOG_FILE.exists():
    return DEFAULT_WORKFLOW_CATALOG
try:
    with open(WORKFLOW_CATALOG_FILE) as f:
        ...
except (json.JSONDecodeError, OSError, KeyError) as e:
    ...
```

If `WORKFLOW_CATALOG_FILE` pointed to a directory (e.g. `catalog.json` created as a directory), `open()` raised `IsADirectoryError` on POSIX. Furthermore, if `WORKFLOW_CATALOG_FILE` contained invalid non-UTF8 binary bytes, `open()` raised `UnicodeDecodeError`, which could bypass `OSError` depending on environment context and crash the endpoint with HTTP 500.

## Fix

Update `load_workflow_catalog()` to check `.is_file()`, specify `encoding="utf-8"`, and include `UnicodeError` in the caught exception tuple:

```python
if not WORKFLOW_CATALOG_FILE.is_file():
    return DEFAULT_WORKFLOW_CATALOG
try:
    with open(WORKFLOW_CATALOG_FILE, encoding="utf-8") as f:
        ...
except (json.JSONDecodeError, OSError, UnicodeError, KeyError) as e:
    logger.warning("Failed to load workflow catalog from %s: %s", WORKFLOW_CATALOG_FILE, e)
    return DEFAULT_WORKFLOW_CATALOG
```

## Threat model (honest scoping)

This is a workflow catalog file reader safety fix. It guarantees that directory paths and corrupted binary files fall back to `DEFAULT_WORKFLOW_CATALOG` gracefully without raising unhandled 500 server errors.

## Verification

Added unit test in `tests/test_workflows.py`:

- `test_load_workflow_catalog_handles_directory_and_unicode_error`
  - Verified directory catalog paths and binary file contents fall back to default catalog cleanly.

- `pytest tests/test_workflows.py` passed (46 passed in 0.60s).

PR #2114, opened 2026-07-25
fix(dashboard-api): resolve n8n workflow identity without ambiguity

tang-vu:
## Why this matters

The dashboard identifies catalog workflows in n8n by name. The previous first-substring rule made the result depend on n8n list order: with catalog `Daily Digest`, an earlier user workflow named `Daily Digest Extended` could be shown, inspected, or deleted instead of the exact installed workflow. If several partial candidates existed, ODS still chose one arbitrarily.

This is a reachable destructive path through authenticated `DELETE /api/workflows/{id}` and `POST /disable`; the same identity is also consumed by the workflow list and executions endpoint.

## Root cause and invariant

Root cause: list, disable, and executions originally used inconsistent containment scans, then this PR centralized the scan but still returned the first containment candidate.

Invariant: all consumers resolve the same n8n identity; a unique exact case-insensitive name wins regardless of response ordering, a unique containment match remains available for older templates whose catalog and imported names differ, and ambiguous identities never authorize a mutation.

## What changed

- Route list, disable, and executions through one `find_installed_n8n_workflow()` helper.
- Prefer one exact case-insensitive match over partial matches.
- Preserve the legacy compatibility fallback only when containment identifies exactly one workflow.
- Refuse duplicate exact names and multiple partial matches instead of selecting by n8n response order.
- Add public-route regressions proving list/disable parity, exact-target deletion, refusal of ambiguous deletion, and executions parity.

## Overlap check

Searched open and closed PRs semantically for workflow substring/exact-name matching and scanned every open PR changing `routers/workflows.py`.

- #2749 prefers an exact name only inside `_remove_workflow`, but retains arbitrary substring fallback and leaves list/executions on different selection behavior. This older #2114 is the canonical broader change: it now ports the exact-match advantage, rejects ambiguity, and applies one invariant to all three consumers with route-level tests.
- #2748 only bounds the executions `limit` parameter.
- #2055, #2187, #2343, #2423, #2469, #2473, #2495, and #2964 cover catalog I/O, dependencies, response validation, metadata, or n8n activation contracts; none owns workflow identity selection.

## Focused validation

- `cd ods/extensions/services/dashboard-api && pytest tests/test_workflows.py -q` — **51 passed**
- `git diff --check` — passed (Git emitted only the existing Windows LF/CRLF checkout notice)

The regression exercises the authenticated HTTP deletion boundary and asserts the exact n8n ID in the outgoing DELETE URL. No live n8n instance was used; mocked transport proves selection and mutation targeting, not live n8n compatibility.

## Design, compatibility, and rollback

Unique containment is retained because imported template names can legitimately be shorter than catalog display names. Ambiguous names are reported as not installed rather than guessed; this is deliberately fail-closed for deletion. Operators can disambiguate by renaming duplicate/prefix-colliding n8n workflows.

Rollback is a normal revert of the two commits, but restores order-dependent identity selection and the wrong-workflow deletion risk. Human review is requested because this protects a destructive lifecycle action.

### DIFF
--- a/ods/extensions/services/dashboard-api/routers/workflows.py
+++ b/ods/extensions/services/dashboard-api/routers/workflows.py
@@ -26,6 +26,27 @@ def _validate_workflow_id(workflow_id: str) -> None:
         raise HTTPException(status_code=400, detail="Invalid workflow ID format")
 
 
+def find_installed_n8n_workflow(catalog_name: str, n8n_workflows: list[dict]) -> dict | None:
+    """Return the n8n workflow that corresponds to *catalog_name*, if any.
+
+    n8n stores whatever name the imported template carried, which is not
+    guaranteed to equal the catalog name, so the match is a containment test in
+    both directions. Every caller must use the same rule: the list endpoint
+    reporting a workflow as installed while disable/executions cannot find it
+    leaves the UI offering a Disable action that always 404s.
+    """
+    target = (catalog_name or "").lower()
+    if not target:
+        return None
+    for workflow in n8n_workflows:
+        name = (workflow.get("name") or "").lower()
+        if not name:
+            continue
+        if target in name or name in target:
+            return workflow
+    return None
+
+
 def load_workflow_catalog() -> dict:
     """Load workflow catalog from JSON file."""
     if not WORKFLOW_CATALOG_FILE.exists():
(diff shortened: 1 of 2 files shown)


Lightheartdevs:
Thanks for making the n8n workflow install lookup consistent across list, disable, and executions.

Decision point:
- The code change itself looks coherent: `ods/extensions/services/dashboard-api/routers/workflows.py` routes all three call sites through one `find_installed_n8n_workflow()` helper, and `ods/extensions/services/dashboard-api/tests/test_workflows.py` covers the shorter-n8n-name case across list, disable, and executions.
- One required check is red: `distro: opensuse-tw`. The job failed before repo tests ran because Docker could not pull `opensuse/tumbleweed:latest` from Docker Hub after retries (`context deadline exceeded` / timeout awaiting headers). This looks like external registry or runner flake rather than a PR regression.
- Quickest path is likely to rerun that failed distro job, then merge if it comes back green and no maintainer sees a reason to require broader workflow/n8n validation.

What I checked:
- Reviewed the workflow router diff, the new regression tests, and the current check rollup.
- Inspected the failed openSUSE job log; all other checks shown for the PR are green, including API, Python lint, mypy, integration smoke, other distro smoke jobs, and macOS smoke.

Lightheartdevs:
Thanks for making the n8n workflow installed-state lookup consistent across list, disable, and executions.

What looks good:
- `ods/extensions/services/dashboard-api/routers/workflows.py` now routes all three call sites through `find_installed_n8n_workflow()`, so the UI should not advertise a Disable action that the mutation path cannot resolve.
- The tests cover the shorter-n8n-name case across list, disable, and executions, which is the exact mismatch described in the PR.
- The existing broad containment heuristic is still a product tradeoff, but this PR does not broaden the list endpoint's matching rule; it makes the downstream actions agree with it.

What I checked:
- Diff/static review of `routers/workflows.py` and `tests/test_workflows.py`.
- `gh pr checks 2114`: current head is green across dashboard `api`, integration smoke, Python lint/type, matrix smoke, and related gates.
- `git diff --check origin/main...refs/remotes/pr/2114`.

Thanks again for closing the UI/action mismatch.

PR #2115, opened 2026-07-25
fix(dashboard-api): stop templates rejecting extensions with no declared GPU backends

tang-vu:
## Problem

`service.gpu_backends` is **optional** in `service-manifest.v1.json` (`required` is `id, name, port, type, category`). Two places disagree about what an undeclared list means.

`config.py` loads such a service as supported everywhere:

```python
supported = service.get("gpu_backends", ["amd", "nvidia", "apple"])   # inclusion decision
...
"gpu_backends": service.get("gpu_backends", []),                        # what it stores
```

`routers/extensions.py` reads the stored `[]` as "no declared restriction":

```python
gpu_backends = ext.get("gpu_backends", [])
if gpu_backends and "all" not in gpu_backends and GPU_BACKEND not in gpu_backends:
    return "incompatible"
```

`routers/templates.py` read the same `[]` as "supports nothing":

```python
gpu_backends = service_config.get("gpu_backends", ["amd", "nvidia", "apple"])  # dead default
if "all" in gpu_backends or GPU_BACKEND in gpu_backends:
    return None
return f"requires one of {gpu_backends}; ..."   # → "requires one of []"
```

The `.get` default never fires, because config.py always writes the key. So for an extension whose manifest omits `gpu_backends`:

- Extensions page → usable
- `POST /api/templates/{id}/preview` → listed under **incompatible**, with warning `requires one of []; current backend is nvidia`
- `POST /api/templates/{id}/apply` → **skipped**, and the same rule aborts activation of any template service that depends on it with a 424

All bundled manifests happen to declare the field, so this hits user/third-party extensions installed from the library or authored locally.

## Fix

Treat an empty or absent list as unrestricted in `_gpu_backend_error`, matching `_compute_extension_status`. A declared list is still enforced unchanged.

## Tests

- `test_gpu_backend_error_treats_undeclared_backends_as_unrestricted` — unit coverage of both directions.
- `test_template_preview_does_not_flag_undeclared_backends_incompatible` — the extension lands in `to_enable` with no warning.

Red on the pre-fix router (2 failures), green after; the other 31 tests in the file are unaffected.

### DIFF
--- a/ods/extensions/services/dashboard-api/routers/templates.py
+++ b/ods/extensions/services/dashboard-api/routers/templates.py
@@ -67,8 +67,13 @@ def _gpu_backend_error(service_id: str) -> str | None:
     if not service_config or GPU_BACKEND == "apple":
         return None
 
-    gpu_backends = service_config.get("gpu_backends", ["amd", "nvidia", "apple"])
-    if "all" in gpu_backends or GPU_BACKEND in gpu_backends:
+    # gpu_backends is optional in the manifest schema, and both config.py and
+    # the catalog generator store an undeclared list as []. An empty list means
+    # "no declared restriction", exactly as _compute_extension_status treats it
+    # — reading it as "supports nothing" would make the Extensions page and the
+    # template gate disagree about the same extension.
+    gpu_backends = service_config.get("gpu_backends") or []
+    if not gpu_backends or "all" in gpu_backends or GPU_BACKEND in gpu_backends:
         return None
     return f"requires one of {gpu_backends}; current backend is {GPU_BACKEND}"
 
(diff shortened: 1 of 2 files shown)


Lightheartdevs:
Thanks for lining up the template GPU compatibility gate with the Extensions page behavior.

What looks good:
- `ods/extensions/services/dashboard-api/routers/templates.py` now treats an empty or missing `gpu_backends` list as unrestricted, matching the manifest schema and `_compute_extension_status`.
- Declared backend restrictions are still enforced, so this fixes the undeclared-extension false negative without weakening explicit compatibility rules.
- `ods/extensions/services/dashboard-api/tests/test_templates.py` covers both the unit helper behavior and the template preview path that previously put usable extensions into `incompatible`.

What I checked:
- Reviewed the templates router diff and the service-manifest compatibility assumptions from the repo architecture.

PR #2116, opened 2026-07-25
fix(deps): resolve image ARGs the way the builder does

tang-vu:
## Problem

`_resolve_vars()` consulted the inline fallback **before** the declared `ARG` value:

```python
default = match.group("default")
if default is not None:
    return default          # ← wins even when the ARG declares a value
if name in defaults:
    return defaults[name]
```

Docker resolves it the other way round. Given

```dockerfile
ARG RUNTIME_TAG=1.2.3
FROM example/runtime:${RUNTIME_TAG:-9.9.9}
```

the build pulls `example/runtime:1.2.3`, but the gate reported `example/runtime:9.9.9`. The consequence is worse than a false failure: whoever adds such a `FROM` is told to record `9.9.9` in `dependency-lock.json`, the gate then passes, and the lock permanently documents an image the build never pulls — the documented-vs-built drift this gate exists to catch.

No `FROM` in the tree uses a `:-` fallback today, so this is a latent inversion in the release gate rather than a live mismatch — but the gate is exactly the place that must not be wrong when someone does.

## Fix

Prefer the declared value, preserving shell substitution semantics:

| expression | `TAG` declared | result |
|---|---|---|
| `${TAG:-d}` | `1.2.3` | `1.2.3` |
| `${TAG:-d}` | `""` | `d` (`:-` substitutes on empty) |
| `${TAG-d}` | `""` | `""` (plain `-` only on unset) |
| `${TAG:-d}` | undeclared | `d` |
| `${TAG}` | undeclared | unresolved → existing "does not resolve to a default" error |

## Tests

`tests/test-dependency-pins.py` gains `test_declared_arg_wins_over_inline_fallback` (the table above) and `test_dockerfile_from_uses_declared_arg_value` (end-to-end through `_dockerfile_image_refs`).

Red on the pre-fix script, green after; the 7 existing tests and `scripts/check-dependency-pins.py` on the real tree still pass.

### DIFF
--- a/ods/scripts/check-dependency-pins.py
+++ b/ods/scripts/check-dependency-pins.py
@@ -64,13 +64,25 @@ def _rel(path: Path, root: Path = ROOT) -> str:
 
 
 def _resolve_vars(value: str, defaults: dict[str, str]) -> str:
+    """Expand ``${VAR}`` / ``${VAR:-fallback}`` the way the builder does.
+
+    A declared value wins over an inline fallback: with ``ARG TAG=1.2.3``,
+    Docker builds ``base:${TAG:-9.9.9}`` as ``base:1.2.3``. Returning the
+    fallback first made this gate record and enforce an image tag the build
+    never pulls — the exact documented-vs-built drift it exists to catch.
+    """
+
     def replace(match: re.Match[str]) -> str:
         name = match.group("braced") or match.group("plain") or ""
         default = match.group("default")
+        declared = defaults.get(name)
+        # ``:-`` substitutes when unset *or* empty; ``-`` only when unset.
+        if declared is not None and (declared != "" or match.group("op") == "-"):
+            return declared
         if default is not None:
             return default
-        if name in defaults:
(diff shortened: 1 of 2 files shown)


Lightheartdevs:
Thanks for correcting the dependency-pin resolver to match Docker's `ARG` substitution semantics.

What looks good:
- `ods/scripts/check-dependency-pins.py` now prefers declared `ARG` values over inline fallbacks, while still preserving the `:-` versus `-` empty-value distinction.
- The added tests in `ods/tests/test-dependency-pins.py` cover the direct resolver table and the end-to-end Dockerfile `FROM` parsing path, which is the right contract for this gate.
- The change is confined to the dependency-lock validation surface and reduces false guidance that could make `dependency-lock.json` document an image the build never pulls.

What I checked:
- Reviewed the resolver diff, the Dockerfile image-ref path, and the regression tests.
- GitHub checks are green, including Python lint, mypy, integration smoke, distro smoke, and macOS smoke.

PR #2120, opened 2026-07-25
feat(mode): add mesh as a valid ODS_MODE (stub)

Arifuzzamanjoy:
## Summary

Adding any mesh capability means touching the `ODS_MODE` enum, the CLI, `mode-switch.sh`, `ods-doctor`, and the compose resolver all at once. There was no mode reserved for distributed reasoning (only local, cloud, hybrid, lemonade). This PR does that plumbing once, as a stub, so the actual routing work later doesn't have to re-touch the enum and CLI.

`mesh` is local-family: each node runs its own local llama-server and reasons peer-to-peer through LiteLLM, so it resolves exactly like `hybrid` (keeps the local llama-server overlay, uses LiteLLM as the gateway for peer routing) and never layers the cloud overlay. cloud is the no-gpu path to external APIs, so mesh is basically the opposite of cloud. The mode CLI, `mode-switch.sh`, `ods-doctor`, the `.env` schema enum, and the dashboard-api / host-agent mode allowlists all recognize it now. Added mesh assertions to `test-linux-cloud-mode.sh` and mesh coverage to the doctor install-diagnosis tests so compose resolution and the inference contract can't silently break.

Nothing changes for local, cloud, hybrid, or lemonade. Unknown modes still fail closed.

Context: groundwork for planned distributed-inference work, discussed with the maintainer. This PR only reserves the mode and wires plumbing, no routing logic.

## User-facing behavior

Setting `ODS_MODE=mesh` today reserves the mode as local-family. It resolves like `hybrid` — local llama-server stays up, LiteLLM acts as the gateway for peer routing. No distributed routing lands in this PR; that's the follow-up. No cloud overlay, no external-API path.

## Changed surface

Mode routing / CLI / validation / doctor inference contract.

## Risk and validation

Risk level: Low

### DIFF
--- a/ods/.env.schema.json
+++ b/ods/.env.schema.json
@@ -18,12 +18,13 @@
     },
     "ODS_MODE": {
       "type": "string",
-      "description": "LLM backend mode: local, cloud, hybrid, or lemonade (AMD)",
+      "description": "LLM backend mode: local, cloud, hybrid, lemonade (AMD), or mesh (experimental)",
       "enum": [
         "local",
         "cloud",
         "hybrid",
-        "lemonade"
+        "lemonade",
+        "mesh"
       ],
       "default": "local"
     },
--- a/ods/bin/ods-host-agent.py
+++ b/ods/bin/ods-host-agent.py
@@ -184,8 +184,8 @@ ALWAYS_ON_SERVICES: frozenset = frozenset({
 })
 USER_EXTENSIONS_DIR: Path = Path()
 EXTENSIONS_DIR: Path = Path()
-_ODS_MODES = frozenset({"local", "cloud", "hybrid", "lemonade"})
-_LOCAL_MODEL_MODES = frozenset({"local", "hybrid", "lemonade"})
+_ODS_MODES = frozenset({"local", "cloud", "hybrid", "lemonade", "mesh"})
+_LOCAL_MODEL_MODES = frozenset({"local", "hybrid", "lemonade", "mesh"})
 _MODEL_TIER_RE = re.compile(r"^[A-Z0-9_]{1,32}$")
 _MODEL_TIERS = frozenset({
     "0", "1", "2", "3", "4", "ARC", "ARC_LITE",
--- a/ods/extensions/services/dashboard-api/config.py
+++ b/ods/extensions/services/dashboard-api/config.py
@@ -25,8 +25,8 @@ EXTENSIONS_DIR = Path(
 
 DEFAULT_SERVICE_HOST = os.environ.get("SERVICE_HOST", "host.docker.internal")
 GPU_BACKEND = os.environ.get("GPU_BACKEND", "nvidia")
-ODS_MODES = frozenset({"local", "cloud", "hybrid", "lemonade"})
-LOCAL_MODEL_MODES = frozenset({"local", "hybrid", "lemonade"})
+ODS_MODES = frozenset({"local", "cloud", "hybrid", "lemonade", "mesh"})
+LOCAL_MODEL_MODES = frozenset({"local", "hybrid", "lemonade", "mesh"})
 LLM_CONTRACT_ROUTES = frozenset({"gateway", "direct"})
 LLM_CONTRACT_PINNING = frozenset({"none", "dynamic"})
 
--- a/ods/ods-cli
+++ b/ods/ods-cli
@@ -3525,12 +3525,16 @@ cmd_mode() {
             mv "$litellm_disabled" "$litellm_cf"
             success "Auto-enabled litellm for $mode mode"
         fi
-        # Check for API keys
-        local has_keys=false
-        grep -q "^ANTHROPIC_API_KEY=." .env 2>/dev/null && has_keys=true
-        grep -q "^OPENAI_API_KEY=." .env 2>/dev/null && has_keys=true
-        if [[ "$has_keys" == "false" ]]; then
-            warn "No API keys found in .env — add ANTHROPIC_API_KEY or OPENAI_API_KEY"
+        # Cloud and hybrid reach external providers, so warn when no provider
+        # key is set. mesh routes peer traffic through the local LiteLLM gateway
+        # and needs no external API key, so skip the warning for it.
+        if [[ "$mode" == "cloud" || "$mode" == "hybrid" ]]; then
+            local has_keys=false
+            grep -q "^ANTHROPIC_API_KEY=." .env 2>/dev/null && has_keys=true
+            grep -q "^OPENAI_API_KEY=." .env 2>/dev/null && has_keys=true
+            if [[ "$has_keys" == "false" ]]; then
+                warn "No API keys found in .env — add ANTHROPIC_API_KEY or OPENAI_API_KEY"
+            fi
         fi
     fi
 
--- a/ods/ods-cli
+++ b/ods/ods-cli
@@ -850,7 +850,7 @@ get_compose_flags() {
                     _external_lemonade_active="true"
                 fi
             fi
-            if [[ "$_ods_mode" == "cloud" || "$_ods_mode" == "mesh" ]] && \
+            if [[ "$_ods_mode" == "cloud" ]] && \
                { [[ "$cached" == *"docker-compose.cpu.yml"* ]] || [[ "$cached" == *"compose.local.yaml"* ]] || [[ "$cached" != *"docker-compose.cloud.yml"* ]]; }; then
                 stale=1
             fi
@@ -858,7 +858,7 @@ get_compose_flags() {
                { [[ "$cached" == *"compose.local.yaml"* ]] || [[ "$cached" == *"docker-compose.amd.yml"* ]] || [[ "$cached" != *"docker-compose.cloud.yml"* ]] || [[ "$cached" != *"docker-compose.lemonade-external.yml"* ]]; }; then
                 stale=1
             fi
-            if [[ "$_ods_mode" != "cloud" ]] && [[ "$_ods_mode" != "mesh" ]] && [[ "$_external_lemonade_active" != "true" ]] && \
+            if [[ "$_ods_mode" != "cloud" ]] && [[ "$_external_lemonade_active" != "true" ]] && \
                [[ "$cached" == *"docker-compose.cloud.yml"* ]]; then
                 stale=1
             fi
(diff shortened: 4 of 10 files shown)


Lightheartdevs:
Thanks for reserving the `mesh` mode through the same mode/compose surfaces that already own local, cloud, hybrid, and Lemonade behavior.

Quickest path to merge:
- Rerun the cancelled `Matrix Smoke / distro: opensuse-tw` check for this head. The code change is small and the other checks are green, but the check rollup still contains a cancelled matrix leg, so the PR does not have a clean CI receipt yet.
- Keep the stub contract explicit in `ods/scripts/resolve-compose-stack.sh` and `ods/tests/test-linux-cloud-mode.sh`: `ODS_MODE=mesh` currently renders the cloud/LiteLLM overlay and intentionally does not introduce distributed routing behavior in this PR.

What I checked:
- Reviewed the PR body, changed files, and patch for `ods/.env.schema.json`, `ods/ods-cli`, `ods/scripts/mode-switch.sh`, `ods/scripts/ods-doctor.sh`, `ods/scripts/resolve-compose-stack.sh`, and the cloud-mode regression test.
- Checked GitHub status rollup: Dashboard, Ruff, ShellCheck, compose validation, env validation, Linux integration, macOS smoke, and most distro matrix legs passed; `distro: opensuse-tw` was cancelled.
- Tried to run `ods/tests/test-linux-cloud-mode.sh` locally, but this Windows runner only exposes the WSL `bash.exe` relay and WSL has no `/bin/bash`, so local shell execution was unavailable here.

Thanks again for keeping this as plumbing-only groundwork.

Arifuzzamanjoy:
Pushed a fresh head (28f7892), CI is green now. The opensuse-tw cancellation was a mirror timeout on base package install, not a code failure, and it passed this run.

Rebased onto current main and folded in one small fix: the compose-cache check in ods-cli now treats mesh as cloud-family, so mesh's cloud overlay cache isn't marked stale every run (matches resolve-compose-stack.sh, which already renders the cloud overlay for mesh).

Stub contract unchanged: ODS_MODE=mesh renders the cloud/LiteLLM overlay, no distributed routing here. That's the follow-up.

Lightheartdevs:
Thanks for tightening the `mesh` stub after the earlier review and getting the compose-cache behavior aligned with the cloud/LiteLLM overlay.

Quickest path to merge:
- Update the runtime mode allowlists in `ods/extensions/services/dashboard-api/config.py` and `ods/bin/ods-host-agent.py`. Both still list supported modes as `local`, `cloud`, `hybrid`, and `lemonade`, so `ODS_MODE=mesh` normalizes to `unknown` in dashboard-api/host-agent paths even though the schema, CLI, and resolver now accept it. `mesh` can stay outside the local-model-mode sets if model activation should remain disabled there.
- Extend the `ods-doctor` inference contract in `ods/scripts/ods-doctor.sh` to treat `mesh` as the same cloud-family/LiteLLM route this PR documents. Right now `expected_gateway` becomes `litellm`, but the missing cloud overlay and local llama-server route diagnostics still only run for `cloud`, so a broken `mesh` install can pass doctor without the blocker that `cloud` would get. A focused assertion in `ods/tests/test-ods-doctor-install-diagnoses.sh` would pin this.
- Update the top-level CLI help in `ods/ods-cli` and tighten `ods/tests/contracts/test-ods-cli-contracts.py`. `cmd_help` still advertises `mode [local|cloud|hybrid]`, and the contract now allows either the old or new help text; requiring `mode [local|cloud|hybrid|mesh]` would keep user-facing help from drifting.

What I checked:
- Reviewed the PR body, the existing discussion, and the current-head patch for all 7 changed files.
- Checked current GitHub checks: Linux integration, ShellCheck, compose validation, env validation, Ruff, dashboard jobs, PowerShell lint, macOS smoke, and all distro matrix legs are green; only Claude-review skip jobs are skipped.
- Ran `git diff --check origin/main...refs/remotes/codex/pr-2120` and `python ods/tests/contracts/test-ods-cli-contracts.py` locally on the fetched PR worktree. Shell tests could not be run on this Windows runner because the available `bash` shim points to a WSL install without `/bin/bash`.

Thanks again for pushing the stub forward; the remaining work is mostly making every mode-aware boundary agree on what `mesh` means today.

Arifuzzamanjoy:
thanks for both reviews. pushed an update, but  it changes direction so wanted to explain rather than slip it in.

had mesh pointed the wrong way. a mesh node runs its own local model and reasons peer-to-peer over litellm. cloud is the no-gpu path to external apis, so mesh is basically the opposite of cloud. rendering the cloud overlay was profiling out the llama-server a mesh node is supposed to contribute.

so mesh now reserves as local-family, resolves like hybrid: local llama-server + litellm as the gateway for peer routing. also covers your two follow-ups. config.py/host-agent recognize mesh now (no more unknown) and it's a local-model mode, and the cli help + contract require the mesh variant.

left ods-doctor as-is: base already recognizes mesh + expects a litellm gateway with no cloud-overlay requirement, which is right for local-family. can make it explicit if you'd rather.

same minimal scope otherwise, local/cloud/hybrid/lemonade untouched. this is the dreamreason groundwork we discussed, now on the right base.

Lightheartdevs:
Thanks for reworking the `mesh` stub around local llama-server plus LiteLLM after the previous review.

Quickest path to merge:
- `ods/scripts/ods-doctor.sh`: add an explicit `mesh` inference-contract branch. The current code accepts `mesh` and reports `expected_gateway=litellm`, but the actual diagnostics still only check cloud-mode local-llama misroutes and local-mode LiteLLM/cloud-overlay drift. A broken mesh install with `LLM_API_URL=http://llama-server:8080` or `HERMES_LLM_BASE_URL` bypassing LiteLLM can pass without the blocker this mode needs.
- Add focused doctor coverage for `ODS_MODE=mesh`, ideally near the existing doctor/install diagnosis tests: local inference overlay present, cloud overlay absent, and `LLM_API_URL`/`HERMES_LLM_BASE_URL` routed through LiteLLM.
- Update the PR description so the summary and user-facing behavior match the latest local-family semantics. It still describes mesh as cloud-like/no user-visible difference from cloud, while the current code and contributor note now make it local inference plus a LiteLLM peer gateway.
- Rerun the red `distro: opensuse-tw` matrix leg after the code/doc updates. The log looks infrastructure-shaped, ending with `context canceled` during zypper package install, but the rollup is still red.

What I checked:
- Reviewed the full current diff and discussion for `.env` schema, `ods-cli`, `mode-switch.sh`, `resolve-compose-stack.sh`, `ods-doctor.sh`, dashboard-api/host-agent mode allowlists, and the cloud-mode contract test.
- Ran `git diff --check origin/main...refs/remotes/codex/pr-2120`; it passed.
- Ran `python ods/tests/contracts/test-ods-cli-contracts.py`; it passed. Local `bash ods/tests/test-linux-cloud-mode.sh` could not run because this runner's `bash` is a WSL relay without `/bin/bash`.

Thanks again for keeping the stub small; the remaining work is making the runtime diagnostics agree with the new mode contract.

Arifuzzamanjoy:
thanks. added an explicit mesh branch to ods-doctor. it blocks the misroute cases you flagged now instead of passing silently.

added tests next to the existing ones. broken fixture asserts the blockers fire, healthy one stays clean.

updated the PR description to the local-family wording. pushed as 486dbec so opensuse-tw reruns.

Lightheartdevs:
Thanks for carving out `mesh` as an explicit mode instead of leaving distributed-inference work to overload an existing cloud/local path.

Quickest path to merge:
- In `ods/scripts/resolve-compose-stack.sh`, handle `ODS_MODE=mesh` with `TIER=CLOUD` explicitly. The current resolver still takes the cloud branch when `tier == "CLOUD"`, so switching an existing cloud install to mesh can regenerate `docker-compose.cloud.yml` even though the new doctor check flags that exact state as `ODS-RUNTIME-MESH-CLOUD-OVERLAY`. Either let `mesh` override the `CLOUD` tier in the resolver, or have `ods mode mesh` / `mode-switch.sh` reset the persisted tier/backend to a local-family value before `_regenerate_compose_flags`.
- Add a regression in `ods/tests/test-linux-cloud-mode.sh` for `--ods-mode mesh --tier CLOUD` or for an `ods mode mesh` fixture starting from `TIER=CLOUD`. The current mesh resolver test only exercises `--tier LOCAL`, so it misses the cloud-to-mesh transition path.
- In `ods/ods-cli`, keep the LiteLLM gateway setup for mesh but split the cloud API-key warning out of the generic non-local branch. Right now `ods mode mesh` without `ANTHROPIC_API_KEY` or `OPENAI_API_KEY` prints the cloud-key warning even though the PR now describes mesh as local nodes routing through LiteLLM.

What I checked:
- GitHub checks are green for this head, including shell/Python lint, dashboard checks, compose validation, distro smoke, env validation, and secret scan.
- `gh pr diff` for the 10-file PR plus static review of `ods-cli`, `mode-switch.sh`, `resolve-compose-stack.sh`, `ods-doctor.sh`, and the added mesh tests.
- Existing tests cover the enum acceptance, the local-tier mesh compose shape, and doctor diagnoses, but not the cloud-tier to mesh transition or the CLI warning path.

Thanks again for pushing the mode plumbing forward.

Arifuzzamanjoy:
thanks . went with the resolver option: mesh no longer takes the cloud branch when tier is CLOUD, so a cloud→mesh switch falls through to the local overlay instead of docker-compose.cloud.yml. agrees with the doctor check now.

added the --ods-mode mesh --tier CLOUD regression, and split the api-key warning so mesh keeps the litellm setup but doesn't nag for a cloud key.

pushed as SHA.

Lightheartdevs:
Thanks for sticking with the `mesh` mode plumbing and tightening it around the local-node plus LiteLLM gateway contract.

What looks good:
- The current head keeps `mesh` aligned across the mode-aware boundaries: `.env` schema, `ods mode`, `mode-switch.sh`, dashboard-api config, host-agent config, compose resolution, and doctor diagnostics all now recognize the mode.
- The resolver now handles the cloud-to-mesh transition explicitly: `ODS_MODE=mesh` overrides a persisted `TIER=CLOUD`, keeps a local llama-server overlay, includes LiteLLM, and avoids `docker-compose.cloud.yml`.
- The added doctor checks make the stub contract enforceable by flagging cloud-overlay drift, missing local inference overlays, and direct llama-server routes that bypass the LiteLLM peer gateway.

What I checked:
- Reviewed the PR body, existing discussion, and the current 10-file GitHub PR diff for mode/schema/CLI/resolver/doctor/test changes.
- Checked GitHub status for this head: shell/Python lint, type check, dashboard jobs, compose validation, env validation, Linux integration, distro matrix, macOS smoke, PowerShell lint, and secret scan are green.
- Ran `git diff --check origin/main...origin/pr/2120`; it passed.

Arifuzzamanjoy:
rebased onto current main. one conflict in resolve-compose-stack.sh: aafd0b9d added `and not external_llm` to the two overlay guards, same lines mesh went into. kept both clauses.

also fixed a gap i hit while re-testing. at TIER=CLOUD mesh kept the local llama-server but picked up zero compose.local.yaml overlays, so a converted cloud→mesh node had no depends_on wiring where a fresh install has five. doctor missed it too, since the overlay it checks for was still there. both guards exempt mesh from the cloud-tier skip now, plus an assertion at CLOUD tier.

local/cloud/hybrid/lemonade unchanged.

PR #2122, opened 2026-07-26
fix(dashboard-api): bound model picks to the usable share of unified memory

tang-vu:
## Problem

`_usable_model_memory_gb()` treats the whole reported memory pool as available to model weights unless the backend is `apple` or the GPU name literally contains `strix-halo`:

```python
if backend == "apple" or "strix-halo" in normalize_key(gpu_info.name):
    return max(total_gb * 0.55, 2.0)
return total_gb
```

But `gpu.py` already decides this properly and records it. It flags an AMD APU as unified whenever the GTT pool dwarfs the VRAM carve-out, then reports `memory_type="unified"` with whatever `sysfs` `product_name` says as the name — commonly `AMD Radeon 780M Graphics`, `AMD Radeon Graphics`, etc. (`gpu.py:104-140`).

Those hosts fall through to `return total_gb`, so a 32 GB shared pool is sized as if it were 32 GB of dedicated VRAM. The ranker then recommends a model whose weights plus KV cache leave nothing for the OS, Docker services and the runtime.

Two places in the same file already disagreed: `_matching_runtime_profile()` derives unified-ness from three signals including `memory_type`, while `_usable_model_memory_gb()` only had two. The installer's own selector (`scripts/select-model.py:134-144`) treats `memory_type == "unified"` as authoritative and applies the 55% share — so dashboard and installer recommend different models on the same host.

## Fix

Extract one `_has_unified_memory()` predicate that trusts the reported `memory_type` first (keeping backend/name as fallbacks for surrogate `GPUInfo` records built from runtime contracts), and use it for both the runtime-profile match and the capacity calculation.

## Tests

`tests/test_performance_oracle.py`:
- `test_pre_download_ranker_bounds_unreported_unified_memory_apu` — a 32 GB `memory_type="unified"` APU named `AMD Radeon 780M Graphics` no longer picks the 24 GB-class model. **Fails on `main`** (`assert 'qwen3.5-32b-q4' == 'qwen3.5-12b-q4'`).
- `test_pre_download_ranker_uses_full_pool_for_discrete_memory` — the same 32 GB pool reported as `discrete` still picks the large model, proving `memory_type` is the discriminator and not just a smaller cap.

Full `dashboard-api` suite: 1881 passed (the one failure is a pre-existing Windows symlink-privilege test, unrelated).

### DIFF
--- a/ods/extensions/services/dashboard-api/performance_oracle.py
+++ b/ods/extensions/services/dashboard-api/performance_oracle.py
@@ -685,18 +685,30 @@ def _selector_required_memory_gb(model: dict[str, Any]) -> float:
     return round(max(declared, size_gb + context_kv_gb), 2)
 
 
+def _has_unified_memory(gpu_info: Optional[GPUInfo]) -> bool:
+    """Return whether the detected GPU shares its memory pool with the host.
+
+    The reported `memory_type` is the authoritative signal: `gpu.py` sets it to
+    "unified" for Apple Silicon and for any AMD APU whose GTT pool dwarfs its
+    carve-out VRAM, whatever the sysfs product name happens to be. Backend and
+    name are kept as fallbacks for surrogate GPUInfo records built from runtime
+    contracts that predate the field.
+    """
+    if not gpu_info:
+        return False
+    return (
+        normalize_key(gpu_info.gpu_backend) == "apple"
+        or normalize_key(getattr(gpu_info, "memory_type", "")) == "unified"
+        or "strix-halo" in normalize_key(gpu_info.name)
+    )
+
+
 def _matching_runtime_profile(model: dict[str, Any], gpu_info: Optional[GPUInfo],
                               system_ram_gb: int | None = None) -> dict[str, Any] | None:
     if not gpu_info:
         return None
     backend = normalize_key(gpu_info.gpu_backend)
-    memory_type = (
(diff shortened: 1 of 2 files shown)


Lightheartdevs:
Thanks for tightening the dashboard model picks on unified-memory APUs; this closes a practical mismatch for hosts that report a generic AMD name but a shared memory pool.

What looks good:
- `ods/extensions/services/dashboard-api/performance_oracle.py` now treats `GPUInfo.memory_type == "unified"` as the authoritative signal, matching `gpu.py` detection and `scripts/select-model.py` capacity sizing while keeping the Apple/name fallbacks.
- The same `_has_unified_memory()` predicate feeds runtime-profile matching and usable-memory sizing, which reduces drift inside the dashboard API.
- `ods/extensions/services/dashboard-api/tests/test_performance_oracle.py` covers the important discriminator: the same 32 GB pool is bounded for `memory_type="unified"` and left full-size for discrete memory.

What I checked:
- Static review of the diff against `gpu.py`, `scripts/select-model.py`, and the existing `rank_pre_download_models()` callers.
- `git diff --check origin/main...pr/2122` was clean.
- GitHub checks are green, including Dashboard API, Python lint/type check, Test Linux, Matrix Smoke, env validation, ShellCheck, and secret scan.

PR #2123, opened 2026-07-26
fix(installer): stop the selector allowlist dropping catalog runtime tuning

tang-vu:
## Problem

`scripts/select-model.py` forwards a matched runtime profile's `env` block verbatim into its `--env` output:

```python
for key, value in (runtime_profile.get("env") or {}).items():
    if value is not None:
        env[str(key)] = value
```

The installer then loads that output through `load_model_selector_env_from_output` (`installers/phases/02-detection.sh:559`, `installers/macos/install-macos.sh:1387`), which re-exports only an allowlisted set of keys.

The catalog declares four distinct env keys across its runtime profiles:

```
LLAMA_ARG_CACHE_TYPE_K, LLAMA_ARG_CACHE_TYPE_V, LLAMA_ARG_FLASH_ATTN, LLAMA_PARALLEL
```

`LLAMA_PARALLEL` was the one missing from the allowlist. It is declared by the `nvidia-8gb-64k-q4-kv` profile on both `qwen3-4b-instruct-2507-q4` and `qwen3.5-4b-q4` (`config/model-library.json`), and it is a real consumed variable — `docker-compose.base.yml:49` passes it as llama.cpp's `--parallel`. So on an 8 GB NVIDIA host the profile's concurrency choice never reached `.env`; `installers/phases/06-directories.sh:755` wrote the `${LLAMA_PARALLEL:-1}` fallback instead of the profile's value. That profile's 64 GB-context Q4 KV budget is sized around a specific slot count, and it was silently not being applied.

## Fix

Add `LLAMA_PARALLEL` to `load_model_selector_env_from_output`.

## Tests

The interesting failure here is the *class*, not the instance — the allowlist and the catalog are two lists that have to stay in sync and nothing checked that. `tests/test-safe-env.sh` (already in `make test`) gains a check that derives every `env` key any runtime profile in `config/model-library.json` declares and asserts each one survives the loader:

```
Test 13: model selector loader accepts every catalog runtime-profile env key
[FAIL] LLAMA_PARALLEL is declared by a catalog runtime profile but dropped by the model selector allowlist
```

That is the failure on `main`; it passes with the fix. A future profile that adds tuning can no longer be dropped silently.

Also green: `tests/test-tier-map.sh` (128 passed), `bash -n` on both changed files.


Lightheartdevs:
Thanks for closing the selector-env allowlist gap around `LLAMA_PARALLEL`; this is a small change in the right high-risk path.

What looks good:
- `ods/lib/safe-env.sh` now allows `LLAMA_PARALLEL`, which is emitted by catalog runtime profiles and consumed by `ods/docker-compose.base.yml` as llama.cpp's `--parallel` argument.
- The new `ods/tests/test-safe-env.sh` coverage derives every runtime-profile `env` key from `ods/config/model-library.json`, so the selector allowlist should fail loudly if future catalog tuning adds another key.
- The change is narrowly scoped to selector output loading and does not broaden general `.env` parsing.

What I checked:
- Static review of `load_model_selector_env_from_output`, `scripts/select-model.py --env` output, the catalog runtime-profile `env` blocks, and compose/runtime consumption.
- `git diff --check origin/main...pr/2123` was clean.
- GitHub checks are green, including Test Linux, Matrix Smoke, ShellCheck, Dashboard, Python lint, env validation, PowerShell lint, and secret scan.

tang-vu:

Merged current upstream/main and kept both the catalog-driven selector allowlist invariant and upstream's safe double-quoted escape regression.

- New head: 82960eb0`n- ash tests/test-safe-env.sh: 14 tests passed
- ash tests/test-tier-map.sh: 129 passed, 0 failed
- Bash syntax, ShellCheck, and git diff --check: passed

ods/tests/test-safe-env.sh
```diff
@@ -141,7 +141,34 @@ load_env_file "$tmpdir/.env-quotes"
...
+        for profile in model.get("runtime_profiles") or []:
+            keys.update((profile.get("env") or {}).keys())
+print("\n".join(sorted(keys)))
+PY
+)"
+[[ -n "$_profile_env_keys" ]] || fail "no runtime-profile env keys found in config/model-library.json"
+while IFS= read -r _key; do
+    _key="${_key%$'\r'}"
+    [[ -n "$_key" ]] || continue
+    unset "$_key" 2>/dev/null || true
```
ahmadisaeed403-art, line 165: Hard-rule violation: the root `CLAUDE.md` says Bash must let failures surface and explicitly forbids both `2>/dev/null` and `|| true`. This new test suppresses `unset` failures, so a catalog key that is invalid/readonly could be hidden instead of making the invariant test fail. Replace this line with `unset "$_key"`; all current catalog keys are ordinary writable identifiers, and a future invalid key should fail loudly.

ahmadisaeed403-art (changes requested):
One merge blocker:

- Hard-rule violation: the new test adds `unset "$_key" 2>/dev/null || true`, while root `CLAUDE.md` explicitly forbids silent catches (`2>/dev/null` and `|| true`). Use plain `unset "$_key"`; an invalid or readonly future catalog key should fail the invariant test visibly. Exact fix is inline.

Nits:
- [nit] The PR description says `LLAMA_PARALLEL` occurs on two profiles, but the current catalog has three: `ministral3-8b-instruct-2512-q4`, `qwen3-4b-instruct-2507-q4`, and `qwen3.5-4b-q4`.
- [nit] The post-conflict line citations are stale: the current selector-loader calls are at detection line 563 and macOS line 1402, and the `.env` fallback write is at directories line 887.

The production allowlist addition and catalog-derived coverage are otherwise correct. All required checks are green.

PR #2124, opened 2026-07-26
fix(dashboard-api): stop recommending models the catalog has retired

tang-vu:
## Problem

`config/model-library.json` keeps superseded entries so an existing install keeps working and the UI can still describe what is on disk, and marks **32 of its 52 models** `install_recommendation: false` so nothing steers a new user onto them.

`scripts/select-model.py` honors that flag (`install_recommendation_allowed()`). The dashboard never saw it:

- `normalize_catalog_entry()` did not copy `install_recommendation` into the oracle shape at all, so the field was gone before any ranking happened.
- `rank_pre_download_models()` only filtered on `gguf_url`.

So the two rankers disagree on identical hardware:

| GPU | `select-model.py --installable-only` | dashboard `rank_pre_download_models` (top 3) |
|---|---|---|
| 4 GB | `qwen3.5-2b` | `qwen2.5-coder-1.5b-128k-q4` ❌, `qwen2.5-coder-3b-128k-q4` ❌, `granite3.3-2b-instruct-q4` ❌ |
| 8 GB | `qwen3.5-9b` | `qwen2.5-coder-3b-128k-q4` ❌, `qwen2.5-coder-1.5b-128k-q4` ❌, `qwen3.5-9b-q4` |
| 12 GB | — | `phi4-q4`, `qwen2.5-coder-3b-128k-q4` ❌, `qwen3.5-9b-q4` |
| 24 GB | — | `qwen3.5-27b-q4`, `deepseek-r1-32b-q4`, `mistral-nemo-12b-instruct-q4` ❌ |

(❌ = `install_recommendation: false`)

It is not a near miss. Retired entries carry `Code` (4.4) and `Long Context` specialties, which outweigh `General` (3.8) in the scoring table both rankers share — so on the smaller cards they took the *top* recommendation slot precisely because they were retired-but-unfiltered.

`recommended_entry` feeds the "Recommended" badge and `recommendationAlternatives` in the models payload, so this is what a fresh dashboard user is told to install.

## Fix

- `_install_recommendation_flag()` carries the flag through `normalize_catalog_entry()`, mirroring `value_enabled()` in `select-model.py` so a string `"false"` reads the same in both.
- `rank_pre_download_models()` skips retired entries, and degrades in the same order as `select-model.py`: relax the fit first, then the retirement flag, then the profile — a host that fits nothing still gets a concrete suggestion.

Retired models stay listed in the library payload; only the recommendation changes. After the fix the dashboard's top pick matches the installer exactly (4 GB → `qwen3.5-2b-q4`, 8 GB → `qwen3.5-9b-q4`).

## Tests

`tests/test_performance_oracle.py`:
- `test_pre_download_ranker_skips_retired_catalog_entries` — retired entry scores higher (Code beats General, larger file) so only the flag can keep it out. **Fails on `main`.**
- `test_real_catalog_recommendations_are_never_retired_entries` — runs the shipped catalog across 4/8/12/24 GB. **Fails on `main`** with `4096MB GPU was offered retired entries: ['granite3.3-2b-instruct-q4', 'qwen2.5-coder-1.5b-128k-q4', 'qwen2.5-coder-3b-128k-q4']`.
- `test_pre_download_ranker_falls_back_to_retired_entry_when_nothing_else_fits` — pins the degrade path so the filter can't turn into "no suggestion at all".

Full `dashboard-api` suite: 1882 passed (the one failure is a pre-existing Windows symlink-privilege test, unrelated).

### DIFF
--- a/ods/extensions/services/dashboard-api/performance_oracle.py
+++ b/ods/extensions/services/dashboard-api/performance_oracle.py
@@ -204,6 +204,17 @@ def _install_recommendation_flag(raw: dict[str, Any]) -> bool:
     return normalize_key(raw.get("install_recommendation", True)) not in {"", "0", "false", "off", "no"}
 
 
+def _agent_recommendation_allowed(model: dict[str, Any]) -> bool:
+    """Keep fleet-blocked models visible without promoting them as picks."""
+    compatibility = model.get("app_compatibility")
+    if not isinstance(compatibility, dict):
+        return True
+    viability = compatibility.get("agent_viability")
+    if isinstance(viability, dict):
+        viability = viability.get("status")
+    return normalize_key(viability) != "not-agent-viable"
+
+
 def normalize_catalog_entry(raw: dict[str, Any]) -> dict[str, Any] | None:
     """Convert config/model-library.json entries to the oracle shape."""
     if not isinstance(raw, dict):
@@ -1197,6 +1208,8 @@ def rank_pre_download_models(catalog: list[dict[str, Any]], gpu_info: Optional[G
             continue
         if not model.get("install_recommendation", True):
             continue
+        if not _agent_recommendation_allowed(model):
+            continue
         if not _family_allowed_for_profile(model, normalized_profile):
             continue
         runtime_profile = _matching_runtime_profile(model, gpu_info, system_ram_gb)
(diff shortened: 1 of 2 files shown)


Lightheartdevs:
Thanks for keeping the dashboard recommendation path aligned with the catalog's retired-model contract.

What looks good:
- `normalize_catalog_entry()` now preserves `install_recommendation` with the same false-value semantics as `scripts/select-model.py`, so the dashboard ranker can see the catalog's source-of-truth flag.
- `rank_pre_download_models()` filters retired models for normal recommendations while still keeping a last-resort fallback when the catalog has no recommendable fit, which preserves a useful dashboard response.
- The tests cover both a targeted high-scoring retired model and the shipped catalog across 4/8/12/24 GB envelopes, so this should catch the actual user-facing regression.

What I checked:
- Static review of the diff against `scripts/select-model.py`'s `install_recommendation_allowed()` and the dashboard payload path that fills `recommendationAlternatives`.
- `git diff --check origin/main...pr/2124` was clean.
- GitHub checks are green, including Dashboard API, Python lint/type check, Test Linux, Matrix Smoke, env validation, ShellCheck, PowerShell lint, and secret scan.

PR #1481, opened 2026-05-27
feat(tier-map): add JETSON_ORIN_NANO tier and jetson backend

matedev01:
## Summary

Phase 2 of issue #195 milestone 1. Stacks on #1479 (Phase 1 detection). Adds the `JETSON_ORIN_NANO` tier and a `jetson` backend contract — pure data, no compose changes, no runtime impact.

## Why

PR #1479 plumbs `GPU_BACKEND=jetson` through detection but the tier-map has no entry, so a user running `--tier JETSON_ORIN_NANO` today gets `error "Invalid tier"`. This PR closes that gap and lets the installer resolve a sane model and context size for Orin Nano hardware.

## Tier choices

Sized for the 8 GB unified-memory budget on Orin Nano (Ampere sm_87). Leaves ~5 GB free for KV cache + open-webui + dashboard-api + LiteLLM after the model loads.

| Profile | Model | Quant | Size on disk | Context | Rationale |
|---|---|---|---|---|---|
| qwen   | qwen3.5-2b      | Q4_K_M | ~1.5 GB  | 8K | Same model as Tier 0 — proven small-footprint default; comfortable headroom |
| gemma4 | gemma-4-e2b-it  | Q4_K_M | ~2.81 GB | 8K | Matches the gemma4 Tier 1 model; reduced context vs Tier 1 to keep KV cache room |

Both set `N_GPU_LAYERS=99` — the Tegra iGPU shares system RAM with the CPU, so partial offload has no upside on unified memory.

Larger Orin Nano variants and Orin NX/AGX are deliberately excluded; this milestone is "Orin Nano end-to-end" per the maintainer's 2026-05-10 triage on #195.

## Files changed

| File | Change |
|---|---|
| `installers/lib/tier-map.sh` | `JETSON_ORIN_NANO` case added to both `set_qwen_tier_config()` and `set_gemma4_tier_config()` switches. Validation error lists at lines 194 + 307 updated. `tier_to_model()` extended in both profile branches so `dream model swap` resolves correctly |
| `config/backends/jetson.json` (new) | Mirrors `nvidia.json` — same `llama-server` contract on port 8080, same provider URL. The runtime difference (JetPack-pinned image, Tegra container runtime) lives in `docker-compose.jetson.yml`, follow-up PR |
| `tests/test-tier-map.sh` | Added 13 assertions covering both qwen and gemma4 paths for the new tier; extended the GGUF_URL coverage loop to include `JETSON_ORIN_NANO` |

## Test plan

```bash
[...]

PR #2126, opened 2026-07-26
fix(models): persist and honor the resolved auto profile

tang-vu:
## Summary

- persist `MODEL_PROFILE_EFFECTIVE` as the installer-owned receipt on Linux, macOS, and Windows
- make the dashboard rank `MODEL_PROFILE=auto` from that exact receipt, with a compatibility fallback for pre-upgrade installs
- align the standalone catalog selector with the canonical tier rule: cloud/tier 0 use Qwen; every other tier uses Gemma 4
- expose the receipt in the env schema while keeping it read-only in Dashboard Settings

## Why this matters

The installer resolves `MODEL_PROFILE=auto` from the hardware tier, but previously persisted only the raw value. The dashboard therefore had to infer the selected family from backend presence. That is observably wrong for CPU hosts assigned a non-zero tier from system RAM, and the standalone selector also disagreed with the shell and PowerShell tier maps for AMD.

The production contract is now explicit: tier resolution produces `MODEL_PROFILE_EFFECTIVE`, every platform persists it, the dashboard consumes it, and a subsequent installer run is the rollback/reconciliation path. Existing installs without the receipt retain the prior cloud/GPU approximation until their next update rather than losing recommendations.

## Behavioral invariant

For one install transaction, tier-map selection, catalog selection, persisted state, and dashboard recommendation alternatives must use the same model family. Explicit `qwen`/`gemma4` requests remain authoritative; only `auto` consumes the receipt.

## Overlap check

Searched open and closed PRs for `MODEL_PROFILE_EFFECTIVE`, `MODEL_PROFILE=auto`, effective-profile persistence, and the changed installer/dashboard files. This remains the canonical PR for dashboard auto-profile parity. Open #1481 also mentions `MODEL_PROFILE_EFFECTIVE`, but changes the retired `dream-server/` Jetson tier path and does not persist or consume the `ods/` receipt; its scope is independent.

## Tests

- `bash tests/test-tier-map.sh` ? 135 passed, including AMD tier 2 and NVIDIA tier 0 selector parity plus all three env generators
- `bash tests/test-macos-installer-transitions.sh` ? passed; generated `.env` contains the effective receipt
- `bash tests/smoke/installer-env-smoke.sh` ? generated `.env` validated against the shipped schema, survived rerun, and persisted the receipt; 12 relevant checks passed. The script's final function-inventory probe still reports the pre-existing undefined `ai_err` call in phase 07 (unrelated to this diff).
- `python -m pytest extensions/services/dashboard-api/tests/test_performance_oracle.py extensions/services/dashboard-api/tests/test_settings_env.py -q` ? 130 passed
- `bash tests/test-validate-env.sh` ? 43 passed
- Windows PowerShell parse check ? passed
- `powershell -File tests/test-windows-catalog-selector.ps1` ? passed
- Python compile, Bash syntax, JSON parse, and `git diff --check` ? passed

## Tradeoffs and rollback

The receipt is deliberately installer-owned and read-only in the dashboard. Pre-upgrade installs use the legacy approximation so this change is backward compatible. Rolling back the code leaves the extra schema-valid env key unused; rerunning an older installer removes or ignores it according to that installer's normal env-generation behavior.

Generated with Codex


Lightheartdevs:
Thanks for tightening the dashboard's `MODEL_PROFILE=auto` handling; this is exactly the path that needs to stay aligned with install-time selection.

Decision needed:
- This matches `scripts/select-model.py` when `auto` is passed directly: NVIDIA/Apple/SYCL resolve to `gemma4`, while AMD/CPU resolve to `qwen`.
- The installed path currently does something different before `select-model.py` runs: `ods/installers/phases/02-detection.sh` passes `--profile "${MODEL_PROFILE_EFFECTIVE:-...}"`, and `ods/installers/lib/tier-map.sh` / `ods/installers/windows/lib/tier-map.ps1` resolve `MODEL_PROFILE=auto` to `gemma4` for every non-cloud, non-tier-0 tier. Phase 06 then persists only raw `MODEL_PROFILE=auto` plus a comment for the effective profile.
- Because `build_models_payload()` reads the raw persisted profile, this PR can make `recommendationAlternatives` use the qwen lane on AMD/CPU installs where the installer may already have selected the Gemma lane via `MODEL_PROFILE_EFFECTIVE`. That may be the desired product decision, but it should be made deliberately across installer + selector + dashboard rather than only in the dashboard API.

What would make this ready:
- Pick one canonical `auto` rule and encode it in all three places: shell/PowerShell tier maps, `scripts/select-model.py`, and `performance_oracle.py`.
- If the effective install-time profile remains distinct from raw `MODEL_PROFILE`, persist it as a real env key or otherwise pass it into the dashboard instead of relying on an `.env` comment.
- Add an AMD/CPU `MODEL_PROFILE=auto` regression case that proves the dashboard recommendation alternatives match the actual install-time selector result, not only a direct call to `rank_pre_download_models()`.

What I checked:
- Static review of `performance_oracle.py`, `tests/test_performance_oracle.py`, `scripts/select-model.py`, Linux phase 02/06, and the Windows tier-map/env-generator path.
- `git diff --check origin/main...pr/2126` was clean.
- GitHub checks are green, including Dashboard API, Python lint/type check, Test Linux, Matrix Smoke, env validation, ShellCheck, PowerShell lint, and secret scan.

Thanks again for pushing on the alignment here.

tang-vu:
Pushed a correction — the advisory was right, and the fix was encoding the wrong rule.

The `auto` rule I copied came from `scripts/select-model.py:effective_profile()`, which is backend-based (`apple`/`nvidia`/`sycl` → gemma4, AMD/CPU → qwen). That is not what an install resolves:

- `installers/lib/tier-map.sh:effective_model_profile()` and `installers/windows/lib/tier-map.ps1` are **tier**-based — `CLOUD` and tier `0` → qwen, every other tier → gemma4, **AMD included**.
- `installers/phases/02-detection.sh:552` passes `--profile "${MODEL_PROFILE_EFFECTIVE:-...}"`, i.e. already resolved. `effective_profile()`'s `profile != "auto"` guard then returns it untouched, so its backend split only fires when the selector is invoked directly and never on an install.

So the original diff would have put AMD hosts on the qwen lane while the installer had put them on Gemma — exactly the mismatch this PR set out to remove. Corrected to follow the tier rule: `gemma4` unless `ODS_MODE=cloud` or no GPU backend was detected (the observable half of tier 0, since the tier itself is not persisted).

Tests updated accordingly — `amd` and `sycl` now assert the Gemma lane, and the qwen case is keyed on a missing backend rather than on AMD. Verified red against the backend-based helper:

```
FAILED tests/test_performance_oracle.py::test_auto_profile_resolves_to_gemma4_on_every_detected_backend
```

`45 passed` in `test_performance_oracle.py`, `129 passed` with `test_models.py`.

This keeps the change inside the dashboard and does not touch installer behavior. The `select-model.py` / `tier-map.sh` divergence is real but pre-existing and unreachable on the installed path — worth its own issue rather than a silent behavior change here.

Lightheartdevs:
Thanks for tightening the dashboard `MODEL_PROFILE=auto` path so recommendations follow the installer state instead of drifting into the wrong family.

Decision point:
- `ods/extensions/services/dashboard-api/performance_oracle.py` now resolves persisted `MODEL_PROFILE=auto` through installer tier-map semantics: cloud or no detected backend stays in the qwen lane, while a detected backend resolves to gemma4. That lines up with `installers/lib/tier-map.sh` after phase 06 persists the raw `auto` value.
- This intentionally differs from a direct `scripts/select-model.py --profile auto --backend amd` invocation, which would still resolve AMD to qwen. If maintainers accept the tier-map as the dashboard source of truth for installed systems, this looks ready; if the standalone selector should converge too, align or document that before merge.

What I checked:
- Diff/static review of `performance_oracle.py` and `tests/test_performance_oracle.py` against `origin/main`.
- `gh pr checks 2126`: current head is green across dashboard `api`, integration smoke, Python lint/type, matrix smoke, and related gates.
- `git diff --check origin/main...refs/remotes/pr/2126`.

tang-vu:
Addressed the remaining cross-component decision in commit `9eeb0ee0`.

The dashboard no longer approximates the install-time tier decision on upgraded installs. Linux, macOS, and Windows now persist `MODEL_PROFILE_EFFECTIVE`; the dashboard consumes that installer-owned read-only receipt; and `scripts/select-model.py --profile auto` now follows the same cloud/tier-0 vs non-zero-tier rule as both tier maps. The fallback heuristic remains only for older installs that have not yet regenerated `.env`.

Boundary receipts: tier/selector 135 passed, dashboard/settings 130 passed, macOS transitions passed, Linux generated `.env` validates and survives rerun, validate-env 43 passed, and Windows parser/catalog-selector checks passed. The PR body now records the exact production handoff, compatibility behavior, overlap search, and the unrelated pre-existing `ai_err` inventory failure from the broad local smoke script.

PR #2127, opened 2026-07-26
fix(compose): keep llama.cpp runtime tunables on the Intel and Arc overlays

tang-vu:
## Problem

Docker Compose **replaces** a `command` sequence when merging files — it does not append to the base file's. So any overlay that overrides `llama-server`'s command has to repeat every flag it still wants.

`docker-compose.arc.yml` and `docker-compose.intel.yml` re-declare the command to set their own `--n-gpu-layers ${N_GPU_LAYERS:-99}` and `--ctx-size ${CTX_SIZE:-32768}`, and in doing so silently drop three flags the base file sets:

```
base:  --model … --host 0.0.0.0 --port 8080 --n-gpu-layers 999 --ctx-size …        --batch-size ${LLAMA_BATCH_SIZE:-2048} --threads ${LLAMA_THREADS:-4}        --parallel ${LLAMA_PARALLEL:-1} --metrics

arc:   --model … --host 0.0.0.0 --port 8080 --n-gpu-layers … --ctx-size … --metrics
intel: --model … --host 0.0.0.0 --port 8080 --n-gpu-layers … --ctx-size … --metrics
```

All three are documented, schema-validated operator tunables:

| variable | `.env.schema.json` |
|---|---|
| `LLAMA_BATCH_SIZE` | "llama.cpp prompt processing batch size…" (default 2048) |
| `LLAMA_THREADS` | "CPU threads for llama.cpp non-GPU work." (default 4) |
| `LLAMA_PARALLEL` | "llama.cpp `--parallel` request slots…" (default 1) |

They are listed in `.env.example:193-195`, and `installers/phases/06-directories.sh:755` writes `LLAMA_PARALLEL` into every generated `.env`. On an Intel or Arc install, setting any of them did nothing — the variable was never referenced by the resolved command — while every other backend honored it. `docker-compose.cpu.yml` keeps all three, so this is specific to the two SYCL overlays.

## Fix

Restore the three flags on both overlays. Each keeps its own `--n-gpu-layers` / `--ctx-size` choices; only the dropped tunables come back.

## Tests

New `tests/contracts/test-llama-runtime-tunables.py`, wired into `make test` beside the other overlay contracts. It compares every overlay whose `llama-server` command is a llama.cpp invocation against the base file's flag set, and additionally checks each tunable is still wired to the `.env` variable the schema documents (an overlay may pick a different default, but not hardcode the value). The AMD overlays are skipped by design — their `llama-server` runs Lemonade (`serve …`), a different binary with its own argument surface.

On `main` it reports exactly the regression:

```
[FAIL] llama.cpp runtime tunable parity
  - docker-compose.arc.yml: llama-server command drops --batch-size; compose replaces the whole list, so it is lost on this backend
  - docker-compose.arc.yml: llama-server command drops --threads; …
  - docker-compose.arc.yml: llama-server command drops --parallel; …
  - docker-compose.intel.yml: llama-server command drops --batch-size; …
  - docker-compose.intel.yml: llama-server command drops --threads; …
  - docker-compose.intel.yml: llama-server command drops --parallel; …
```

Also green: `tests/contracts/test-overlay-map-coherence.sh`, `scripts/validate-golden-paths.py`.

### DIFF
--- a/ods/Makefile
+++ b/ods/Makefile
@@ -56,6 +56,7 @@ test: ## Run unit and contract tests
 	@echo "=== Overlay/plist contracts ==="
 	@bash tests/contracts/test-overlay-map-coherence.sh
 	@bash tests/contracts/test-plist-log-paths.sh
+	@python3 tests/contracts/test-llama-runtime-tunables.py
 	@echo ""
 	@echo "=== safe .env loading (quote/CRLF handling) ==="
 	@bash tests/test-safe-env.sh
--- a/ods/docker-compose.arc.yml
+++ b/ods/docker-compose.arc.yml
@@ -55,6 +55,15 @@ services:
       - "${N_GPU_LAYERS:-99}"
       - --ctx-size
       - "${CTX_SIZE:-32768}"
+      # Compose replaces `command` wholesale rather than merging it, so the
+      # documented runtime tunables have to be repeated here or they go inert
+      # on this overlay.
+      - --batch-size
+      - "${LLAMA_BATCH_SIZE:-2048}"
+      - --threads
+      - "${LLAMA_THREADS:-4}"
+      - --parallel
+      - "${LLAMA_PARALLEL:-1}"
       - --metrics
     deploy:
       resources:
--- a/ods/docker-compose.intel.yml
+++ b/ods/docker-compose.intel.yml
@@ -37,6 +37,15 @@ services:
       - "${N_GPU_LAYERS:-99}"
       - --ctx-size
       - "${CTX_SIZE:-32768}"
+      # Compose replaces `command` wholesale rather than merging it, so the
+      # documented runtime tunables have to be repeated here or they go inert
+      # on this overlay.
+      - --batch-size
+      - "${LLAMA_BATCH_SIZE:-2048}"
+      - --threads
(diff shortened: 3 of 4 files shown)


Lightheartdevs:
Thanks for fixing the Intel and Arc compose overlay tunable parity.

What looks good:
- `ods/docker-compose.arc.yml` and `ods/docker-compose.intel.yml` restore the base llama.cpp `--batch-size`, `--threads`, and `--parallel` flags with env-backed values while preserving each overlay's own `--n-gpu-layers` and `--ctx-size` choices.
- `ods/tests/contracts/test-llama-runtime-tunables.py` matches the repo's compose-layering risk: overlays that replace `command` must keep the base flag contract, while Lemonade/AMD overlays are skipped because they use a different binary surface.
- Current CI is green after the retrigger commit, including `Validate Docker Compose`, shell lint, Python lint/type-check, dashboard API/frontend, and the multi-distro matrix.

What I checked:
- Static diff review of both compose overlays and the new runtime-tunable parity test.
- `gh pr checks 2127`: all checks passing.
- Local execution was not available here because this runner lacks PyYAML and usable Bash, so I relied on CI for the new contract.

PR #2128, opened 2026-07-26
fix(windows): enable the llama.cpp metrics endpoint on native launches

tang-vu:
## Problem

llama.cpp serves `/metrics` only when `llama-server` is started with `--metrics`; the endpoint is off by default.

Every launch path passes it except the two Windows PowerShell ones:

| launcher | `--metrics` |
|---|---|
| `docker-compose.base.yml:50` | ✅ |
| `docker-compose.cpu.yml` / `.arc.yml` / `.intel.yml` | ✅ |
| `bin/ods-host-agent.py:10098` | ✅ |
| `installers/windows/install-windows.ps1:705` | ❌ |
| `installers/windows/ods.ps1:1366` | ❌ |

Two dashboard features scrape that endpoint:

- `get_llama_metrics()` (`dashboard-api/helpers.py:387`) reads `tokens_predicted_total` / `tokens_predicted_seconds_total` for the tokens/sec display. Without `--metrics` the GET fails, and the guard at `helpers.py:487` raises `"llama-server metrics response has no token counter"` — the reading falls back to zero.
- `_extract_llama_cpp_prometheus_counters()` (`routers/usage.py:435`) returns `None` when both token counters are 0, so the Usage page shows no local-runtime counters.

So on a Windows native Vulkan install both read empty — indistinguishable from an idle server — while the exact same llama.cpp build under compose reports normally. It is silent: nothing logs, nothing errors, the numbers are just always 0.

## Fix

Pass `--metrics` from both PowerShell launchers, alongside the `--n-gpu-layers` / `--ctx-size` they already set.

Scoped deliberately to `--metrics`. The PowerShell launchers also omit `--batch-size` / `--threads`, but so does `ods-host-agent.py` — that is a consistent choice across the native launchers, not a divergence, so I left it alone.

## Tests

New `tests/contracts/test-llama-metrics-flag.sh`, wired into `make test` beside the other Windows contracts. It walks every known llama-server launch path — the four compose files that define their own command list plus the three native launchers — and asserts each enables the endpoint.

On `main`:

```
[FAIL] installers/windows/install-windows.ps1 starts llama-server without --metrics; /metrics stays off and the dashboard reads 0 tok/s on this platform
[FAIL] installers/windows/ods.ps1 starts llama-server without --metrics; /metrics stays off and the dashboard reads 0 tok/s on this platform
[FAIL] llama.cpp --metrics contract (2 launcher(s) affected)
```

All seven pass with the fix. Both `.ps1` files re-verified with `[System.Management.Automation.Language.Parser]::ParseFile` (0 parse errors); `bash -n` clean on the new test.


Lightheartdevs:
Thanks for closing the Windows native metrics gap for llama-server.

What looks good:
- `ods/installers/windows/install-windows.ps1` and `ods/installers/windows/ods.ps1` now pass `--metrics` on native llama.cpp launches, matching the compose, host-agent, and macOS paths that the dashboard metrics readers already depend on.
- The change is deliberately scoped to the metrics endpoint and does not mix in unrelated Windows launcher tunables.
- `ods/tests/contracts/test-llama-metrics-flag.sh` walks the compose and native launchers so future launcher additions should not silently lose `/metrics` again.

What I checked:
- Static diff review of the two PowerShell launchers and the new metrics contract.
- `gh pr checks 2128`: all checks passing, including shell lint, PowerShell lint on Ubuntu/Windows, dashboard API/frontend, env validation, and multi-distro smoke.

PR #2130, opened 2026-07-26
fix(macos): honour LLAMA_PARALLEL on native llama-server launches

tang-vu:
> Stacked on #2127 — the diff here is the last commit only. Without it the new contract test also flags `docker-compose.arc.yml` / `.intel.yml`, which is exactly what that PR fixes.

## Problem

`LLAMA_PARALLEL` is a documented operator tunable — `.env.schema.json` ("llama.cpp `--parallel` request slots"), `.env.example:195` ("Concurrent request slots (increase for multi-user)") — and `installers/phases/06-directories.sh:755` writes it into **every** generated `.env`.

None of the three macOS native launch paths passed `--parallel`:

| launcher | reads `LLAMA_PARALLEL` |
|---|---|
| `docker-compose.base.yml` / `.cpu.yml` | ✅ |
| `bin/ods-host-agent.py:10096` | ✅ |
| `installers/windows/install-windows.ps1:722` | ✅ |
| `installers/windows/ods.ps1:1377` | ✅ |
| `installers/macos/install-macos.sh` | ❌ |
| `installers/macos/ods-macos.sh` | ❌ |
| `scripts/bootstrap-upgrade.sh` (macOS hot-swap + rollback) | ❌ |

Each of the three builds a `llama_args` array with `--host/--port/--model/--ctx-size/--n-gpu-layers/--reasoning-format/--metrics`, then appends optional flags for `LLAMA_ARG_FLASH_ATTN`, `LLAMA_ARG_CACHE_TYPE_K/V`, `LLAMA_ARG_N_CPU_MOE`, `LLAMA_ARG_SPEC_TYPE`, `LLAMA_ARG_SPEC_DRAFT_N_MAX` — but never `LLAMA_PARALLEL`.

llama.cpp defaults `--parallel` to 1. So an Apple Silicon operator raising the slot count for multi-user use edits `.env`, restarts, and gets no change and no error — while the same setting works on Docker, on the host agent, and on Windows.

## Fix

Read it alongside the flags each launcher already reads, using that launcher's existing pattern (`ENV_`-prefixed var in `ods-macos.sh`, `grep '^KEY=' "$ENV_FILE"` in the other two).

## Tests

New `tests/contracts/test-llama-parallel-flag.sh`, wired into `make test`. It walks all ten llama.cpp launch paths.

One detail worth flagging: the bash launchers are matched on the `llama_args+=(--parallel` **append form**, not a bare `grep -- --parallel`. `scripts/bootstrap-upgrade.sh` also embeds a Windows PowerShell block that has always passed `--parallel` via `$env:ODS_WIN_PARALLEL`, so a plain grep passes on that file today and would never have caught the macOS block. I hit exactly that false pass while writing this and tightened the matcher.

On the base branch:

```
[FAIL] installers/macos/install-macos.sh builds its llama_args array without --parallel; LLAMA_PARALLEL is inert on this platform
[FAIL] installers/macos/ods-macos.sh builds its llama_args array without --parallel; LLAMA_PARALLEL is inert on this platform
[FAIL] scripts/bootstrap-upgrade.sh builds its llama_args array without --parallel; LLAMA_PARALLEL is inert on this platform
[FAIL] LLAMA_PARALLEL contract (3 launcher(s) affected)
```

All ten pass with the fix. Also green: `tests/test-macos-env-quote-strip.sh` (9 passed), `tests/contracts/test-llama-runtime-tunables.py`, `bash -n` on all three changed scripts.

### DIFF
--- a/ods/Makefile
+++ b/ods/Makefile
@@ -63,6 +63,7 @@ test: ## Run unit and contract tests
 	@bash tests/contracts/test-overlay-map-coherence.sh
 	@bash tests/contracts/test-plist-log-paths.sh
 	@python3 tests/contracts/test-llama-runtime-tunables.py
+	@bash tests/contracts/test-llama-parallel-flag.sh
 	@echo ""
 	@echo "=== safe .env loading (quote/CRLF handling) ==="
 	@bash tests/test-safe-env.sh
--- a/ods/Makefile
+++ b/ods/Makefile
@@ -57,6 +57,7 @@ test: ## Run unit and contract tests
 	@bash tests/contracts/test-overlay-map-coherence.sh
 	@bash tests/contracts/test-plist-log-paths.sh
 	@python3 tests/contracts/test-llama-runtime-tunables.py
+	@bash tests/contracts/test-llama-parallel-flag.sh
 	@echo ""
 	@echo "=== safe .env loading (quote/CRLF handling) ==="
 	@bash tests/test-safe-env.sh
--- a/ods/Makefile
+++ b/ods/Makefile
@@ -56,6 +56,7 @@ test: ## Run unit and contract tests
 	@echo "=== Overlay/plist contracts ==="
 	@bash tests/contracts/test-overlay-map-coherence.sh
 	@bash tests/contracts/test-plist-log-paths.sh
+	@python3 tests/contracts/test-llama-runtime-tunables.py
 	@echo ""
 	@echo "=== safe .env loading (quote/CRLF handling) ==="
 	@bash tests/test-safe-env.sh
--- a/ods/docker-compose.arc.yml
+++ b/ods/docker-compose.arc.yml
@@ -55,6 +55,15 @@ services:
       - "${N_GPU_LAYERS:-99}"
       - --ctx-size
       - "${CTX_SIZE:-32768}"
+      # Compose replaces `command` wholesale rather than merging it, so the
+      # documented runtime tunables have to be repeated here or they go inert
+      # on this overlay.
+      - --batch-size
+      - "${LLAMA_BATCH_SIZE:-2048}"
+      - --threads
+      - "${LLAMA_THREADS:-4}"
+      - --parallel
+      - "${LLAMA_PARALLEL:-1}"
       - --metrics
     deploy:
       resources:
--- a/ods/docker-compose.intel.yml
+++ b/ods/docker-compose.intel.yml
@@ -37,6 +37,15 @@ services:
       - "${N_GPU_LAYERS:-99}"
       - --ctx-size
       - "${CTX_SIZE:-32768}"
+      # Compose replaces `command` wholesale rather than merging it, so the
+      # documented runtime tunables have to be repeated here or they go inert
+      # on this overlay.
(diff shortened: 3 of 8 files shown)


Lightheartdevs:
Thanks for narrowing the `LLAMA_PARALLEL` gap on the macOS native launch paths.

Quickest path to merge:
- `#2130` current head `3fa25754f828fdf2760e952f72fbfacaa87abdab` still has a failing `distro: opensuse-tw` matrix-smoke check. If that was the same transient mirror/registry issue later cleared on the stacked heads, rerun or refresh this PR so this head has its own green receipt before merge.
- The code diff in `ods/installers/macos/install-macos.sh`, `ods/installers/macos/ods-macos.sh`, and `ods/scripts/bootstrap-upgrade.sh` looks coherent: each native macOS launcher reads `LLAMA_PARALLEL` through its existing env-loading path and appends `--parallel` to the llama args.
- `ods/tests/contracts/test-llama-parallel-flag.sh` targets the macOS `llama_args+=` append form, which is important because `bootstrap-upgrade.sh` also contains a Windows block that would make a loose grep pass.

What I checked:
- Static diff review of the three macOS launcher additions and the new `LLAMA_PARALLEL` contract.
- `gh pr checks 2130`: all checks pass except `distro: opensuse-tw`.

Lightheartdevs:
Thanks for tightening `LLAMA_PARALLEL` coverage across the native macOS and Intel compose launch paths.

Quickest path to merge:
- `ods/scripts/bootstrap-upgrade.sh:2974` still uses an inline rollback `exec $LLAMA_SERVER_BIN` that rebuilds the macOS native llama-server argv without `--parallel $_parallel`. If the new model fails during hot-swap, rollback starts the previous model with `LLAMA_PARALLEL` dropped, so the setting is still inert on that launch path.
- Refactor the rollback restart to reuse the same `_llama_args` with only `--model` swapped to `$_old_model_path`, or append `--parallel $_parallel` and the same optional tunables to the rollback exec. Add a contract assertion for the rollback block specifically; the current grep sees the successful-launch `_llama_args+=(--parallel ...)` and misses lines 2974-2980.

What I checked:
- Diff/static review of the macOS launchers, Intel overlays, and new llama tunable contracts.
- `gh pr checks 2130`: current head is green, including compose validation, shellcheck, matrix smoke, and integration smoke.
- `git grep -n -- '--parallel' refs/remotes/pr/2130 -- ods/scripts/bootstrap-upgrade.sh` shows `--parallel` in the Windows block and success-path `_llama_args`, but not the macOS rollback exec at lines 2974-2980.

Thanks again for pushing the parity contract forward.

tang-vu:
Addressed the outstanding macOS rollback advisory in commit b8af050f. The rollback now copies the complete verified `_llama_args` array and swaps only the `--model` value, so `LLAMA_PARALLEL` and every other optional native tunable survive a failed hot-swap. The contract test now asserts the rollback uses that preserved argv. Focused validation: `bash -n` on both changed scripts, `bash tests/contracts/test-llama-parallel-flag.sh` (all launch and rollback assertions pass), and `git diff --check`.

PR #2131, opened 2026-07-26
fix(macos): honour the long-context llama.cpp tunables on native launches

tang-vu:
> Stacked on #2130 (which is stacked on #2127) — the diff here is the last commit only.

## Problem

`.env.schema.json` documents two flags explicitly for long-context profiles:

```json
"LLAMA_ARG_NO_CACHE_PROMPT": { "description": "Optional llama.cpp boolean flag (--no-cache-prompt) used by specific long-context runtime profiles." },
"LLAMA_ARG_CHECKPOINT_EVERY_N_TOKENS": { "description": "Optional llama.cpp long-context checkpoint interval. Some experimental profiles set -1 to disable checkpoints." }
```

Every platform honours them except macOS:

| path | mechanism |
|---|---|
| Docker | `docker-compose.base.yml:61-62` passes both into the container env; llama.cpp reads `LLAMA_ARG_*` natively |
| host agent | `bin/ods-host-agent.py` `optional_args` table maps both |
| Windows installer / CLI | `install-windows.ps1`, `ods.ps1` append both |
| Windows hot-swap | `bootstrap-upgrade.sh` PowerShell block, lines 1181-1182 |
| **macOS installer** | ❌ |
| **macOS CLI** | ❌ |
| **macOS hot-swap** | ❌ |

The three macOS launchers read every *other* optional flag from `.env` — `--flash-attn`, `--cache-type-k/v`, `--n-cpu-moe`, `--spec-type`, `--spec-draft-n-max` — and stop there. The native launchers get no environment passthrough either (`ods-macos.sh` exports `.env` keys under an `ENV_` prefix; the other two read the file with `grep`), so on Apple Silicon a long-context profile setting either flag silently runs without it.

## Fix

Append both from the same `.env` each launcher already reads, treating `--no-cache-prompt` as the valueless boolean it is and honouring the same `0/false/off/no` spellings the Windows launchers accept.

## Tests

New `tests/contracts/test-llama-macos-tunables.sh`, wired into `make test`. Rather than checking one flag at a time it asserts the **whole documented tunable set** (8 flags) against each of the three macOS launchers, so the next flag added to the schema cannot land on Docker and Windows only.

As in #2130, `scripts/bootstrap-upgrade.sh` is matched on the `llama_args+=(` append form — it embeds a Windows PowerShell block that already handles both flags, so a bare grep for the flag name passes on that file today and would miss the macOS block entirely.

On the base branch:

```
[FAIL] installers/macos/install-macos.sh never appends --checkpoint-every-n-tokens (LLAMA_ARG_CHECKPOINT_EVERY_N_TOKENS); the setting is inert on macOS
[FAIL] installers/macos/install-macos.sh never appends --no-cache-prompt (LLAMA_ARG_NO_CACHE_PROMPT); the setting is inert on macOS
[FAIL] installers/macos/ods-macos.sh never appends --checkpoint-every-n-tokens …
[FAIL] installers/macos/ods-macos.sh never appends --no-cache-prompt …
[FAIL] scripts/bootstrap-upgrade.sh never appends --checkpoint-every-n-tokens …
[FAIL] scripts/bootstrap-upgrade.sh never appends --no-cache-prompt …
```

All three pass with the fix. Also green: `tests/contracts/test-llama-parallel-flag.sh`, `tests/test-macos-env-quote-strip.sh` (9 passed), `tests/test-macos-download-retries.sh`, `bash -n` on all three changed scripts.


Lightheartdevs:
Thanks for carrying the macOS long-context tunables through the native launchers.

What looks good:
- `ods/installers/macos/install-macos.sh`, `ods/installers/macos/ods-macos.sh`, and `ods/scripts/bootstrap-upgrade.sh` now append the missing llama.cpp tunables from the same `.env` sources they already use for the other native flags.
- The boolean `--no-cache-prompt` handling avoids enabling it for the documented falsey values, while `--checkpoint-every-n-tokens` stays value-bearing, matching the Windows/native patterns already in the tree.
- The new contract tests check the macOS `llama_args+=` append form rather than a loose grep, which should catch the exact Apple Silicon drift this fixes. The stacked head also includes the Intel/Arc overlay parity base and has green CI.

What I checked:
- Static diff review of the macOS launcher blocks, `ods/tests/contracts/test-llama-macos-tunables.sh`, `ods/tests/contracts/test-llama-parallel-flag.sh`, and `ods/tests/contracts/test-llama-runtime-tunables.py`.
- `gh pr checks 2131`: all checks passing, including compose validation, shell lint, dashboard API/frontend, multi-distro smoke, and PowerShell lint.
- Local execution was not available here because this runner lacks usable Bash and PyYAML, so I relied on CI for the new contract tests.

tang-vu:

Merged current upstream/main and retained both this branch's optional macOS llama.cpp tunables and upstream's N_GPU_LAYERS lifecycle in the install and bootstrap launchers.

- New head: 121f0a1f`n- macOS tunable contract: 3 launchers passed all 8 tunables
- LLAMA_PARALLEL contract: all 10 launch paths passed
- compose runtime tunables, GPU-layer contract, bootstrap hot-swap (35 checks), macOS update verification, syntax and ShellCheck: passed

Merge order remains #2130 first, then #2131; #2131 carries the additional long-context tunables.

PR #2132, opened 2026-07-26
fix(windows): apply LLAMA_REASONING on native llama-server launches

tang-vu:
## Problem

`.env.schema.json` documents the setting and why it exists:

> `LLAMA_REASONING` — "llama.cpp reasoning/thinking mode: off (default) | auto | on. **Off prevents thinking models from consuming the entire token budget on internal reasoning.**"

llama.cpp has its own default for `--reasoning-format`, so a launcher that never passes the flag does not merely ignore the operator's setting — it leaves reasoning at whatever the build prefers, which is exactly the outcome the ODS default exists to prevent.

Coverage today:

| launcher | applies `LLAMA_REASONING` |
|---|---|
| `docker-compose.base.yml:52` | ✅ via `LLAMA_ARG_REASONING` container env |
| `bin/ods-host-agent.py:10097` | ✅ |
| `installers/macos/install-macos.sh` / `ods-macos.sh` | ✅ |
| `scripts/bootstrap-upgrade.sh` — **Windows** PowerShell block, line 1173 | ✅ |
| `installers/windows/install-windows.ps1` | ❌ |
| `installers/windows/ods.ps1` | ❌ |

Note the third and fourth rows: Windows already has one launcher that handles this correctly and two that don't. That asymmetry within the same platform is what makes it an oversight rather than a deliberate exclusion.

## Fix

Read and map the value exactly the way `bootstrap-upgrade.sh`'s Windows block already does — `off` → `none`, `on` → `deepseek`, anything else passes through, absent defaults to `off` — so all three Windows paths agree with each other and with macOS, the host agent and Docker.

**Behaviour change worth calling out:** on a Windows native install this moves reasoning from the llama.cpp build default to `--reasoning-format none`. That is the documented ODS default and what every other platform already does, but it is a real change for anyone who was relying on the unset behaviour — flagging it rather than burying it.

## Tests

New `tests/contracts/test-llama-reasoning-format.sh`, wired into `make test`. It requires each native launcher to pass **both** `--reasoning-format` and to read `LLAMA_REASONING` — the `.env` values (`off`/`on`/`auto`) are not llama.cpp's vocabulary, so passing the flag without the mapping input would be just as broken. Docker is asserted on its own route (the `LLAMA_ARG_REASONING` env forward) since it has no argv flag.

On `main`:

```
[FAIL] installers/windows/install-windows.ps1 starts llama-server without --reasoning-format; LLAMA_REASONING is ignored and thinking mode falls back to the llama.cpp default
[FAIL] installers/windows/ods.ps1 starts llama-server without --reasoning-format; …
```

All seven pass with the fix. Also green: `tests/test-windows-installer-flags.sh` (30 passed), `tests/contracts/test-windows-llm-model-readiness.sh` (14 passed). Both `.ps1` files re-verified with `[System.Management.Automation.Language.Parser]::ParseFile` (0 parse errors).

### DIFF
--- a/ods/Makefile
+++ b/ods/Makefile
@@ -52,6 +52,7 @@ test: ## Run unit and contract tests
 	@bash tests/contracts/test-windows-llm-model-readiness.sh
 	@bash tests/contracts/test-windows-lemonade-swap-wait.sh
 	@bash tests/contracts/test-windows-hermes-config-patching.sh
+	@bash tests/contracts/test-llama-reasoning-format.sh
 	@echo ""
 	@echo "=== Overlay/plist contracts ==="
 	@bash tests/contracts/test-overlay-map-coherence.sh
--- a/ods/installers/windows/install-windows.ps1
+++ b/ods/installers/windows/install-windows.ps1
@@ -715,6 +715,18 @@ if ($dryRun) {
                     $parts = $_ -split '=', 2
                     $_llamaEnv[$parts[0].Trim()] = $parts[1].Trim().Trim('"')
                 }
+                # Map the .env values (off/on/auto) onto llama-server's own
+                # vocabulary, the same way scripts/bootstrap-upgrade.sh does for
+                # its Windows hot-swap. Defaulting to off keeps thinking models
+                # from spending the whole token budget on internal reasoning.
+                $_reasoning = $_llamaEnv["LLAMA_REASONING"]
+                if (-not $_reasoning) { $_reasoning = "off" }
+                switch ($_reasoning) {
+                    "off"   { $_reasoningFmt = "none" }
+                    "on"    { $_reasoningFmt = "deepseek" }
+                    default { $_reasoningFmt = $_reasoning }
+                }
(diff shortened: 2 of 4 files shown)


Lightheartdevs:
Thanks for tightening the Windows native llama-server launch parity around reasoning mode.

What looks good:
- `ods/installers/windows/install-windows.ps1` and `ods/installers/windows/ods.ps1` now map `LLAMA_REASONING` through the same off/on/auto vocabulary as the existing Windows hot-swap path, closing a real platform divergence without changing Docker or macOS behavior.
- `ods/tests/contracts/test-llama-reasoning-format.sh` is wired into `make test` and checks both the argv flag and the setting source, which is the right guard for this regression.
- Current CI is green, including shell lint, PowerShell lint on Ubuntu/Windows, dashboard API/frontend, env validation, and multi-distro smoke.

What I checked:
- Static diff review of the two PowerShell launcher paths and the new contract test.
- `gh pr checks 2132`: all checks passing.

PR #2139, opened 2026-07-26
fix(dashboard-api): bound array length and metadata count during GGUF inspection

vaibhavsrv:
## Problem

When inspecting untrusted or corrupted GGUF model files, `_skip_value()` and `_read_array()` in `gguf_inspector.py` looped over `range(length)` for array items without checking if `length` exceeds the total metadata buffer size (`len(reader.data)`). Similarly, `metadata_count` in `inspect_gguf()` ran `range(metadata_count)` unbounded. A corrupted or crafted file with an astronomical header length (e.g. 2^63 - 1) caused CPU resource exhaustion loops.

The inspector checked for max array nesting depth (`_MAX_ARRAY_DEPTH`) but did not validate that parsed array lengths or metadata count headers fit within the size of the read buffer.

## Fix

Bound array `length` and `metadata_count` against `len(reader.data)` before initiating loops in `_skip_value()`, `_read_array()`, and `inspect_gguf()`. If an array length or metadata count exceeds buffer size, raise a `ValueError`, allowing `inspect_gguf()` to fail closed safely with a degraded result.

## Verification

Added `test_oversized_metadata_count_degrades_gracefully` to `ods/extensions/services/dashboard-api/tests/test_gguf_inspector.py`. Ran `pytest` locally: 28/28 passed.

### DIFF
--- a/ods/extensions/services/dashboard-api/gguf_inspector.py
+++ b/ods/extensions/services/dashboard-api/gguf_inspector.py
@@ -134,6 +134,8 @@ def _skip_value(reader: _Reader, value_type: int, depth: int = 0) -> None:
             raise ValueError("GGUF array nesting too deep")
         item_type = reader.unpack("<I")
         length = reader.unpack("<Q")
+        if length > len(reader.data):
+            raise ValueError(f"GGUF array length {length} exceeds file buffer size")
         for _ in range(length):
             _skip_value(reader, item_type, depth + 1)
         return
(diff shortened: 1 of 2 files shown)


Lightheartdevs:
Thanks for hardening the GGUF inspector against corrupted length fields; that is the right failure mode for model files the dashboard should treat as untrusted input.

What looks good:
- `ods/extensions/services/dashboard-api/gguf_inspector.py` bounds `metadata_count` and array lengths before entering parser loops, so oversized headers degrade through the existing `ValueError` path instead of burning CPU.
- The readable/unknown fallback behavior stays consistent with the existing truncated-file and non-GGUF tests.
- `ods/extensions/services/dashboard-api/tests/test_gguf_inspector.py` adds focused coverage for an oversized metadata count and keeps the assertion on safe degradation rather than a hard crash.

What I checked:
- Diff/static review of `gguf_inspector.py` and `test_gguf_inspector.py`.
- Ran `python -m py_compile` on the changed Python files in a detached worktree for `353e9b4`.
- GitHub checks are green for the head, including API, integration smoke, lint, env validation, and distro/macOS/Linux smoke. Local pytest was not available in this Windows runner.

Thanks again for making this parser more defensive.

PR #2141, opened 2026-07-26
fix(dashboard-api): bound model context size to positive integers in model info helper

vaibhavsrv:
## Problem

`get_model_info()` in `helpers.py` parses `MAX_CONTEXT` or `CTX_SIZE` from `.env`. If `MAX_CONTEXT` is set to `0` or a negative integer (e.g. `MAX_CONTEXT=0`), `int(...)` evaluated to `0`, causing `ModelInfo` context length to be zero or negative. Downstream model calculations and context checks failed or divided by zero when using non-positive context limits.

The helper caught `ValueError` and `TypeError` for non-numeric values (falling back to 32768) but did not check if the parsed integer was non-positive (`<= 0`).

## Fix

Add a check `if context <= 0: context = 32768` in `get_model_info()`, ensuring context length is strictly positive.

## Verification

Added `test_non_positive_context_falls_back_to_default` to `ods/extensions/services/dashboard-api/tests/test_helpers.py`. Ran `pytest` locally: 105/105 passed.

### DIFF
--- a/ods/extensions/services/dashboard-api/helpers.py
+++ b/ods/extensions/services/dashboard-api/helpers.py
@@ -845,6 +845,8 @@ def get_model_info() -> Optional[ModelInfo]:
                 # the guard already used in routers/models.py.
                 try:
                     context = int(env_values.get("MAX_CONTEXT") or env_values.get("CTX_SIZE") or 32768)
+                    if context <= 0:
+                        context = 32768
                 except (TypeError, ValueError):
                     context = 32768
 
--- a/ods/extensions/services/dashboard-api/tests/test_helpers.py
+++ b/ods/extensions/services/dashboard-api/tests/test_helpers.py
@@ -102,6 +102,14 @@ class TestGetModelInfo:
         assert info is not None
         assert info.context_length == 32768
 
+    def test_non_positive_context_falls_back_to_default(self, install_dir):
+        env_file = install_dir / ".env"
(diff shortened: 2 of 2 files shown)


Lightheartdevs:
Thanks for tightening the model-info context guard; this is a small fix in a path the dashboard can hit repeatedly.

What looks good:
- `ods/extensions/services/dashboard-api/helpers.py` now treats non-positive `MAX_CONTEXT`/`CTX_SIZE` the same as non-numeric values, keeping `ModelInfo.context_length` safely positive instead of propagating `0` or negatives.
- `ods/extensions/services/dashboard-api/tests/test_helpers.py` adds a focused regression for `CTX_SIZE=0`, matching the existing `.env` parsing test style without widening unrelated behavior.
- The change is narrow, follows the existing fallback default of `32768`, and does not alter model-name, size, or quantization inference.

What I checked:
- Diff/static review of `helpers.py` and `test_helpers.py`.
- Ran `python -m py_compile` on the changed Python files in a detached worktree for `8d5bc3d`.
- GitHub checks are effectively green: lint, API, integration smoke, distro smoke, macOS/Linux smoke, and env validation jobs show success. `gh pr checks` still displayed `tier-2-env-validation` as pending, but the Actions job API shows a success conclusion and completion timestamp. Local pytest was not available in this Windows runner.

Thanks again for closing this edge case.

PR #2151, opened 2026-07-26
fix(performance-oracle): handle unicode decoding errors safely when reading env file

vaibhavsrv:
## Problem

`read_env_file_value()` in `performance_oracle.py` read environment configuration lines from `.env` using `env_path.read_text(encoding="utf-8")`. If `.env` contained invalid non-UTF8 binary data or corrupted bytes, `read_text()` raised an uncaught `UnicodeDecodeError` because `except OSError:` only caught filesystem OS errors.

`UnicodeDecodeError` inherits from `UnicodeError` -> `ValueError` -> `Exception`, not `OSError`, allowing binary file decoding errors to escape exception handling.

## Fix

Update `read_env_file_value()` exception tuple to `except (OSError, UnicodeError):`, ensuring corrupted `.env` file reads fall back safely to empty values without raising uncaught exceptions.

## Verification

Added `test_read_env_file_value_handles_unicode_decode_error` to `ods/extensions/services/dashboard-api/tests/test_performance_oracle.py`. Ran `pytest` locally: 43/43 passed.

### DIFF
--- a/ods/extensions/services/dashboard-api/performance_oracle.py
+++ b/ods/extensions/services/dashboard-api/performance_oracle.py
@@ -142,7 +142,7 @@ def read_env_file_value(key: str, install_dir: str | Path) -> str:
         for line in env_path.read_text(encoding="utf-8").splitlines():
             if line.startswith(f"{key}="):
                 return line.split("=", 1)[1].strip().strip("\"'")
-    except OSError:
+    except (OSError, UnicodeError):
         pass
     return ""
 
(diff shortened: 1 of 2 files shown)


Lightheartdevs:
Thanks for hardening the performance oracle `.env` reader; model ranking should fail soft around bad local config files.

What looks good:
- `ods/extensions/services/dashboard-api/performance_oracle.py` keeps the existing env-var precedence and normal parsing behavior while adding `UnicodeError` to the file-read fallback path.
- The new test in `ods/extensions/services/dashboard-api/tests/test_performance_oracle.py` writes invalid UTF-8 bytes directly, so it covers the exact decode failure that previously escaped `except OSError`.
- This is a narrow dashboard-api robustness fix with no change to model scoring, catalog normalization, or published-evidence matching.

What I checked:
- Reviewed the title/body, changed files, and diff against current `main`; there were no existing discussion comments or inline review threads for this PR.
- Ran `py -3.13 -m pytest tests/test_performance_oracle.py::test_read_env_file_value_handles_unicode_decode_error -q` from `ods/extensions/services/dashboard-api`: 1 passed.

PR #2152, opened 2026-07-26
fix(config): specify utf-8 encoding when reading extension manifest files

vaibhavsrv:
## Problem

`_read_manifest_file()` in `config.py` read JSON/YAML extension manifest files using `path.read_text()` without an explicit `encoding="utf-8"` parameter. On platforms or environments with non-UTF-8 default locale encodings (such as Windows default CP1252 or GBK environments), reading manifests containing UTF-8 characters raised `UnicodeDecodeError` or corrupted non-ASCII text fields (e.g. extension names, descriptions, icons).

`read_text()` used the host system's default locale encoding instead of forcing UTF-8 encoding.

## Fix

Pass `encoding="utf-8"` to `path.read_text()` in `_read_manifest_file()`, ensuring cross-platform UTF-8 manifest parsing.

## Verification

Added `test_read_manifest_file_uses_utf8_encoding` to `ods/extensions/services/dashboard-api/tests/test_config.py`. Ran `pytest` locally: 43/43 passed.

### DIFF
--- a/ods/extensions/services/dashboard-api/config.py
+++ b/ods/extensions/services/dashboard-api/config.py
@@ -157,7 +157,7 @@ def _apply_host_native_llm_service_override(
 
 def _read_manifest_file(path: Path) -> dict[str, Any]:
     """Load a JSON or YAML extension manifest file."""
-    text = path.read_text()
+    text = path.read_text(encoding="utf-8")
     if path.suffix.lower() == ".json":
         data = json.loads(text)
     else:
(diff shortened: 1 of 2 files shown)


Lightheartdevs:
Thanks for tightening manifest decoding; extension metadata is exactly the sort of cross-platform boundary where an implicit locale can bite Windows users.

What looks good:
- `ods/extensions/services/dashboard-api/config.py` now reads extension manifests with explicit UTF-8 decoding before JSON/YAML parsing, so non-ASCII manifest fields no longer depend on the host locale.
- `ods/extensions/services/dashboard-api/tests/test_config.py` adds a focused byte-level regression that proves UTF-8 text like `Prénom Test` survives the loader.
- The change stays scoped to the manifest file reader and does not alter schema validation, GPU filtering, or compose discovery behavior.

What I checked:
- Reviewed the title/body, changed files, and diff against current `main`; there were no existing discussion comments or inline review threads for this PR.
- Ran `py -3.13 -m pytest tests/test_config.py::test_read_manifest_file_uses_utf8_encoding -q` from `ods/extensions/services/dashboard-api`: 1 passed.

PR #2155, opened 2026-07-26
fix(models): parse float timeout environment variables safely with fallback defaults

vaibhavsrv:
## Problem

In `routers/models.py`, module-level timeout constants (`_MODEL_DISCOVERY_TIMEOUT_SECONDS`, `_AGENT_MODEL_STATUS_CACHE_TTL_SECONDS`, `_STALE_TERMINAL_DOWNLOAD_STATUS_SECONDS`, and `_STALE_ACTIVE_BOOTSTRAP_STATUS_SECONDS`) parsed environment variables directly with `float(os.environ.get(...))`. If any of these environment variables were set to invalid non-numeric strings (e.g. `DASHBOARD_MODEL_DISCOVERY_TIMEOUT="invalid"`), non-positive numbers, or non-finite positive floats (`inf`, `Infinity`, `1e309`), module import raised an uncaught `ValueError` or set infinite timeouts that disabled fallback logic.

Direct `float()` calls lacked exception handling for `ValueError` and `TypeError` when converting process environment strings, and did not check for finite range bounds (`math.isfinite(val) and val > 0`).

## Fix

Introduce `_safe_float_env(key, default)` helper that verifies `math.isfinite(val) and val > 0`, safely parsing environment float variables and falling back to default values when invalid, non-positive, or non-finite values are encountered.

## Verification

Extended `test_safe_float_env_handles_invalid_and_non_positive_values` in `ods/extensions/services/dashboard-api/tests/test_models.py` with non-finite (`inf`, `1e309`) and valid positive float cases. Ran `pytest` locally: 85/85 passed.

### DIFF
--- a/ods/extensions/services/dashboard-api/routers/models.py
+++ b/ods/extensions/services/dashboard-api/routers/models.py
@@ -82,16 +82,18 @@ _HF_AVATAR_CACHE_MAX_ENTRIES = 512
 _HF_AVATAR_CACHE: dict[tuple[str, str], tuple[float, str | None]] = {}
 _HF_AVATAR_CACHE_LOCK = threading.Lock()
 _IMPORTED_MODELS_LOCK = threading.Lock()
-_MODEL_DISCOVERY_TIMEOUT_SECONDS = float(os.environ.get("DASHBOARD_MODEL_DISCOVERY_TIMEOUT", "15.0"))
-_AGENT_MODEL_STATUS_CACHE_TTL_SECONDS = float(
-    os.environ.get("DASHBOARD_AGENT_MODEL_STATUS_CACHE_TTL", "0.5")
-)
-_STALE_TERMINAL_DOWNLOAD_STATUS_SECONDS = float(
-    os.environ.get("DASHBOARD_STALE_TERMINAL_DOWNLOAD_STATUS_SECONDS", "1800")
-)
-_STALE_ACTIVE_BOOTSTRAP_STATUS_SECONDS = float(
-    os.environ.get("DASHBOARD_STALE_ACTIVE_BOOTSTRAP_STATUS_SECONDS", "900")
-)
+def _safe_float_env(key: str, default: float) -> float:
+    try:
+        val = float(os.environ.get(key, str(default)))
+        return val if val > 0 else default
+    except (ValueError, TypeError):
+        return default
+
+
+_MODEL_DISCOVERY_TIMEOUT_SECONDS = _safe_float_env("DASHBOARD_MODEL_DISCOVERY_TIMEOUT", 15.0)
+_AGENT_MODEL_STATUS_CACHE_TTL_SECONDS = _safe_float_env("DASHBOARD_AGENT_MODEL_STATUS_CACHE_TTL", 0.5)
+_STALE_TERMINAL_DOWNLOAD_STATUS_SECONDS = _safe_float_env("DASHBOARD_STALE_TERMINAL_DOWNLOAD_STATUS_SECONDS", 1800.0)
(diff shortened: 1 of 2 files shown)


Lightheartdevs:
Thanks for hardening the dashboard-api model timeout parsing path; preventing a bad env value from taking down import is the right direction.

Quickest path to merge:
- In `ods/extensions/services/dashboard-api/routers/models.py`, make `_safe_float_env` reject non-finite positive floats such as `inf`, `Infinity`, and overflowed `1e309`, not just invalid strings and non-positive values. These constants feed `asyncio.wait_for(...)` timeouts plus stale-status/cache TTL checks, so an infinite value can disable the fallback behavior this helper is meant to preserve. A small `math.isfinite(val) and val > 0` guard would keep the accepted range explicit.
- In `ods/extensions/services/dashboard-api/tests/test_models.py`, extend `test_safe_float_env_handles_invalid_and_non_positive_values` with a positive non-finite case (`inf` or `1e309`) and a valid positive case, so the helper documents both fallback and pass-through behavior.

What I checked:
- Reviewed the PR metadata, changed files, and diff against `origin/main`.
- Checked call sites for `_MODEL_DISCOVERY_TIMEOUT_SECONDS`, `_AGENT_MODEL_STATUS_CACHE_TTL_SECONDS`, `_STALE_TERMINAL_DOWNLOAD_STATUS_SECONDS`, and `_STALE_ACTIVE_BOOTSTRAP_STATUS_SECONDS`.
- `python -m py_compile routers/models.py tests/test_models.py` passed in a detached worktree at this head.
- `git diff --check origin/main...origin/pr/2155` passed.
- I could not run the targeted pytest locally because this runner's Python environment is missing `pytest`; GitHub's dashboard API, lint, smoke, env, and security checks are green.

Thanks again for pushing this reliability fix forward.

vaibhavsrv:
Thanks @Lightheartdevs for the sharp catch regarding non-finite positive floats!

I've updated `routers/models.py` and `tests/test_models.py`:
1. **Added `math.isfinite` Check**: Updated `_safe_float_env` in `routers/models.py` to check `math.isfinite(val) and val > 0`, ensuring non-finite values (`inf`, `Infinity`, overflowed floats like `1e309`) fall back safely to default timeout values.
2. **Extended Test Coverage**: Extended `test_safe_float_env_handles_invalid_and_non_positive_values` in `tests/test_models.py` to cover non-finite floats (`inf`, `1e309`) and valid positive numbers (`30.0`).
3. **Updated PR Description**: Updated the PR body to document the `math.isfinite` check.

Local verification:
- `pytest ods/extensions/services/dashboard-api/tests/test_models.py` passed cleanly (85/85 passed).

Appreciate the review!

Lightheartdevs:
Thanks for tightening up the model timeout env parsing and for folding in the non-finite float follow-up.

What looks good:
- ods/extensions/services/dashboard-api/routers/models.py now routes all four module-level timeout/TTL constants through _safe_float_env, so invalid strings, non-positive values, NaN/Infinity, and overflowed floats fall back before import-time configuration can break the router or disable stale/cache behavior.
- ods/extensions/services/dashboard-api/tests/test_models.py covers invalid, negative, non-finite (inf, 1e309), and valid positive values, matching the regression from the previous advisory.
- The change is tightly scoped to dashboard-api model timeout parsing and does not touch installer, compose, auth, or runtime service wiring.

What I checked:
- Reviewed the PR metadata, prior advisory thread, current diff, changed files, and call sites for the four timeout constants.
- gh pr checks 2155 is green for Dashboard API/frontend, Test Linux, lint, type check, secret scan, env validation, and matrix smoke.
- In a detached worktree at 7a8f706, git diff --check origin/main...HEAD and python -m py_compile routers/models.py tests/test_models.py passed.
- Local pytest could not run in this runner because pytest is not installed; the GitHub Dashboard API pytest job passed.

Thanks again for turning the follow-up around.

PR #2175, opened 2026-07-26
fix(python): find the real interpreter when Windows Store aliases shadow PATH

tang-vu:
## Problem

`lib/python-cmd.sh` has two interpreter resolvers. Both reject the Windows App Execution Aliases (`%LOCALAPPDATA%\Microsoft\WindowsApps\python*.exe`) — correctly, since those are Store stubs, not interpreters, and the repo already has a contract test pinning that rejection. Only one of them recovers afterwards:

| | rejects Store aliases | falls back to `%LOCALAPPDATA%` install locations |
|---|---|---|
| `ods_detect_python_cmd_with_module` | yes | yes (`_ods_python_windows_candidates`) |
| `ods_detect_python_cmd` | yes | **no — returns 1** |

The aliases ship enabled by default on Windows, so under Git Bash they usually shadow **both** `python3` and `python` on PATH. `_ods_python_runnable` only inspects the first PATH hit for each name, so both probes reject and the generic resolver reports that no Python exists.

Reproduced on a Windows host with a working Python 3.13 install, clean checkout, no local changes:

```
$ command -v python;  command -v python3
/c/Users/<user>/AppData/Local/Microsoft/WindowsApps/python
/c/Users/<user>/AppData/Local/Microsoft/WindowsApps/python3

$ bash -c '. ods/lib/python-cmd.sh; ods_detect_python_cmd'
ERROR: Neither python3 nor python is available/runnable.

$ bash -c '. ods/lib/python-cmd.sh; ods_detect_python_cmd_with_module json'
/c/Users/<user>/AppData/Local/Python/bin/python.exe
```

Same file, same host, same interpreter — the sibling resolver finds it.

## Why it matters on the install path

`Get-UsableWindowsBash` (`installers/windows/install-windows.ps1:143-175`) prefers `Git\bin\bash.exe`, and the Windows installer runs `scripts/bootstrap-upgrade.sh` through it as a Scheduled Task. That script resolves its interpreter with `ods_detect_python_cmd` before rendering runtime configs:

- `scripts/bootstrap-upgrade.sh:1441` — empty resolver result → `ERROR: runtime config renderer is unavailable for Windows Lemonade` and the LiteLLM route regeneration is abandoned.
- `scripts/bootstrap-upgrade.sh:2656` — falls back to the literal `python3`, whose `command -v` hit *is* the Store alias, so the render is attempted against the stub. On failure the swap rolls back and `exit 1`s with status `failed` — after the full model has already been downloaded and verified.
- `scripts/bootstrap-upgrade.sh:3086` — same alias fallback for the Perplexica config update.

Other affected callers include `scripts/ods-doctor.sh:379`, `scripts/preflight-engine.sh:78`, and `extensions/services/token-spy/session-manager.sh`.

## Fix

After the PATH probes, sweep the same `_ods_python_windows_candidates` locations the module-aware resolver already uses (13 lines, `installers/`-free). Precedence is unchanged: `ODS_PYTHON_CMD`, then `ODS_PYTHON_PREFER_SYSTEM`, then PATH `python3`/`python`, and only then the Windows install locations.

Does not touch `ods_detect_python_cmd_with_module`, so it does not overlap the hunk #1934 edits in the same file.

## Verification

New `tests/test-python-cmd-resolver.sh`, gated in `Makefile` (`make test`) and `.github/workflows/test-linux.yml`:

1. both PATH names are Store aliases + real interpreter under `LOCALAPPDATA` → resolver returns it (**fails on `main`: returns empty, exit 1**)
2. a usable PATH `python3` still wins over the fallback (precedence regression guard)
3. nothing installed anywhere → still fails loudly
4. `ODS_PYTHON_CMD` override still wins

```
$ bash tests/test-python-cmd-resolver.sh          # this branch
[PASS] resolver falls back to the LOCALAPPDATA interpreter
[PASS] PATH python3 keeps precedence over the LOCALAPPDATA fallback
[PASS] resolver still reports failure when nothing is installed
[PASS] ODS_PYTHON_CMD override keeps precedence

$ git checkout main -- ods/lib/python-cmd.sh && bash tests/test-python-cmd-resolver.sh
[FAIL] resolver returned '' instead of the LOCALAPPDATA interpreter
```

`bash -n` clean on both shell files; `test-linux.yml` parses.

### DIFF
--- a/.github/workflows/test-linux.yml
+++ b/.github/workflows/test-linux.yml
@@ -195,6 +195,9 @@ jobs:
       - name: Issue-to-PR Security Contract Tests
         run: bash tests/test-issue-to-pr-security.sh
 
+      - name: Shared Python Resolver Tests
+        run: bash tests/test-python-cmd-resolver.sh
+
       - name: Upload Installer Simulation Artifacts
         uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7
         with:
--- a/ods/Makefile
+++ b/ods/Makefile
@@ -60,6 +60,9 @@ test: ## Run unit and contract tests
 	@echo "=== safe .env loading (quote/CRLF handling) ==="
 	@bash tests/test-safe-env.sh
 	@echo ""
+	@echo "=== shared Python interpreter resolver ==="
+	@bash tests/test-python-cmd-resolver.sh
+	@echo ""
 	@echo "=== QR code helper LAN detection ==="
 	@bash tests/test-qrcode-helper.sh
 	@echo ""
--- a/ods/lib/python-cmd.sh
+++ b/ods/lib/python-cmd.sh
@@ -115,6 +115,19 @@ ods_detect_python_cmd() {
         return 0
     fi
 
+    # On Windows Git Bash the App Execution Aliases usually shadow both names on
+    # PATH, so the checks above reject python3 and python even when a real
+    # interpreter is installed under LOCALAPPDATA. Scan the same install
+    # locations the module-aware resolver uses instead of giving up.
+    local candidate
+    while IFS= read -r candidate; do
+        if _ods_python_runnable "$candidate"; then
+            _ods_python_cmd_cached="$candidate"
+            printf '%s' "${_ods_python_cmd_cached}"
+            return 0
+        fi
+    done < <(_ods_python_windows_candidates)
+
     echo "ERROR: Neither python3 nor python is available/runnable." >&2
     return 1
 }
--- a/ods/tests/test-python-cmd-resolver.sh
+++ b/ods/tests/test-python-cmd-resolver.sh
@@ -1,0 +1,80 @@
+#!/usr/bin/env bash
+# Test lib/python-cmd.sh: ods_detect_python_cmd interpreter discovery.
+#
+# The resolver deliberately rejects the Windows App Execution Aliases
+# (%LOCALAPPDATA%\Microsoft\WindowsApps\python*.exe) because they are Store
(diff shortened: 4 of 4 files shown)


Lightheartdevs:
Thanks for closing the gap between the two Python resolvers; the Windows Git Bash alias case is a real installer recovery edge.

What looks good:
- `ods/lib/python-cmd.sh` now gives `ods_detect_python_cmd` the same Windows `LOCALAPPDATA` candidate sweep that the module-aware resolver already had, while preserving the existing precedence for `ODS_PYTHON_CMD`, system preference, and usable PATH interpreters.
- The fix is narrowly placed after the PATH probes, so it only helps when `python3`/`python` are unavailable or rejected as WindowsApps aliases and does not weaken the alias rejection contract.
- `ods/tests/test-python-cmd-resolver.sh` covers both shadowed PATH names plus a real local install, PATH precedence, a genuinely missing interpreter, and the explicit override; the test is also wired into `ods/Makefile` and `.github/workflows/test-linux.yml`.

What I checked:
- Reviewed the PR title/body, changed files, full diff, resolver call sites, and the existing WindowsApps hardening contract in `ods/tests/contracts/test-installer-hardening.sh`.
- In a detached worktree at this head under Git Bash: `git diff --check origin/main...HEAD`, `bash -n lib/python-cmd.sh`, `bash -n tests/test-python-cmd-resolver.sh`, and `bash tests/test-python-cmd-resolver.sh` passed.
- `gh pr checks 2175` is green for Test Linux `integration-smoke`, Matrix Smoke, ShellCheck, Dashboard frontend/api, Python lint, env validation, secret scan, and PowerShell lint. The Claude Review Summary job is skipped, not failing.

PR #2177, opened 2026-07-26
fix(dashboard): open the service UI path from the Settings routing table

tang-vu:
## Problem

`RouteRow` in the Settings routing table builds its link from the port alone:

```js
const href = getExternalUrl(service.port)   // http://<host>:<port>
```

Every service whose UI does not live at the port root therefore opens the wrong page:

| service | manifest `ui_path` | Settings link lands on |
|---|---|---|
| Qdrant (`:6333`) | `/dashboard` | version JSON blob |
| LiteLLM (`:4000`) | `/ui/` | `{"detail":"Not Found"}` |
| Token Spy (`:9092`) | `/dashboard` | not the dashboard |

(`privacy-shield` and `tts` declare `/docs`, `embeddings` `/info` — same shape.)

## The path is already available client-side

Two things already handle this correctly, which is what makes the Settings row the outlier:

- **Sidebar Quick Links** — `plugins/registry.js:69` appends `link.ui_path` whenever it is not `/`.
- **dashboard-api** — `_service_public_url` (`main.py:627`) resolves the manifest `ui_path` and `_serialize_services` ships it to the client as each service's `url`/`href`. Settings already consumes that payload (`/api/settings/summary` → `setServices(statusData.services)`), it just ignored the field.

The server builds that URL against `127.0.0.1`, so the host still has to come from the page — that is why the fix takes only the path and keeps `getDashboardHost()`. This keeps remote access (dashboard opened from another device on the LAN) working, which is the reason `getExternalUrl` exists in the first place.

## Fix

```js
const getServiceUiPath = (service) => { /* pathname (+search) from service.url/href */ }
...
const base = getExternalUrl(service.port)
const href = base ? `${base}${getServiceUiPath(service)}` : null
```

Root `ui_path` contributes nothing (no trailing `/` appended), a missing/unparseable `url` degrades to the previous behaviour, and portless services stay unlinked.

## Verification

New case in `src/pages/Settings.test.jsx` (existing harness, no fixture changes):

```
$ npx vitest run src/pages/Settings.test.jsx --no-cache     # this branch
Tests  9 passed (9)

$ git checkout main -- src/pages/Settings.jsx && npx vitest run src/pages/Settings.test.jsx --no-cache
× routing table links open the service UI path, not the bare port root
Tests  1 failed | 8 passed (9)
```

Asserts `/dashboard` and `/ui/` survive on the browser hostname, a `/` path does not become a trailing segment, and a portless service renders no link. `npm run lint` clean.

### DIFF
--- a/ods/extensions/services/dashboard/src/pages/Settings.jsx
+++ b/ods/extensions/services/dashboard/src/pages/Settings.jsx
@@ -95,6 +95,21 @@ const getErrorText = (err) => (
 const getDashboardHost = () => (typeof window !== 'undefined' ? window.location.hostname : 'localhost')
 const getExternalUrl = (port) => (port ? `http://${getDashboardHost()}:${port}` : null)
 
+// /api/status already resolves each service's manifest ui_path into `url`, but
+// against 127.0.0.1. Take the path from there and keep the browser's hostname
+// so a service whose UI is not at the port root (LiteLLM /ui/, Qdrant and
+// Token Spy /dashboard) opens its UI instead of the bare API root.
+const getServiceUiPath = (service) => {
+  const declared = service?.url || service?.href
+  if (!declared) return ''
+  try {
+    const { pathname, search } = new URL(declared)
+    return `${pathname === '/' ? '' : pathname}${search}`
+  } catch {
+    return ''
+  }
+}
+
 const todayKey = () => {
   const now = new Date()
   const month = String(now.getMonth() + 1).padStart(2, '0')
@@ -861,7 +876,8 @@ function RouteStatusCard({ tone, label, count, description }) {
 }
 
 function RouteRow({ service }) {
-  const href = getExternalUrl(service.port)
+  const base = getExternalUrl(service.port)
+  const href = base ? `${base}${getServiceUiPath(service)}` : null
   const healthy = service.status === 'healthy'
   const degraded = service.status === 'degraded'
   const dot = healthy ? 'bg-emerald-400' : degraded ? 'bg-amber-400' : 'bg-red-400'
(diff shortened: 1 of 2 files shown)


Lightheartdevs:
Thanks for fixing the Settings routing table path handling; this closes a real papercut for services whose UI lives below the port root.

What looks good:
- `ods/extensions/services/dashboard/src/pages/Settings.jsx` now takes only the pathname/search from the service `url`/`href` while preserving the browser hostname, which matches the Dashboard's LAN-access assumption and the existing sidebar quick-link behavior.
- `ods/extensions/services/dashboard/src/pages/Settings.test.jsx` covers Qdrant `/dashboard`, LiteLLM `/ui/`, a root-path Open WebUI link, and a portless internal service, so the previous bare-port behavior is pinned.
- The change is tightly scoped to Settings route links and does not touch dashboard-api serialization, compose, installer, auth, or runtime service wiring.

What I checked:
- Reviewed the PR title/body, changed files, full diff, `RouteRow` context, manifest `ui_path` conventions, and dashboard-api `_service_public_url` behavior.
- In a detached worktree at this head: `git diff --check origin/main...HEAD`, `npm ci`, `npx vitest run src/pages/Settings.test.jsx --no-cache`, and `npm run lint` passed.
- `gh pr checks 2177` is green for Dashboard frontend/api, Test Linux `integration-smoke`, Matrix Smoke, ShellCheck, Python lint, env validation, secret scan, and PowerShell lint. The Claude Review Summary job is skipped, not failing.

tang-vu:
Superseded by the shared serviceUrl helper now on main. Settings already calls serviceUrl(service), which preserves public_url and falls back through external_port/port plus ui_path, with coverage in src/lib/serviceUrls.test.js. Closing instead of force-pushing an empty rebased branch.

PR #2178, opened 2026-07-26
fix(windows): size the disk preflight to the selected model

tang-vu:
## Problem

The Windows disk gate in phase 04 is a static per-tier table:

```powershell
$_minDiskGB = switch ($selectedTier) {
    "NV_ULTRA" { 100 }  "SH_LARGE" { 100 }  "SH_COMPACT" { 50 }
    "4" { 50 }  "3" { 35 }  "2" { 30 }  "1" { 25 }  "0" { 15 }  "CLOUD" { 10 }
}
```

Those numbers were sized for the tier map's own model. The catalog selector (`Resolve-CatalogModelRecommendation`) runs in phase 02, before this check, and fits candidates against **memory**, not disk — so it can promote a considerably larger entry and still write it into `$tierConfig`.

Worked example on tier 4 (>=40GB VRAM, minimum 50GB):

| | |
|---|---|
| selected catalog entry | `size_mb: 48500` |
| disk needed (model + Docker image layers) | ceil(48500/1024) + 15 = **63GB** |
| preflight requirement | **50GB** |

Preflight reports "Disk: 55 GB free OK (>= 50 GB for Tier 4)", then the download runs out of space. Tier 3 has the same shape: minimum 35GB, and a 24GB pick needs 39GB.

## Linux and macOS already do this

Both other platforms re-check against the resolved size with identical arithmetic — Windows is the only one that skips it:

- `installers/phases/04-requirements.sh:139-148` — `_model_disk_gb=$(( (LLM_MODEL_SIZE_MB + 1023) / 1024 )); _model_needed_gb=$(( _model_disk_gb + 15 ))`
- `installers/macos/install-macos.sh:1405-1428` — same, with the comment *"Prefer the catalog selector's exact model size when available; it can choose entries that do not fit the older tier-map filename heuristics."*

## Fix

`Get-WindowsModelDiskRequirementGB -ModelSizeMB <n> -TierMinimumGB <n>` returns `max(tier minimum, ceil(MB/1024) + 15)`:

- the existing tier floor is never lowered, only raised
- no resolved size (CLOUD, or `ODS_DISABLE_CATALOG_MODEL_SELECTOR=true`) keeps the tier minimum, matching the `-gt 0` guards on Linux/macOS
- the raise is printed so the number in the failure message is explainable

## Verification

New `tests/test-windows-disk-preflight.ps1`, gated in `lint-powershell.yml` next to the existing port-preflight contract and following the same AST-extraction harness:

```
[PASS] unknown model size keeps the tier minimum
[PASS] negative model size keeps the tier minimum
[PASS] small model does not weaken the tier minimum
[PASS] tier 3 requirement follows a 24GB model
[PASS] tier 4 requirement follows a 48500MB model
[PASS] model size rounds up to whole GB
[PASS] tier 4 minimum (50GB) is raised to 63GB for a 48500MB model
[PASS] phase 04 applies the requirement to the selected model
```

Run on Windows PowerShell 5.1 (the edition the installer uses). Rounding is asserted against the Linux integer-division result for the same inputs (48500 → 63, 21110 → 36, 24576 → 39). The tier-4 case reads the minimum out of the phase's own switch rather than hardcoding 50, so it tracks the table.

On `main` the suite fails: the helper does not exist, and the two phase-text assertions confirm the phase never reads `$tierConfig.ModelSizeMB`.

PSScriptAnalyzer (`Error,Warning`, repo settings) clean on `04-requirements.ps1`.

### DIFF
--- a/.github/workflows/lint-powershell.yml
+++ b/.github/workflows/lint-powershell.yml
@@ -88,6 +88,10 @@ jobs:
         shell: pwsh
         run: ./tests/test-windows-python-resolver.ps1
 
+      - name: Windows Disk Preflight Contract
+        shell: pwsh
+        run: ./tests/test-windows-disk-preflight.ps1
+
       - name: Windows Runtime Recovery Contract
         shell: pwsh
         run: ./tests/contracts/test-windows-runtime-recovery.ps1
--- a/ods/installers/windows/phases/04-requirements.ps1
+++ b/ods/installers/windows/phases/04-requirements.ps1
@@ -158,6 +158,28 @@ function Test-WindowsODSLemonadeOwnsPort {
     }).Count
 }
 
+function Get-WindowsModelDiskRequirementGB {
+    <#
+    .SYNOPSIS
+        Disk requirement for the tier, raised to fit the selected model.
+    .DESCRIPTION
+        The per-tier minimums assume the tier map's own model. The catalog
+        selector can promote a larger one, so the download would exceed a
+        minimum that already passed. Mirrors the model-aware recheck in
+        installers/phases/04-requirements.sh and installers/macos/
+        install-macos.sh: model size rounded up to GB plus 15GB for Docker
+        image layers. Never returns less than the tier minimum.
+    #>
+    param(
+        [int]$ModelSizeMB,
+        [int]$TierMinimumGB
+    )
+
+    if ($ModelSizeMB -le 0) { return $TierMinimumGB }
+    $modelGB = [int][Math]::Ceiling($ModelSizeMB / 1024.0)
+    return [Math]::Max($TierMinimumGB, $modelGB + 15)
+}
+
 function Stop-WindowsODSLemonadePortConflicts {
     <#
     .SYNOPSIS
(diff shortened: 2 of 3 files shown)


Lightheartdevs:
Thanks for closing the Windows preflight gap around catalog-selected model sizes.

What looks good:
- `ods/installers/windows/phases/04-requirements.ps1` now reads `$tierConfig.ModelSizeMB` after phase 02 selection and raises, but never lowers, the tier disk floor.
- `Get-WindowsModelDiskRequirementGB` mirrors the Linux/macOS arithmetic: selected model size rounded up to GB plus 15 GB for Docker/image overhead.
- `ods/tests/test-windows-disk-preflight.ps1` exercises disabled/unknown selector size, smaller models, tier 3 and tier 4 promoted models, rounding, and verifies the phase actually applies the helper.

What I checked:
- Static diff review of the Windows phase, the new PowerShell contract, and the `lint-powershell.yml` wiring.
- Compared the helper against the existing Linux/macOS model-aware disk requirement called out in the PR.
- CI is green for this head, including PowerShell lint/contracts on both `ubuntu-latest` and `windows-latest`, plus the broader repo checks.

Issue #2149, opened 2026-07-26
Bootstrap GGUF download: support resume, improve progress reporting, and preserve partial downloads


TeamLider9141:
# Bootstrap GGUF download: support resume, improve progress reporting, and preserve partial downloads

First of all, thank you for building ODS. I've been testing the Linux installer (v2.5.3) on my machine and really like the overall experience.

While installing, I ran into a few UX issues around the bootstrap GGUF model download. None of these are critical bugs, but I think they would noticeably improve the installation experience, especially for users with slower or unstable internet connections.

---

## Environment

- Ubuntu 24.04.4 LTS
- AMD Ryzen 7 5800U
- 16 GB RAM
- ODS v2.5.3
- Tier 1 (AMD)

---

# 1. Resume partial downloads instead of restarting

### Current behaviour

The installer downloads the bootstrap model as:

```
Qwen3.5-2B-Q4_K_M.gguf.part
```

If the installer is interrupted (for example, pressing **Ctrl+C**) and started again, it verifies the partial file, reports a SHA256 mismatch, deletes it, and starts downloading from the beginning.

Since the file was intentionally interrupted, it isn't actually corrupt—it's simply incomplete.

For users on slower or unstable connections this means downloading hundreds of megabytes again.

### Suggested improvement

If a `.part` file already exists:

```
Found partial download.
Resuming download...
```

Resume the download (for example using `curl -C -`), then verify the SHA256 checksum only after the download has completed.

Suggested flow:

```
.part exists
        │
        ▼
Resume download
        │
        ▼
Download finished
        │
        ▼
Verify SHA256
        │
        ▼
Rename .part → .gguf
```

---

# 2. More informative download progress

During the bootstrap download the installer mostly shows a spinner and rotating messages.

The messages are nice, but it's difficult to know whether the download is progressing normally or appears stuck.

Something like this would be much more helpful:

```
Downloading bootstrap model...

397 MB / 1221 MB (32%)

Speed: 5.4 MB/s

ETA: 2m 48s
```

Even a simple progress bar would make a big difference.

---

# 3. Differentiate incomplete downloads from corrupted files

Currently interrupted downloads are reported as:

```
Model file is corrupt.
```

That wording can be confusing because an interrupted download isn't necessarily corrupted.

Something like this may better describe the situation:

```
Partial download detected.
Resuming...
```

or

```
Incomplete download found.
```

[...]

PR #2179, opened 2026-07-26
feat(installer): show download progress and resume state for the model fetch

tang-vu:
Addresses the reportable parts of #2149 (points 1, 3, 5, 6).

## Problem

The model download is the longest step of a Linux install and the least legible one. `installers/phases/11-services.sh` runs curl in the background with **all output redirected to `logs/model-download.log`**, so the terminal only ever shows a spinner plus rotating lore:

```bash
curl -fSL -C - ... -o "$GGUF_DIR/$GGUF_FILE.part" "$GGUF_URL" >> ".../model-download.log" 2>&1 &
spin_task $dl_pid "Downloading $GGUF_FILE"
```

A 1.5GB fetch over a hotspot and a socket that stalled ten minutes ago look exactly the same. From the issue:

> When I was testing the installer over a mobile hotspot, this was the first thing I looked for. Knowing that the downloaded size is increasing gives confidence that everything is working correctly.

The resume story has the same shape. curl already resumes (`-C -` into the same `.part`), and nothing on the failure path deletes the partial — the behaviour is correct, it is just never communicated, so an interrupted install looks like it threw the bytes away.

macOS already says all of this: `installers/macos/lib/ui.sh::download_with_progress` renders `--progress-bar` and, on failure, prints *"Re-run the installer to resume the download."*

## Change

**`installers/lib/ui.sh`**
- `download_part_bytes` — size of the in-flight `.part` (GNU `stat -c%s`, then BSD `stat -f%z`); a not-yet-created file reads 0 so it can be polled from the first tick.
- `format_download_progress` — `397 MB / 1221 MB (32%)`, or plain `397 MB` when no expected size is known. Pure, so it is unit-testable.
- `spin_task` — two **optional** args (part file, expected MB). When present the label carries the live counter; existing 2-arg callers are unchanged.

**`installers/phases/11-services.sh`**
- announce a resumed partial before downloading: `Found partial download: 397 MB / 1500 MB (26%). Resuming...`
- pass the `.part` to the spinner
- on total failure: `Partial download preserved: 397 MB / 1500 MB (26%). Re-run the installer to resume it.`

**`installers/lib/bootstrap-model.sh`**
- `BOOTSTRAP_GGUF_SIZE_MB=1500`. `LLM_MODEL_SIZE_MB` describes the *full* model, but fast-start installs are fetching the 2B bootstrap one — using it would have shown `397 MB / 18600 MB (2%)` for a 1.5GB file. The value is the size `tier-map.sh` already declares for the same GGUF, and it only drives display: a stale estimate degrades the percentage, never the install. The percentage is clamped at 100 so drift cannot render `118%`.

Result:

```
  ▸ Found partial download: 397 MB / 1500 MB (26%). Resuming...
  ⠹ [01:12] Downloading Qwen3.5-2B-Q4_K_M.gguf — 612 MB / 1500 MB (40%)
```

## Deliberately not included

- **speed / ETA** — the issue's own fallback position is that the byte count alone is what it needs; deriving a stable rate from a 1s poll adds a second state machine to a spinner for less signal.
- **`HF_TOKEN` rate-limit notice** (issue point 4) — belongs with the download helper, not this retry loop.
- Point 2's exact layout — kept inside the existing spinner rather than replacing it with a new multi-line renderer.

Happy to follow up on any of these if you'd like them here instead.

## Verification

`tests/bats-tests/ui.bats` (gated in CI as *BATS Unit Tests*), 39 pass:

```
ok 32 format_download_progress: reports downloaded MB when the size is unknown
ok 33 format_download_progress: adds total and percent when the size is known
ok 34 format_download_progress: never renders more than 100% for a stale estimate
ok 35 format_download_progress: treats missing or non-numeric input as zero
ok 36 download_part_bytes: reads the size of an in-flight download
ok 37 download_part_bytes: reads zero before curl creates the file
ok 38 spin_task: shows the download counter when a part file is supplied
ok 39 spin_task: keeps the plain label when no part file is supplied
```

Test 38 asserts the rendered label against a real background task and a real `.part`; on `main` it fails because `spin_task` ignores the extra arguments. The spinner tests drive `spin_task` with a redirect rather than `run`/`$( )`, since both fork a subshell that cannot `wait` on the test shell's child.

Also re-ran the existing suites that source this library: `test-docker-image-pull-retry.sh`, `test-phase11-local-build-failure.sh`, `test-install-readiness-summary.sh` — all pass. `bash -n` clean on all four files.


Lightheartdevs:
Thanks for making the long model download legible without changing the installer?s retry and resume behavior.

What looks good:
- `ods/installers/lib/ui.sh` keeps existing two-argument `spin_task` callers compatible and only adds the live counter when a `.part` path is passed.
- `ods/installers/phases/11-services.sh` uses the selected model size for normal downloads and the new bootstrap-specific size for fast-start, so the displayed denominator does not accidentally use the full model size during bootstrap.
- The retry, `curl -C -`, `.part` preservation, final `mv`, and SHA verification paths remain intact; this is mostly visibility around an existing resumable flow.

What I checked:
- Static diff review of `ui.sh`, `11-services.sh`, `bootstrap-model.sh`, and `tests/bats-tests/ui.bats`.
- Checked existing `spin_task` callers to confirm the new optional parameters do not require call-site changes.
- CI is green for this head: BATS-backed installer tests through the repo workflows, ShellCheck, matrix smoke, dashboard/API jobs, PowerShell, and secret scan.

gabsprogrammer:
Implemented on current `main` at `22f179a2`.

### Changes
- Aligned the pinned bootstrap GGUF and Tier 0 metadata with the actual Hugging Face object size: `1,280,835,840` bytes (`1,221 MiB`). A completed transfer now reports `1221 MB / 1221 MB (100%)` instead of stopping near 81%.
- Added explicit installer-owned download state for the active PID, partial file, and expected size.
- A confirmed double `Ctrl+C` now terminates only the installer-owned download child, waits for it to exit, and reports the preserved `.part` file for the next resumable run.
- Centralized the resumable receipt used by cancellation and exhausted retries.
- Preserved the existing fallback contract: the Hugging Face fallback is still entered only after a failed primary transfer, not after an unrelated post-download move failure.

### Regression coverage
- Exact pinned artifact completion renders 100%.
- Tier 0 bootstrap size metadata remains synchronized.
- Partial and missing `.part` states are handled correctly.
- Cancellation test starts a real child process, terminates it, clears the PID, and proves the resumable receipt.

### Validation
- Focused BATS: `82/82` passed.
- Tier-map suite: `128/128` passed.
- Docker image pull retry: `13/13` passed.
- Phase 11 local-build failure cases passed.
- Installer contract/hardening commands completed successfully.
- Touched Bash parsers, `git diff --check`, and a Tier 0 installer dry-run passed.
- Hugging Face response headers verified the pinned artifact size and SHA.

The six failures seen in the broad local BATS run are existing Windows/WSL fixture and CRLF differences; the same six reproduce unchanged on a clean current-main worktree. The published CI matrix is the authoritative cross-platform check for those lanes.

PR #2180, opened 2026-07-26
feat(dashboard): add the Workflows page the n8n docs already describe

tang-vu:
## Why

`extensions/services/n8n/README.md` tells operators:

> Templates are stored in `config/n8n/` as JSON files and **can be installed with one click from the Workflows page**.

There is no Workflows page. The backend has been complete for a while and nothing in the dashboard calls it:

| piece | state |
|---|---|
| `routers/workflows.py` | catalog, n8n status, enable, disable, executions — with tests |
| `config/n8n/catalog.json` | **18 workflows across 7 categories**, shipped |
| dashboard | zero references to `/api/workflows` |

So the templates ODS ships can only be installed with `curl` (which is exactly what the README's own examples fall back to) or by importing JSON by hand in n8n. This adds the missing surface; no backend changes.

## What

`src/pages/Workflows.jsx`, built against the existing response contract:

- catalog list with live state from n8n — `available` / `installed` / `active` — plus category, setup time, and execution count
- category filter chips built from the API's own `categories` map rather than a second hardcoded list
- **install / remove** call the existing `POST /api/workflows/{id}/enable|disable`, then re-read `/api/workflows`. n8n owns install state, so a locally-guessed update would drift from it — the test asserts the refetch happens.
- a workflow whose dependency is down cannot be installed and names the service that has to come up first, mirroring the `400 Missing dependencies: …` the API would otherwise return after a pointless round trip
- when n8n is unreachable (`n8nAvailable: false`) the page says so and disables installs instead of letting every click fail
- API errors surface their `detail` rather than failing silently

Route registration in `plugins/core.js` (order 2.2, next to Extensions/Integrations). The route is **always registered** so direct links and bookmarks resolve; the sidebar entry only appears once n8n is actually deployed, following the existing "GPU Monitor only on multi-GPU" pattern.

## Verification

```
$ npx vitest run --no-cache
Test Files  28 passed (28)
     Tests  259 passed (259)

$ npm run lint     # clean
$ npm run build    # ✓ built in 12.07s
```

New tests:

- `src/pages/Workflows.test.jsx` — listing and status badges, category filter, install round-trip (asserts `POST` + the catalog refetch), remove round-trip, blocked dependency disables the button and explains why, n8n-unreachable banner, catalog fetch failure
- `src/plugins/core.test.js` — the route is registered regardless of context; the sidebar entry is hidden with no n8n and when n8n is `not_deployed`, and present once n8n is deployed even while `unhealthy`/`down` (so the page is reachable precisely when it can explain itself)

## Notes

- `GET /api/workflows/{id}/executions` is deliberately left unused for now — a run-history drawer is a second interaction surface and belongs in its own change once this lands.
- Icons come from the catalog's `icon` field through a small map, falling back to `Workflow` for names the dashboard does not bundle.

### DIFF
--- a/ods/extensions/services/dashboard/src/pages/Workflows.jsx
+++ b/ods/extensions/services/dashboard/src/pages/Workflows.jsx
@@ -1,0 +1,224 @@
+import {
+  AlertCircle, Bot, Brain, Check, Code, Database, FileText, Image, Loader2,
+  MessageSquare, Mic, RefreshCw, Search, Shield, Trash2, Workflow, Zap,
+} from 'lucide-react'
+import { useCallback, useEffect, useState } from 'react'
+
+// Auth: nginx injects "Authorization: Bearer ${DASHBOARD_API_KEY}" for /api/
+// requests, so relative URLs need no explicit auth here (see Extensions.jsx).
+const fetchJson = async (url, options = {}, ms = 8000) => {
+  const c = new AbortController()
+  const t = setTimeout(() => c.abort(), ms)
+  try {
+    return await fetch(url, { ...options, signal: c.signal })
+  } finally {
+    clearTimeout(t)
+  }
+}
+
+// Catalog entries carry an icon name; anything unknown falls back to Workflow.
+const ICON_MAP = {
+  Bot, Brain, Code, Database, FileText, Image, MessageSquare, Mic,
+  Search, Shield, Workflow, Zap,
+}
+
+const STATUS_STYLES = {
+  active: 'bg-green-500/20 text-green-400',
+  installed: 'bg-blue-500/20 text-blue-400',
+  available: 'border border-theme-border text-theme-text-muted',
+}
+
+const STATUS_LABELS = {
+  active: 'Active',
+  installed: 'Installed',
+  available: 'Available',
+}
+
+export default function Workflows() {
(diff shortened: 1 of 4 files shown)


Lightheartdevs:
Thanks for wiring the shipped n8n templates into a real dashboard page rather than leaving that workflow stuck in docs and curl examples.

What looks good:
- `ods/extensions/services/dashboard/src/pages/Workflows.jsx` stays aligned with the existing `routers/workflows.py` contract: catalog read from `/api/workflows`, enable/disable via the existing POST endpoints, dependency state respected before install, and a refetch after n8n changes own the source of truth.
- The route registration in `plugins/core.js` keeps direct `/workflows` links valid while only showing the sidebar item when `n8n` is deployed, matching the repo?s pattern for conditional navigation.
- The tests cover catalog rendering, category filtering, enable/disable round trips, disabled installs when dependencies or n8n are unavailable, API error surfacing, and sidebar gating.

What I checked:
- Static diff review of `Workflows.jsx`, `Workflows.test.jsx`, `plugins/core.js`, and `plugins/core.test.js`.
- Compared the UI assumptions against the current dashboard-api workflow router and catalog response shape.
- CI is green for this head: frontend lint/test/build, dashboard-api tests, Ruff, ShellCheck, matrix smoke, PowerShell, and secret scan.

PR #2181, opened 2026-07-26
feat(dashboard): surface the Privacy Shield toggle and counters

tang-vu:
## Why

`extensions/services/privacy-shield/README.md`:

> Privacy Shield status is displayed in the ODS dashboard:
> - Service health indicator
> - **Enable/disable toggle**
> - **Statistics (requests processed, PII items scrubbed)**

Only the first one exists, and only because it falls out of the generic services list. The dashboard makes no request to `/api/privacy-shield/status`, `/stats`, or `/toggle` — all three endpoints are backend-only. Privacy Shield also ships headless (it is in the Extensions page's `HEADLESS_EXTENSIONS` set), so there is no service link that could stand in: starting, stopping, or inspecting the shield needs a terminal today.

## What

`components/PrivacyShieldCard.jsx`, mounted under the services panel on the Dashboard:

- current state plus the API's own `message`, with **Start / Stop** driving `POST /api/privacy-shield/toggle`
- counters from `/stats`: PII scrubbed, active sessions, cache state, port
- renders nothing — and issues no request — when `privacy-shield` is not deployed

Two contract details the card respects rather than assumes:

| endpoint behaviour | what the card does |
|---|---|
| `/toggle` answers **200** with `{"success": false, "message": …}` for host-agent failures (`AgentUnavailable`, `AgentTimeout`, …) | reads success from the payload, not the HTTP status |
| `/stats` answers **200** with `{"error": …}` when the shield is unreachable or `SHIELD_API_KEY` is unset | degrades counters to em dashes and leaves the toggle usable, instead of blanking the card |

Counters are limited to what `/stats` actually returns (`cache_enabled`, `cache_size`, `active_sessions`, `total_pii_scrubbed`). The README also advertises "requests processed", which the endpoint does not expose — I did not invent it.

## Verification

```
$ npx vitest run --no-cache
Test Files  29 passed (29)
     Tests  266 passed (266)

$ npm run lint     # clean
$ npm run build    # ✓ built
```

`components/PrivacyShieldCard.test.jsx` covers: not-deployed renders nothing and fetches nothing; counters render; the stop round-trip asserts `POST` with `{enable:false}` **and** the status refetch; the start affordance appears when the shield is down; a `200` + `success:false` toggle surfaces the host-agent message; missing stats keep the card usable; a failed status read is reported.

Placement note: the Dashboard was the only uncontended home that is always reachable — the shield has no page of its own and no service link. Happy to move it to Settings or a dedicated surface if you would rather it live elsewhere.

### DIFF
--- a/ods/extensions/services/dashboard/src/components/PrivacyShieldCard.jsx
+++ b/ods/extensions/services/dashboard/src/components/PrivacyShieldCard.jsx
@@ -1,0 +1,131 @@
+import { Loader2, Shield, ShieldOff } from 'lucide-react'
+import { useCallback, useEffect, useState } from 'react'
+
+// Auth: nginx injects the dashboard API key for /api/ requests, so relative
+// URLs need no explicit header here (see Extensions.jsx).
+const fetchJson = async (url, options = {}, ms = 8000) => {
+  const c = new AbortController()
+  const t = setTimeout(() => c.abort(), ms)
+  try {
+    return await fetch(url, { ...options, signal: c.signal })
+  } finally {
+    clearTimeout(t)
+  }
+}
+
+// /api/privacy-shield/stats answers 200 with an {error} body when the shield
+// is unreachable or SHIELD_API_KEY is unset, so "no stats" is a normal state.
+const readStats = (payload) => (
+  payload && !payload.error && typeof payload.total_pii_scrubbed === 'number' ? payload : null
+)
+
+export default function PrivacyShieldCard({ deployed = false }) {
+  const [status, setStatus] = useState(null)
+  const [stats, setStats] = useState(null)
+  const [error, setError] = useState(null)
+  const [notice, setNotice] = useState(null)
+  const [busy, setBusy] = useState(false)
+
+  const load = useCallback(async () => {
+    if (!deployed) return
+    try {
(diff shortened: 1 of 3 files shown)


Lightheartdevs:
Thanks for giving Privacy Shield an operator-facing surface instead of leaving the toggle and counters as backend-only endpoints.

What looks good:
- `ods/extensions/services/dashboard/src/components/PrivacyShieldCard.jsx` follows the existing `/api/privacy-shield/status`, `/stats`, and `/toggle` contracts, including `200` + `success:false` toggle failures and `{error}` stats responses.
- The Dashboard mount is gated on `privacy-shield` being deployed, so stacks without the headless service avoid the extra status/stats calls.
- The new component test covers the important downgrade paths: not deployed, missing stats, failed status read, start/stop request body, status refetch, and host-agent failure messaging.

What I checked:
- Static diff review of `PrivacyShieldCard.jsx`, `PrivacyShieldCard.test.jsx`, and the `Dashboard.jsx` mount.
- Compared against the current dashboard-api Privacy Shield router contract in `routers/privacy.py` and the existing headless-extension handling.
- CI is green for this head: frontend lint/test/build, dashboard-api tests, Ruff, ShellCheck, matrix smoke, PowerShell, and secret scan.

PR #2182, opened 2026-07-26
fix(perplexica): detect Lemonade runtime on AMD local installs

tang-vu:
## Problem

Five places decide which model id Perplexica must be pointed at. Three of them resolve it with a rule that can never fire:

```bash
# installers/phases/12-health.sh:311, installers/phases/13-summary.sh:313,
# scripts/repair/repair-perplexica.sh:26
_perplexica_backend="$(printf '%s' "${LLM_BACKEND:-${AMD_INFERENCE_RUNTIME:-}}" | tr '[:upper:]' '[:lower:]')"
if [[ "$_perplexica_backend" == "lemonade" ]]; then
    ..._MODEL="extra.$GGUF_FILE"
fi
```

`${LLM_BACKEND:-...}` only falls through when `LLM_BACKEND` is empty, and phase 06 always writes it:

```bash
# installers/phases/06-directories.sh:701,703
LLM_BACKEND=$(if [[ "$ODS_MODE_VALUE" == "lemonade" ]]; then echo "lemonade"; else echo "llama-server"; fi)
AMD_INFERENCE_RUNTIME=$(if [[ "$LEMONADE_EXTERNAL_VALUE" == "true" || ( "$GPU_BACKEND" == "amd" && "${ODS_MODE:-local}" == "local" ) ]]; then echo "lemonade"; else echo ""; fi)
```

So on a plain **AMD GPU install in the default local mode** the env is `LLM_BACKEND=llama-server` **and** `AMD_INFERENCE_RUNTIME=lemonade` — the `AMD_INFERENCE_RUNTIME` branch is dead code and all three conclude "not Lemonade".

The two implementations that get it right check the variables independently:

| Writer | Rule |
|---|---|
| `extensions/services/perplexica/sync-model-config.js:6-9` | `AMD_INFERENCE_RUNTIME \|\| LLM_BACKEND \|\| ODS_MODE` |
| `scripts/bootstrap-upgrade.sh:3108` | `[[ "$_runtime_for_perplexica" == "lemonade" \|\| "$_llm_backend_for_perplexica" == "lemonade" ]]` |

## Impact

On AMD the three broken copies produce a self-reinforcing loop:

1. **Phase 12** seeds `defaultChatModel` with the bare GGUF id — an id Lemonade does not serve.
2. The container entrypoint runs `sync-model-config.js` on every start and **corrects** it to the Lemonade id.
3. **Phase 13** compares the (now correct) stored value against its own wrong expectation, decides the config is `needed`, and runs the repair script:

```bash
# installers/phases/13-summary.sh:329-336
if [[ "$_perplexica_status" == "needed" ]]; then
    ai_warn "Perplexica config incomplete — running auto-setup..."
    ... bash "$INSTALL_DIR/scripts/repair/repair-perplexica.sh" ...
```

4. **The repair script** writes the broken id back, and reports `Perplexica configured`.

The same wrong id is written by the manual path documented in `docs/MODEL-MANAGEMENT.md:352` and `extensions/services/perplexica/README.md:45`.

A second mismatch on the external-Lemonade path: phase 06 can discover a real Lemonade model id into `LEMONADE_MODEL` (`06-directories.sh:418-424`), which `sync-model-config.js` and `bootstrap-upgrade.sh` both prefer. The three shell copies ignored it and always expected `extra.<GGUF_FILE>`, so seeding and validation were wrong there too.

## Fix

All three call sites now apply the rule the other two implementations already use: Lemonade when *either* variable says so, preferring `LEMONADE_MODEL` over the `extra.<GGUF_FILE>` fallback. No new abstraction — the logic stays inline next to its use, matching how `bootstrap-upgrade.sh` already carries its own copy.

## Tests

Extends `tests/test-perplexica-entrypoint.py`, which is already gated in CI (`.github/workflows/test-linux.yml:151`).

The three new tests slice the model-id block out of each shell file and execute it under `bash` against a fixed env matrix, so they assert real resolution rather than grepping for text:

| Case | `LLM_BACKEND` | `AMD_INFERENCE_RUNTIME` | `LEMONADE_MODEL` | expected id |
|---|---|---|---|---|
| AMD local | `llama-server` | `lemonade` | – | `extra.Modern-Model.gguf` |
| external Lemonade | `lemonade` | `lemonade` | `Qwen3-8B-GGUF` | `Qwen3-8B-GGUF` |
| NVIDIA/Apple/CPU | `llama-server` | – | – | `Modern-Model.gguf` |

Verified red on the pre-fix sources:

```
AssertionError: amd_local_runs_lemonade_under_llama_server_backend:
  expected extra.Modern-Model.gguf, got Modern-Model.gguf
```

Full module passes after the fix (`python3 tests/test-perplexica-entrypoint.py`, exit 0).

### DIFF
--- a/ods/installers/phases/12-health.sh
+++ b/ods/installers/phases/12-health.sh
@@ -387,9 +387,16 @@ if $DOCKER_CMD inspect ods-perplexica &>/dev/null; then
     PERPLEXICA_MODEL="${LLM_MODEL:-default}"
     if [[ -n "${GGUF_FILE:-}" ]]; then
         PERPLEXICA_MODEL="$GGUF_FILE"
-        _perplexica_backend="$(printf '%s' "${LLM_BACKEND:-${AMD_INFERENCE_RUNTIME:-}}" | tr '[:upper:]' '[:lower:]')"
-        if [[ "$_perplexica_backend" == "lemonade" ]]; then
-            PERPLEXICA_MODEL="extra.$GGUF_FILE"
+        # Lemonade serves the model under a separate id. An AMD local install
+        # runs Lemonade while LLM_BACKEND stays "llama-server", so the runtime
+        # and the backend have to be checked independently — same rule as
+        # scripts/bootstrap-upgrade.sh and the container-side
+        # extensions/services/perplexica/sync-model-config.js.
+        _perplexica_runtime="$(printf '%s' "${AMD_INFERENCE_RUNTIME:-}" | tr '[:upper:]' '[:lower:]')"
+        _perplexica_backend="$(printf '%s' "${LLM_BACKEND:-}" | tr '[:upper:]' '[:lower:]')"
+        if [[ "$_perplexica_runtime" == "lemonade" || "$_perplexica_backend" == "lemonade" ]]; then
+            PERPLEXICA_MODEL="${LEMONADE_MODEL:-}"
+            [[ -n "$PERPLEXICA_MODEL" ]] || PERPLEXICA_MODEL="extra.$GGUF_FILE"
         fi
     fi
     PERPLEXICA_LLM_BASE_URL="${LLM_API_URL:-http://llama-server:8080}"
--- a/ods/installers/phases/13-summary.sh
+++ b/ods/installers/phases/13-summary.sh
@@ -310,9 +310,17 @@ if ! $DRY_RUN; then
         _perplexica_model="${LLM_MODEL:-qwen3-30b-a3b}"
         if [[ -n "${GGUF_FILE:-}" ]]; then
             _perplexica_model="$GGUF_FILE"
-            _perplexica_backend="$(printf '%s' "${LLM_BACKEND:-${AMD_INFERENCE_RUNTIME:-}}" | tr '[:upper:]' '[:lower:]')"
-            if [[ "$_perplexica_backend" == "lemonade" ]]; then
-                _perplexica_model="extra.$GGUF_FILE"
+            # Lemonade serves the model under a separate id, so the expected
+            # route differs from the bare GGUF name. An AMD local install runs
+            # Lemonade while LLM_BACKEND stays "llama-server", so both
+            # variables have to be consulted independently — same rule as
+            # scripts/bootstrap-upgrade.sh and the container-side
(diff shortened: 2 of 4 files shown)


Lightheartdevs:
Thanks for taking on the Perplexica model-id path for AMD local installs; this is a subtle place for installer and runtime state to disagree.

What looks good:
- `ods/installers/phases/12-health.sh`, `ods/installers/phases/13-summary.sh`, and `ods/scripts/repair/repair-perplexica.sh` now use the same independent `AMD_INFERENCE_RUNTIME` / `LLM_BACKEND` rule as the container sync and bootstrap-upgrade paths.
- The fix preserves the discovered `LEMONADE_MODEL` preference and only falls back to `extra.<GGUF_FILE>` when that runtime id is absent.
- `ods/tests/test-perplexica-entrypoint.py` executes sliced shell blocks for AMD local, external Lemonade, and regular llama-server cases, which is stronger than a text-only grep for this regression.

What I checked:
- Static diff review against the five Perplexica model-id writers described in the PR.
- GitHub checks are green, including shell lint, `integration-smoke`, `api`, `frontend`, type check, env validation, and the distro/macOS smoke matrix.
- I did not run a live AMD/Perplexica install in this automation; that remains release-grade validation outside this advisory.

gabsprogrammer:
Rebased onto current `main` at `d6b7efdc` and resolved the conflict in `ods/tests/test-perplexica-entrypoint.py` additively:

- preserved the current-main search-adapter regression coverage;
- preserved the PR's AMD local, external Lemonade, and regular llama-server model-ID matrix;
- retained all three installer/repair execution tests in the standalone test entrypoint.

Validation on exact head `8fcf6656357fc8f3541774ccc7cdf9426dd1d20d`:

- `python -m pytest ods/tests/test-perplexica-entrypoint.py -q` -> `20 passed`
- `tests/test-installer-context-parity.sh` -> `67 passed`
- Ruff and `py_compile` -> passed
- `bash -n` for all three changed shell scripts -> passed
- ShellCheck at the repository's blocking severity -> passed
- `git diff --check` -> passed

The PR is mergeable again. GitHub CI is still completing on the rebased head; no check has failed so far.

PR #2185, opened 2026-07-26
fix(macos): write the OpenCode config.json compat file

tang-vu:
## Problem

OpenCode reads `config.json`, not `opencode.json`. Two of the three platform writers say so and act on it; macOS does not.

**Linux** — `installers/phases/07-devtools.sh:270-271`

```bash
# OpenCode reads config.json, not opencode.json — always sync
cp "$OPENCODE_CONFIG_DIR/opencode.json" "$OPENCODE_CONFIG_DIR/config.json"
```

**Windows** — `installers/windows/lib/opencode-config.ps1:139,183-184`

```powershell
$_ocCompatConfigFile = Join-Path $ConfigDir "config.json"
...
[System.IO.File]::WriteAllText($_ocConfigFile, $_configJson, $_utf8NoBom)
[System.IO.File]::WriteAllText($_ocCompatConfigFile, $_configJson, $_utf8NoBom)
```

**macOS** — `installers/macos/install-macos.sh:2701` calls `_write_macos_opencode_config` with a single path:

```bash
|| ! _write_macos_opencode_config \
     "$OPENCODE_CONFIG_DIR/opencode.json" \
     "$_opencode_model" "$_opencode_base_url" "$_opencode_api_key" \
     "${MAX_CONTEXT:-32768}"
```

and the writer only ever produces that one file. Nothing else on macOS writes `config.json` — `scripts/bootstrap-upgrade.sh` has a sync path for Windows only (`sync_windows_opencode_config`, line 500).

The directory is identical on all three platforms (`$HOME/.config/opencode` — `installers/macos/lib/constants.sh:64`, `installers/phases/07-devtools.sh:139`), so this is not a path-layout difference.

## Impact

A macOS install ends with OpenCode configured only in a file OpenCode does not read. The LaunchAgent (`installers/macos/install-macos.sh:2718+`) starts `opencode web`, it finds no `config.json`, and the `llama-server` provider, model id, base URL and API key the installer just wrote and verified are never loaded — while the installer prints:

```
OpenCode configured for <model> at <base-url>
```

The Windows compat write and the explicit Linux comment are the evidence that both files are intended; macOS is the odd one out.

## Fix

`_write_macos_opencode_config` derives the compat path next to the config path, renders the document once, and writes both through the same atomic `0600` replace it already used. The post-write verification now runs per file instead of only on the first, so a failure names which file failed.

No behaviour change on Linux or Windows.

## Tests

Extends `tests/test-installer-context-parity.sh` — the existing cross-platform installer parity suite, run by `make test` / `make gate`.

- Three static assertions pin the "writes config.json" contract for the Linux, Windows and macOS writers, so the next platform writer cannot quietly skip it.
- One behavioural check lifts `_write_macos_opencode_config` out of the installer, points its interpreter at the runner's `python3`, and runs it against a temp dir: both files must exist, be byte-identical, and the `config.json` must carry the active `model` and `options`. Skips cleanly when no `python3` is present.

Verified red against the pre-fix installer:

```
PASS: Linux OpenCode writer syncs config.json
PASS: Windows OpenCode writer syncs config.json
FAIL: macOS OpenCode writer syncs config.json
```

Green after: **73 passed**.

### DIFF
--- a/ods/installers/macos/install-macos.sh
+++ b/ods/installers/macos/install-macos.sh
@@ -420,18 +420,24 @@ HERMES_AUTH_VERIFY_PY
 
 _write_macos_opencode_config() {
     local config_path="$1" model_name="$2" base_url="$3" api_key="$4" context_length="$5"
+    # OpenCode reads config.json, not opencode.json, so the same document has
+    # to land in both files — matching installers/phases/07-devtools.sh on
+    # Linux and installers/windows/lib/opencode-config.ps1 on Windows.
+    local compat_path
+    compat_path="$(dirname "$config_path")/config.json"
     mkdir -p "$(dirname "$config_path")"
     ODS_OPENCODE_MODEL="$model_name" \
     ODS_OPENCODE_BASE_URL="$base_url" \
     ODS_OPENCODE_API_KEY="$api_key" \
     ODS_OPENCODE_CONTEXT="$context_length" \
-        /usr/bin/python3 - "$config_path" <<'OPENCODE_CONFIG_PY'
+        /usr/bin/python3 - "$config_path" "$compat_path" <<'OPENCODE_CONFIG_PY'
 import json
 import os
 import sys
 from pathlib import Path
 
 path = Path(sys.argv[1])
+compat_path = Path(sys.argv[2])
 try:
     data = json.loads(path.read_text(encoding="utf-8")) if path.exists() else {}
 except (OSError, ValueError):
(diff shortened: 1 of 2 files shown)


Lightheartdevs:
Thanks for closing the macOS OpenCode compatibility gap and lining it up with the Linux and Windows writers.

What looks good:
- `ods/installers/macos/install-macos.sh` now renders the OpenCode config once and atomically writes both `opencode.json` and the `config.json` file OpenCode actually reads, preserving the existing `0600` behavior.
- The post-write verification now checks each target file, so a partial compat write fails with the file name instead of silently leaving OpenCode unconfigured.
- `ods/tests/test-installer-context-parity.sh` adds both cross-platform static checks and a behavioral extraction of the macOS writer.

What I checked:
- Static diff review against the existing Linux `07-devtools.sh` sync and Windows `opencode-config.ps1` dual-write pattern.
- GitHub checks are green, including shell lint, `integration-smoke`, `macos-smoke`, PowerShell lint, and the dashboard API/frontend jobs.
- I did not run a live macOS OpenCode launch in this automation; this advisory is based on the targeted writer test plus CI.

PR #2187, opened 2026-07-26
fix(workflows): resolve dependency aliases before health-checking

tang-vu:
## Problem

`check_workflow_dependencies` resolves catalog dependency names through a two-entry hardcoded dict, then looks the result up in `SERVICES`:

```python
# routers/workflows.py:67-86
_DEP_ALIASES = {"ollama": "llama-server"}
...
for dep in deps:
    resolved = _DEP_ALIASES.get(dep, dep)
    if resolved in health_cache:
        results[dep] = health_cache[resolved]
    elif resolved in SERVICES:
        status = await check_service_health(resolved, SERVICES[resolved])
        ...
    else:
        results[dep] = True          # <-- never probed
```

`SERVICES` is keyed by manifest **service id**, but the catalog names services the way the CLI does — where aliases are first-class (`ods logs kokoro`, `ods restart stt`). A dependency that is an alias misses `SERVICES` entirely and lands in the `else` branch, which reports it as satisfied without any probe.

`config/n8n/catalog.json` has exactly one such name:

```json
{ "id": "voice-to-voice", "dependencies": ["whisper", "llama-server", "kokoro"] }
```

and `extensions/services/tts/manifest.yaml:7-9`:

```yaml
  id: tts
  aliases: [kokoro]
```

Every other dependency across the 18 catalog entries (`llama-server`, `qdrant`, `whisper`, `litellm`, `comfyui`, `privacy-shield`, `hermes`, `openclaw`) is a real service id.

## Impact

Kokoro TTS was always reported healthy for `voice-to-voice`:

- `GET /api/workflows` returns `dependencyStatus.kokoro = true` and `allDependenciesMet = true`, so the UI shows the dependency green with TTS stopped or never enabled.
- `POST /api/workflows/{id}/enable` only refuses when a dependency is `False` (`workflows.py:179-182`), so the workflow installs into n8n regardless.

The user gets a workflow that reports ready and fails at its TTS step at runtime, with nothing in the dashboard pointing at the cause.

## Fix

Aliases are already declared in the manifests, so this is a lookup that should be derived from data rather than hardcoded (the `ollama` entry stays — it is a legacy catalog name and *not* a manifest alias of `llama-server`, whose alias list is `[llm]`).

- `config.py` keeps `aliases` on the loaded service config and exposes `SERVICE_ALIASES`, a reversed alias → id map built by `build_service_aliases()`. A real service id is never shadowed by another service claiming it as an alias.
- `check_workflow_dependencies` resolves through `_DEP_ALIASES` first, then `SERVICE_ALIASES`.

The permissive `else: results[dep] = True` fallback is left as-is — flipping unknown names to "unmet" would be a separate behavioural decision about catalog entries that name non-services.

## Tests

`tests/test_workflows.py` (gated by `pytest tests/` in `.github/workflows/dashboard.yml:60`):

1. `test_check_workflow_dependencies_probes_service_behind_manifest_alias` — a dependency named `kokoro` must probe `tts` and report `False` when it is down; asserts the health check was actually awaited, so "assumed met" cannot pass.
2. `test_build_service_aliases_reverses_manifest_aliases` and `..._never_shadows_a_real_service_id` — pure-function coverage of the new map.
3. `test_shipped_catalog_dependencies_all_resolve_to_a_known_service` — reads the shipped catalogue and the real manifests and fails if any dependency names neither a service id nor an alias, so a future catalog entry cannot re-enter the silent branch.

Verified red by reverting only the resolution line (keeping the new `SERVICE_ALIASES` seam, so the failure is the defect and not a missing import):

```
>       assert result["kokoro"] is False
E       assert True is False
FAILED tests/test_workflows.py::test_check_workflow_dependencies_probes_service_behind_manifest_alias
1 failed, 3 passed
```

Green after: `tests/test_workflows.py` **49 passed**, `tests/test_config.py` **42 passed**.

### DIFF
--- a/ods/extensions/services/dashboard-api/config.py
+++ b/ods/extensions/services/dashboard-api/config.py
@@ -253,6 +253,7 @@ def load_extension_manifests(
                     "ui_path": service.get("ui_path", "/"),
                     "external_link": bool(service.get("external_link", True)),
                     "container_name": service.get("container_name", f"ods-{service_id}"),
+                    "aliases": [str(a) for a in (service.get("aliases") or [])],
                     "depends_on": service.get("depends_on", []),
                     "category": service.get("category", "optional"),
                     "host_network": bool(service.get("host_network", False)),
@@ -294,6 +295,22 @@ def load_extension_manifests(
 
 MANIFEST_SERVICES, MANIFEST_FEATURES, MANIFEST_ERRORS = load_extension_manifests(EXTENSIONS_DIR, GPU_BACKEND)
 SERVICES = MANIFEST_SERVICES
+
+
+def build_service_aliases(services: dict[str, dict[str, Any]]) -> dict[str, str]:
+    """Reverse the manifest `aliases` lists into alias -> service id."""
+    return {
+        alias: service_id
+        for service_id, config in services.items()
+        for alias in config.get("aliases", [])
+        if alias not in services
+    }
+
+
+# The CLI and the service manifests treat aliases as first-class service names
+# (`ods logs kokoro`), so anything that accepts a user- or catalog-supplied
+# service name has to resolve them before looking SERVICES up.
+SERVICE_ALIASES = build_service_aliases(SERVICES)
 if not SERVICES:
     logger.error("No services loaded from manifests in %s — dashboard will have no services", EXTENSIONS_DIR)
 
--- a/ods/extensions/services/dashboard-api/routers/workflows.py
+++ b/ods/extensions/services/dashboard-api/routers/workflows.py
@@ -9,7 +9,7 @@ import aiohttp
 from fastapi import APIRouter, Depends, HTTPException
 
 from config import (
-    SERVICES, WORKFLOW_DIR, WORKFLOW_CATALOG_FILE,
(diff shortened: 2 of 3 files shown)


Lightheartdevs:
Thanks for tightening the workflow dependency health path around manifest aliases, especially for the `kokoro`/`tts` catalog case.

What looks good:
- `ods/extensions/services/dashboard-api/config.py` now keeps manifest `aliases` and builds a data-derived `SERVICE_ALIASES` map instead of adding another hardcoded workflow exception.
- `ods/extensions/services/dashboard-api/routers/workflows.py` resolves legacy dependency aliases first, then manifest aliases, so the permissive unknown-dependency branch no longer masks the shipped `kokoro` dependency.
- `ods/extensions/services/dashboard-api/tests/test_workflows.py` covers both the alias map contract and the actual health probe for an alias-backed dependency.

What I checked:
- Static diff review against the shipped n8n workflow catalog and service manifest model.
- GitHub checks are green, including `api`, `frontend`, `integration-smoke`, lint, env validation, and the distro/macOS smoke matrix.

PR #2188, opened 2026-07-26
fix(windows): write ODS_DEVICE_NAME and HOST_LAN_IP to .env

tang-vu:
## Problem

Diffing the three `.env` writers by key, two are written by **Linux and macOS but never by Windows**:

| Key | `phases/06-directories.sh` | `macos/lib/env-generator.sh` | `windows/lib/env-generator.ps1` |
|---|---|---|---|
| `ODS_DEVICE_NAME` | line 620 | line 460 (+ backfill at 283) | — |
| `HOST_LAN_IP` | line 592 | line 456 (+ backfill at 316) | — |

Neither string appears anywhere under `installers/windows/`.

Both keys are optional in `.env.schema.json`, so nothing fails loudly — every consumer just takes its fallback.

## Impact

The macOS generator documents both failure modes in the comments on its backfill blocks, because macOS shipped without these keys once and had to repair existing installs:

**`ODS_DEVICE_NAME`** — `installers/macos/lib/env-generator.sh:280-282`

> `ODS_DEVICE_NAME` backfill: older macOS installs omitted this key, so magic links defaulted to `auth.ods.local`/`chat.ods.local` and **collided with every other default install on the LAN**.

The Caddyfile vhosts are all `{$ODS_DEVICE_NAME:ods}.local`, so a Windows box answers to `ods.local` and hands out owner cards on `auth.ods.local` — indistinguishable from any other default install on the same network.

**`HOST_LAN_IP`** — `installers/macos/lib/env-generator.sh:305-313`

> the fresh-install heredoc populates `HOST_LAN_IP` when `BIND_ADDRESS=0.0.0.0` was pre-set, so openclaw can extend `allowedOrigins` for LAN clients. Pre-existing installs that opted into LAN mode … have no `HOST_LAN_IP` — **openclaw then rejects LAN client requests until a manual .env edit**.

Windows is exactly that case permanently: `-Lan` writes `BIND_ADDRESS=0.0.0.0` (`env-generator.ps1:812`), openclaw is part of the Windows stack (`install-windows.ps1:2382`, `OPENCLAW_TOKEN` at `env-generator.ps1:924`), and `extensions/services/openclaw/compose.yaml:22` passes `HOST_LAN_IP=${HOST_LAN_IP:-}` — which resolves to empty forever.

## Fix

Two helpers plus three lines in the heredoc, mirroring the other two platforms:

- `Get-WindowsODSDeviceName` sanitises `$env:COMPUTERNAME` with the same rules as the bash sanitisers (lowercase → collapse non-alphanumerics to hyphens → trim to 32 → alphanumeric first/last), falling back to `ods`.
- `Get-WindowsODSHostLanIp` reuses the loopback/APIPA filter `Write-SuccessCard` already applies in `installers/windows/lib/ui.ps1:389-394`, so both surfaces report the same address.
- `BIND_ADDRESS` is hoisted into `$bindAddress` (same value, computed once) so `HOST_LAN_IP` detection can be gated on it. It stays empty on loopback installs, keeping the compose `${HOST_LAN_IP:-}` fallback meaningful.
- Both keys go through `Get-EnvOrNew`, which returns a non-empty persisted value — so a re-run backfills an older `.env` without discarding an operator override, matching Linux `_env_get` and the macOS backfill.

## Tests

New `tests/test-windows-env-device-identity.ps1`, wired into `.github/workflows/lint-powershell.yml` next to the other PowerShell contract tests (runs on `windows-latest`).

- Seven `Get-WindowsODSDeviceName` cases (plain, punctuation, empty, hyphen-only, 40-char truncation) — each asserted against the **`ODS_DEVICE_NAME` pattern read out of `.env.schema.json`**, not a copy of it, so the sanitiser cannot drift from the schema. The 40-char case pins that truncation does not leave a trailing hyphen.
- `Get-WindowsODSHostLanIp` must return a plain IPv4 or `""`, never loopback or APIPA — tolerant of whatever adapters the runner has.
- Static checks that the generator emits both keys, gates `HOST_LAN_IP` on `BIND_ADDRESS=0.0.0.0`, and routes both through `Get-EnvOrNew`.

On upstream the test fails immediately — the helper does not exist and neither key is emitted:

```
Get-WindowsODSDeviceName : The term 'Get-WindowsODSDeviceName' is not recognized...
CategoryInfo : ObjectNotFound: (Get-WindowsODSDeviceName:String)
```

Green after, run locally on Windows PowerShell 5.1 (exit 0):

```
[PASS] Get-WindowsODSDeviceName produces schema-valid names
[PASS] Get-WindowsODSHostLanIp returns a LAN address or an empty string
[PASS] Windows .env carries ODS_DEVICE_NAME and HOST_LAN_IP
[PASS] windows env device identity
```


Lightheartdevs:
Thanks for bringing the Windows `.env` identity keys up to parity with Linux and macOS.

What looks good:
- `Get-WindowsODSDeviceName` mirrors the existing sanitizer shape and preserves the `.env.schema.json` contract for proxy vhosts and magic-link URLs.
- `HOST_LAN_IP` is populated only for `BIND_ADDRESS=0.0.0.0`, uses the same LAN-address filter as the success card, and both new keys preserve existing operator values through `Get-EnvOrNew`.
- `ods/tests/test-windows-env-device-identity.ps1` checks helper behavior, schema compatibility, LAN-IP filtering, and static generator wiring, and it is now run by the PowerShell lint workflow.

What I checked:
- GitHub checks are green for this head, including Windows and Ubuntu PowerShell lint, shell/Python lint, dashboard checks, distro smoke, env validation, and secret scan.
- Static diff review of `ods/installers/windows/lib/env-generator.ps1`, the new contract test, and the workflow placement.

Thanks again for taking on the Windows LAN identity path here.

tang-vu:

Merged current upstream/main and resolved the PowerShell workflow conflict by retaining both the Windows device-identity contract and upstream's dotenv serializer contract.

- New head: 539bb3f1`n- device identity contract: 4 passed
- dotenv serializer contract: 6 passed
- PowerShell parse, workflow YAML parse, and git diff --check: passed

PR #2189, opened 2026-07-26
fix(linux): honour OPENCODE_PORT in the OpenCode systemd unit

tang-vu:
## Problem

`OPENCODE_PORT` is the documented way to move the OpenCode Web UI:

- `.env.schema.json:677` declares it,
- `config/ports.json:127` maps it (`{"env_var": "OPENCODE_PORT", "external_default": 3003, "service_id": "opencode", "compose_managed": false}`),
- `extensions/services/opencode/manifest.yaml:13` names it as `external_port_env`, which is what `config.py` reads to build the OpenCode link shown in the dashboard.

macOS and Windows launch OpenCode on it:

| Platform | Launcher |
|---|---|
| macOS | `installers/macos/lib/constants.sh:65` + `install-macos.sh:2929,3165` use `$OPENCODE_PORT` |
| Windows | `installers/windows/phases/07-devtools.ps1:120` — `opencode.exe web --port $($script:OPENCODE_PORT)` |
| Linux | `opencode/opencode-web.service` — `ExecStart=__OPENCODE_BIN__ serve --port 3003 --hostname 127.0.0.1` |

The Linux unit hard-codes the port, and `installers/phases/07-devtools.sh:281-287` only substitutes `__HOME__`, `__OPENCODE_BIN__` and `__OPENCODE_BIN_DIR__` — there is no port placeholder to fill.

## Impact

On Linux, setting `OPENCODE_PORT` in `.env` moves everything *except* the listener:

- `SERVICES["opencode"]["external_port"]` picks up the new value, so the dashboard tile and the sidebar link point at the new port and 404;
- the service keeps serving on 3003;
- a port conflict on 3003 has no resolution — the one knob the schema, `ports.json` and the manifest all advertise does nothing.

Nothing warns: the installer prints `OpenCode Web UI service installed (user-level, port 3003)` regardless of what `.env` says.

## Fix

- The unit carries `__OPENCODE_PORT__` alongside the placeholders it already has.
- Phase 07 reads `OPENCODE_PORT` from `.env` (same `grep | cut | tr` idiom the phase already uses for `ODS_DEVICE_NAME` at line 523), validates it as a 1–65535 integer, and substitutes it.
- A present-but-invalid value falls back to 3003 **with a warning**, so a typo is visible instead of silently ignored. A missing key falls back silently — 3003 is the schema default.
- The success line and the dry-run line now report the port actually used.

Phase 06 still does not write `OPENCODE_PORT` to `.env` on Linux; that is unchanged and harmless, since the fallback equals the schema default. This PR only makes an explicitly-set value take effect.

## Tests

Extends `tests/test-linux-opencode-path.sh` — the existing home for this unit's templating contract — and wires it into `make test` and `.github/workflows/test-linux.yml` (it was previously in neither).

Static:
- the unit templates the port and contains no `serve --port <number>` literal;
- the phase substitutes the placeholder.

Behavioural — the test lifts the phase's own resolution + rendering block out of `07-devtools.sh`, runs it against a temp `INSTALL_DIR` with a stubbed `_sed_i`/`ai_warn`, and inspects the rendered unit:

| `.env` | rendered `ExecStart` |
|---|---|
| `OPENCODE_PORT=3103` | `--port 3103` |
| key absent | `--port 3003` |
| `OPENCODE_PORT=not-a-port` | `--port 3003` |

plus an assertion that no placeholder survives rendering.

Verified red against the pre-fix sources:

```
--- OPENCODE_PORT override ---
[FAIL] systemd service must template the listen port, not hard-code it
```

Green after (all 7 new checks pass, full file green). `bash -n` clean.

### DIFF
--- a/.github/workflows/test-linux.yml
+++ b/.github/workflows/test-linux.yml
@@ -153,6 +153,9 @@ jobs:
       - name: CPU-Only Path Tests
         run: bash tests/test-cpu-only-path.sh
 
+      - name: Linux OpenCode Path Tests
+        run: bash tests/test-linux-opencode-path.sh
+
       - name: Hermes Slash Worker Prune Tests
         run: |
           bash tests/test-hermes-slash-worker-prune.sh
--- a/ods/Makefile
+++ b/ods/Makefile
@@ -40,6 +40,7 @@ test: ## Run unit and contract tests
 	@echo "=== Linux install preflight ==="
 	@bash tests/test-linux-install-preflight.sh
 	@bash tests/test-linux-host-agent-stop.sh
+	@bash tests/test-linux-opencode-path.sh
 	@bash tests/test-preflight-dashboard-port.sh
 	@echo ""
 	@echo "=== ods-doctor rootless subordinate IDs ==="
--- a/ods/installers/phases/07-devtools.sh
+++ b/ods/installers/phases/07-devtools.sh
@@ -17,7 +17,7 @@
 ods_progress 42 "devtools" "Installing developer tools"
 if $DRY_RUN; then
     log "[DRY RUN] Would install AI developer tools (Claude Code, Codex CLI, OpenCode)"
-    log "[DRY RUN] Would configure OpenCode for local llama-server (user-level systemd service on port 3003)"
+    log "[DRY RUN] Would configure OpenCode for local llama-server (user-level systemd service on \${OPENCODE_PORT:-3003})"
     log "[DRY RUN] Would install ODS host agent systemd service (system-mode, port 7710)"
     log "[DRY RUN] Would install ODS mDNS announcer systemd service (if zeroconf available)"
 else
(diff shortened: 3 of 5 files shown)


Lightheartdevs:
Thanks for making the Linux OpenCode unit honor the documented port override.

What looks good:
- `ods/opencode/opencode-web.service` now templates the listen port, and phase 07 substitutes `__OPENCODE_PORT__` from `.env` with a 1..65535 validation fallback.
- The dry-run and success messages report the effective port, so operators can see whether the override actually took.
- `ods/tests/test-linux-opencode-path.sh` covers the override, missing-key fallback, invalid-value fallback, and unsubstituted-placeholder cases, and the test is now part of both `make test` and Linux CI.

What I checked:
- GitHub checks are green for this head, including shell lint, `integration-smoke`, distro smoke, dashboard checks, env validation, PowerShell lint, and secret scan.
- Static diff review of `ods/installers/phases/07-devtools.sh`, the systemd unit template, and the extracted rendering test.

Thanks again for cleaning up this Linux/macOS/Windows parity issue.

PR #2191, opened 2026-07-26
fix(audit): check the feature display fields the dashboard renders

tang-vu:
> Consistency fix, not a user-visible crash — three places disagree about which fields a manifest feature must carry, and the gate that should have caught it checks the smallest set.

## Problem

dashboard-api treats five fields as a feature's expected metadata:

```python
# config.py — on every startup, for every manifest feature
missing = [f for f in ("description", "icon", "category", "setup_time", "priority") if f not in feature]
if missing:
    logger.warning("Feature '%s' in %s missing optional fields: %s", feature["id"], path, ", ".join(missing))
```

and the API renders two of them straight into the UI:

```python
# routers/features.py:108
"setupTime": feature.get("setup_time", "Unknown"),
```
```jsx
// components/FeatureDiscovery.jsx:50
Setup time: {topSuggestion.setupTime}
```

`scripts/audit-extensions.py` — the contract gate, run in CI (`Extension Audit`) and by `ods audit` — checks a **different, smaller** set:

```python
for required in ("id", "name", "description", "category", "priority"):
```

`icon` and `setup_time` are validated by nothing. A manifest can ship, pass the audit, log five warnings on every dashboard-api start, and render `Setup time: Unknown` on the Features page.

## Scope of the drift today

Across all 60 shipped manifests, exactly one omits `setup_time`: `extensions/library/services/localai/manifest.yaml`, on all five of its features. Every other service sets it, and every service uses one value for all of its own features (`invokeai` → `~5 minutes` ×3, `text-generation-webui` → `~3 minutes` ×3).

No manifest omits `icon`.

## Fix

- The auditor now reports missing `icon` / `setup_time` as **warnings**, not errors. Warning is the right severity: `config.py` calls them optional and a third-party extension that omits them must stay installable, while `--strict` (used for release gating) still turns them into a failure.
- The five LocalAI features get `setup_time: ~5 minutes`, matching the per-service convention and its closest peer by image weight (`invokeai`).
- The "valid fixture" in the audit test now carries `icon` and `setup_time`, since it is the reference example of a complete manifest.

Deliberately **not** promoted to errors, and no third-party manifest becomes uninstallable.

## Tests

Extends `tests/test-extension-audit.sh` (gated by the `Extension Audit` step in `.github/workflows/test-linux.yml`) with case 8, which strips `icon` and `setup_time` from the valid fixture and asserts:

1. a non-strict audit still succeeds — the new checks are warnings, not errors;
2. at least two warnings are reported;
3. a `feature-field-missing` warning names `setup_time`, so the message stays actionable.

Verified red against the pre-fix auditor:

```
PASS  missing display fields do not fail a non-strict audit
FAIL  audit did not warn about missing feature display fields
FAIL  no feature-field-missing warning mentions setup_time
Results: 16 passed, 2 failed
```

Green after: **18 passed, 0 failed**, and `python3 scripts/audit-extensions.py --project-dir .` reports `27 services, 0 errors, 0 warnings, result=pass`.

### DIFF
--- a/ods/config/extensions-catalog.json
+++ b/ods/config/extensions-catalog.json
@@ -1,5 +1,5 @@
 {
-  "generated_at": "2026-05-24T03:20:28Z",
+  "generated_at": "2026-07-26T18:06:12Z",
   "schema_version": "1.0.0",
   "extensions": [
     {
@@ -1660,7 +1660,8 @@
           "enabled_services_all": [
             "localai"
           ],
-          "priority": 5
+          "priority": 5,
+          "setup_time": "~5 minutes"
         },
         {
           "id": "localai-image-generation",
@@ -1677,7 +1678,8 @@
           "enabled_services_all": [
             "localai"
           ],
-          "priority": 6
+          "priority": 6,
+          "setup_time": "~5 minutes"
         },
         {
           "id": "audio-generation",
@@ -1694,7 +1696,8 @@
           "enabled_services_all": [
             "localai"
           ],
-          "priority": 7
+          "priority": 7,
+          "setup_time": "~5 minutes"
         },
         {
           "id": "video-generation",
@@ -1711,7 +1714,8 @@
           "enabled_services_all": [
             "localai"
           ],
-          "priority": 8
+          "priority": 8,
+          "setup_time": "~5 minutes"
         },
         {
           "id": "localai-voice-cloning",
@@ -1728,7 +1732,8 @@
           "enabled_services_all": [
             "localai"
           ],
-          "priority": 9
+          "priority": 9,
+          "setup_time": "~5 minutes"
         }
       ]
     },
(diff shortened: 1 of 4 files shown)


Lightheartdevs:
Thanks for tightening the extension audit/display-field contract here.

What looks good:
- `ods/scripts/audit-extensions.py` now checks the two fields the dashboard already renders or warns about, `icon` and `setup_time`, as warnings. That keeps third-party extensions installable while still letting strict release gates catch drift.
- `ods/extensions/library/services/localai/manifest.yaml` and the regenerated `ods/config/extensions-catalog.json` bring the shipped LocalAI feature metadata back into sync.
- `ods/tests/test-extension-audit.sh` covers the intended severity split: non-strict audit succeeds, warnings are emitted, and the `setup_time` warning stays actionable.

What I checked:
- GitHub checks are green for this head, including lint, dashboard API/frontend, Extension Audit through `integration-smoke`, distro smoke, env validation, type check, and secret scan.
- Static diff review of `audit-extensions.py`, the LocalAI manifest/catalog output, and the audit fixture coverage.

Thanks again for closing this consistency gap.

PR #2192, opened 2026-07-27
fix(token-spy): stop the postgres SSE cursor from dropping rows

### LINKED ISSUE #2173: token-spy: the Postgres SSE cursor pages by UUID, so /events silently drops usage rows
Follow-up split out of #2025, which fixes the SSE cursor for the SQLite backend only.

## The SQLite bug (fixed in #2025)

`query_recent_events()` paged with `WHERE id > ? ORDER BY timestamp DESC`. `id` is the `AUTOINCREMENT` key, so ordering by `timestamp DESC` made the caller's cursor walk backwards and re-send each batch every poll. Ordering by `id ASC` fixes it: the cursor advances in insertion order.

## The Postgres bug (not fixed)

`db_postgres.py:query_recent_events()` has the same ordering shape:

```sql
WHERE r.tenant_id = %s
AND r.id > %s
ORDER BY r.timestamp DESC
LIMIT %s
```

but `r.id` is a UUID, not a monotonic key (`register_uuid`, `after_id: Optional[UUID]`). So this is worse than a re-send:

- `id > cursor` compares UUIDs bytewise. For random UUIDs that ordering is unrelated to insertion order.
- `event_stream()` in `main.py` is shared by both backends and advances `last_id` to `max(...)` of the batch. Python's `UUID` supports comparison, so this silently "works" — but the max of a set of random UUIDs is an arbitrary value, not the newest row.
- Net effect: every subsequent poll selects an arbitrary subset. New rows whose UUID sorts **below** the cursor are never streamed — permanently lost from `/events` — while older rows whose UUID sorts above it are re-sent.

The SQLite `ORDER BY id ASC` fix does not transfer, because on Postgres `id` carries no insertion order to begin with.

## Why it is not in #2025

The fix needs a monotonic cursor column, and that is a schema question this repo cannot answer from the inside: `init_db()` only ensures the tenant row exists and assumes `tenants` / `requests` / `agents` are already provisioned externally. There is no migration or `CREATE TABLE` for them anywhere in the repo.

Likely options, in maintainer-decision order:

1. Add a `BIGSERIAL` (or `bigint GENERATED ALWAYS AS IDENTITY`) sequence column to `requests` and page on it — matches the SQLite semantics exactly.
2. Page on `(timestamp, id)` as a composite keyset cursor, ordered ASC. Works without a schema change but needs `event_stream()` to carry a tuple cursor and a tiebreak on equal timestamps.
3. Switch `id` to UUIDv7 (time-ordered), which makes the existing `id >` comparison correct.

Option 1 or 2 also means `event_stream()` stops assuming the cursor is a single comparable scalar shared across backends.

## Reproducing

Needs `DB_BACKEND=postgres` against a provisioned schema — not covered by the current test suite, which exercises the SQLite path only. Worth adding a Postgres cursor test alongside whichever fix lands.

AmirF194:
## Summary

PostgreSQL `requests.id` values are random UUIDv4 values. The previous SSE query used `id > after_id`, so paging order had no relationship to insertion order and `/token_events` silently lost usage rows.

This change gives PostgreSQL a stable composite cursor while preserving the existing SQLite path and the external SSE payload.

Closes #2173.

## Root Cause

The stream and query had three related consistency defects:

1. UUID comparison was used as insertion order.
2. The initial page was emitted newest-first, leaving the stream cursor on the oldest event and replaying the initial backlog on the next poll.
3. Every poll re-read the cursor row. If retention deleted that row, the stream returned no events forever.

## Implementation

- Page PostgreSQL rows by `(timestamp, id)` with matching ascending filter and ordering.
- Return the newest initial window in chronological order, so the cursor finishes at the newest emitted event.
- Carry the composite cursor privately as `_cursor`; it is not serialized into the public SSE event.
- Fall back to the row ID for SQLite and older backend responses.
- Preserve legacy PostgreSQL UUID cursors with tenant-scoped lookup.
- Add a dedicated GitHub Actions job backed by PostgreSQL 16.

No schema, Compose, installer, API authentication, or public event format is changed.

## Invariants

| Invariant | Evidence |
|---|---|
| Every row after the cursor is delivered once in keyset order | Live PostgreSQL forward-poll regression |
| Initial connection returns the newest bounded window without replay | Initial-window regression |
| Deleting the cursor source row does not freeze an active stream | Retention regression |
| Equal timestamps are deterministic | UUID tie-break regression |
| A legacy cursor cannot resolve through another tenant | Tenant-isolation regression |
| SQLite behavior remains selected by the existing backend switch | `main.py` ID fallback; token-spy suite |
| Private cursor metadata is not exposed over SSE | Existing explicit `event_data` serialization |

## Validation

Exact head: `cedb76effe8e9a8daebd5ff423c8d6710954b51f`

```text
PostgreSQL 16 container:
python -m pytest ods/extensions/services/token-spy/tests/test_recent_events_cursor_postgres.py -q
5 passed

Token Spy test suite:
python -m pytest ods/extensions/services/token-spy/tests -q
20 passed, 5 skipped

Python:
python -m py_compile <changed Python files>
ruff check <changed Python files> --select E,F,W --ignore E501,E701,E731,E741,E402
passed

Workflow YAML parse:
passed

git diff --check:
passed
```

The five skipped tests in the general local suite are existing environment-dependent tests. The PostgreSQL cursor suite itself ran without skips against the disposable PostgreSQL 16 container.

## Risk And Compatibility

- **Clean install / upgrade / rerun:** no schema or generated configuration change.
- **Service disabled:** no effect while token-spy is not running.
- **Partial failure / restart:** the cursor is connection-local; a reconnect intentionally receives the newest bounded snapshot, matching existing behavior.
- **Rollback:** reverting these commits restores the previous query without data migration.
- **Linux / Windows / macOS:** application-level Python/PostgreSQL behavior is platform-neutral; the new CI lane runs on Linux and does not alter platform installers.

## Residual Limitation

The composite key assumes database timestamps do not move backward. Removing that assumption requires a persistent monotonic sequence and a schema migration, which is intentionally outside this focused fix.

### DIFF
--- a/.github/workflows/test-linux.yml
+++ b/.github/workflows/test-linux.yml
@@ -8,6 +8,46 @@ on:
       - master
 
 jobs:
+  token-spy-postgres:
+    runs-on: ubuntu-latest
+    services:
+      postgres:
+        image: postgres:16-alpine
+        env:
+          POSTGRES_DB: tokenspy
+          POSTGRES_USER: tokenspy
+          POSTGRES_PASSWORD: testpass
+        ports:
+          - 5432:5432
+        options: >-
+          --health-cmd "pg_isready -U tokenspy -d tokenspy"
+          --health-interval 5s
+          --health-timeout 5s
+          --health-retries 10
+    defaults:
+      run:
+        working-directory: ods/extensions/services/token-spy
+    env:
+      DB_HOST: 127.0.0.1
+      DB_PORT: "5432"
+      DB_NAME: tokenspy
+      DB_USER: tokenspy
+      DB_PASSWORD: testpass
+    steps:
+      - name: Checkout
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
+
+      - name: Setup Python
+        uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v6
+        with:
+          python-version: "3.12"
+
+      - name: Install test dependencies
+        run: python -m pip install "psycopg2-binary>=2.9.9,<3.0.0" "pytest>=8,<9"
+
+      - name: Test Postgres SSE cursor
+        run: python -m pytest tests/test_recent_events_cursor_postgres.py -v
+
   integration-smoke:
     runs-on: ubuntu-latest
     defaults:
--- a/ods/extensions/services/token-spy/db_postgres.py
+++ b/ods/extensions/services/token-spy/db_postgres.py
@@ -6,9 +6,9 @@ Set DB_BACKEND=postgres to use this module.
 
 import os
 import logging
-from datetime import date, timedelta
+from datetime import date, datetime, timedelta
 from decimal import Decimal
-from typing import Optional
+from typing import Optional, Tuple
 from uuid import UUID, uuid4
 
 from psycopg2.extras import RealDictCursor, register_uuid
@@ -34,6 +34,8 @@ SINGLE_TENANT_SLUG = os.environ.get("SINGLE_TENANT_SLUG", "default")
 _pool: Optional[pool.ThreadedConnectionPool] = None
 _tenant_id: Optional[UUID] = None
 _agent_cache: dict[str, UUID] = {}
+
+EventCursor = Tuple[datetime, UUID]
 
 
 def _get_pool() -> pool.ThreadedConnectionPool:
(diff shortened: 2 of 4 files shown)


Lightheartdevs:
Thanks for taking on the Postgres side of the token-spy SSE cursor; the UUID-ordering failure mode is easy to miss.

Decision point:
- The `db_postgres.py` change looks directionally right: resolving `after_id` to its `(timestamp, id)` tuple and paging forward with `ORDER BY timestamp ASC, id ASC` makes the filter agree with the cursor order instead of comparing random UUIDs.
- The open question is validation policy. This head currently has no GitHub checks reported, and the new regression file skips unless a live Postgres is configured. I think the maintainer call is whether the manual live-Postgres evidence in the PR body is enough for this optional backend fix, or whether we want a CI-backed Postgres service lane before merge.

What I checked:
- Reviewed the PR body, empty discussion, two-file diff, and surrounding `/token_events` cursor loop in `ods/extensions/services/token-spy/main.py`.
- Compared the changed Postgres query against the current token-spy backend shape, default SQLite deployment, and `DB_BACKEND=postgres` override path.
- Ran `git diff --check origin/main...HEAD` and `python -m py_compile ods/extensions/services/token-spy/db_postgres.py ods/extensions/services/token-spy/tests/test_recent_events_cursor_postgres.py` in a temporary PR worktree; both passed.
- `gh pr checks` reports no checks for this branch. I could not run the new pytest locally because this runner does not have `pytest`/`psycopg2-binary` installed or a local Postgres service configured.

Thanks again for pushing the backend-specific fix forward.

gabsprogrammer:
Updated at exact head `cedb76effe8e9a8daebd5ff423c8d6710954b51f`. The PR now includes a PostgreSQL 16 CI lane and fixes the additional replay and retention failure modes found during adversarial review. Local PostgreSQL receipt: 5/5 cursor regressions passed; full token-spy suite: 20 passed, 5 existing environment-dependent skips. The public SSE payload and SQLite fallback remain unchanged.

gabsprogrammer:
Final validation receipt for `cedb76effe8e9a8daebd5ff423c8d6710954b51f`: the new `token-spy-postgres` job passed against PostgreSQL 16, `integration-smoke` passed, and the complete GitHub rollup is green across Python, shell, PowerShell, dashboard, env validation, macOS smoke, and the distro matrix. The branch is mergeable; only the repository's required human approvals remain.

AmirF194:
Thanks for the review and for hardening this with the Postgres CI lane and the extra replay/retention fixes, appreciate it getting over the line.

PR #2195, opened 2026-07-27
fix(privacy-shield): scrub card numbers that are not 16 digits

tang-vu:
## Problem

The Privacy Shield credit-card detector only recognises one PAN shape:

```python
'credit_card': re.compile(r'\b(?:\d{4}[-\s]?){3}\d{4}\b'),
```

and the Luhn gate hard-rejects anything that is not exactly 16 digits:

```python
if len(digits) != 16:
    return False
```

Real card numbers are 13-19 digits — Amex is 15, Diners is 14, Visa issues both 13 and 19. Those PANs are handed to the upstream provider in cleartext, which is exactly what the shield exists to prevent. `PII_COVERAGE.md` already advertised `{13,16}` coverage.

Verified against the deployed module before the change:

```
'Amex 3782 822463 10005 exp'  -> 'Amex 3782 822463 10005 exp'      # untouched
'Amex 378282246310005'        -> 'Amex 378282246310005'            # untouched
'Diners 3056 930902 5904'     -> 'Diners 3056 <PII_phone_...>'     # partial leak
```

The Diners case is worse than a plain miss: the PAN contains a phone-shaped 10-digit run, and because `phone` ran before `credit_card` the tail was tokenized as a phone number while the issuer digits stayed in cleartext — the payload still *looks* scrubbed.

## Fix

- Match the two groupings that occur in practice — 4-4-4-N for the even-width brands and 4-6-N for Amex/Diners — across the whole 13-19 digit range.
- Widen the Luhn gate to 13-19 digits. Luhn still carries the false-positive control, so short digit runs (phones, SSNs) stay below the floor and unchecksummed runs are still skipped.
- Move `credit_card` ahead of the shorter numeric patterns so it claims an overlapping match instead of losing the tail to `phone`.
- Correct the `PII_COVERAGE.md` row to describe what the detector actually does.

## Tests

`extensions/services/privacy-shield/tests/test_pii_scrubber.py` — new cases for Amex/Diners/13-digit/19-digit PANs (grouped, dashed and plain), the Luhn length window, the phone/SSN non-regression floor, and the card-wins-over-phone overlap.

```
# before: 9 failed, 30 passed
# after:  42 passed
python -m pytest tests/test_pii_scrubber.py tests/test_key_management.py -q
```

### DIFF
--- a/ods/extensions/services/privacy-shield/PII_COVERAGE.md
+++ b/ods/extensions/services/privacy-shield/PII_COVERAGE.md
@@ -15,7 +15,7 @@ Privacy Shield provides PII (Personally Identifiable Information) detection and 
 | Email addresses | `\b[A-Za-z0-9._%+-]+@[A-Za-z0-9.-]+\.[A-Z|a-z]{2,}\b` | user@example.com |
 | Phone numbers (US) | `\b\d{3}[-.]?\d{3}[-.]?\d{4}\b` | 555-123-4567 |
 | Social Security Numbers | `\b\d{3}-\d{2}-\d{4}\b` | 123-45-6789 |
-| Credit card numbers | `\b(?:\d[ -]*?){13,16}\b` | 4111 1111 1111 1111 |
+| Credit card numbers | 13-19 digit PANs in 4-4-4-N or 4-6-N grouping, gated by a Luhn checksum | 4111 1111 1111 1111, 3782 822463 10005 |
 | IP addresses | `\b\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}\b` | 192.168.1.1 |
 
 ### Limitations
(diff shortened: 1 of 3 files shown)


Lightheartdevs:
Thanks for tightening the Privacy Shield card scrubber; this closes a concrete leak for valid non-16-digit PANs and the phone-overlap case.

What looks good:
- `pii_scrubber.py` now matches the 13-19 digit PAN range with practical 4-4-4-N and 4-6-N groupings, while keeping Luhn as the false-positive gate.
- Moving credit cards before shorter numeric patterns fixes the Diners-style overlap where the phone detector could mask only the tail.
- Tests cover Amex, Diners, 13/19-digit Visa, bad checksum/length bounds, phone/SSN floors, and the overlap behavior.

What I checked:
- Reviewed `PII_COVERAGE.md`, `pii_scrubber.py`, and the new scrubber test cases; also checked the PR is mergeable against current `main`.
- Verified all GitHub checks are green, including Dashboard/API, Python Ruff/mypy, Test Linux `integration-smoke`, Matrix Smoke, ShellCheck, and secret scan.
- Ran local focused checks: `python -m py_compile ...` and a direct `PIIDetector` smoke for Amex, Diners-overlap, and phone-floor cases.

Thanks again for pushing the privacy behavior in the right direction.

PR #2197, opened 2026-07-27
fix(env): declare every key the compose stack interpolates

tang-vu:
## Problem

`validate-env.sh` treats the schema as the complete vocabulary of `.env`:

```bash
for key in "${!ENV_MAP[@]}"; do
    if [[ -z "${SCHEMA_KEY_SET[$key]-}" ]]; then
        unknown+=("$key")
    fi
done
...
log_error "Unknown keys not defined in schema:"
```
(exit 2)

But 18 keys the shipped compose files interpolate have no schema entry:

```
LLAMA_ARC_IMAGE                                ODS_REMOTE_PROVIDER_SSH_RESTART_COOLDOWN
LLAMA_TAG                                      ODS_REMOTE_PROVIDER_SSH_START_GRACE
ODS_CONFIG_DIR                                 ODS_REMOTE_PROVIDER_SSH_STOP_TIMEOUT
ODS_DATA_DIR                                   ODS_REMOTE_PROVIDER_SSH_TUNNEL_HEALTH_TIMEOUT
ODS_FLEET_PROBE_KEY                            ODS_REMOTE_PROVIDER_SSH_TUNNEL_HEALTH_URL
ODS_REMOTE_PROVIDER_MAX_BODY_BYTES             ODS_REMOTE_PROVIDER_UPSTREAM_TIMEOUT
ODS_REMOTE_PROVIDER_SSH_RECONCILE_INTERVAL     ODS_ROUTER_INTERNAL_KEY
ONEAPI_VERSION                                 ODS_TALK_HERMES_TIMEOUT
RAG_EMBEDDING_ENGINE                           RAG_EMBEDDING_MODEL_AUTO_UPDATE
```

Every one is written as `${KEY:-default}`, which is the compose way of saying "set this in .env to override". An operator who raises `ODS_REMOTE_PROVIDER_UPSTREAM_TIMEOUT` for a slow provider, or points `ODS_DATA_DIR` at another disk, gets their next `ods-cli` run rejected with "Unknown keys not defined in schema" — for a variable the shipped compose file reads.

The remote-provider block is the largest gap: the egress service and the SSH tunnel supervisor expose nine tunables between them and none of them are declarable.

## Fix

- Declare all 18 keys with the defaults their compose interpolation already uses, placed next to their existing peers (remote-provider knobs after `REMOTE_ODS_PEER_URL`, image/build args after `LEMONADE_SERVER_IMAGE`, and so on).
- Mark `ODS_ROUTER_INTERNAL_KEY` and `ODS_FLEET_PROBE_KEY` `"secret": true` so `ods config show` masks them like every other credential.
- Type the numeric knobs as `integer` with a floor, so a typo is caught as a range error rather than passed to the container.

No behaviour changes in compose or the installer — this only teaches the validator about keys the stack already reads.

## Tests

New `ods/tests/test-compose-env-schema-parity.sh`, wired into `make test`, walks every root and installer compose file and asserts each `${KEY}` interpolation has a schema entry.

```
# on main
[FAIL] compose reads keys .env.schema.json does not declare ... (18 listed)
Passed: 1  Failed: 1   -> exit 1

# with this change
[PASS] all 115 compose-interpolated keys are declared in .env.schema.json
[PASS] .env.schema.json parses as JSON
Passed: 2  Failed: 0
```


Lightheartdevs:
Thanks for closing the compose/.env schema gap; this is exactly the kind of guard that prevents user-tunable runtime knobs from breaking later `ods-cli` validation.

What looks good:
- The schema additions match keys already interpolated by root/installer compose files, and sensitive keys `ODS_ROUTER_INTERNAL_KEY` and `ODS_FLEET_PROBE_KEY` are marked secret.
- The numeric remote-provider and Talk timeouts are typed as integers with minimums, so bad overrides fail validation early.
- The new `ods/tests/test-compose-env-schema-parity.sh` is wired into `make test` and catches future compose/schema drift.

What I checked:
- Reviewed `.env.schema.json`, `ods/Makefile`, and `ods/tests/test-compose-env-schema-parity.sh` against the current compose validation and `validate-env.sh` behavior.
- Verified all GitHub checks are green, including Validate .env Schema tiers 0-4, Test Linux `integration-smoke`, Matrix Smoke, ShellCheck, Dashboard, and secret scan.
- Ran local checks: `C:\Program Files\Git\bin\bash.exe tests/test-compose-env-schema-parity.sh` from `ods/` (`Passed: 2 Failed: 0`) and a JSON schema spot-check for the secret/integer entries.

Thanks again for adding the regression guard with the fix.

PR #2198, opened 2026-07-27
fix(voice): stop uninstalled LiveKit reporting voice as unavailable

tang-vu:
## Problem

`GET /api/voice/status` builds a status map and then folds every entry of it into one verdict:

```python
services_status["livekit"] = {"status": "not_configured"}   # the default branch
...
all_healthy = all(s.get("status") == "healthy" for s in services_status.values())
```

LiveKit is optional and is not in `SERVICES` on a stock install — the endpoint writes `not_configured` for it itself, then fails its own check against that value. So `available` is `False` on **every** install that did not add LiveKit, however healthy Whisper and Kokoro are.

That value is user-visible. `SuccessValidation.jsx` re-runs this endpoint for the first-boot "Voice I/O" check:

```js
testUrl: '/api/voice/status'
...
const success = Boolean(result.success ?? result.available)
```

The card seeds the check as *passed* from the service list, then flips it to *failed* as soon as the user presses "Run tests" — and `onAllPassed` never fires.

## Fix

Base the verdict on the voice stack ODS actually installs:

- Whisper (stt) and Kokoro (tts) must both be healthy — they are what the product ships.
- Any optional service the operator *did* install still has to be healthy, so a broken LiveKit is still reported.
- A service reporting `not_configured` sits the verdict out instead of vetoing it.

## Tests

New `extensions/services/dashboard-api/tests/test_voice.py`:

- healthy stt+tts with no LiveKit → `available: true` (fails on main)
- required service down → `available: false`
- required service not configured → `available: false`
- LiveKit installed but down → `available: false`
- LiveKit installed and healthy → `available: true`

```
# before: 1 failed, 4 passed
# after:  5 passed
python -m pytest tests/test_voice.py -q
python -m pytest tests/test_routers.py tests/test_main.py -q   # 126 passed
```

### DIFF
--- a/ods/extensions/services/dashboard-api/routers/voice.py
+++ b/ods/extensions/services/dashboard-api/routers/voice.py
@@ -10,6 +10,13 @@ logger = logging.getLogger(__name__)
 
 router = APIRouter(tags=["voice"])
 
+# Whisper (STT) and Kokoro (TTS) are the voice stack ODS ships, so both have to
+# be healthy for voice to be usable. LiveKit is an optional add-on that is
+# absent from SERVICES unless an operator installs it — an install without it
+# is complete, not degraded, and must not drag the verdict down.
+REQUIRED_VOICE_SERVICES = ("stt", "tts")
+NOT_CONFIGURED = "not_configured"
+
 
 @router.get("/api/voice/status")
 async def voice_status(api_key: str = Depends(verify_api_key)):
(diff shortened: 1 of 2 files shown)


Lightheartdevs:
Thanks for tightening the Voice I/O status path here; this fixes a real first-boot false negative without changing the service contract.

What looks good:
- `routers/voice.py` now separates required shipped voice services (`stt`, `tts`) from optional LiveKit, so a stock install can report `available: true` while an installed-but-unhealthy LiveKit still fails the verdict.
- The new `tests/test_voice.py` covers the key matrix: stock healthy voice, missing/down required service, LiveKit down, and LiveKit healthy.
- The change stays inside the dashboard-api status endpoint and does not broaden install/compose behavior.

What I checked:
- Reviewed the PR body, diff, and empty PR discussion.
- Verified all GitHub checks are green, including Dashboard `api`, Test Linux `integration-smoke`, Python Ruff/mypy, Matrix Smoke, and secret scan.
- Ran local focused checks: `python -m py_compile ...`, AST parse, and a direct stubbed `voice_status` smoke for stock, missing required, and LiveKit-down cases.

Thanks again for closing the loop on this setup-path false negative.

PR #2199, opened 2026-07-27
ci: run the whole remote-provider contract suite

tang-vu:
## Problem

The remote-provider subsystem ships four contract tests. CI runs two of them, and `make test` runs none:

| test | `make test` | `test-linux.yml` |
|---|---|---|
| `test-remote-provider-egress-policy.py` | no | yes |
| `test-remote-provider-egress-service.py` | no | yes |
| `test-remote-provider-cli.py` | **no** | **no** |
| `test-remote-provider-ssh-tunnel-service.py` | **no** | **no** |

The two files nothing runs carry the checks with the sharpest teeth — 600 lines of them:

- `ods-cli` still implements the `remote-provider` lifecycle command surface
- the SSH supervisor starts a single `ssh` child **without a shell**
- it stops that child when secret custody disappears
- it restarts when the route argv changes
- it honours the restart cooldown after a child exit
- the service source never names a public secret

A regression in any of those lands green today.

## Fix

Run all four from both entry points, in the same block as their existing siblings. All four are stdlib-only source and fixture inspections — no docker, network, or GPU — so they cost about a second.

## Verification

```
tests/contracts/test-remote-provider-egress-policy.py         PASS
tests/contracts/test-remote-provider-egress-service.py        PASS
tests/contracts/test-remote-provider-cli.py                   PASS
tests/contracts/test-remote-provider-ssh-tunnel-service.py    PASS
```

`make -n test` expands cleanly and the workflow still parses as YAML.

### DIFF
--- a/.github/workflows/test-linux.yml
+++ b/.github/workflows/test-linux.yml
@@ -167,6 +167,8 @@ jobs:
           python3 tests/contracts/test-network-exposure-contracts.py
           python3 tests/contracts/test-remote-provider-egress-policy.py
           python3 tests/contracts/test-remote-provider-egress-service.py
+          python3 tests/contracts/test-remote-provider-cli.py
+          python3 tests/contracts/test-remote-provider-ssh-tunnel-service.py
 
       - name: Linux install preflight tests
         run: |
--- a/ods/Makefile
+++ b/ods/Makefile
@@ -30,6 +30,12 @@ test: ## Run unit and contract tests
 	@bash tests/test-hermes-slash-worker-prune.sh
 	@python3 tests/contracts/test-hermes-worker-guardrails.py
 	@echo ""
+	@echo "=== Remote provider contracts ==="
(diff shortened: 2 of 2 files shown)


Lightheartdevs:
Thanks for tightening the remote-provider contract coverage; the SSH supervisor and CLI command surface are the right things to keep under CI.

What looks good:
- `.github/workflows/test-linux.yml` now runs the previously unwired `test-remote-provider-cli.py` and `test-remote-provider-ssh-tunnel-service.py` beside the existing remote-provider policy/service contracts.
- `ods/Makefile` adds the full four-file remote-provider contract block to the local `make test` path, so contributor and CI coverage now match.
- The added tests are stdlib/source/fixture inspections, so the coverage gain is meaningful without introducing Docker, network, or GPU cost.

What I checked:
- Static review of the workflow and `Makefile` additions against the current remote-provider tests under `ods/tests/contracts/`.
- The PR body reports all four remote-provider contract files passing and `make -n test` expanding cleanly.

PR #2201, opened 2026-07-27
test(installer): assert the docker temp-file invariant where it now lives

tang-vu:
## Problem

`tests/test-tmpfile-cleanup.sh` **fails on main**:

```
2. Explicit Cleanup Tests
  Phase 05 has explicit cleanup after success...     ✓ PASS
  Phase 05 has explicit cleanup in error path...     ✗ FAIL
✗ Some tests failed (8 passed, 1 failed)
```

The assertion is anchored on the *caller*:

```bash
grep -B 2 'error "Docker installation failed' 05-docker.sh | grep -q "rm -f.*tmpfile"
```

That held only while the `mktemp`, the `curl`, and the failure branch sat inline next to the `error` call. The download has since moved into `_docker_install_from_script()`, which removes the temp file on **both** paths and returns a code the callers turn into an error. The grep window now contains only the caller's error line, so the check fails against correct code.

Nothing runs the suite — it is in neither `make test` nor any workflow — so the stale assertion has been sitting red unnoticed, and the invariant it was meant to protect (a shell script downloaded from `get.docker.com` never outlives the function that fetched it) has been unguarded the whole time.

## Fix

Assert the invariant where it now lives. Extract the helper body, then require:

- it still `mktemp`s the install script,
- every `return` inside it is preceded by `rm -f "$tmpfile"`,
- the body's last statement is that removal (the success path).

Then run the suite from `make test` and the Linux workflow so it cannot rot again.

## Verification

Mutation-tested against the real phase file:

| mutation | result |
|---|---|
| delete the error-path `rm -f` | ✗ FAIL — `return without cleanup: 4:         return 1` |
| delete the success-path `rm -f` | ✗ FAIL |
| unmodified | ✓ All tests passed (9/9) |

No change to `installers/phases/05-docker.sh` — the code was already right; the test was measuring the wrong thing.

### DIFF
--- a/.github/workflows/test-linux.yml
+++ b/.github/workflows/test-linux.yml
@@ -162,6 +162,7 @@ jobs:
         run: |
           bash tests/contracts/test-installer-contracts.sh
           bash tests/test-phase03-multigpu-tty.sh
+          bash tests/test-tmpfile-cleanup.sh
           bash tests/contracts/test-preflight-fixtures.sh
           bash tests/test-preflight-resolver-bug.sh
           python3 tests/contracts/test-network-exposure-contracts.py
--- a/ods/Makefile
+++ b/ods/Makefile
@@ -85,6 +85,9 @@ test: ## Run unit and contract tests
 	@echo "=== ods config show secret-mask matrix ==="
 	@bash tests/test-ods-config-secret-mask.sh
 	@echo ""
+	@echo "=== installer temp-file cleanup ==="
+	@bash tests/test-tmpfile-cleanup.sh
+	@echo ""
 	@echo "=== Manifest schema source of truth ==="
 	@bash tests/test-validate-manifest-schema.sh
 	@echo ""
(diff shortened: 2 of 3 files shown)


Lightheartdevs:
Thanks for moving the Docker temp-file assertion to the helper that actually owns the file lifecycle.

What looks good:
- `ods/tests/test-tmpfile-cleanup.sh` now checks `_docker_install_from_script()` directly for the `mktemp`, error-return cleanup, and success-path cleanup invariant.
- The suite is wired into both `ods/Makefile` and `.github/workflows/test-linux.yml`, which matches the repo’s high-risk installer validation expectations.
- No installer behavior changes are bundled with the test fix, keeping the PR tightly scoped.

What I checked:
- Static review of the diff plus surrounding `ods/installers/phases/05-docker.sh` helper/caller structure on `origin/main`.
- The PR body reports mutation checks for deleting both error-path and success-path cleanup.

PR #2203, opened 2026-07-27
fix(preset): validate and publish imported archives transactionally

tang-vu:
## Summary

- keep the original case-arm assertions and wire the preset import/export suite into `make test`
- validate every tar member from one open archive handle before extraction
- accept exactly one preset root containing regular files/directories; reject traversal, links, special files, and multi-root archives
- extract into a private directory inside the preset store, validate required files, then publish by same-filesystem rename
- restore the previous preset if publication fails
- preserve valid relative archive paths without changing the CLI working directory

## Why this matters

`ods preset import` is a production CLI path that accepts an operator-supplied archive. It previously derived the preset name from only the first member, removed an existing preset before validating the replacement, and extracted the whole archive directly into `PRESETS_DIR`.

A reproducible archive whose first root was `good/` and whose second root was `victim/` returned success while overwriting an existing `presets/victim/marker.txt`. A same-root archive missing `meta.txt` deleted the operator's previous preset before failing. A valid relative archive path also failed because the command checked it before `cd $PRESETS_DIR` and extracted it after that directory change.

Root cause: archive inspection, extraction, validation, and publication were interleaved against the live preset store.

## Behavioral invariant

A preset import may mutate exactly one target preset only after every archive member and both required files have been validated. Any rejection leaves all presets byte-identical. If final publication fails, the previous target is restored before the command returns non-zero.

## Implementation

The CLI uses the installer-guaranteed Python runtime and `tarfile` to inspect and extract from the same open handle. Member paths must be relative and remain under one top-level root; only regular files and directories are accepted. The staged tree must contain `meta.txt` and `extensions.list` before the overwrite prompt appears.

Staging lives inside `PRESETS_DIR`, so target replacement uses same-filesystem renames. The previous target is first moved into the private staging directory; a failed publish moves it back. If recovery itself fails, cleanup is disarmed and the recovery path is printed instead of deleting the last copy.

## Overlap check

Searched open and closed PRs for preset import, archive traversal, multi-root archives, overwrite validation, and the changed production file `ods/ods-cli`.

- merged #111 introduced preset import/export
- merged #739 fixed relative output handling for preset export
- no open or closed PR validates all preset archive members or stages import publication
- restore-focused #2346 and #2720 operate on `ods-restore.sh`, not the preset command or preset store

This updates the existing canonical tang-vu PR instead of opening another PR under the current backlog gate.

## Regression coverage

`tests/test-preset-import-export.sh` now executes the installed CLI boundary and proves:

- a multi-root archive is rejected without changing an unrelated preset
- an invalid replacement leaves the old preset byte-identical
- a valid relative archive imports successfully
- an archive containing a symbolic link is rejected before publication
- a valid replacement removes stale files and publishes the staged tree
- an injected final-move failure restores the previous preset and cleans staging

The earlier mutation-resistant case-arm assertions remain, and the suite runs from `make test`.

## Validation

- `bash tests/test-preset-import-export.sh` — 18 passed
- `bash -n ods-cli tests/test-preset-import-export.sh` — passed
- `shellcheck --rcfile /dev/null --severity=error ods-cli tests/test-preset-import-export.sh` — passed
- `make lint` — passed
- `bash tests/test-cli-enable-compose-reconciliation.sh` — passed
- `bash tests/test-ods-list-json-escaping.sh` — passed
- `bash tests/test-phase-c-p1.sh` — 7 passed, 0 failed, 9 auth-dependent skips
- `git diff --check` — passed
- GitHub CI at `b810091d` — all required checks passed, including integration-smoke, API/frontend, ShellCheck, PowerShell on Windows and Ubuntu, and the 10-distro matrix
- `bash tests/contracts/test-installer-contracts.sh` passed the affected CLI and preceding platform contracts, then reached the existing Windows-checkout CRLF failure at `Hermes template bounds each model turn`; neither Hermes file is changed here, and #2715 covers repository-wide LF normalization

## Cross-platform behavior and rollback

The archive policy and Python extraction path are shared by Linux, macOS, and WSL. The shell publication path uses Bash 4, portable `mktemp`, `mv`, and `rm` forms without GNU-only flags. No persisted format changes are introduced: exported `.tar.gz` presets remain compatible.

Reverting this PR restores direct extraction into the live preset store and the pre-validation delete behavior; no data migration rollback is required.

### DIFF
--- a/ods/Makefile
+++ b/ods/Makefile
@@ -93,6 +93,7 @@ test: ## Run unit and contract tests
 	@echo ""
 	@echo "=== CLI preset restore user extensions ==="
 	@bash tests/test-cli-preset-restore-user-extensions.sh
+	@bash tests/test-preset-import-export.sh
 	@echo ""
 	@echo "=== bootstrap OpenClaw compose-guard regression ==="
 	@bash tests/test-bootstrap-openclaw-compose-guard.sh
--- a/ods/Makefile
+++ b/ods/Makefile
@@ -109,6 +109,7 @@ test: ## Run unit and contract tests
 	@echo ""
 	@echo "=== CLI preset restore user extensions ==="
 	@bash tests/test-cli-preset-restore-user-extensions.sh
+	@bash tests/test-preset-import-export.sh
 	@echo ""
 	@echo "=== bootstrap OpenClaw compose-guard regression ==="
 	@bash tests/test-bootstrap-openclaw-compose-guard.sh
--- a/ods/ods-cli
+++ b/ods/ods-cli
@@ -3081,6 +3081,118 @@ validate_preset_compatibility() {
     return 0
 }
 
+# Validate and extract one preset archive into a private staging directory.
+# The same open tar handle is used for inspection and extraction so the
+# archive cannot change between those two operations. Only regular files and
+# directories under one top-level preset root are accepted; links and special
+# files are unnecessary for preset data and could escape later reads.
+_extract_preset_archive() {
+    local archive="$1"
+    local destination="$2"
+    local py
+    py=$(command -v python3 || command -v python || true)
+    [[ -n "$py" ]] || {
+        log_error "Python is required to validate preset archives"
+        return 1
+    }
+
+    "$py" - "$archive" "$destination" <<'PYEOF'
+import sys
+import tarfile
+from pathlib import Path, PurePosixPath
+
+archive = Path(sys.argv[1]).resolve(strict=True)
+destination = Path(sys.argv[2])
+
+try:
+    with tarfile.open(archive, mode="r:gz") as bundle:
+        members = bundle.getmembers()
+        if not members:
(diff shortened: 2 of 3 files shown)


Lightheartdevs:
Thanks for making the preset import/export assertions semantic instead of line-window based.

What looks good:
- `ods/tests/test-preset-import-export.sh` now inspects the actual `ods-cli` case arm instead of assuming implementation lines stay within a fixed grep window.
- The export relative-path assertion now matches the behavior it is meant to guard: archiving from `$PRESETS_DIR` rather than accidentally depending on nearby line count.
- Wiring the suite into `ods/Makefile` means this regression test will run with the existing contributor `make test` path.

What I checked:
- Static review of the changed test helper against the current `export|e)` and `import|i)` arms in `ods/ods-cli`.
- The PR body documents the main-branch failure and mutation check for deleting the export `cd $PRESETS_DIR` guard.

PR #2204, opened 2026-07-27
test(gpu): make the Apple-branch suite hermetic on hosts that have nvidia-smi

tang-vu:
## Problem

`tests/test-gpu-apple.sh` opens with a hermeticity promise:

```
# External binaries (sysctl, system_profiler, nvidia-smi, curl, docker) are
# stubbed via a PATH-prepended fixture directory so the test is hermetic and
# runs on any host.
```

and then leaves `nvidia-smi` to chance:

```bash
# Deliberately do NOT stub nvidia-smi. ...
# For negative tests we rely on nvidia-smi being absent from PATH too.
```

On any NVIDIA developer box or GPU runner it is not absent, so the negative case measures the host instead of `ods-cli` and fails against correct code:

```
── 6. gpu status (nvidia, negative) ──
  ✓ PASS nvidia backend does not produce Apple-branch output
  ✗ FAIL nvidia backend fallthrough warning missing; output: ━━━ GPU Status (1 GPU) ━━━
    0    NVIDIA GeForce GTX 106 1.7 / 3.0 GB   14%   50°C   10W

Result: 20 passed, 1 failed
```

The suite is in neither `make test` nor any workflow, so the host-dependence has never been surfaced.

## Fix

A stub cannot express *absent* here — `ods-cli` branches on `command -v nvidia-smi`, which would find one. So drop every PATH entry that provides GPU tooling (`nvidia-smi`, `rocm-smi`, `rocminfo`) instead of relying on the host not to have it. Both the apple branch and the two negative cases then see the no-tooling state they assert, on any machine.

The header is corrected to describe what the fixture actually does, and the suite is wired into `make test` — a suite nothing runs is where this drifted in the first place.

## Verification

On a host with a real `nvidia-smi`:

```
# before
Result: 20 passed, 1 failed

# after
Result: 21 passed, 0 failed
```

No change to `ods-cli` — the CLI already emitted the fallthrough warning; the test just never got to see it.

### DIFF
--- a/ods/Makefile
+++ b/ods/Makefile
@@ -84,6 +84,7 @@ test: ## Run unit and contract tests
 	@echo ""
 	@echo "=== ods config show secret-mask matrix ==="
 	@bash tests/test-ods-config-secret-mask.sh
+	@bash tests/test-gpu-apple.sh
 	@echo ""
 	@echo "=== Manifest schema source of truth ==="
 	@bash tests/test-validate-manifest-schema.sh
--- a/ods/tests/test-gpu-apple.sh
+++ b/ods/tests/test-gpu-apple.sh
@@ -13,9 +13,9 @@
 # Verifies each subcommand under GPU_BACKEND=apple and includes negative
 # cases asserting nvidia/amd backends do NOT take the apple path.
 #
-# External binaries (sysctl, system_profiler, nvidia-smi, curl, docker) are
-# stubbed via a PATH-prepended fixture directory so the test is hermetic and
-# runs on any host.
+# External binaries (sysctl, system_profiler, curl, docker) are stubbed via a
+# PATH-prepended fixture directory, and GPU tooling (nvidia-smi, rocm-smi) is
+# removed from PATH entirely, so the test is hermetic and runs on any host.
 #
 # Usage: ./tests/test-gpu-apple.sh
 # ============================================================================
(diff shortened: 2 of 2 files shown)


Lightheartdevs:
Thanks for taking on the Apple GPU test hermeticity; wiring the suite into `make test` is the right direction.

Quickest path to merge:
- In `ods/tests/test-gpu-apple.sh`, avoid removing whole `PATH` entries that contain `nvidia-smi`/`rocm-smi`. On many NVIDIA Linux hosts those tools live in `/usr/bin` and often `/bin` via usrmerge, which also provide `bash`; `run_ods_cli` then executes `ods-cli` through `#!/usr/bin/env bash` under `PATH=$STUB_PATH` and can fail before reaching the GPU assertions.
- Keep a stable shell/runtime path while still making GPU tools absent. A small fix would be to add a `bash` shim in `$STUB_BIN` that execs the current interpreter, invoke `$BASH $ODS_CLI` directly, or add equivalent self-checks that `PATH=$STUB_PATH command -v bash` succeeds while `command -v nvidia-smi`, `rocm-smi`, and `rocminfo` fail.

What I checked:
- Static review of `ods/tests/test-gpu-apple.sh` and the new `path_without_gpu_tools` / `run_ods_cli` interaction.
- PR checks are currently green, but this edge is specific to the NVIDIA-host scenario the PR is trying to make hermetic, so CI without `nvidia-smi` does not exercise it.

Thanks again for chasing down the host-dependent test failure.

tang-vu:
﻿Addressed the outstanding PATH hermeticity advisory in the latest commit:

- add a `bash` symlink in the fixture stub directory before filtering GPU-tool directories
- assert the filtered PATH still resolves Bash
- assert `nvidia-smi`, `rocm-smi`, and `rocminfo` variants remain absent
- keep `ods-cli` execution through its normal `/usr/bin/env bash` shebang

Local verification: `bash tests/test-gpu-apple.sh` ? 21 passed, 0 failed.

PR #2212, opened 2026-07-27
fix(security): pin validated remote-provider addresses

### LINKED ISSUE #2211: sec: DNS Rebinding SSRF (TOCTOU) in remote-provider egress proxy
### Summary

A Time-of-Check to Time-of-Use (TOCTOU) vulnerability exists in the egress proxy server of the `remote-provider-egress` service. While the egress proxy attempts to validate that direct provider hosts do not resolve to private or loopback IP addresses, the hostname is resolved twice: once during validation, and again when the HTTP request is executed.

### Vulnerable Code Pointers

In `ods/extensions/services/remote-provider-egress/app/main.py`:

```python
@app.api_route("/{full_path:path}", methods=["GET", "POST", ...])
async def forward(full_path: str, request: Request) -> Response:
    path = "/" + full_path
    try:
        route = _load_route()
        validate_direct_provider_resolution(route) # CHECK TIME: Resolves and checks IPs
        secret = read_provider_secret(SECRET_PATH)
        ...
        upstream_request = prepare_upstream_request(
            ...
            route=route,
            ...
        )
    except EgressError as exc:
        return _error_response(exc)
    client = _http_client()
    try:
        # USE TIME: httpx.AsyncClient resolves the hostname in upstream_request.url again!
        upstream = await client.request(
            upstream_request.method,
            upstream_request.url,
            ...
        )
```

### Attack Scenario (DNS Rebinding)

1. An attacker configures a malicious provider URL pointing to their own DNS server: `http://rebind.attacker.com/v1`.
2. When the egress proxy calls `validate_direct_provider_resolution(route)`, the DNS resolves to a safe public IP (e.g. `8.8.8.8`). The check passes successfully.
3. During the actual request execution (`client.request`), the DNS TTL expires and the next resolution returns a private IP (e.g. `127.0.0.1` or the container-internal IP of `remote-provider-ssh-tunnel` / `dashboard-api`).
4. The proxy forwards the request to the internal endpoint, bypassing security boundaries (SSRF).

### Remediation

Pin the connection to the validated IP address resolved during check time instead of resolving it again. This can be done by rewriting the outgoing request host to the validated IP, and setting the original host in the `Host` HTTP header:

```python
# Pin request to the validated IP address
resolved_ip = resolved_addresses[0]

# Rewrite request target to use IP directly and set Host header
```

### 

A user configures a connection to a remote provider through the settings dashboard (or via CLI) to offload model inference or sync state.

If a malicious user configures a provider URL pointing to a domain they control (e.g., `http://rebind.attacker.com/v1`), they can exploit this issue.

Once the route is enabled, standard chat and capability requests sent by the system (via LiteLLM/Hermes) to the remote provider are forwarded by the egress proxy. The proxy resolves the hostname again during request execution, allowing a DNS rebinding attack to redirect the connection to loopback or private services (e.g., the local Docker daemon on port `2375`, the SSH supervisor control API, or internal database ports) that were intended to be protected by the private IP filter.

### Proposed Solution

_No response_

### Alternatives Considered

_No response_

IronicRayquaza:
## Summary

Closes #2211 by removing the DNS time-of-check/time-of-use gap in the direct remote-provider egress path.

A direct provider hostname is now resolved and policy-validated once per request. The outbound socket connects to that validated IP instead of resolving the hostname again, while TLS SNI and the HTTP `Host` authority remain bound to the original provider URL.

## Security Contract

| Invariant | Enforcement |
|---|---|
| A rejected DNS answer is never contacted | Every resolved address must pass the existing global-address policy before request preparation |
| The network connection uses the validated answer | Direct URLs are rewritten to the selected validated IPv4/IPv6 address |
| TLS identity remains the configured provider | Original hostname is passed as `sni_hostname` |
| HTTP virtual-host routing remains correct | Original authority is sent in `Host`, including an explicit provider port |
| Different TLS identities never share a connection | HTTP clients are partitioned by scheme, SNI hostname, and port |
| DNS rotation cannot create an unbounded client map | Address changes for one TLS identity share HTTPX's bounded origin pool |
| Environment proxies cannot bypass pinning or receive provider credentials | Egress clients use `trust_env=False` |
| Redirects cannot introduce a second destination | Existing `follow_redirects=False` behavior is preserved |
| SSH transport behavior is unchanged | DNS pinning is applied only to direct transport |

## Changes

- Return validated direct-provider addresses from the runtime resolution check.
- Rewrite direct upstream URLs to the selected validated address, including bracketed IPv6 formatting.
- Preserve SNI and `Host` independently so non-default ports remain correct.
- Standardize streaming and non-streaming sends through HTTPX request extensions.
- Isolate reusable clients by TLS identity and close all clients during shutdown.
- Remove unrelated search-probe commits from the branch; the diff is now limited to three egress files.

## Regression Coverage

- Public IPv4 and IPv6 answers are accepted and pinned.
- Private, loopback, link-local, multicast, mixed unsafe, and empty DNS answers fail closed.
- SSH routes skip direct DNS resolution and continue through the tunnel boundary.
- Non-default provider ports are retained in `Host` but excluded from SNI.
- Two hostnames on the same IP receive separate connection pools.
- Address rotation for one hostname retains a bounded pool identity.
- Secret custody, hop-by-hop header filtering, streaming behavior, and disabled-route failures remain covered.

## Validation

- `python ods/tests/contracts/test-remote-provider-egress-service.py`
- `python ods/tests/contracts/test-remote-provider-egress-policy.py`
- `python ods/tests/contracts/test-network-exposure-contracts.py`
- `python ods/tests/contracts/test-remote-provider-ssh-tunnel-service.py`
- `python ods/tests/test-render-runtime-configs.py`
- `python ods/tests/test-runtime-config-wiring.py`
- `python -m pytest ods/extensions/services/dashboard-api/tests/test_remote_provider_status.py -q` (`26 passed`)
- Focused Ruff and Python compile checks
- `docker compose --env-file ods/.env.example -f ods/docker-compose.base.yml config --quiet`
- Local TLS integration receipt: connection opened to the pinned IP, certificate validated for the configured hostname, SNI observed as the configured hostname, and `Host` observed with the explicit port.
- `git diff --check osmantic/main...HEAD`

## Residual Limitation

The first validated DNS address is selected for a request. If that public address is unreachable while a later validated answer is reachable, the request fails normally instead of retrying another address. This is a bounded availability limitation, not an SSRF bypass; address fallback is intentionally outside this security fix.

### DIFF
--- a/ods/bin/remote_provider/egress.py
+++ b/ods/bin/remote_provider/egress.py
@@ -12,7 +12,7 @@ from dataclasses import dataclass
 from collections.abc import Callable
 from pathlib import Path
 from typing import Any, Mapping
-from urllib.parse import urlsplit
+from urllib.parse import urlsplit, urlunsplit
 
 from .policy import (
     DEFAULT_POLICY_PATH,
@@ -78,6 +78,9 @@ class UpstreamRequest:
     requested_model: str
     provider_model: str
     stream: bool
+    tls_server_name: str = ""
+    host_header: str = ""
+    connection_key: str = ""
 
 
 def _disabled_route(mode: str = "cloud") -> dict[str, Any]:
@@ -325,6 +328,7 @@ def prepare_upstream_request(
     route: Mapping[str, Any],
     provider_secret: str,
     max_body_bytes: int = DEFAULT_MAX_BODY_BYTES,
+    resolved_addresses: list[str] | None = None,
 ) -> UpstreamRequest:
     """Validate and rewrite one OpenAI-compatible request for the provider."""
     required_method = FORWARD_PATHS.get(path)
@@ -348,6 +352,33 @@ def prepare_upstream_request(
     if not isinstance(provider, Mapping):
         raise EgressError(503, "invalid_route", "provider route is missing")
     upstream_base_url = upstream_base_url_for_route(route)
+    tls_server_name = ""
+    host_header = ""
+    connection_key = ""
+    if route.get("transport") == "direct" and resolved_addresses:
+        parts = urlsplit(upstream_base_url)
+        tls_server_name = parts.hostname or ""
+        ip_address = resolved_addresses[0]
+        if ":" in ip_address and not ip_address.startswith("["):
+            ip_address = f"[{ip_address}]"
+        netloc = ip_address
+        if parts.port is not None:
+            netloc = f"{ip_address}:{parts.port}"
+        original_host = tls_server_name
+        if ":" in original_host and not original_host.startswith("["):
+            original_host = f"[{original_host}]"
+        host_header = original_host
+        if parts.port is not None:
+            host_header = f"{original_host}:{parts.port}"
+        tls_port = parts.port or (443 if parts.scheme == "https" else 80)
+        connection_key = f"{parts.scheme}://{tls_server_name}:{tls_port}"
+        upstream_base_url = urlunsplit((
+            parts.scheme,
+            netloc,
+            parts.path,
+            parts.query,
+            parts.fragment
+        ))
     try:
         payload = json.loads(body.decode("utf-8"))
     except (UnicodeDecodeError, ValueError) as exc:
@@ -366,6 +397,9 @@ def prepare_upstream_request(
         requested_model=requested_model,
         provider_model=provider_model,
         stream=bool(payload.get("stream")),
+        tls_server_name=tls_server_name,
+        host_header=host_header,
+        connection_key=connection_key,
     )
 
 
--- a/ods/extensions/services/remote-provider-egress/app/main.py
+++ b/ods/extensions/services/remote-provider-egress/app/main.py
@@ -97,10 +97,20 @@ def _load_route() -> dict[str, Any]:
     return route_from_state(load_route_state(ROUTE_PATH), policy=policy)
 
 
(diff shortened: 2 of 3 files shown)


gabsprogrammer:
Rebased this fix onto current `main` and removed the unrelated search-probe history. The PR is now a one-commit, three-file security patch.

Additional review fixes included:
- preserve explicit provider ports in the HTTP `Host` authority while keeping SNI hostname-only;
- partition reusable clients by TLS identity so different hostnames on one IP cannot reuse a TLS connection;
- keep DNS address rotation inside a bounded per-identity pool;
- disable environment proxy inheritance so pinned requests and provider credentials cannot be redirected outside the egress contract;
- close direct-provider clients during shutdown.

Validation on head `ac601f44287717c69a26e72e2ee55dc885e295ce` is complete. All 28 required GitHub checks pass, including Linux integration, dashboard API/frontend, distro and macOS smoke, PowerShell, ShellCheck, Ruff, mypy, env validation, and secret scan. Focused direct/SSH egress contracts, compose rendering, runtime config tests, and a local pinned-IP TLS/SNI/Host integration receipt also pass.

No reproducible defect remains in the #2211 scope. The PR is mergeable and is blocked only by the repository's required approving reviews.

PR #2221, opened 2026-07-28
fix(linux): repair rootless bind-mount ownership

### LINKED ISSUE #1702: Enhancement: Support Docker rootless mode — fix data directory ownership for non-root container users
## Summary

When ODS is installed on a host running **Docker rootless** (via `rootlesskit`), containers that run as a non-root user (n8n, hermes, whisper, tts, token-spy, privacy-shield, ape, langfuse) fail to start with `EACCES: permission denied` errors. The data directories created by the installer are owned by the host user (e.g. UID 1000), but in rootless mode, container UIDs are remapped to a different range (starting at 100000), so non-root container users cannot write to them.

## Environment

- **OS:** Ubuntu (Linux)
- **Docker:** rootless mode (`rootlesskit` + `dockerd-rootless.sh`)
- **subuid/subgid:** `maxs:100000:65536`
- **ODS version:** `ods-cli v2.5.3`
- **Install path:** `~/ods` (owned by `maxs`, UID 1000)

## Problem

In Docker rootless mode, UID mapping works as follows:

| Container UID | Host UID |
|---|---|
| 0 (root) | 1000 (host user) |
| 1000 (node, ubuntu, appuser, odser) | 100000 |
| 10000 (hermes) | 109000 |

The installer creates data directories under `~/ods/data/` owned by the host user (UID 1000). Containers running as **root** (UID 0) work fine because root in the container maps to the host user. However, containers running as a **non-root user** fail because their container UID maps to a host UID (100000+) that has no write permission on the directory.

## Symptoms

**n8n:**
```
No encryption key found - Auto-generating and saving to: /home/node/.n8n/config
Error: EACCES: permission denied, open '/home/node/.n8n/config'
    at writeFileSync (node:fs:2437:20)
    at InstanceSettings.save (...)
```

**Hermes:**
```
mkdir: cannot create directory '/opt/data': Permission denied
Fixing ownership of /opt/data to hermes (10000)
Warning: chown failed (rootless container?) — continuing anyway
Dropping root privileges
mkdir: cannot create directory '/opt/data': Permission denied
```

## Affected containers

| Service | Container user | Container UID |
|---|---|---|
| n8n | node | 1000 |
| hermes | hermes | 10000 |
| whisper | ubuntu | 1000 |
| tts | appuser | 1000 |
| token-spy | odser | 1000 |
| privacy-shield | — | 1000 |
| ape | ape | 100 |
| langfuse-web | nextjs | 1001 |

Containers running as root (open-webui, litellm, qdrant, embeddings, perplexica, searxng, llama-server, dashboard, dashboard-api) are **not** affected.

## Reproduction

1. Install Docker rootless on a Linux host (e.g. `dockerd-rootless.sh` + `rootlesskit`).
2. Ensure `subuid`/`subgid` is configured: `username:100000:65536`.
3. Run the ODS installer as a non-root user.
4. Start ODS: `ods start`
5. Observe that n8n and hermes fail to start with `EACCES` / `Permission denied` errors.

## Workaround

Manually fix ownership of each affected data directory using a container with root privileges (which maps to the host user in rootless mode):

```bash
# n8n (container UID 1000)
docker run --rm --user 0:0 -v ~/ods/data/n8n:/data alpine sh -c "chown -R 1000:1000 /data"

# hermes (container UID 10000)
docker run --rm --user 0:0 -v ~/ods/data/hermes:/data alpine sh -c "chown -R 10000:10000 /data"

# whisper, tts, token-spy, privacy-shield (container UID 1000)
for svc in whisper tts token-spy privacy-shield; do
  docker run --rm --user 0:0 -v ~/ods/data/$svc:/data alpine sh -c "chown -R 1000:1000 /data"
done

# ape (container UID 100)
docker run --rm --user 0:0 -v ~/ods/data/ape:/data alpine sh -c "chown -R 100:100 /data"

# langfuse (container UID 1001)
docker run --rm --user 0:0 -v ~/ods/data/langfuse:/data alpine sh -c "chown -R 1001:1001 /data"
```

Then restart: `ods restart`

## Suggested fix

The installer (or `ods start` / `ods restart`) should detect rootless mode and fix data directory ownership automatically. Possible approaches:

### Option A — Detect rootless and chown after directory creation

After creating data directories in `06-directories.sh`, check if Docker is running in rootless mode and, for each service that uses a non-root container user, chown the corresponding data directory to the container's UID:

```bash
# Detect rootless mode
_is_rootless_docker() {
    docker info --format '{{.SecurityOptions}}' 2>/dev/null | grep -q rootless
}

if _is_rootless_docker; then
    # Fix ownership for non-root containers
    _chown_data_dir "n8n" 1000
    _chown_data_dir "hermes" 10000
    _chown_data_dir "whisper" 1000
    _chown_data_dir "tts" 1000
    _chown_data_dir "token-spy" 1000
    _chown_data_dir "privacy-shield" 1000
    _chown_data_dir "ape" 100
    _chown_data_dir "langfuse" 1001
fi
```

Where `_chown_data_dir` uses a helper container (since the host user may not have sudo):

```bash
_chown_data_dir() {
    local svc="$1" uid="$2"
    local dir="$INSTALL_DIR/data/$svc"
    [[ -d "$dir" ]] || return 0
    docker run --rm --user 0:0 -v "$dir:/data" alpine \
        sh -c "chown -R ${uid}:${uid} /data" 2>/dev/null || true
}
```

### Option B — Use `docker run --user 0:0` for the chown in the entrypoint

Services like hermes already attempt a `chown` in their entrypoint (`Fixing ownership of /opt/data to hermes (10000)`), but it fails because the container has already dropped root privileges or because rootless remapping prevents it. Running the chown **before** dropping privileges, or using `user: "0:0"` in the compose file for the initial setup step, would fix this.

### Option C — Document the rootless requirement

At minimum, detect rootless mode during `ods doctor` / `ods preflight` and warn the user that non-root containers need manual permission fixes, with the workaround commands printed.

## Additional context

- Docker rootless is increasingly common for security best practices (running the Docker daemon as a non-root user).
- The `hermes` entrypoint already tries to chown `/opt/data` but fails with `Warning: chown failed (rootless container?)` — the detection is there but the fix is incomplete.
- This affects both fresh installs and reinstalls.

gabsprogrammer:
Fixes #1702

## Summary

Docker rootless remaps container UIDs into the caller's subordinate-ID namespace. Installer-created bind mounts can therefore be readable by the host while remaining unwritable by non-root container users.

This change adds one fail-closed ownership contract for clean install, rerun, CLI start/restart, dashboard-driven late enable, and Langfuse post-install. It repairs only active, known ODS bind mounts and leaves rootful Docker, macOS, Windows, and unsupported services unchanged.

Supersedes the conflicted approaches in #1708 and #1709 without widening the Host Agent listener.

## Root Cause

Host-side `chown 1000:1000` is not equivalent to container UID `1000` under rootless Docker. Ownership must be applied from inside the rootless user namespace. Broad recursive repair is also unsafe for stacks such as Langfuse, where Postgres and ClickHouse use different UIDs.

## Behavior And Invariants

|---|---|
| Rootful and non-Linux installs do not mutate | Detection returns before image pulls or filesystem mutation |
| Docker detection errors fail closed | An unavailable daemon is not treated as rootful Docker |
| Only enabled built-ins are repaired | Active Compose flags and a fixed service allowlist select targets |
| Missing active directories are safe | They are created inside the rootless namespace before ownership repair |
| Disabled services remain untouched | Their directories are neither created nor repaired |
| Late extension enable is covered | Host Agent invokes a targeted repair before Compose start |
| Start and restart are covered | `ods-cli` verifies ownership before either lifecycle operation |
| Reruns repair stale descendants | Stopped services repeat recursive correction so nested ownership cannot hide behind correct root metadata |
| Running workloads are protected | Correct metadata is a no-op; mismatches fail closed with recovery instructions |
| Paths cannot escape `data/` | Existing and dangling symlinks, traversal, and resolved out-of-tree targets are rejected |
| Results are proven | UID, GID, and required mode are read back after repair |
| User overrides match Compose | Numeric, quoted numeric, empty, and quoted-empty `UID` / `GID` values follow Compose defaults |
| Partial failures remain visible | Installer, CLI, hooks, and Host Agent stop before Compose start |

## Service Contracts

- APE: `100:100`
- Token Spy, Whisper, NVIDIA ComfyUI: `1000:1000`
- n8n and Privacy Shield: `${UID:-1000}:${GID:-1000}`
- Hermes: `${UID:-10000}:${GID:-10000}`, mode `0700`
- Langfuse Postgres: `70:70`
- Langfuse ClickHouse: `101:101`

AMD ComfyUI is intentionally excluded because its persistent paths use a different root-owned runtime contract. OpenClaw, Hermes Proxy, and root-owned services retain their existing in-container setup.

## Security And Compatibility

- Uses the pinned `busybox:1.36.1` helper with `--user 0:0`, `--network none`, and `--pull never` during mutation.
- Pulls the helper only when native Linux rootless Docker is active and the image is absent.
- Restricts directory creation and ownership repair to resolved descendants of `data/`.
- Does not set `ODS_AGENT_BIND=0.0.0.0`, infer a LAN address, alter low-port forwarding, or change Tailscale/proxy behavior.
- Preserves `DOCKER_HOST` and `XDG_RUNTIME_DIR` only for extension hooks that must address the caller's rootless daemon.

## Recovery For Existing Installs

```bash
cd ~/ods
./ods-cli stop
./ods-cli repair rootless-ownership
./ods-cli start
```

No reinstall or data deletion is required.

## Validation

Validated on head `71a1ded3a434744e2fbe9ca0f0b2925b21143557` against `main` `d6b7efdc2f9466ea018a8e41dab55a24115262e6`:

```text
bash ods/tests/test-rootless-docker-ownership.sh
  25 passed

python -m pytest ods/extensions/services/dashboard-api/tests/test_host_agent.py -q
  256 passed

bash ods/tests/test-ods-doctor-rootless-subid.sh
  10 passed

bash ods/tests/test-update-rollback-contract.sh
bash ods/tests/test-rollback-compose-stack.sh
  passed

shellcheck --norc -e SC1090,SC1091 ods/lib/rootless-ownership.sh
python -m ruff check ods/bin/ods-host-agent.py
bash -n <changed shell plus macOS installer scripts>
PowerShell parser <all Windows installer scripts>
python -m py_compile ods/bin/ods-host-agent.py
git diff --check
  passed

Docker Compose render:
  base + NVIDIA: passed
  base + AMD: passed
  base + Apple: passed
  base + CPU: passed
```

The installer environment smoke also passed clean generation, second run, schema, dependency, and NVIDIA Compose checks. Its unrelated `ai_err` function-resolution failure reproduces on current `main`. The broad installer contract's unrelated Hermes template-bound failure also reproduces on a clean `main` worktree.

## Residual Limits

- This Windows workstation cannot provide a physical native-Linux rootless Docker receipt. Hermetic tests cover namespace behavior and lifecycle integration, but a real rootless install/start/restart/late-enable run remains the final release-environment receipt.
- Ownership changes are intentionally persistent filesystem state. If a later service fails, ODS reports the failure but does not reverse a successful ownership correction.
- Arbitrary third-party extensions are not inferred automatically; only built-ins with documented runtime UIDs are repaired.
- Rootless Host Agent networking, privileged ports, Tailscale, and proxy exposure are separate concerns and remain out of scope.

## Release Lane

- [ ] Next-minor work

## Changed Surface

- [ ] Dashboard UI
- [x] Dashboard API / Host Agent
- [ ] Docker Compose / manifests
- [x] Documentation and regression tests

### DIFF
--- a/ods/Makefile
+++ b/ods/Makefile
@@ -46,6 +46,8 @@ test: ## Run unit and contract tests
 	@echo ""
 	@echo "=== ods-doctor rootless subordinate IDs ==="
 	@bash tests/test-ods-doctor-rootless-subid.sh
+	@echo "=== Docker rootless bind-mount ownership ==="
+	@bash tests/test-rootless-docker-ownership.sh
 	@echo ""
 	@echo "=== AMD/Lemonade contracts ==="
 	@bash tests/contracts/test-amd-lemonade-contracts.sh
--- a/ods/bin/ods-host-agent.py
+++ b/ods/bin/ods-host-agent.py
@@ -3106,10 +3106,58 @@ def _precreate_data_dirs(service_id: str):
                 logger.warning("Failed to pre-create %s: %s", dir_path, e)
 
 
+_ROOTLESS_BIND_OWNERSHIP_SERVICES = {
+    "ape",
+    "comfyui",
+    "hermes",
+    "langfuse",
+    "n8n",
+    "privacy-shield",
+    "token-spy",
+    "whisper",
+}
+
+
+def _repair_rootless_data_ownership(service_id: str) -> None:
+    """Apply the built-in rootless bind-mount ownership contract before start."""
+    if platform.system() != "Linux" or service_id not in _ROOTLESS_BIND_OWNERSHIP_SERVICES:
+        return
+
+    helper = INSTALL_DIR / "lib" / "rootless-ownership.sh"
+    if not helper.is_file():
+        raise RuntimeError(f"Rootless ownership helper not found: {helper}")
+    bash = _find_usable_bash()
+    if not bash:
+        raise RuntimeError("Bash is required for Docker rootless ownership repair")
+
+    try:
+        result = subprocess.run(
+            [bash, str(helper), str(INSTALL_DIR), service_id],
+            cwd=str(INSTALL_DIR),
+            env=os.environ.copy(),
+            capture_output=True,
+            text=True,
+            timeout=SUBPROCESS_TIMEOUT_START,
+        )
+    except (OSError, subprocess.SubprocessError) as exc:
+        raise RuntimeError(
+            f"Rootless ownership repair could not run for {service_id}: {exc}"
+        ) from exc
+    if result.returncode != 0:
+        detail = (result.stderr or result.stdout or "unknown error").strip()
+        raise RuntimeError(
+            f"Rootless ownership repair failed for {service_id}: {detail[-500:]}"
+        )
+
+
 def docker_compose_action(service_id: str, action: str) -> tuple:
     flags = resolve_compose_flags()
     if action == "start":
         _precreate_data_dirs(service_id)
+        try:
+            _repair_rootless_data_ownership(service_id)
+        except RuntimeError as exc:
+            return False, str(exc)
         cmd = ["docker", "compose"] + flags + ["up", "-d", service_id]
     elif action == "stop":
         cmd = ["docker", "compose"] + flags + ["stop", service_id]
@@ -3835,6 +3883,10 @@ def _run_post_install_hook(service_id: str, ext_dir: Path) -> tuple[bool, str]:
         "GPU_BACKEND": GPU_BACKEND,
         "HOOK_NAME": "post_install",
     }
+    for runtime_key in ("DOCKER_HOST", "XDG_RUNTIME_DIR"):
+        runtime_value = os.environ.get(runtime_key, "")
+        if runtime_value:
+            hook_env[runtime_key] = runtime_value
     bash = _find_usable_bash()
     if not bash:
         msg = "post_install hook requires a usable Bash runtime. Install Git Bash or run ODS through WSL/Linux."
@@ -6151,6 +6203,10 @@ class AgentHandler(BaseHTTPRequestHandler):
             "GPU_BACKEND": GPU_BACKEND,
             "HOOK_NAME": hook_name,
         }
+        for runtime_key in ("DOCKER_HOST", "XDG_RUNTIME_DIR"):
+            runtime_value = os.environ.get(runtime_key, "")
+            if runtime_value:
+                hook_env[runtime_key] = runtime_value
         bash = _find_usable_bash()
         if not bash:
             json_response(self, 500, {
@@ -6285,6 +6341,16 @@ class AgentHandler(BaseHTTPRequestHandler):
                 # Step 3: Start
                 _write_progress(service_id, "starting", "Starting container...")
                 _precreate_data_dirs(service_id)
+                try:
+                    _repair_rootless_data_ownership(service_id)
+                except RuntimeError as exc:
+                    _write_progress(
+                        service_id,
+                        "error",
+                        "Installation failed",
+                        error=str(exc),
+                    )
+                    return
                 start_result = subprocess.run(
                     ["docker", "compose"] + flags + ["up", "-d", service_id],
                     cwd=str(INSTALL_DIR), capture_output=True, text=True,
--- a/ods/docs/LINUX-TROUBLESHOOTING-GUIDE.md
+++ b/ods/docs/LINUX-TROUBLESHOOTING-GUIDE.md
@@ -162,6 +162,34 @@ docker run --rm hello-world
 
 ---
 
+## Rootless Docker bind-mount ownership
+
+**Symptoms:** An ODS container repeatedly exits with `Permission denied` while
+writing under `data/`, but the same install works with rootful Docker.
+
+ODS prepares active bind mounts inside Docker's rootless user namespace. It
+does not apply host-side numeric `chown` values that would map to the wrong
+container identity. Existing rootless installs can be repaired without a
+reinstall:
+
+```bash
+cd ~/ods
+./ods-cli stop
+./ods-cli repair rootless-ownership
+./ods-cli start
+```
+
+The repair is limited to active ODS services with a documented container UID.
+It rejects symlink escapes, verifies the resulting ownership, and refuses to
+recursively change a mismatched directory while its container is running.
+The first repair may pull the pinned `busybox:1.36.1` helper image; offline
+repairs require that image to be present already.
+
+This command does not configure low-port forwarding or Tailscale networking.
+Those are separate rootless Docker concerns.
+
+---
+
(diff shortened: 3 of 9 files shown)


gabsprogrammer:
Validation update: all reported GitHub checks passed on head 3f155464ba57887f4d2fcc263bf21db5e8b6894c, including ShellCheck, dashboard API/frontend, mypy, Ruff, secret scan, PowerShell lint, Linux integration smoke, macOS smoke, env validation, and the full distro matrix. The remaining release-grade limitation is the native-Linux rootless Docker hardware receipt already recorded in the PR body.

gabsprogrammer:
## Final adversarial audit receipt

Audited final head `71a1ded3a434744e2fbe9ca0f0b2925b21143557` against `main` `d6b7efdc2f9466ea018a8e41dab55a24115262e6`.

### Confirmed fixes added during the audit

- fail closed when Docker rootless state cannot be determined;
- create a missing directory only for an active known service, inside the rootless namespace;
- reject both existing and dangling symlinks before filesystem mutation;
- treat empty and quoted-empty `UID` / `GID` overrides like Compose `${VAR:-default}`;
- repair stopped-service descendants on rerun even when the root directory metadata already matches;
- retain the no-mutation guard for running services and verify UID/GID/mode after every repair.

### Final validation

- rootless ownership regression suite: **25 passed**
- Host Agent tests: **256 passed**
- rootless subordinate-ID doctor tests: **10 passed**
- installer hardening, update rollback, and Compose rollback contracts: **passed**
- ShellCheck, Ruff, Python compile, Bash parser, Windows PowerShell parser, and `git diff --check`: **passed**
- real Compose rendering for NVIDIA, AMD, Apple, and CPU overlays: **passed**
- GitHub rollup on this exact SHA: **28 passed, 4 intentionally skipped, 0 failed, 0 pending**

The broad installer smoke's `ai_err` failure and the broad installer contract's Hermes template-bound failure both reproduce unchanged on current `main`; neither is introduced by this PR.

### Independent review

A separate edge-case pass reproduced the stale-descendant rerun defect described above and added its minimal regression. After that correction, the review could not reproduce another in-scope defect. No speculative changes were added.

The remaining release limitation is the physical native-Linux rootless Docker receipt already documented in the PR body; this Windows host cannot supply that hardware/environment proof.

Issue #2248, opened 2026-07-29
HIGH: Temp file cleanup gaps in ods-cli (missing EXIT trap)


patil2001:
## Description

In `ods/ods-cli:756-757`, temp file cleanup has two issues:

```bash
_compose_log=$(mktemp)
trap 'rm -f "$_compose_log"' INT TERM
```

1. `mktemp` without a `-t` prefix: on systems where `$TMPDIR` is unset, the file is created in the current working directory
2. The `trap` only covers `INT` and `TERM` but not `EXIT` — if the function returns normally, the temp file leaks

## Impact

Temp file accumulation; potential information disclosure if temp file is predictable.

## Suggested Fix

```bash
_compose_log=$(mktemp -t ods-compose.XXXXXXXX)
trap 'rm -f "$_compose_log"' EXIT INT TERM
```

## Severity
HIGH — Resource leak

gabsprogrammer:
Closing after reproducing the actual lifecycle on current `main` (`06d87b12`).

This report does not describe a remaining bug:

- `_compose_run_with_summary()` removes `_compose_log` on the successful path at `ods/ods-cli:1027-1031`.
- On failure, retaining the file is intentional and documented at `ods/ods-cli:904-911`; the CLI prints its path so the operator can inspect the complete Compose error.
- The `INT`/`TERM` trap is cleared on both normal returns (`1030` and `1099`).
- A Bash `EXIT` trap runs when the shell exits, not when this function returns, so adding `EXIT` would not provide the claimed normal-return cleanup. It would instead install a shell-global trap around a function-local variable.
- `mktemp` without a template already creates the file in the platform temporary directory on the supported GNU/BSD implementations; it does not create a predictable file in the working directory.

The proposed #2252 change therefore does not fix the reported condition and weakens the function's deliberate failure-evidence contract. If a bounded retention policy for failed Compose logs is desired later, that is a separate lifecycle feature, not this resource-leak claim.

PR #2252, opened 2026-07-29
fix: bash security hardening — quoting, glob safety, and temp file traps

patil2001:
## Summary
Three bash hardening fixes found during code review:

### 1. Unquoted variable expansion in detection.sh
- **File:** ods/installers/lib/detection.sh:726
- `for mod_path in /lib/modules/${kver}/updates/dkms/nvidia*.ko*` - `${kver}` unquoted
- Fix: Quote the variable to prevent word splitting

### 2. Glob expansion in packaging.sh
- **File:** ods/installers/lib/packaging.sh:206
- `pkgs+=($resolved)` triggers glob expansion on resolved package names
- Fix: Wrap with `set -f` / `set +f` to disable glob expansion

### 3. Missing EXIT trap in ods-cli
- **File:** ods/ods-cli:930-931
- `trap 'rm -f "$_compose_log"' INT TERM` missing EXIT signal
- Fix: Add EXIT so temp files are cleaned up on normal function return
- Also added `-t` prefix to mktemp for deterministic temp directory

## Testing
- `make lint` passes (shell syntax check)
- No functional change expected

### DIFF
--- a/ods/installers/lib/detection.sh
+++ b/ods/installers/lib/detection.sh
@@ -723,7 +723,7 @@ fix_nvidia_secure_boot() {
 
     # Sign every nvidia DKMS module (handles .ko, .ko.zst, .ko.xz)
     local signed_count=0
-    for mod_path in /lib/modules/${kver}/updates/dkms/nvidia*.ko*; do
+    for mod_path in /lib/modules/"${kver}"/updates/dkms/nvidia*.ko*; do
         [[ -f "$mod_path" ]] || continue
         case "$mod_path" in
             *.zst)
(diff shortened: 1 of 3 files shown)


gabsprogrammer:
The `ods-cli` portion should not be merged as currently described. Issue #2248 has been closed after verifying the function lifecycle: success already removes the log, failure intentionally preserves it and prints the path, and an `EXIT` trap does not execute on function return. It would add a shell-global trap without solving a normal-return leak.

The `detection.sh` quoting change addresses a real but narrowly scoped path-splitting hardening case; it should be reviewed independently from the invalid temp-file premise and the unrelated packaging change.

PR #2266, opened 2026-07-29
fix(upgrade): atomically replace model-update .env files

### LINKED ISSUE #2241: HIGH: Non-atomic .env file updates cause crash-corruption risk
## Description

In `ods/scripts/bootstrap-upgrade.sh:1242-1258` (4 occurrences) and `ods/scripts/upgrade-model.sh:246`, all `.env` update sites use a non-atomic write pattern:

```bash
awk '...' "$ENV_FILE" > "${ENV_FILE}.tmp" \
    && cat "${ENV_FILE}.tmp" > "$ENV_FILE" \
    && rm -f "${ENV_FILE}.tmp"
```

The `cat tmp > file` step is non-atomic. If the system crashes between the `cat` operation writing to `$ENV_FILE` and the `rm -f`, the `.env` file is left either partially written or truncated.

## Impact

Corrupted `.env` file after crash/power loss during model swap or bootstrap, rendering the stack unable to start until manually repaired.

## Suggested Fix

Replace `cat tmp > file && rm tmp` with `mv tmp file` which is atomic on the same filesystem:

```bash
mv "${ENV_FILE}.tmp" "$ENV_FILE"
```

## Severity
HIGH — Data corruption risk

sricursion:
Fixes #2241

## Summary

Atomically replace model-update `.env` files so a failed or interrupted update cannot truncate an existing configuration. The target keeps its existing POSIX metadata or Windows private ACL.

## Changes

- Render updates to a restricted same-directory temporary file and replace the target atomically.
- Preserve symlink targets while rejecting link cycles.
- Use native `File.Replace` on Windows Git Bash and preserve the existing `mv` path elsewhere.
- Route both existing-key replacement and missing-key append through the helper; do not start the runtime after a failed write.
- Add regression coverage and run it on Linux, macOS, and Windows CI.

## Validation

- `bash ods/tests/test-atomic-env-writes.sh` ? passed (success, restrictive temporary permissions/DACLs, failure cleanup, symlink/cycle behavior, and startup-abort coverage).
- `bash -n ods/scripts/bootstrap-upgrade.sh ods/scripts/upgrade-model.sh ods/tests/test-atomic-env-writes.sh` ? passed.
- `bash ods/tests/test-bootstrap-upgrade-hotswap-contract.sh` ? passed.
- `bash ods/tests/test-network-timeouts.sh` ? 17 passed.
- `make lint` ? passed.
- `make smoke` ? passed.
- YAML parse smoke for changed workflow files ? passed.
- `git diff --check origin/main...HEAD` ? passed.
- `make test` was rerun from a native Linux checkout; pre-existing tests passed through its installer-contract step, then the local environment stopped at `jq is required`, so a full-suite pass is not claimed.

### DIFF
--- a/.github/workflows/lint-powershell.yml
+++ b/.github/workflows/lint-powershell.yml
@@ -96,3 +96,8 @@ jobs:
         if: runner.os == 'Windows'
         shell: pwsh
         run: ./tests/contracts/test-install-footprint-windows.ps1
+
+      - name: Atomic Environment Write Tests
+        if: runner.os == 'Windows'
+        shell: bash
+        run: bash tests/test-atomic-env-writes.sh
--- a/.github/workflows/matrix-smoke.yml
+++ b/.github/workflows/matrix-smoke.yml
@@ -224,6 +224,9 @@ jobs:
       - name: macOS Dispatch Smoke
         run: bash tests/smoke/macos-dispatch.sh
 
+      - name: Atomic Environment Write Tests
+        run: /bin/bash tests/test-atomic-env-writes.sh
+
       - name: Bash 3.2 syntax check (catches declare -g, associative arrays, etc.)
         run: |
           echo "=== Checking installer scripts for Bash 3.2 compatibility ==="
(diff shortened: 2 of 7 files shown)


Lightheartdevs (approved):

PR #2269, opened 2026-07-30
fix(installer): escape Hermes fallback substitutions

### LINKED ISSUE #2243: HIGH: Sed injection in Hermes fallback patcher (phase 11)
## Description

In `ods/installers/phases/11-services.sh:740-745`, the fallback `sed -i.bak` path (used when Python is unavailable) applies sed replacements to Hermes config templates without escaping variables:

```bash
sed -i.bak \
    -e "s|^  default: \"qwen3.5-9b\"|  default: \"$_hermes_model\"|" \
    -e "s|^  context_length: .*|  context_length: ${_hermes_context}|" \
    ...
    "$_hermes_tpl"
```

The primary Python patcher path (line 737) is safe because it uses structured arguments. However, this fallback does not apply `_sed_escape` to `$_hermes_model`, `$_hermes_context`, or `$_hermes_request_timeout`. If any of these contain `/`, `\`, `&`, or `|`, the sed command fails or produces corrupt output.

## Impact

Hermes config template corruption; Hermes may fail to start or serve requests if model names or config values contain sed-special characters.

## Suggested Fix

Apply `_sed_escape` from bootstrap-upgrade.sh to all interpolated values:

```bash
local _hermes_model_sed
_hermes_model_sed=$(printf '%s' "$_hermes_model" | sed 's/[\/&|]/\&/g')
# Use _sed variants in sed -e strings
```

## Severity
HIGH — Config corruption

sricursion:
Fixes #2243

## Summary

Harden the non-Python Hermes template fallback so model names containing sed replacement characters remain data instead of breaking or altering the generated config.

Supported installs normally use the structured Python patcher. When that helper is unavailable, the degraded fallback now either produces the intended values or leaves the original template intact.

## Changes

- Escape backslashes, ampersands, and the sed delimiter in model substitutions.
- Reject nonnumeric context and timeout values and multiline model input before editing.
- Use a per-process backup and restore it if sed fails.
- Verify the resulting model and context with fixed-string, whole-line matching.
- Add executable regression coverage for metacharacter preservation, invalid structure rejection, and backup cleanup.

## Validation

- `bash -n ods/installers/phases/11-services.sh ods/tests/test-installer-context-parity.sh` - passed.
- `(cd ods && bash tests/test-installer-context-parity.sh)` - 74 passed.
- `(cd ods && make lint)` - passed.
- `(cd ods && make smoke)` - passed.

### DIFF
--- a/ods/installers/phases/11-services.sh
+++ b/ods/installers/phases/11-services.sh
@@ -204,7 +204,7 @@ _phase11_model_file_valid() {
 
 _phase11_patch_hermes_with_sed() {
     local template_path="$1" model="$2" context_length="$3" request_timeout_seconds="$4"
-    local model_sed backup_path
+    local model_yaml model_sed backup_path
 
     [[ -f "$template_path" ]] || return 1
     [[ "$context_length" =~ ^[0-9]+$ ]] || return 1
@@ -213,7 +213,10 @@ _phase11_patch_hermes_with_sed() {
         *$'\n'*|*$'\r'*) return 1 ;;
     esac
 
-    model_sed="$(printf '%s' "$model" | sed 's/[\\&|]/\\&/g')" || return 1
+    # The replacement is parsed twice: first by sed, then as a YAML
+    # double-quoted scalar. Serialize for YAML before escaping sed metacharacters.
+    model_yaml="$(printf '%s' "$model" | sed 's/\\/\\\\/g; s/"/\\"/g')" || return 1
+    model_sed="$(printf '%s' "$model_yaml" | sed 's/[\\&|]/\\&/g')" || return 1
     backup_path="${template_path}.bak.$$"
 
     if sed -i".bak.$$" \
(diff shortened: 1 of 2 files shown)


gabsprogrammer:
Independent edge-case review found one additional reproducible defect: sed metacharacters were escaped, but double quotes and backslashes could still render an invalid YAML scalar. Head e2ed9e49 now serializes the model for YAML before escaping the sed replacement. The regression includes quotes, backslashes, ampersands, and the sed delimiter; an actual PyYAML round trip preserved the original value. bash syntax and diff checks pass. The full context-parity script remains green in Linux CI; this Windows checkout hits an unrelated CRLF-sensitive baseline assertion before the new case.

gabsprogrammer:
Validation update for final head `be20a349`:

- The YAML serialization helper is now shared by both the write and verification paths, so model IDs containing quotes, backslashes, `&`, or the sed delimiter no longer produce invalid YAML or a false post-write mismatch.
- `tests/test-installer-context-parity.sh`: **76/76 passed** on the final head.
- `bash -n` and `git diff --check` passed.
- The final GitHub rollup is green: **28 successful, 0 failed, 0 pending**.

This supersedes the earlier note tied to `e2ed9e49`.

Lightheartdevs (approved):

PR #2290, opened 2026-07-30
fix(config): preserve unmatched env quotes

### LINKED ISSUE #2279: MEDIUM - .env parser in helpers.py strips quotes incorrectly with strip("\"'")
## Summary

`ods/extensions/services/dashboard-api/helpers.py` (line 825) and `ods/extensions/services/dashboard-api/config.py` (line 101) parse .env values using `value.strip("\"'")`. This removes **any** combination of `"` and `'` from both ends, rather than only matching quote pairs.

## Example

A value like `it's` becomes `it` after stripping.

## Impact

Incorrect parsing of environment variable values can lead to subtle configuration bugs and data corruption.

## Locations (4 occurrences)

1. `config.py` — line 101
2. `helpers.py` — line 825
3. `main.py` — unknown line
4. `updates.py` — unknown line

## Fix

Only strip matching quote pairs `"..."` or `'...'` rather than blindly removing all quote characters.

gabsprogrammer:
## Summary

ODS had several independent Dashboard API readers using `str.strip("\"'")` for persisted `.env` values. That operation removes arbitrary quote characters from both ends, so malformed or intentional unmatched values were silently mutated.

This change defines one dependency-free quote contract and applies it consistently across settings, runtime/model metadata, GPU assignment, performance estimates, and update/version readers:

- trim surrounding whitespace
- remove exactly one complete matching outer pair (`"..."` or `'...'`)
- preserve internal, unmatched, mixed, and additional quote characters
- preserve existing missing/empty-value and source-precedence behavior

> [!NOTE]
> The issue's `it's -> it` example is not reproducible because the apostrophe is not at an edge. The underlying defect is still confirmed: for example, `value'` currently becomes `value`.

Fixes #2279

## Data flow and invariants

| Surface | Producer | Consumer | Preserved invariant |
|---|---|---|---|
| Persisted settings | installer / Settings API writes `.env` | settings parser and live config readers | One matching pair is syntax; unmatched quotes remain data |
| Model state | model activation writes `GGUF_FILE` / `LLM_MODEL` | dashboard model cards and performance oracle | Model identifiers are never silently truncated |
| GPU state | installer / reassignment writes `GPU_ASSIGNMENT_JSON_B64` | GPU topology decoder | Payload bytes remain unchanged except one valid wrapper pair |
| Version state | installer writes `ODS_VERSION` | version API and update dry-run | Normal and dry-run readers share the same quote behavior |
| Process fallback | Compose supplies environment values | performance oracle | Existing file-over-process precedence and explicit empty values remain unchanged |

## Changed surface

- Added `env_values.strip_matching_quotes()` as the single quote-boundary helper.
- Replaced destructive or duplicated parsing in all affected Dashboard API consumers.
- Added reader-level regressions for configuration, GPU, models, performance, settings, installed version, and update dry-run.

No Compose, installer, schema, frontend, network, secret, or service lifecycle contract changes are introduced.

## Validation receipt

- `python -m pytest ...test_env_values.py ...test_updates.py -q` -> **453 passed**
- `python -m py_compile` on all changed production modules -> **passed**
- `python -m ruff check ... --select E,F,W --ignore E501,E701,E731,E741,E402` -> **passed**
- `bash ods/tests/test-safe-env.sh` -> **12 passed**
- `bash ods/tests/test-macos-env-quote-strip.sh` -> **9 passed**
- `git diff --check origin/main...HEAD` -> **passed**

The Windows endpoint shell test could not be counted locally because its `/tmp/*.ps1` handoff is not executable by native Windows PowerShell from this host. This PR changes only dependency-free Python parsing and does not alter that script or the Windows installer.

## Adversarial review

The second review started from the final diff without edits and exercised empty pairs, paired single/double quotes, internal apostrophes, unmatched leading/trailing quotes, mixed delimiters, repeated pairs, and surrounding whitespace. It also reproduced the old mutation independently.

**Result:** no remaining reproducible defect was found inside this PR's scope.

## Limits

This deliberately does not turn the Dashboard API into a full dotenv/shell parser. Escape expansion, inline comments, interpolation, and multiline values retain their existing behavior.

### DIFF
--- a/ods/extensions/services/dashboard-api/config.py
+++ b/ods/extensions/services/dashboard-api/config.py
@@ -10,6 +10,8 @@ from urllib.parse import urlparse
 
 import yaml
 
+from env_values import strip_matching_quotes
+
 logger = logging.getLogger(__name__)
 
 # --- Paths ---
@@ -98,7 +100,7 @@ def _find_env_file_value(key: str) -> tuple[bool, str]:
         for line in env_path.read_text(encoding="utf-8").splitlines():
             if line.startswith(f"{key}="):
                 found = True
-                value = line.split("=", 1)[1].strip().strip("\"'")
+                value = strip_matching_quotes(line.split("=", 1)[1])
     except (OSError, UnicodeError):
         pass
     return found, value
--- a/ods/extensions/services/dashboard-api/env_values.py
+++ b/ods/extensions/services/dashboard-api/env_values.py
@@ -1,0 +1,14 @@
+"""Small, dependency-free helpers for values read from ODS ``.env`` files."""
+
+
+def strip_matching_quotes(value: str) -> str:
+    """Trim whitespace and remove exactly one matching outer quote pair.
+
+    ODS writes shell-compatible values that may be wrapped in single or
+    double quotes. Unmatched or mixed quotes are data, not delimiters, and
+    must survive reads unchanged.
+    """
+    value = value.strip()
+    if len(value) >= 2 and value[0] == value[-1] and value[0] in {"'", '"'}:
+        return value[1:-1]
+    return value
--- a/ods/extensions/services/dashboard-api/gpu.py
+++ b/ods/extensions/services/dashboard-api/gpu.py
@@ -9,6 +9,7 @@ import subprocess
 from pathlib import Path
 from typing import Optional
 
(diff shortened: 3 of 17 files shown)


Lightheartdevs (approved):

PR #2277, opened 2026-07-30
fix: shell quoting, schema types, macOS compat, race condition, DOCKER_CMD, minLength

patil2001:
## Summary

Fixes 7 issues found during codebase audit:

1. **#2273** - Unquoted SSID/WPA passphrase in `ap-mode.sh` → hostapd.conf malformed if SSID contains spaces
2. **#2274** - `WEBUI_AUTH` schema declares `boolean` but Compose reads as `string` → validation failures
3. **#2275** - `mktemp` template on macOS/BSD fails because `XXXXXX` must be at end of template
4. **#2276** - `LLAMA_NO_METAL=1` inside Docker container on Apple is dead code, misleading
5. **#2271** - Race in `session-cleanup.sh`: file deleted before sessions.json updated → gateway errors
6. **#2270** - `DOCKER_CMD` custom values (nerdctl, podman) silently ignored → wrong binary used
7. **#2272** - `minLength: 10` allows sentinel `not-needed` to pass as valid API key

## Changes

| File | Fix |
|------|-----|
| `scripts/ap-mode.sh` | Quote SSID and passphrase in hostapd.conf |
| `.env.schema.json` | WEBUI_AUTH: boolean→string enum; minLength 10→12 |
| `installers/phases/03-features.sh` | Remove `.json` suffix from mktemp template |
| `installers/phases/05-docker.sh` | Word-split DOCKER_CMD instead of hardcoding fallback |
| `scripts/session-cleanup.sh` | Atomic `os.replace()` for sessions.json write |
| `docker-compose.apple.yml` | Remove dead LLAMA_NO_METAL=1 env var |

PR #2291, opened 2026-07-30
fix(sessions): commit index before pruning files

### LINKED ISSUE #2272: Race condition in session-cleanup.sh between file deletion and sessions.json update
## Problem

In `ods/scripts/session-cleanup.sh` (lines 141-171), the script deletes bloated session files (`rm -f "$f"`, line 141) in a loop, collecting IDs to wipe. Only **after** the loop does it update `sessions.json` (line 149). If the script crashes between deletion and JSON update (e.g., signal, disk-full, or other error), the session file is already deleted but `sessions.json` still references it. The gateway will then fail trying to load a non-existent session file.

## Impact

Gateway crashes with missing session file errors when the cleanup script is interrupted mid-execution, leading to service disruption until the stale `sessions.json` entry is manually removed.

## Previous attempt

The earlier fix only moved Python detection before the wipe loop, but the crash window between deletion and JSON update still exists.

## Fix

Fixed in working tree by using atomic `os.replace()` write to ensure `sessions.json` is updated atomically.

gabsprogrammer:
## Summary

Fixes #2272.

The cleanup previously deleted an oversized active `.jsonl` file before removing its entry from `sessions.json`. A signal, disk error, or process failure in that window left a dangling index entry and could break the gateway on its next session read.

This change makes the lifecycle fail closed and commits it in the safe direction:

1. Resolve Python and validate the complete JSON index before any cleanup mutation.
2. Collect oversized active sessions without deleting them.
3. Re-read the latest index, remove all selected references, write a same-directory temporary file, flush it, preserve its mode, and atomically replace `sessions.json`.
4. Delete only the files that are no longer referenced.

A failure before the atomic replace keeps both the old reference and file. A failure after the replace can only leave an unreferenced file, which is safe and is removed by the next run.

## AI Assistance

Static analysis, edge-case enumeration, and regression-test drafting were assisted. The author reviewed the implementation, selected the scope, reproduced the original failure invariant, and ran the validation listed below.

## Release Lane

Stable hotfix reason:

```text
```

## Changed Surface

- [ ] Docs only
- [ ] Tests only
- [ ] Dashboard UI

## Behavior And Invariants

| Requirement | Implementation | Evidence |
|---|---|---|
| Never leave a reference to a file deleted by this cleanup | Index commit precedes file removal | Signal injection kills cleanup immediately after `rm`; file-or-no-reference invariant passes |
| Do not mutate on invalid/partial JSON | Full Python parse and object-root validation happen before debris or session deletion | Malformed, zero-byte, and non-object fixtures fail closed |
| Do not partially wipe when index commit fails | Files remain queued until the atomic replace succeeds | Injected commit failure preserves both file and reference |
| Preserve unrelated/current entries | Commit re-reads the latest index and removes only matching session IDs | Concurrent-addition test preserves the new entry |
| Preserve filesystem mode | Replacement inherits the original mode | `0640` fixture remains `0640` |
| Keep repeated cleanup safe | Removed entries/files are absent and a second run succeeds | Idempotent rerun test |
| Handle multiple and non-word IDs safely | Bash arrays and argv transport replace word-splitting strings | Multiple targets and a session ID containing spaces pass |

## Risk And Validation

- Risk level: Medium
- Validation run:
  - [x] `git diff --check`
  - [ ] Not required because:

Commands/results:

```text
bash -n ods/scripts/session-cleanup.sh
bash -n ods/tests/test-session-cleanup.sh
bash ods/tests/test-session-cleanup.sh
  19 passed, 0 failed

Independent edge matrix:
  PASS fail-closed roots
  PASS multi-target, spaced ID, mode, rerun
  PASS latest-index reread preserves concurrent addition

git diff --check
  passed
```

ShellCheck reports only the two pre-existing findings in the untouched final listing block: SC2012 and SC2034.

## Operational Change Check

The affected operation is deterministic filesystem cleanup. Scoped fault injection exercises the relevant interruption and write-failure boundaries without requiring model or GPU hardware.

## Platform Scope

The timer installs this Bash script through the Linux OpenClaw/AMD tuning lifecycle. Windows and macOS installer, Compose, dashboard, and runtime paths are unchanged. The Python transaction uses only standard-library filesystem APIs and a same-directory replacement.

## Confirmed Bugs

- Confirmed: deletion-before-index-update interruption window from #2272.
- Confirmed during reproduction: malformed or partially written `sessions.json` was treated as an empty active set by the grep parser, allowing every `.jsonl` file to be classified as inactive.

## Risks And Limitations

- The gateway does not participate in a shared filesystem lock. This change minimizes the transaction and re-reads the latest index before replacement, but it does not introduce a cross-process locking protocol that the gateway itself does not honor.
- An interruption after index replacement may leave an orphaned `.jsonl`; this is intentionally the safe failure direction and the next cleanup removes it as inactive.
- Inactive-file pruning behavior is otherwise unchanged.

## Notes For Reviewers

PR #2277 adds atomic replacement but retains the unsafe delete-before-index order. This PR is intentionally limited to the session cleanup transaction and its focused regression coverage.

### DIFF
--- a/ods/scripts/session-cleanup.sh
+++ b/ods/scripts/session-cleanup.sh
@@ -5,8 +5,9 @@
 #
 # Prevents context overflow crashes by automatically managing
 # session file lifecycle. When a session file exceeds the size
-# threshold, it's deleted and its reference removed from
-# sessions.json, forcing the gateway to create a fresh session.
+# threshold, its reference is atomically removed from sessions.json
+# before the file is deleted, forcing the gateway to create a fresh
+# session without leaving a dangling index entry on interruption.
 #
 # The agent doesn't notice — it just gets a clean context window.
 # ═══════════════════════════════════════════════════════════════
@@ -36,7 +37,7 @@ usage() {
     echo "  MAX_SIZE       Max session file size in bytes (default: 256000)"
     echo ""
     echo "Exit: 0 on success or when paths are missing (skipped with a log message);"
-    echo "      1 when no usable Python is found (refused before touching any session)."
+    echo "      1 when Python or sessions.json validation fails before any session mutation."
 }
 
 case "${1:-}" in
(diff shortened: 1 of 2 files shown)


gabsprogrammer:
Final-head validation for `a9bec434`:

- Reproduced a Windows/Git-Bash edge case where native Python returned CRLF and the trailing `\r` caused the active session ID to be treated as inactive during pruning.
- The session-index reader now normalizes the trailing carriage return before comparison.
- `tests/test-session-cleanup.sh`: **19/19 passed**, including atomic index commit, prune failure, active-session preservation, CRLF input, and idempotent rerun coverage.
- The final GitHub rollup is green: **28 successful, 0 failed, 0 pending**.

Lightheartdevs (approved):

PR #2292, opened 2026-07-30
fix(llama): respect GPU layer placement

### LINKED ISSUE #2255: PERF: hardcoded --n-gpu-layers 999 causes OOM on 4GB VRAM cards; no env override
## Description

`--n-gpu-layers 999` is hardcoded in 6 locations across the codebase with no environment variable override:

1. `ods/docker-compose.base.yml:41` — `"999"` (no env var)
2. `ods/scripts/bootstrap-upgrade.sh:1203` — `"--n-gpu-layers", "999"`
3. `ods/scripts/bootstrap-upgrade.sh:2978` — `--n-gpu-layers 999`
4. `ods/scripts/bootstrap-upgrade.sh:3026` — `--n-gpu-layers 999`
5. `ods/installers/windows/ods.ps1:1839` — `"--n-gpu-layers", "999"`
6. `ods/installers/windows/install-windows.ps1:717` — `"--n-gpu-layers", "999"`

On GPUs with 4GB VRAM (GTX 1650, RTX 3050, RTX 4050), offloading ALL layers to GPU causes OOM for any model larger than ~3.5 GB. Even with flash attention, the KV cache at 16K context consumes ~1-2GB VRAM.

## Impact

- Model loading fails silently or Docker container OOM-kills on 4GB VRAM cards
- User cannot tune the GPU/CPU split without modifying compose files
- On 8GB system RAM + 4GB VRAM, the hardcoded `999` prevents running larger models at all

## Suggested Fix

1. Make `--n-gpu-layers` configurable via `N_GPU_LAYERS` env var in `docker-compose.base.yml`:
```yaml
- --n-gpu-layers
- "${N_GPU_LAYERS:-999}"
```

2. Calculate a safe default in the installer based on model_size / VRAM ratio:
   - If model + context > VRAM: `N_GPU_LAYERS = floor(total_layers * VRAM_GB * 0.85 / model_size_GB)`
   - Fallback: `N_GPU_LAYERS=16` for cards with < 6GB VRAM

3. Apply the same env var pattern in bootstrap-upgrade.sh and Windows installer scripts.

## Files to change
- `ods/docker-compose.base.yml:41`
- `ods/docker-compose.cpu.yml:18`
- `ods/scripts/bootstrap-upgrade.sh:1203,2978,3026`
- `ods/installers/windows/ods.ps1:1839`
- `ods/installers/windows/install-windows.ps1:717`
- `ods/.env.example` (document the new variable)

### LINKED ISSUE #2259: PERF: no auto VRAM-layers calculation causes trial-and-error OOM for 4GB cards
## Description

There is no code anywhere in the codebase that calculates how many GPU layers a given model can fit in available VRAM. The installer always sets `--n-gpu-layers 999` regardless of model size or VRAM capacity.

For effective GPU offload on 4GB VRAM:
- Gemma-4-E2B (2.81 GB / ~26 layers): ~99% offload works (~2.8 GB in VRAM)
- Qwen3.5-9B (5.76 GB / ~40 layers): only ~50% offload (~20 layers, ~2.8 GB in VRAM)
- Qwen3.5-4B (2.87 GB / ~28 layers): ~99% offload works

A formula like `layers_to_offload = floor(total_layers * (VRAM_GB * 0.85 - context_overhead_GB) / model_size_GB)` would calculate a safe default.

## Impact
- Users cannot run larger models without trial-and-error OOM
- 4GB VRAM users forced to reduce `--n-gpu-layers` manually
- Model swap becomes risky (new model may OOM with 999 layers)

## Suggested Fix

Add a function in `ods/installers/lib/tier-map.sh` or `ods/installers/lib/detection.sh`:

```bash
calc_safe_n_gpu_layers() {
    local total_layers=$1   # From GGUF metadata
    local model_size_mb=$2  # From tier-map config
    local vram_mb=$3        # From GPU detection
    local ctx_mb=$4         # Estimated KV cache at target context
    
    local available_vram=$((vram_mb - ctx_mb - 256))  # Reserve 256MB for overhead
    if [[ $available_vram -le 0 ]]; then
        echo "0"  # CPU only
        return
    fi
    
    # Proportional offload
    local layers=$(( total_layers * available_vram / model_size_mb ))
    [[ $layers -gt $total_layers ]] && layers=$total_layers
    [[ $layers -lt 0 ]] && layers=0
    echo "$layers"
}
```

Export as `N_GPU_LAYERS_AUTO` and use it as default in compose when `N_GPU_LAYERS` is unset.

## Files to change
- `ods/installers/lib/tier-map.sh` or `ods/installers/lib/detection.sh` (add calc function)
- `ods/installers/phases/02-detection.sh` (call calculator, export N_GPU_LAYERS_AUTO)
- `ods/docker-compose.base.yml:41` (use ${N_GPU_LAYERS:-${N_GPU_LAYERS_AUTO:-999}})
- `ods/.env.example` (document N_GPU_LAYERS_AUTO)

gabsprogrammer:
## Summary

Closes #2255 and #2259.

ODS forced `--n-gpu-layers 999` in the shared Compose path and every native llama.cpp launcher. Current llama.cpp treats that as an explicit operator choice, so its memory fitter cannot reduce GPU placement when a model, context, or KV cache no longer fits. The result is a startup failure instead of the memory-aware placement that llama.cpp already provides.

This change makes `N_GPU_LAYERS=auto` the default, preserves explicit `all` or non-negative layer-count overrides, and carries the same value through clean installs, reruns, upgrades, rollback launchers, Settings, and runtime recreation.

### Behavior and invariants

|---|---|
| Default to memory-aware placement | Shared Compose and native launchers use `auto` when the key is absent or empty |
| Preserve operator intent | `all` and integer values survive installer reruns and Settings serialization |
| Keep CPU-only deployments CPU-only | The CPU overlay remains pinned to `0` |
| Preserve backend-specific Intel defaults | Intel and Arc tier selection can continue to set `99` |
| Keep every runtime path aligned | Compose, Linux host agent, Windows native, macOS native, bootstrap upgrade, and rollback consume the same key |
| Reject malformed launch arguments | Schema, Settings API, and `validate-env.sh` accept only `auto`, `all`, or a non-negative integer |
| Avoid unrelated lifecycle changes | Model identity, context, ports, health checks, and service topology are unchanged |

## AI Assistance

AI-assisted repository search, edge-case enumeration, and regression-test drafting were used. The author reviewed the implementation, selected the validation scope, reproduced the contract failure, and is responsible for the final diff.

## Release Lane

Stable hotfix reason:

```text
```

## Changed Surface

- [ ] Docs only
- [ ] Tests only
- [ ] Dashboard UI

## Risk And Validation

- Risk level: High
- Validation run:
  - [x] `git diff --check`
  - [x] Markdown/link sanity for docs
  - [ ] Not required because:

Commands/results:

```text
python -m pytest ods/extensions/services/dashboard-api/tests/test_settings_env.py -q
88 passed

python ods/tests/test-gpu-layer-contract.py
PASS: Windows fresh/default, explicit override, rerun preservation, whitespace normalization, missing-key backfill, and empty-key backfill
SKIP: POSIX macOS lifecycle in this Windows checkout

python ods/tests/contracts/test-llama-runtime-tunables.py
PASS: every Compose overlay and native launcher preserves the runtime tunable

bash ods/tests/contracts/test-llama-reasoning-format.sh
PASS

bash ods/tests/contracts/test-llama-metrics-flag.sh
PASS

PowerShell parser: install-windows.ps1, env-generator.ps1, ods.ps1
PASS

python -m ruff check (touched Python files)
PASS

python -m py_compile (touched Python files)
PASS

git diff --check
PASS

Compose rendering verified:
- NVIDIA default -> auto
- NVIDIA explicit override -> requested count
- CPU overlay -> 0
- Intel/Arc defaults and explicit overrides remain valid
```

The macOS generator lifecycle was also exercised manually through Bash with fresh, explicit, persisted, missing, empty, and whitespace-padded values. No physical Apple Silicon or Linux GPU runtime receipt is claimed by this draft.

## Operational Change Check

## Upgrade, failure, and rollback behavior

- Existing valid `N_GPU_LAYERS` values are preserved on rerun. This includes legacy `99` values because old files do not record whether `99` came from the installer or an operator; changing it silently would overwrite operator intent. Those installs can set `auto` through Settings or `.env`, and the apply plan recreates only `llama-server`.
- Older environments without the key are backfilled to `auto`.
- Empty values normalize to `auto`; malformed values fail validation before launch.
- Bootstrap and rollback launchers read the same persisted value, so a failed upgrade does not silently restore the old forced-`999` behavior.
- AMD Lemonade and cloud-provider runtime paths are not changed because they do not launch llama.cpp through this flag.

## Notes For Reviewers

This intentionally uses llama.cpp's upstream memory fitter instead of introducing a second ODS-specific VRAM calculator. The exact pinned llama.cpp builds used by ODS (`b8210`, `b8248`, and `b9014`) were checked and all accept `auto`, `all`, and explicit counts; `auto` activates their native placement fitter. The independent review attempted malformed values, whitespace, missing/empty keys, reruns, explicit overrides, CPU-only rendering, backend overlays, native launchers, and rollback paths. One serialization inconsistency was found and fixed during that review. A separate regression proves that changing `N_GPU_LAYERS` from `99` to `auto` through Settings persists the value and schedules `llama-server` recreation. No other reproducible in-scope defect remained after the final pass.

Closes #2259

### DIFF
--- a/ods/.env.example
+++ b/ods/.env.example
@@ -225,6 +225,7 @@ SYSTEM_RAM_GB=32
 # WHISPER_IMAGE=ghcr.io/speaches-ai/speaches:0.9.0-rc.3-cuda
 
 # llama-server inference tuning (advanced)
+# N_GPU_LAYERS=auto         # auto | all | non-negative layer count
 # LLAMA_BATCH_SIZE=2048      # Batch size for prompt processing (higher = faster prefill)
 # LLAMA_THREADS=4            # CPU threads for non-GPU work
 # LLAMA_PARALLEL=1           # Concurrent request slots (increase for multi-user)
--- a/ods/.env.schema.json
+++ b/ods/.env.schema.json
@@ -400,9 +400,10 @@
       "default": "nvidia"
     },
     "N_GPU_LAYERS": {
-      "type": "integer",
-      "description": "Number of model layers to offload to GPU",
-      "default": 99
+      "type": "string",
+      "description": "llama.cpp GPU layer offload policy: auto, all, or a non-negative layer count. CPU and Intel/Arc overlays may set an explicit numeric value.",
+      "pattern": "^(?:auto|all|[0-9]+)$",
+      "default": "auto"
     },
     "LLAMA_BATCH_SIZE": {
       "type": "integer",
--- a/ods/bin/ods-host-agent.py
+++ b/ods/bin/ods-host-agent.py
@@ -11770,12 +11770,13 @@ def _restart_macos_native_llama_server(
 def _launch_native_llama_server(env_path: Path, llama_bin: Path, llama_log: Path, pid_file: Path):
     """Launch the native (Metal) llama-server process and write its PID file.
 
-    Reads the current .env for GGUF_FILE, CTX_SIZE, and LLAMA_REASONING so
-    the caller only needs to ensure .env is up-to-date before calling.
+    Reads the current .env for model and llama.cpp runtime settings so the
+    caller only needs to ensure .env is up-to-date before calling.
     """
     env = load_env(env_path)
     gguf_file = env.get("GGUF_FILE", "")
     ctx_size = env.get("CTX_SIZE", "32768")
+    gpu_layers = env.get("N_GPU_LAYERS", "").strip() or "auto"
     model_path = INSTALL_DIR / "data" / "models" / gguf_file
     reasoning = env.get("LLAMA_REASONING", "off")
     reasoning_fmt = {"off": "none", "on": "deepseek"}.get(reasoning, reasoning)
@@ -11793,7 +11794,7 @@ def _launch_native_llama_server(env_path: Path, llama_bin: Path, llama_log: Path
         "--host", bind_addr, "--port", str(port),
         "--model", str(model_path),
         "--ctx-size", ctx_size,
-        "--n-gpu-layers", "999",
+        "--n-gpu-layers", gpu_layers,
         "--parallel", env.get("LLAMA_PARALLEL", "1"),
         "--reasoning-format", reasoning_fmt,
         "--metrics",
--- a/ods/docker-compose.base.yml
+++ b/ods/docker-compose.base.yml
@@ -38,7 +38,7 @@ services:
       - --port
       - "8080"
       - --n-gpu-layers
-      - "999"
+      - "${N_GPU_LAYERS:-auto}"
       - --ctx-size
       - "${CTX_SIZE:-16384}"
       - --batch-size
--- a/ods/extensions/services/dashboard-api/settings.py
+++ b/ods/extensions/services/dashboard-api/settings.py
@@ -278,7 +278,13 @@ def _validate_env_values(
             if _normalize_bool(value) is None:
                 issues.append({"key": key, "message": "Must be true or false."})
 
-        if key == "EMBEDDING_MODEL":
+        if key == "N_GPU_LAYERS":
+            if not re.fullmatch(r"(?:auto|all|[0-9]+)", str(value).strip()):
+                issues.append({
+                    "key": key,
+                    "message": "Must be auto, all, or a non-negative whole number.",
+                })
+        elif key == "EMBEDDING_MODEL":
             if _is_unsupported_tei_model_id(value):
                 issues.append({
                     "key": key,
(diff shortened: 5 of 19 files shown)


gabsprogrammer:
Independent edge-case review found one reproducible test-harness defect on Windows: pytest collected the macOS generator case even though the script entry point correctly skipped it, causing the Windows WSL bash shim to receive a Git-Bash-style path. Commit d80b595a aligns the per-test prerequisite checks with the script runner using explicit skips. Runtime behavior is unchanged. Revalidation: 89 passed, 1 skipped; direct Windows generator lifecycle passed; env validation passed 45/45; Ruff, Python compile, and diff checks passed. No additional reproducible defect was found in the GPU-layer lifecycle scope.

Lightheartdevs (approved):

PR #2294, opened 2026-07-31
fix(installer): quote NVIDIA DKMS kernel path

### LINKED ISSUE #2242: HIGH: Unquoted glob path in detection.sh causes word splitting
## Description

In `ods/installers/lib/detection.sh:726`:

```bash
for mod_path in /lib/modules/${kver}/updates/dkms/nvidia*.ko*; do
```

The `${kver}` variable (kernel version) is unquoted. If `$kver` contains spaces or shell-special characters, the glob breaks into multiple arguments via word splitting. Additionally, if `nullglob` is not set and no files match, the literal glob string is used as `mod_path`.

## Impact

DKMS module signing silently skips files or produces errors if kernel version string contains shell-special characters.

## Suggested Fix

Quote the variable and handle nullglob:

```bash
shopt -s nullglob
for mod_path in /lib/modules/"${kver}"/updates/dkms/nvidia*.ko*; do
    [[ -f "$mod_path" ]] || continue
    ...
done
shopt -u nullglob
```

## Severity
HIGH — Silent failures in GPU detection

gabsprogrammer:
Fixes #2242

## Summary

Keep the kernel release returned by `uname -r` as one path segment while discovering NVIDIA DKMS modules for Secure Boot signing.

## Root cause

The module glob expanded `${kver}` without quoting. A release string containing whitespace or shell metacharacters could be split before pathname expansion, so the signing loop would inspect the wrong paths or silently sign no modules.

## Change

- Quote only the `/lib/modules/${kver}/updates/dkms` directory segment.
- Leave `nvidia*.ko*` unquoted so module discovery still uses pathname expansion.
- Retain the existing `[[ -f "$mod_path" ]] || continue` guard for the no-match case without changing the caller's global `nullglob` state.
- Add installer-hardening contracts that require the quoted form and reject the old unquoted form.

## Invariants

- Normal kernel releases resolve to the same DKMS module files.
- `.ko`, `.ko.zst`, and `.ko.xz` discovery remains unchanged.
- No-match behavior remains a clean zero-module iteration.
- No shell option is changed globally.
- This affects only the Linux NVIDIA Secure Boot repair path; Windows, macOS, AMD, Intel, and ordinary NVIDIA startup paths are unchanged.

## Validation

- `bash -n installers/lib/detection.sh tests/contracts/test-installer-hardening.sh`
- `bash tests/contracts/test-installer-hardening.sh` -> passed
- `git diff --check`

## Scope

This intentionally does not include the unrelated packaging and `ods-cli` trap changes from #2252. The latter does not solve a normal-return cleanup problem because an `EXIT` trap is process-scoped, not function-scoped.

### DIFF
--- a/ods/installers/lib/detection.sh
+++ b/ods/installers/lib/detection.sh
@@ -723,7 +723,7 @@ fix_nvidia_secure_boot() {
 
     # Sign every nvidia DKMS module (handles .ko, .ko.zst, .ko.xz)
     local signed_count=0
-    for mod_path in /lib/modules/${kver}/updates/dkms/nvidia*.ko*; do
+    for mod_path in /lib/modules/"${kver}"/updates/dkms/nvidia*.ko*; do
         [[ -f "$mod_path" ]] || continue
         case "$mod_path" in
             *.zst)
--- a/ods/tests/contracts/test-installer-hardening.sh
+++ b/ods/tests/contracts/test-installer-hardening.sh
@@ -666,6 +666,12 @@ PATH="$open_bin:$PATH" bash -c '
   validate_nvidia_blackwell_open_modules
 '
 
+echo "[contract] NVIDIA DKMS module discovery preserves the kernel release as one path segment"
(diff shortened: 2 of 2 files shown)


Lightheartdevs (approved):

PR #2311, opened 2026-08-04
fix(gpu): keep every GPU in the hybrid llama parallelism plan

Hoang130203:
## Summary

`largest_pow2_divisor()` in `scripts/assign_gpus.py` clamped its result up to 2
(`return max(2, p)`), so it could return a tensor group size that does not
divide the GPU count. `select_parallelism()` then derives
`pipeline_parallel_size = n // tensor_parallel_size`, so every odd multi-GPU
subset produced a plan that covers fewer GPUs than the assignment reserved:

| llama GPUs | emitted plan | GPUs covered |
|---|---|---|
| 5 | `hybrid` tp=2, pp=2 | 4 |
| 7 | `hybrid` tp=2, pp=3 | 6 |
| 9 | `hybrid` tp=2, pp=4 | 8 |
| 11 | `hybrid` tp=2, pp=5 | 10 |

The uncovered GPUs stay reserved for `llama_server` (so no other service can
use them) but never take a shard, and the operator sees a `dedicated` strategy
that silently wastes a device. `bin/ods-host-agent.py` consumes this plan
verbatim via `_run_gpu_planner()`, so the same gap reaches model activation.

Reproduced on a fixture already in the repo, before the change:

```text
$ python scripts/assign_gpus.py \
    --topology tests/fixtures/topology_json/nvidia_smi_topo_matrix_8gpus_nv12_full_mesh.json \
    --model-size 100000
llama gpus: 5
mode: hybrid tp: 2 pp: 2
covered: 4 of 5
```

### Fix

- `largest_pow2_divisor()` now returns `1` when no power of 2 divides `n`
  instead of clamping to 2, and documents that `1` means "hybrid cannot cover
  this GPU count".
- Both hybrid call sites go through a new `hybrid_or_pipeline()` helper, which
  falls back to pipeline parallelism (one stage per GPU) when a hybrid layout
  is not representable. Pipeline places every GPU, so no device is dropped.

Even GPU counts are unaffected — they already had a power-of-2 divisor and keep
the exact same plan (verified 2..16).

### Test changes

The existing suite had encoded the bug. `test_model_needs_five_gpus_hybrid`
asserted `tensor_parallel_size == 2` and `pipeline_parallel_size == 2` for a
**five**-GPU subset. Three such assertions are updated to the pipeline
fallback, one hybrid case is re-pointed at a genuine 8-GPU split (tp=2, pp=4),
and a new `TestParallelismCoversEveryGpu` asserts the invariant
`tensor_parallel_size * pipeline_parallel_size == len(gpus)` across the
topology fixtures.

## AI Assistance

AI assisted with reviewing the planner, drafting the regression test, and
wording this description. I read the full diff, confirmed the failing case
against the repo's own fixtures, and ran the validation below.

## Release Lane

Stable hotfix reason:

```text
n/a
```

## Changed Surface

- [ ] Docs only
- [ ] Tests only
- [ ] Dashboard UI

## Risk And Validation

- Risk level: Low
- Validation run:
  - [x] `git diff --check`

Commands/results:

```text
$ python -m pytest tests/test-assign-gpus.py -q
87 passed in 11.69s

# new invariant test against the unpatched planner:
$ git stash -- ods/scripts/assign_gpus.py
$ python -m pytest tests/test-assign-gpus.py::TestParallelismCoversEveryGpu -q
E  AssertionError: nvidia_smi_topo_matrix_8gpus_nv12_full_mesh.json @ 100000MB:
   mode=hybrid covers 4 of 5 assigned GPUs
1 failed, 1 passed

# every GPU count 2..16 covered after the change:
n= 4 hybrid   tp=2 pp=2 covered=4 ok
n= 5 pipeline tp=1 pp=5 covered=5 ok
n= 8 hybrid   tp=2 pp=4 covered=8 ok
n= 9 pipeline tp=1 pp=9 covered=9 ok
n=16 hybrid   tp=4 pp=4 covered=16 ok
```

## Operational Change Check

The planner is a pure function over a topology JSON file; the change only
affects which parallelism mode is emitted for GPU counts that previously
produced an under-covered plan. No installer phase, compose generation, or
host mutation is touched.

## Notes For Reviewers

Odd-count subsets now report `pipeline` instead of `hybrid`. That is the point
of the change — a hybrid layout cannot tile an odd GPU count with a power-of-2
tensor group, and pipeline is the mode the planner already uses everywhere else
when tensor grouping does not apply.

If you would rather keep `hybrid` for counts like 9 (3 x 3), that needs
relaxing the power-of-2 constraint the function documents today. Happy to do
that instead, but it changes tensor-group semantics rather than just fixing
coverage, so I kept this diff to the coverage bug.

### DIFF
--- a/.github/workflows/test-linux.yml
+++ b/.github/workflows/test-linux.yml
@@ -63,7 +63,7 @@ jobs:
           python-version: "3.12"
 
       - name: Install manifest validator dependencies
-        run: python3 -m pip install "PyYAML>=6.0,<7.0" "jsonschema>=4.21,<5.0"
+        run: python3 -m pip install "PyYAML>=6.0,<7.0" "jsonschema>=4.21,<5.0" "pytest>=8,<9"
 
       - name: Docs Link Checks
         run: bash tests/test-doc-links.sh
@@ -195,6 +195,9 @@ jobs:
 
       - name: Perplexica Entrypoint Contract Tests
         run: python3 tests/test-perplexica-entrypoint.py
+
+      - name: Multi-GPU Assignment Planner Tests
+        run: python3 -m pytest tests/test-assign-gpus.py -q
 
       - name: CPU-Only Path Tests
         run: bash tests/test-cpu-only-path.sh
--- a/ods/scripts/assign_gpus.py
+++ b/ods/scripts/assign_gpus.py
@@ -299,7 +299,12 @@ def largest_pow2_divisor(n: int) -> int:
     Find the largest power of 2 p such that:
       - p divides n evenly
       - p <= sqrt(n)  (keeps tensor_size <= pipeline_size for balance)
-    Minimum return value is 2 (hybrid requires at least 2 tensor groups).
+
+    Returns 1 when no power of 2 divides n — every odd GPU count. Callers must
+    read 1 as "a hybrid plan cannot cover this GPU count" and fall back to
+    pipeline. Clamping the result up to 2 instead leaves GPUs out of the plan
+    entirely: hybrid derives pipeline_parallel_size as n // tensor_parallel_size,
+    so an odd n silently yields tensor * pipeline < n (5 GPUs -> 2 x 2).
     """
     p = 1
     while True:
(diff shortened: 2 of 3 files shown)


Hoang130203:
Follow-up commit pushed: `tests/test-assign-gpus.py` was not referenced by any workflow or make target, so the suite covering `scripts/assign_gpus.py` — including the regression test in this PR — never actually ran. Wired it into the Linux test job (`python3 -m pytest tests/test-assign-gpus.py -q`) and added pytest to that job's dependency install. The planner is pure Python over a topology JSON fixture, so it needs no GPU.

Also updating the Changed Surface for this PR to include CI.

PR #2322, opened 2026-08-04
fix(model-router): retain routes on malformed endpoint config

tang-vu:
## Summary
- validate the endpoint document root, endpoint array, entries, and string fields
- retain the last-known-good allowlist when a rewritten config has an invalid shape
- cover malformed roots, entries, and field values

## Test plan
- python -m pytest tests/test_router.py -q (51 passed)


tang-vu:
Rebased onto current upstream/main and reconciled overlap with later #3102. This earlier PR remains canonical because its production validation is a strict superset: it validates root/list/entry containers plus id, baseUrl, and apiKeyEnv field types. I ported the stronger HTTP-boundary regression style from #3102 in commit 0eea6e8e, expanded to ten malformed rewrite shapes including invalid field types. Validation: full model-router suite 61 passed; compileall and git diff --check pass.

PR #2340, opened 2026-08-04
fix(cli): route ods rollback through ods-update.sh snapshot restore

patil2001:
## Summary

`ods rollback` never worked. The CLI's `cmd_rollback` read `$INSTALL_DIR/.last-backup-id`, a file nothing ever writes (repo-wide grep finds only the read/rm in `cmd_rollback`), and it delegated the actual restore to `ods-restore.sh`, which reads backups from `$ODS_DIR/.backups` while pre-update snapshots are created in `data/backups` by `ods-update.sh`. So the update flow told users "Run 'ods rollback'" after a failed recreation, and the command always failed with "No rollback point found". This change rewrites the CLI's `cmd_rollback` to delegate to `ods-update.sh rollback`, which already resolves the most recent `pre-update-*` snapshot, stops the stack, restores the snapshot (`.env`, compose overlays, per-extension configs, `.version`), clears stale compose-flags, and recreates services with the restored overlay set.

## AI Assistance

## Release Lane

Stable hotfix reason:

```text
```

## Changed Surface

- [ ] Docs only
- [ ] Tests only
- [ ] Dashboard UI
- [x] Installer
- [x] Core runtime
- [ ] Frontend
- [ ] Backend

## Validation

- `bash -n ods/ods-cli` - passed.

### DIFF
--- a/ods/ods-cli
+++ b/ods/ods-cli
@@ -4691,22 +4691,11 @@ cmd_restore() {
 
 cmd_rollback() {
     check_install
-    cd "$INSTALL_DIR"
-    load_env
 
-    if [[ ! -f "$INSTALL_DIR/.last-backup-id" ]]; then
-        error "No rollback point found. Rollback is only available after a failed update."
+    if [[ ! -x "$INSTALL_DIR/ods-update.sh" ]]; then
+        error "ods-update.sh not found or not executable"
     fi
 
-    local backup_id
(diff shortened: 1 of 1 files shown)


patil2001:
Motivation for this one: fix(cli): route ods rollback through ods-update.sh snapshot restore. This is a contained fix for an edge case that would otherwise leave the stack in the wrong state on that path.

PR #2341, opened 2026-08-04
fix(api): tolerate null release body in release manifest

patil2001:
## Summary

GitHub's releases API sends `"body": null` for releases without release notes. In `get_release_manifest`, `r.get("body", "")` returns `None` for a key present with a null value, so the changelog truncation `None[:500]` raised `TypeError`, turning `GET /api/releases/manifest` into a 500. The changelog now coalesces the body to an empty string before truncating. Added a regression test with a null-body release.

## AI Assistance

## Release Lane

Stable hotfix reason:

```text
```

## Changed Surface

- [ ] Docs only
- [ ] Tests only
- [ ] Dashboard UI
- [ ] Installer
- [ ] Core runtime
- [ ] Frontend
- [x] Backend

## Validation

- `python -m py_compile extensions/services/dashboard-api/routers/updates.py` - passed.
- `cd extensions/services/dashboard-api && pytest tests/test_updates.py` - passed (28 passed).

### DIFF
--- a/ods/extensions/services/dashboard-api/routers/updates.py
+++ b/ods/extensions/services/dashboard-api/routers/updates.py
@@ -236,7 +236,7 @@ async def get_release_manifest():
             raise httpx.HTTPError(f"unexpected releases response: {type(releases).__name__}")
         return {
             "releases": [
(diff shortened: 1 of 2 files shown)


patil2001:
Adding context for reviewers: fix(api): tolerate null release body in release manifest. This is a contained fix for an edge case that would otherwise leave the stack in the wrong state on that path.

patil2001:
Posting the rationale: fix(api): tolerate null release body in release manifest. This is a contained fix for an edge case that would otherwise leave the stack in the wrong state on that path.

PR #2342, opened 2026-08-04
fix(api): skip empty choices chunks in vision stream

patil2001:
## Summary

The vision chat SSE stream crashed mid-response on role/keepalive chunks. `delta = chunk.get("choices", [{}])[0]` only uses the `[{}]` fallback when the `choices` key is absent; when the backend sends `"choices": []`, the `.get` returns the empty array and `[0]` raises `IndexError`, killing the stream and the accumulated reply. The code now coalesces a falsy `choices` value to `[{}]` before indexing. Added a regression test that feeds an empty-choices chunk followed by a content chunk and asserts the stream completes.

## AI Assistance

## Release Lane

Stable hotfix reason:

```text
```

## Changed Surface

- [ ] Docs only
- [ ] Tests only
- [ ] Dashboard UI
- [ ] Installer
- [ ] Core runtime
- [ ] Frontend
- [x] Backend

## Validation

- `python -m py_compile extensions/services/dashboard-api/routers/talk.py` - passed.
- `cd extensions/services/dashboard-api && pytest tests/test_talk.py` - passed (40 passed).

### DIFF
--- a/ods/extensions/services/dashboard-api/routers/talk.py
+++ b/ods/extensions/services/dashboard-api/routers/talk.py
@@ -230,7 +230,7 @@ async def _stream_vision_chat(image_bytes: bytes, content_type: str, prompt_text
                         chunk = json.loads(payload_str)
                     except json.JSONDecodeError:
                         continue
-                    delta = chunk.get("choices", [{}])[0].get("delta", {})
+                    delta = (chunk.get("choices") or [{}])[0].get("delta", {})
                     text = delta.get("content")
                     if isinstance(text, str) and text:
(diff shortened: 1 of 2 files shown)


patil2001:
Motivation for this one: fix(api): skip empty choices chunks in vision stream. This is a contained fix for an edge case that would otherwise leave the stack in the wrong state on that path.

patil2001:
The why behind this change: fix(api): skip empty choices chunks in vision stream. This is a contained fix for an edge case that would otherwise leave the stack in the wrong state on that path.

PR #2343, opened 2026-08-04
fix(api): tolerate null statistics from n8n in workflows list

patil2001:
## Summary

`GET /api/workflows` 500'd when an installed n8n workflow reports `"statistics": null`. `installed.get("statistics", {})` returned `None` for a key present with a null value, and the subsequent `.get("executions")` raised `AttributeError`, taking down the whole workflow catalog response. The code now coalesces null `statistics` and `executions` to `{}` before reading `total`. Added a regression test with a null-statistics workflow.

## AI Assistance

## Release Lane

Stable hotfix reason:

```text
```

## Changed Surface

- [ ] Docs only
- [ ] Tests only
- [ ] Dashboard UI
- [ ] Installer
- [ ] Core runtime
- [ ] Frontend
- [x] Backend

## Validation

- `python -m py_compile extensions/services/dashboard-api/routers/workflows.py` - passed.
- `cd extensions/services/dashboard-api && pytest tests/test_workflows.py` - passed (46 passed).

### DIFF
--- a/ods/extensions/services/dashboard-api/routers/workflows.py
+++ b/ods/extensions/services/dashboard-api/routers/workflows.py
@@ -135,7 +135,7 @@ async def api_workflows(api_key: str = Depends(verify_api_key)):
 
         executions = 0
         if installed:
-            executions = installed.get("statistics", {}).get("executions", {}).get("total", 0)
+            executions = ((installed.get("statistics") or {}).get("executions") or {}).get("total", 0)
 
         workflows.append({
             "id": wf["id"],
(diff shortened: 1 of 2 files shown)


patil2001:
Posting the rationale: fix(api): tolerate null statistics from n8n in workflows list. This is a contained fix for an edge case that would otherwise leave the stack in the wrong state on that path.

patil2001:
Why this is necessary: fix(api): tolerate null statistics from n8n in workflows list. This is a contained fix for an edge case that would otherwise leave the stack in the wrong state on that path.

PR #2344, opened 2026-08-04
fix(api): read user-data files as utf-8 with error tolerance

patil2001:
## Summary

Three dashboard-api reads of user-written files used locale-default encoding and only guarded against `OSError`/`JSONDecodeError`, so a single non-UTF-8 byte in those files raised `UnicodeDecodeError` (a `ValueError` subclass) that escaped the guard and crashed the endpoint: `_read_installed_version` on `.env`, `get_model_info` on `.env`, and `get_bootstrap_status` on `bootstrap-status.json`. These files can be produced by Windows editors that emit legacy encodings. The reads now specify `encoding="utf-8", errors="replace"` at the I/O boundary. Added regression tests that plant a 0xFF byte in both `.env` and `bootstrap-status.json` and assert the callers no longer raise.

## AI Assistance

AI-assisted repository search and regression-test drafting. The author reviewed the implementation, selected the validation scope, and is responsible for the final diff.

## Release Lane

Stable hotfix reason:

```text
```

## Changed Surface

- [ ] Docs only
- [ ] Tests only
- [ ] Dashboard UI
- [ ] Installer
- [ ] Core runtime
- [ ] Frontend
- [x] Backend

## Validation

- `python -m py_compile extensions/services/dashboard-api/main.py extensions/services/dashboard-api/helpers.py` - passed.
- `cd extensions/services/dashboard-api && pytest tests/test_helpers.py -k "non_utf8"` - passed (2 passed).

### DIFF
--- a/ods/extensions/services/dashboard-api/helpers.py
+++ b/ods/extensions/services/dashboard-api/helpers.py
@@ -818,7 +818,7 @@ def get_model_info() -> Optional[ModelInfo]:
     if env_path.exists():
         try:
             env_values = {}
-            with open(env_path) as f:
+            with open(env_path, encoding="utf-8", errors="replace") as f:
                 for line in f:
                     if "=" not in line or line.lstrip().startswith("#"):
                         continue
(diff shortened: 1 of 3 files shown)


patil2001:
The why behind this change: fix(api): read user-data files as utf-8 with error tolerance. This is a contained fix for an edge case that would otherwise leave the stack in the wrong state on that path.

patil2001:
Adding context for reviewers: fix(api): read user-data files as utf-8 with error tolerance. This is a contained fix for an edge case that would otherwise leave the stack in the wrong state on that path.

PR #2345, opened 2026-08-04
fix(preflight): fail loudly when the log cannot be created

patil2001:
## Summary

`ods-preflight.sh` runs under `set -euo pipefail`, and its first write to the log (`echo "" > "$LOG_FILE"`) aborts the script with no explanation when the install dir isn't writable — e.g. a root-owned install run by a normal user. The redirect now fails loudly: the script verifies the log can be created first and prints a clear error to stderr before exiting 1, so the failure is diagnosed instead of silently dying.

## AI Assistance

## Release Lane

Stable hotfix reason:

```text
```

## Changed Surface

- [ ] Docs only
- [ ] Tests only
- [ ] Dashboard UI
- [x] Installer
- [x] Core runtime
- [ ] Frontend
- [ ] Backend

## Validation

- `bash -n ods/ods-preflight.sh` - passed.
- Redirect-failure harness (unwritable log path) - exited 1 with clear ERROR message.

### DIFF
--- a/ods/ods-preflight.sh
+++ b/ods/ods-preflight.sh
@@ -97,7 +97,13 @@ pass() { log "${GREEN}✓${NC} $1"; PASS=$((PASS+1)); }
 fail() { log "${RED}✗${NC} $1"; FAIL=$((FAIL+1)); }
 warn() { log "${YELLOW}⚠${NC} $1"; WARN=$((WARN+1)); }
 
-echo "" > "$LOG_FILE"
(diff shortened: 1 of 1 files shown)


patil2001:
Why this is necessary: fix(preflight): fail loudly when the log cannot be created. This is a contained fix for an edge case that would otherwise leave the stack in the wrong state on that path.

patil2001:
Motivation for this one: fix(preflight): fail loudly when the log cannot be created. This is a contained fix for an edge case that would otherwise leave the stack in the wrong state on that path.

PR #2346, opened 2026-08-04
fix(restore): reject archives whose root doesn't match the backup id

patil2001:
## Summary

`extract_backup` created `$BACKUP_ROOT/<id>` and then unpacked the archive with `tar xzf ... -C $BACKUP_ROOT`, trusting the archive's top-level directory name. A renamed or corrupt archive whose root differs from the requested backup id extracted into `$BACKUP_ROOT/<something-else>`, leaving the caller to restore from an empty directory it had just created — a silent no-op restore. Archives are created by `ods-backup.sh` as `tar czf <id>.tar.gz ... <id>`, so the first path component must equal the backup id. The function now verifies the archive's first entry matches the backup id before extracting and refuses otherwise.

## AI Assistance

## Release Lane

Stable hotfix reason:

```text
```

## Changed Surface

- [ ] Docs only
- [ ] Tests only
- [ ] Dashboard UI
- [x] Installer
- [x] Core runtime
- [ ] Frontend
- [ ] Backend

## Validation

- `bash -n ods/ods-restore.sh` - passed.
- Harness with root=evil archive requested as backup-123 - rejected with clear error, nothing extracted.
- Harness with matching archive - extracted correctly.

### DIFF
--- a/ods/ods-restore.sh
+++ b/ods/ods-restore.sh
@@ -224,6 +224,17 @@ extract_backup() {
             log_error "Backup archive contains unsafe paths (absolute or ../) — refusing to extract" >&2
             return 1
         fi
+        # Archives are created as "<backup_id>.tar.gz" wrapping a top-level
+        # directory named after the backup id (see ods-backup.sh). Refuse an
+        # archive whose root doesn't match, so a renamed or corrupt archive
+        # can't unpack elsewhere and leave the caller restoring from an empty
+        # directory.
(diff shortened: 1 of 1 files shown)


patil2001:
Adding context for reviewers: fix(restore): reject archives whose root doesn't match the backup id. This is a contained fix for an edge case that would otherwise leave the stack in the wrong state on that path.

patil2001:
Posting the rationale: fix(restore): reject archives whose root doesn't match the backup id. This is a contained fix for an edge case that would otherwise leave the stack in the wrong state on that path.

PR #2348, opened 2026-08-04
fix(migrate): stop copying DATA_DIR into itself during backup

patil2001:
## Summary

`cmd_backup` in migrate-config.sh did `cp -r "$DATA_DIR" "$backup_path/data/"` where `backup_path` lives under `DATA_DIR/backups`. GNU cp refuses with "cannot copy a directory into itself", the error was swallowed by `|| cp_exit=$?`, and the script still printed success — so the user-data portion of the backup silently never happened. The copy now iterates `DATA_DIR`'s top-level entries while excluding the `backups` dir, so the data tree is captured without copying the directory into itself.

## AI Assistance

## Release Lane

Stable hotfix reason:

```text
```

## Changed Surface

- [ ] Docs only
- [ ] Tests only
- [ ] Dashboard UI
- [x] Installer
- [x] Core runtime
- [ ] Frontend
- [ ] Backend

## Validation

- `bash -n ods/scripts/migrate-config.sh` - passed.
- Harness with real DATA_DIR tree - backup captured data/n8n and models, no "into itself" refusal.

### DIFF
--- a/ods/scripts/migrate-config.sh
+++ b/ods/scripts/migrate-config.sh
@@ -112,9 +112,15 @@ cmd_backup() {
     cp "${INSTALL_DIR}/docker-compose.yml" "$backup_path/" 2>&1 || cp_exit=$?
     cp -r "${INSTALL_DIR}/config" "$backup_path/" 2>&1 || cp_exit=$?
 
-    # Backup user data references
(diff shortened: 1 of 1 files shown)


patil2001:
Posting the rationale: fix(migrate): stop copying DATA_DIR into itself during backup. This is a contained fix for an edge case that would otherwise leave the stack in the wrong state on that path.

PR #2349, opened 2026-08-04
fix(install): drop dead env keys from offline mode setup

patil2001:
## Summary

Installer phase 09 appended `OFFLINE_MODE`, `DISABLE_TELEMETRY`, `DISABLE_UPDATE_CHECK`, `WEB_SEARCH_ENABLED` and `LOCAL_RAG_ENABLED` to `.env`. None of them exist in `.env.schema.json` and nothing in the codebase reads them (`OFFLINE_MODE` is an installer CLI flag set by `--offline`, not an env key), so every offline install left a `.env` that fails `ods env validate` with unknown-key errors (exit 2). This removes the dead block; offline mode is still marked by the `.offline-mode` marker file, the API-key blanking, and the OpenClaw M1 config.

## AI Assistance

## Release Lane

Stable hotfix reason:

```text
```

## Changed Surface

- [ ] Docs only
- [ ] Tests only
- [ ] Dashboard UI
- [x] Installer
- [x] Core runtime
- [ ] Frontend
- [ ] Backend

## Validation

- `bash -n ods/installers/phases/09-offline.sh` - passed.
- `ods/scripts/validate-env.sh` against a `.env` containing the removed keys - failed (exit 2, all 5 unknown) proving the bug; the phase no longer writes them.

### DIFF
--- a/ods/installers/phases/09-offline.sh
+++ b/ods/installers/phases/09-offline.sh
@@ -30,23 +30,6 @@ elif [[ "$OFFLINE_MODE" == "true" ]] && ! $DRY_RUN; then
     _sed_i 's/^ANTHROPIC_API_KEY=.*/ANTHROPIC_API_KEY=/' "$INSTALL_DIR/.env" 2>/dev/null || true
     _sed_i 's/^OPENAI_API_KEY=.*/OPENAI_API_KEY=/' "$INSTALL_DIR/.env" 2>/dev/null || true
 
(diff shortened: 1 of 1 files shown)


patil2001:
The why behind this change: fix(install): drop dead env keys from offline mode setup. This is a contained fix for an edge case that would otherwise leave the stack in the wrong state on that path.

Issue #2351, opened 2026-08-04
macOS native: post-install swap to Gemma 4 fails (llama.cpp pinned at install time) + stale catalog checksums brick the gemma4 install path


cemphlvn:
**TL;DR:** On macOS native (Metal), Gemma 4 cannot be enabled at all on an existing install — three separate bugs chain together. Each has a verified workaround; all three verified end-to-end on this machine (Gemma 4 E4B and E2B now serve correctly).

**Environment:** ODS CLI (macOS) 2.6.0 · Apple Silicon M4, 16GB · installed with default qwen profile → native llama-server pinned at `b8210` (`lib/constants.sh: LLAMA_CPP_RELEASE_TAG`)

---

### Bug 1 — post-install model swap can't handle newer architectures

Follow the README custom-GGUF flow: put `gemma-4-E4B-it-Q4_K_M.gguf` in `data/models/`, set `GGUF_FILE`/`LLM_MODEL` in `.env`, `./ods-macos.sh restart`.

```
llama_model_load: error loading model architecture: unknown model architecture: 'gemma4'
```

Server exits, stack comes up without inference, LiteLLM returns 500 on every request. No warning that the pinned binary can't serve the requested arch.

**Why:** the gemma4 profile's `LLAMA_CPP_RELEASE_TAG_OVERRIDE=b9014` (PR #855, `installers/macos/lib/tier-map.sh`) only applies at install time. Docker solves this per model via `llama_server_image` in `config/model-library.json`; the macOS native path has no equivalent (`llama_server_image: null`), and `ods-macos.sh update` only pulls Docker images.

**Fix idea:** store the required llama.cpp tag per model family (the tier-map override table already knows it) and check/refresh `bin/llama-server` + dylibs on swap — or at minimum reject the swap with "model needs llama.cpp ≥ b9014, installed is b8210".

### Bug 2 — stale catalog checksums brick the official gemma4 install path

The recommended route (`MODEL_PROFILE=gemma4 ./install-macos.sh --non-interactive`) hard-fails:

1. verifies the **valid** local E4B GGUF against catalog sha → mismatch → deletes it
2. re-downloads 5.3GB from the catalog URL (`resolve/main`)
3. same "mismatch" again → aborts with "file is corrupt. Re-run the installer" → infinite loop

The file isn't corrupt. The catalog pins a sha from an old revision while the URL points at `resolve/main`, and upstream (unsloth) updated the files:

[...]

PR #2352, opened 2026-08-04
fix(models): pin gemma4 E2B/E4B GGUF URLs to a revision and refresh checksums

cemphlvn:
## Summary

`gemma4-e2b-q4` and `gemma4-e4b-q4` in `ods/config/model-library.json` point `gguf_url` at `resolve/main` while pinning a `gguf_sha256` from an older upstream revision. unsloth has since updated both GGUFs, so every download now fails verification: the installer deletes the (valid) file as "corrupt", re-downloads 5.3GB, fails the same check, and aborts — an infinite loop that bricks the `MODEL_PROFILE=gemma4` install path (Bug 2 of #2351).

This PR pins both URLs to the current upstream revision — the pattern the Qwen entries already use — and refreshes `gguf_sha256` / `size_mb` to the values the Hugging Face API reports for those revisions (`GET /api/models/<repo>/tree/main?blobs=true` → `lfs.oid` / `size`).

## AI Assistance

Diagnosed, drafted, and validated with Claude Code; the human author reviewed the diff and the on-hardware validation below.

## Release Lane

(Reviewer may also want this on `release/2.6.x` — it fixes a current-stable user-facing install failure; happy to open a backport PR.)

## Changed Surface

## Risk And Validation

- Risk level: Medium (data-only catalog change; no code paths touched)
- Validation run:
  - [x] `git diff --check`
  - [x] Release-grade fleet or scoped hardware validation

Commands/results:

```text
python3 ods/tests/test-model-library-coverage.py   # exit 0
python3 ods/tests/test-model-library-verdicts.py   # exit 0
json.loads() on the edited file                    # valid

Scoped hardware validation (macOS M4 16GB, native Metal, llama.cpp b9014):
- sha256 of freshly downloaded E4B/E2B GGUFs matches the new pinned values
  (85a896a0… / 740185b2…), and matches the HF API lfs.oid for the pinned revisions
- with corrected checksums the gemma4 install path completes; both models
  load and serve correctly through LiteLLM (model swap smoke: chat completion OK)
```

## Operational Change Check

Note: 56 other catalog entries still use `resolve/main` URLs and are exposed to the same drift; left out of scope here since I could only verify the gemma4 pair end-to-end. #2351 tracks the broader issue.


cemphlvn:
A note on two questions a reviewer might reasonably ask:

**Why replace a readable `main` with a commit hash?**

The old entries asserted two contradictory things at once: `resolve/main` ("always give me the latest bytes" — a mutable pointer) and a fixed `gguf_sha256` ("the bytes must be exactly these" — an immutable expectation). That combination is guaranteed to break on any upstream update, and it did. Since the catalog already verifies checksums, it has already chosen immutability — pinning the URL just completes that choice. It's the same discipline as lockfiles / Docker `@sha256:` digests, and it's the pattern the Qwen entries in this same file already use (e.g. `resolve/f6d5376be…`). Human readability is carried by `name`/`gguf_file`; the URL's job is to be stable. Updating to a newer upstream release becomes a deliberate, reviewable diff (hash + sha + size together) instead of a silent drift.

**Is this drift machine-specific? No — it's verifiable without downloading anything, on any machine:**

```bash
# Current sha256 upstream serves for resolve/main (HF API, lfs.oid):
curl -s "https://huggingface.co/api/models/unsloth/gemma-4-E4B-it-GGUF/tree/main?blobs=true" \
  | jq -r '.[] | select(.path=="gemma-4-E4B-it-Q4_K_M.gguf") | .lfs.oid'
# → 85a896a047553e842f25297ee5b031d64ff30147d9c4af17b1e4b394cd1fab87  (catalog had dff0ffba…)

curl -s "https://huggingface.co/api/models/unsloth/gemma-4-E2B-it-GGUF/tree/main?blobs=true" \
  | jq -r '.[] | select(.path=="gemma-4-E2B-it-Q4_K_M.gguf") | .lfs.oid'
# → 740185b21d22ceb83a11c3aa62ad5842ef32c70f6096d756bbee85a1e4ec34b8  (catalog had ac0069eb…)
```

The upstream cause is visible in both repos' commit history: the tip commit on each is **"Added Gemma official chat template update"** (`0314792d` / `bfc15c38`) — the chat template is embedded in GGUF metadata, so the file bytes (and sha256) changed. The catalog's old hashes belong to a pre-update revision.

And the end-to-end check is deterministic on any OS: `curl -L <pinned url> | shasum -a 256` reproduces the new values exactly (verified here on macOS arm64; the hash math doesn't care about the platform).

Follow-up idea (out of scope here): a small CI test that walks the catalog and compares each `gguf_sha256` against the HF API `lfs.oid` for its URL would catch this class of drift for the remaining 56 `resolve/main` entries before users do.

cemphlvn:
Pushed a correction (62c9f8f): the first commit wrote `size_mb` in MiB; the catalog's convention is decimal MB (cf. `gemma3-4b-it-q4`: `size_bytes` 2489757856 / `size_mb` 2490). Now bytes/10^6 of the pinned revisions: E2B **3107**, E4B **4977**.

Also verified the untouched `quantization` fields against the actual GGUF headers of the pinned files (`general.file_type = 15` → Q4_K_M, `general.architecture = gemma4` for both), so the rest of each entry matches the pinned artifacts.

PR #2353, opened 2026-08-05
fix(remote-provider): log invalid SSH supervisor plans

### LINKED ISSUE #2338: SSH supervisor silently discards ValueError from _combined_ssh_argv leaving tunnel failures undiagnosable
### Description

When `_combined_ssh_argv(plan)` raises a `ValueError` (e.g. because the SSH tunnel plan has incomplete `argv`, a missing `-L` forward flag, or an invalid tunnel entry), the `SshProcessSupervisor.reconcile()` method catches the exception, stops the process, and transitions to an error state—but logs nothing. The original exception message is silently discarded, so operators debugging a broken remote-provider SSH tunnel have no log line explaining why the tunnel refused to start.

### Steps to Reproduce

1. Install ODS with a remote provider configured in SSH transport mode.
2. Corrupt or truncate the `routing-state.json` file so the tunnel plan produces an incomplete `argv` (e.g. remove the `-L` forward entry).
3. Observe the `remote-provider-ssh-tunnel` container logs:

```bash
docker logs remote-provider-ssh-tunnel --follow
```

4. Note that the tunnel silently transitions to `"error"` status with reason `"ssh_plan_unavailable"` but provides no detail about what went wrong.

### Expected Behavior

When `_combined_ssh_argv()` raises a `ValueError`, the supervisor logs a `WARNING` that includes the exception message, for example:

```text id="bx9f3n"
WARNING  remote-provider-ssh-tunnel: ssh plan argv invalid: SSH tunnel argv is missing a local forward
```

This gives operators an actionable log line to diagnose misconfigured routing state without needing to attach a debugger.

### Actual Behavior

The `ValueError` is caught and silently discarded:

```python id="ifd1hk"
try:
    argv = _combined_ssh_argv(plan)
except ValueError:          # <-- exc is thrown away, nothing logged
    self._stop_process_locked()
    ...
```

The container health endpoint returns:

```json id="7l4v2a"
{
  "status": "error",
  "reason": "ssh_plan_unavailable",
  "ready": false
}
```

However, nothing appears in the container logs to explain the root cause.

### Operating System

Any Linux distribution running Docker or Podman with rootless or rootful mode

### GPU

Not applicable — this affects the SSH tunnel supervisor service, not GPU inference

### Docker Version

### VRAM

Not applicable

### Logs

```shell

```

### Screenshots

_No response_

tang-vu:
## Summary
- log the validation error when an SSH supervisor plan cannot produce a safe argv
- keep the existing stopped/error health transition intact
- cover malformed local-forward plans with a regression contract

Closes #2338

## Test plan
- [x] `python ods/tests/contracts/test-remote-provider-ssh-tunnel-service.py`
- [x] `python -m compileall -q ods/extensions/services/remote-provider-ssh-tunnel/app ods/tests/contracts/test-remote-provider-ssh-tunnel-service.py`
- [x] `git diff --check`

Generated with Codex

### DIFF
--- a/ods/extensions/services/remote-provider-ssh-tunnel/app/main.py
+++ b/ods/extensions/services/remote-provider-ssh-tunnel/app/main.py
@@ -301,7 +301,8 @@ class SshProcessSupervisor:
             self._notice_exited_locked(now)
             try:
                 argv = _combined_ssh_argv(plan)
-            except ValueError:
+            except ValueError as exc:
+                LOGGER.warning("ssh plan argv invalid: %s", exc)
                 self._stop_process_locked()
                 self._last_payload = _health_from_plan(
                     _error_plan("ssh_plan_unavailable", secret_dir=self.secret_dir),
--- a/ods/tests/contracts/test-remote-provider-ssh-tunnel-service.py
+++ b/ods/tests/contracts/test-remote-provider-ssh-tunnel-service.py
@@ -10,6 +10,7 @@ import sys
 import tempfile
 from pathlib import Path
 from typing import Iterator
+from unittest.mock import ANY, patch
 
 
 ROOT = Path(__file__).resolve().parents[2]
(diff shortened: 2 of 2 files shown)


bokiko:
Reviewed head `3666fc5f` alongside #2339. The production fix is effectively the same one-line logging change in both PRs; the meaningful difference is regression-test placement.

This PR adds its assertion to `ods/tests/contracts/test-remote-provider-ssh-tunnel-service.py`, which is already executed by `scripts/release-gate.sh`. The contract suite passes at this head, and reverting the logging change makes the new assertion fail as expected. The test also verifies that the actionable `ValueError` detail reaches the warning while the health response remains `ssh_plan_unavailable`.

That makes #2353 the stronger of the two implementations: the regression remains covered by an existing CI/release path without introducing a separate test package. One useful piece from #2339 that could optionally be carried over is its negative case confirming that a disabled plan emits no warning.

The repeated warning on every reconcile cycle for a persistently invalid plan may merit separate throttling work, but it is not a blocker for restoring the missing diagnostic here.

PR #2354, opened 2026-08-05
fix(litellm): finish Token Spy client cleanup on cancellation

### LINKED ISSUE #2336: ODSTokenSpyCallback leaks httpx.AsyncClient connections on LiteLLM worker restart
### Description

`ODSTokenSpyCallback._run()` creates a new `httpx.AsyncClient` inside an `async with` block each time the background worker task is recreated. If the task is cancelled (e.g. during a LiteLLM model swap, gunicorn worker recycle, or an asyncio shutdown race), the `AsyncClient.__aexit__` may never be awaited, leaving TCP sockets and file descriptors open. A new client is created on the next inference event, compounding the leak with every restart cycle.

### Steps to Reproduce

1. Install ODS with Token Spy enabled (`TOKEN_SPY_URL` and `TOKEN_SPY_API_KEY` set).
2. Make the Token Spy endpoint temporarily unreachable to trigger `httpx.HTTPError`.
3. Simultaneously trigger a LiteLLM model swap or worker restart.
4. Repeat steps 2–3 several times under load.
5. Check the open file descriptors on the LiteLLM container:

```bash
ls -la /proc/<pid>/fd | wc -l
```

### Expected Behavior

The `httpx.AsyncClient` is properly closed when the background worker task exits or is cancelled, releasing all TCP connections and file descriptors.

### Actual Behavior

Each worker task restart leaves the previous `httpx.AsyncClient` unclosed. Open TCP connections and file descriptors accumulate over time. Under sustained load, this can exhaust the process file descriptor limit and result in:

```text
OSError: [Errno 24] Too many open files
```

### Operating System

Any (Linux / macOS / Windows — affects all platforms running LiteLLM with Token Spy)

### GPU

Not applicable — this is a software/network resource leak in the LiteLLM callback layer

### Docker Version

### VRAM

Not applicable

### Logs

```shell

```

### Screenshots

_No response_

tang-vu:
## Summary
- close the Token Spy HTTP client explicitly when its background worker exits
- shield socket cleanup so a repeated task cancellation cannot interrupt `aclose()`
- reproduce the double-cancellation race with a focused async regression test

Closes #2336

## Test plan
- [x] `python -m pytest ods/extensions/services/litellm/tests/test_token_spy_callback.py -q`
- [x] `python -m compileall -q ods/extensions/services/litellm`
- [x] `git diff --check`

Generated with Codex


bokiko:
Reviewed the cancellation behavior locally against head `2a19978d`.

Main already closes the `httpx.AsyncClient` after an ordinary single cancellation because `async with` runs `__aexit__`. The reproducible leak is narrower: a second cancellation arriving while `aclose()` is already in progress can interrupt cleanup.

This patch addresses that specific race without introducing shared client state. In the local cancellation harness, main leaked on the repeated-cancellation path, while this head completed cleanup after both the initial and second cancellation. The included regression test fails on the baseline and passes here; the focused tests and current GitHub checks are green.

This is the smaller and safer implementation compared with #2337, which moves client ownership onto the callback instance and depends on an `async_shutdown()` hook that does not appear to be wired into the current lifecycle.

Local evidence supports this patch. Because it touches the LiteLLM inference path, a live-proxy/FD-count exercise remains useful release validation rather than a blocker discovered in this diff. Credit to #2336/#2337 for surfacing the underlying cleanup problem.

PR #2355, opened 2026-08-05
fix(backup): make shared rsync copies additive

### LINKED ISSUE #2304: CRITICAL: rsync --delete destroys backups and live data
## CRITICAL: `--delete` makes full backups self-destructive and wipes live data on restore

`rsync_with_progress()` in `ods/lib/rsync.sh` unconditionally uses `--delete` on every rsync call, which breaks both backup and restore flows.

### Backup flow (`ods/ods-backup.sh`)

1. `backup_user_data()` (line 344) rsyncs each `data/<service>` directory into the **same** backup dir. Each `--delete` iteration deletes the previously backed-up service's directory, so only the last service path survives the loop.
2. `backup_config()` (line 367) then wipes the user-data backup with `--delete` while copying `config/`.
3. `backup_cache()` (line 413) wipes that again while copying `models/`.

Result of `do_backup full`: the backup contains only `models/` — even `manifest.json` is gone. Only the *last* rsync call's content survives.

### Restore flow (`ods/ods-restore.sh`)

`restore_user_data()` at line 413 restores `data/open-webui` with `--delete` against `$ODS_DIR/data/`, deleting the **live** n8n/qdrant/openclaw data directories. This directly contradicts the comment at lines 411-412:

```
# Note: Using -a without --delete to preserve any new files created after backup
```

### Files
- `ods/lib/rsync.sh:39` — `rsync -a --delete --info=progress2 ...`
- `ods/lib/rsync.sh:42` — fallback also uses `--delete`
- `ods/ods-backup.sh:344`, `367`, `413`
- `ods/ods-restore.sh:413`

### Suggested fix
`--delete` must not be baked into the shared helper. Add an opt-in flag (e.g. `rsync_with_progress "$src" "$dest" "label" "" "" "--delete"`) and only pass it where a true mirror-sync is intended, never for multi-target backups of the same dest or for restoring into a live tree.

tang-vu:
## CI follow-up

The first post-reconciliation run passed every repository test but the OpenSUSE matrix leg. That leg failed before checkout assertions because the Tumbleweed OSS mirror could not provide `jq` or `rsync` after three package-manager attempts; an upstream rerun requires repository admin rights. Commit `5173e3be` adds independently useful coverage for the previously unverified legacy rsync fallback and starts a fresh CI receipt without an empty commit.


bokiko:
Reviewed head `41ba3df8` and the current `rsync_with_progress()` call sites.

The patch is sound, with one important scope correction to issue #2304. Replaying the backup layout on disposable fixtures did not reproduce the claim that successive `--delete` calls erase sibling service directories or `manifest.json`. With a non-trailing-slash source, rsync prunes within the transferred subtree rather than deleting sibling paths. Consistently, the PR’s new backup-side assertions also pass on the unmodified baseline.

The narrower restore bug is genuine. `ods-restore.sh:413` uses the shared helper while the surrounding comment explicitly says restoration should preserve files created after the backup. The helper’s unconditional `--delete` violates that contract and removes those newer destination files. The added restore assertion is a real regression test: it fails on baseline and passes at this head.

All five current helper call sites are backup or restore copies with additive intent, so removing the global `--delete` is appropriate. The focused round-trip, backup-integrity, rollback-contract, syntax, ShellCheck, and current GitHub checks are green.

Not validated locally: a live Docker-backed restore, the older macOS rsync fallback path, and cross-filesystem behavior. Local evidence supports this patch while narrowing the rationale to the confirmed restore defect.

tang-vu:
﻿Compared with later overlapping PR #2492. This earlier PR remains canonical.

Why this version is stronger:
- it fixes the confirmed restore behavior directly and keeps the implementation KISS;
- the round-trip test exercises the real backup/restore path, not only an isolated rsync option;
- current restore call sites are additive, so an unused opt-in delete abstraction is not needed.

The original review also distinguished the unconfirmed sibling-deletion report from the real stale-file restore bug. No later change needs to be ported here.

tang-vu:
Backlog triage completed against later #2460. The later PR's independent advantages have now been ported into this earlier canonical PR in commit `1f52f108`: both backup suites are gated by `make test`, and the round-trip asserts `.env` retention in addition to the existing manifest/config/multi-service/model-cache checks. Focused round-trip, integrity, update-backup, syntax, and diff checks pass. The PR body now documents the inaccurate part of issue #2304 separately from the confirmed destructive restore defect.

tang-vu:
CI follow-up: the prior OpenSUSE failure was isolated to the Tumbleweed OSS mirror failing to provide `jq`/`rsync` after three attempts, before repository assertions. Upstream rerun is admin-only. Commit `5173e3be` now closes the remaining validation gap from the earlier review by forcing the legacy rsync fallback (`--info=progress2` unavailable) and proving it copies backup data without deleting a post-snapshot live file. The focused round-trip and syntax checks pass; a fresh full CI receipt is running.

PR #2356, opened 2026-08-05
fix(install): persist host IDs for Compose services

### LINKED ISSUE #2303: MEDIUM: Hermes/n8n UID/GID never resolve to host user
## MEDIUM: Hermes/n8n UID/GID interpolation never resolves to the host user

`${UID:-10000}` / `${GID:-10000}` in the Hermes and n8n compose files silently fall back to the defaults because `UID` and `GID` are **read-only shell variables** — they are never exported into the process environment, so Docker Compose (which interpolates from the *environment*, not shell state) always sees an empty value.

### Effects
- `ods/extensions/services/hermes/compose.yaml:18-19` — `HERMES_UID=${UID:-10000}`, `HERMES_GID=${GID:-10000}`. Files in `./data/hermes` end up owned by uid 10000, requiring `sudo` to manage, even on a normal single-user machine.
- `ods/extensions/services/n8n/compose.yaml:6` — `user: "${UID:-1000}:${GID:-1000}"`. n8n data lands on uid/gid 1000 regardless of the real host user, breaking ownership expectations.

### Note
The repo's own `model-lifecycle-lock.sh:32` demonstrates the correct pattern:
```sh
... ${UID:-$(id -u)} ...
```
`$(id -u)` / `$(id -g)` run inside a `$()` command substitution and produce a real value at compose-interpolation... however for Compose this still needs to come from an exported var.

### Files
- `ods/extensions/services/hermes/compose.yaml:18-19`
- `ods/extensions/services/n8n/compose.yaml:6`
- correct pattern reference: `ods/installers/lib/model-lifecycle-lock.sh:32`

### Suggested fix
Replace the read-only shell vars with values that are actually exported, e.g. set them in the installer/bootstrap:
```sh
export ODS_UID="${ODS_UID:-$(id -u)}"
export ODS_GID="${ODS_GID:-$(id -g)}"
```
and use `${ODS_UID:-10000}` / `${ODS_GID:-10000}` (and `${ODS_UID:-1000}` / `${ODS_GID:-1000}` for n8n) in the compose files — keeping explicit defaults but making the real host user the effective value.

tang-vu:
## Why this matters

Docker Compose interpolates service user IDs from `.env`, but ODS did not persist the installing user's IDs. Hermes, n8n, and Privacy Shield therefore fell back to fixed container IDs, so bind-mounted files could become unwritable to the operator and ownership repair could target identities different from the containers.

The first implementation used `UID`/`GID`; triage found that those names collide with Bash's readonly `UID` when ODS sources `.env`. The canonical fix now uses `ODS_UID`/`ODS_GID` and migrates legacy values.

Closes #2303.

## Root cause and invariant

Root cause: Compose needs explicit exported/persisted IDs, while Bash's special `UID` is not a safe dotenv key.

Invariant: every generated or upgraded install gives Compose and ownership repair the same numeric host identity, preserves an operator override, remains sourceable as Bash dotenv, and keeps legacy `UID`/`GID` installs compatible until migration.

## What changed

- Generate and schema-document `ODS_UID`/`ODS_GID` on Linux and macOS.
- Migrate existing `UID`/`GID` values without writing the readonly Bash key back.
- Use the new keys consistently in Hermes, n8n, Privacy Shield, and both extension compose templates.
- Make rootless ownership prefer the new keys while retaining legacy fallback.
- Align rootful Hermes ownership with the exact persisted IDs and current fail-closed sudo contract.
- Cover generated dotenv, rootless repair, legacy migration, macOS reruns, and Compose rendering.

## Overlap check

Compared open/closed PRs and all open PRs touching the production files. #2885, #2671, closed #2491, and omnibus #2305 address the same UID/GID symptom. This earlier PR remains canonical because it covers Linux and macOS generation, rootful and rootless consumers, all shipped Compose users, legacy migration, schema/templates, and boundary tests. The useful non-readonly-key idea from #2885 was ported here after the original `UID=` implementation failed the sourceable-dotenv smoke. Merged #2221's stricter ownership/sudo behavior is preserved rather than replaced.

## Focused validation

- `bash tests/test-hermes-data-ownership.sh` — passed.
- `bash tests/test-rootless-docker-ownership.sh` — 28 passed, including new and legacy keys.
- `bash tests/test-podman-rootless-contracts.sh` — 24 passed after aligning the rootful ownership assertion with persisted IDs.
- `bash tests/test-safe-env.sh` — 13 passed.
- `bash tests/test-macos-installer-transitions.sh` — 7 passed.
- `bash tests/test-macos-env-quote-strip.sh` — 9 passed.
- `su -s /bin/bash nobody -c 'bash tests/test-installer-noninteractive-sudo.sh'` — 5 passed.
- Combined n8n/Hermes/Privacy Shield `docker compose ... config --no-interpolate` — passed.
- Bash syntax, ShellCheck at error severity, schema parse, and `git diff --check` — passed.
- Installer environment smoke proved generated `.env` sources cleanly, validates against the schema, preserves IDs on rerun, and renders Compose. The suite still ends nonzero on the unrelated `ai_err` undefined-function check already present on `upstream/main` in phase 07; this PR does not claim that baseline failure is fixed.
- GitHub product CI at `4ff4469b` passed: integration smoke, Linux/macOS and 10-distro smoke, API/frontend, Compose, all env tiers, shell/Python/PowerShell lint, and secret scan. The separate automated review workflow was cancelled and requires upstream admin rights to rerun.

## Tradeoffs and rollback

The legacy `UID`/`GID` schema entries remain accepted so existing installs and direct rootless repairs do not break, but newly generated files use only `ODS_UID`/`ODS_GID`. A normal revert restores fixed Compose defaults and the original ownership mismatch; no data format beyond dotenv keys is changed. Runtime container activation was not performed locally, so independent review remains appropriate for the ownership migration.


tang-vu:
﻿Compared with later overlapping PR #2491. This earlier PR remains canonical.

Why this version is stronger:
- it covers both Linux and macOS ownership generation;
- it uses the existing UID/GID contract already declared in .env.schema.json and .env.example;
- Docker Compose consumes the generated .env directly, so introducing parallel ODS_UID/ODS_GID keys would duplicate configuration;
- it includes installer smoke and Hermes ownership regression coverage.

No useful later change needs to be ported.

tang-vu:
Conflict and overlap triage receipt (2026-08-24)

New head 4a902a2 is conflict-free with current main. During validation, the old UID=/GID= approach proved unsafe because Bash treats UID as readonly when sourcing dotenv. The branch now ports the stronger non-readonly naming idea from #2885 into this earlier canonical PR as ODS_UID/ODS_GID, while retaining broader Linux, macOS, rootless, legacy-migration, schema, Compose-user, and template coverage. The PR body now records the comparison and full validation.

Focused results: rootless ownership 28/28, safe-env 13/13, macOS transitions 7/7, macOS quote parsing 9/9, noninteractive sudo 5/5, Compose render/syntax/ShellCheck/diff checks passed. Installer smoke now sources and schema-validates the generated env cleanly; only upstream main's unrelated undefined i_err check remains red.

tang-vu:
CI follow-up (2026-08-24)

The first refreshed run exposed one stale static assertion in the broader installer contract: it still required hardcoded Hermes ownership 10000:10000. Commit `4ff4469b` updates that boundary to require the persisted `_phase06_compose_uid:_phase06_compose_gid` values and the current `ods_sudo` mutation path.

- `bash tests/test-podman-rootless-contracts.sh`: 24 passed
- Bash syntax, ShellCheck at error severity, and `git diff --check`: passed

No production code changed in this follow-up; CI is rerunning on the corrected contract.

tang-vu:
Final CI receipt for `4ff4469b`: all product gates passed, including integration smoke, Linux/macOS and all 10 distro jobs, API/frontend, Compose validation, all env tiers, shell/Python/PowerShell lint, and secret scan. The three non-success entries are only cancelled jobs in the separate automated-review workflow (`Basic Review`, `Detect High-Stakes Changes`, `Review Summary`); GitHub rejected the attempted rerun because it requires upstream admin rights.

PR #2358, opened 2026-08-05
fix(bootstrap): guard recursive install cleanup

### LINKED ISSUE #2301: HIGH: get-ods.sh runs rm -rf on unvalidated install path
## HIGH: `get-ods.sh` runs `rm -rf` on an unvalidated install path

`ODS_INSTALL_DIR` is taken verbatim from the environment (`ods/get-ods.sh:33`) and later passed straight to `remove_install_dir()`, which executes `rm -rf -- "$target_dir"` (line 71) — with no sanity check that the target isn't `/`, `$HOME`, or a home subdirectory.

If the target dir already exists but has no `.env` file it is treated as an "incomplete install", and the script either:

- line 379 — removes it automatically with `--force`, or
- line 387 — removes it after a `[y/N]` prompt.

A typo such as `ODS_INSTALL_DIR=~` (or `/`, `/home`) + answering yes → `rm -rf` of the entire directory tree, and with `sudo` retry at line 77 it can even remove root-owned files it otherwise couldn't.

### Files
- `ods/get-ods.sh:33` — `INSTALL_DIR="${ODS_INSTALL_DIR:-...}"` (unvalidated)
- `ods/get-ods.sh:68-81` — `remove_install_dir()` with `rm -rf -- "$target_dir"` and `sudo -n rm -rf -- "$target_dir"`
- `ods/get-ods.sh:379`, `387` — call sites (auto `--force` and prompted removal)

### Suggested fix
Before removal, assert the resolved path:
- is not `/` or `/root`/`$HOME`
- is not `$HOME` itself or a direct subdir of it (unless it's the actual ODS dir)
- matches an expected basename (e.g. `ods`/`*ods*`) or at least contains a marker like a `.ods`/`.env`-family file before considering it an "incomplete install"

Never call `rm -rf` on a path that isn't known to be the ODS install directory.

tang-vu:
## Summary
- validate and canonicalize an incomplete install path before either normal or sudo recursive cleanup
- reject the filesystem root, top-level system directories, the bootstrap home, symlinks, and unrecognized custom directories
- allow normal ODS-named targets and custom paths carrying an ODS checkout marker pair

Closes #2301

## Test plan
- [x] `bash -n ods/get-ods.sh ods/tests/contracts/test-installer-hardening.sh`
- [x] `bash ods/tests/contracts/test-installer-hardening.sh`
- [x] `git diff --check`

Generated with Codex

### DIFF
--- a/ods/get-ods.sh
+++ b/ods/get-ods.sh
@@ -84,7 +84,7 @@ install_dir_removal_is_safe() {
     [[ "${canonical_target#/}" == */* ]] || return 1
 
     case "$target_name" in
-        ods|ods-*|*-ods|dream-server|dreamserver) return 0 ;;
+        ods|ods-*|*-ods) return 0 ;;
     esac
     [[ -f "$target_dir/ods-cli" && -f "$target_dir/install-core.sh" ]] \
         || [[ -f "$target_dir/docker-compose.base.yml" && -d "$target_dir/installers" ]]
--- a/ods/get-ods.sh
+++ b/ods/get-ods.sh
@@ -65,8 +65,38 @@ format_git_clone_error() {
 }
 
 
+install_dir_removal_is_safe() {
+    local target_dir="${1:-}"
+    local target_parent target_name canonical_parent canonical_target canonical_home
+
+    [[ -n "$target_dir" && -d "$target_dir" && ! -L "$target_dir" ]] || return 1
+    target_parent=$(dirname -- "${target_dir%/}")
+    target_name=$(basename -- "${target_dir%/}")
+    [[ "$target_name" != "." && "$target_name" != ".." ]] || return 1
+    canonical_parent=$(cd -P -- "$target_parent" 2>/dev/null && pwd) || return 1
+    canonical_target="${canonical_parent%/}/$target_name"
+    [[ "$canonical_target" == /* ]] || return 1
+
+    canonical_home=$(cd -P -- "$ODS_BOOTSTRAP_ROOT" 2>/dev/null && pwd) || return 1
+    [[ "$canonical_target" != "$canonical_home" ]] || return 1
+    # Never delete the filesystem root or one of its direct children (such as
+    # /home, /Users, /Volumes, /mnt, or /opt), regardless of its name.
+    [[ "${canonical_target#/}" == */* ]] || return 1
+
+    case "$target_name" in
+        ods|ods-*|*-ods|dream-server|dreamserver) return 0 ;;
+    esac
+    [[ -f "$target_dir/ods-cli" && -f "$target_dir/install-core.sh" ]] \
+        || [[ -f "$target_dir/docker-compose.base.yml" && -d "$target_dir/installers" ]]
+}
+
 remove_install_dir() {
     local target_dir="$1"
+
+    if ! install_dir_removal_is_safe "$target_dir"; then
+        warn "Refusing to recursively remove an unrecognized or protected install path: $target_dir"
+        return 2
+    fi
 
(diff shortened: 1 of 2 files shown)


bokiko:
Reviewed head 3b332ca0 at the function level. The refusal side of install_dir_removal_is_safe() is strong: all 26 tested refusal cases were rejected, including /, direct children of /, $HOME, traversal forms, symlinks, parent-symlink escapes, empty/nonexistent paths, and unrecognized directories.

There is one remaining allow-side gap I believe should be fixed before merge. The basename-only branch at get-ods.sh:87-89:

case "$target_name" in
    ods|ods-*|*-ods) return 0 ;;
esac

accepts the final path component without an ODS marker or location constraint. Unrelated directories such as $HOME/ods-photos, /data/ods-backup, or /tmp/client-ods therefore pass the guard and remain eligible for recursive deletion under --force.

The broad match does not appear necessary for preserving the normal incomplete-install flow. I suggest allowing only the exact canonical default path, $canonical_home/ods, without markers and requiring one of the existing recognized marker pairs for every other custom path.

Useful contract cases would be:

- exact $HOME/ods without markers: allowed;
- $HOME/ods-photos without markers: refused;
- /tmp/client-ods without markers: refused;
- custom path with a recognized marker pair: allowed.

The remaining parts of the patch—protected-path checks, canonical-parent handling, fail-closed behavior, and improved error text—look sound. I would narrow this basename exception before merging.

tang-vu:
﻿Addressed the remaining allow-side review gap in the latest commit:

- only the exact canonical `$ODS_BOOTSTRAP_ROOT/ods` default may be incomplete without markers
- removed basename-only acceptance for `ods-*` and `*-ods`
- custom paths now require one of the existing ODS marker pairs
- added refusal coverage for unrelated `ods-photos` and `client-ods` directories

The installer-hardening contract passed the updated removal-guard section and continued through subsequent contracts; the local full script later stopped on an unrelated missing host `nvidia-ctk` dependency.

I also compared later #2489: its marker requirement is useful, but its strict `$HOME` containment would reject supported custom install roots. This PR now keeps custom-root support while closing the basename deletion gap.

bokiko:
Re-reviewed at the new head, `4719dc22`. This addresses the allow-side gap from my earlier comment precisely: the basename-only exception is gone, the marker-free allowance is now limited to the exact canonical default path (`$canonical_home/ods`), and every other path requires a recognized marker pair.

I verified the change independently against both the previous and current heads using the same function-level fixtures. All previously tested protected, traversal, symlink, unrecognized-path, exact-default, and marked custom-root cases retain their expected behavior. The four targeted unmarked cases—`ods-photos`, an `ods-backup`-style path, `client-ods`, and another custom `*-ods` path—incorrectly pass on the prior head and correctly refuse on this one. The additions to `test-installer-hardening.sh` therefore exercise a genuine regression.

I also compared #2489. Its strict `$HOME` containment would reject valid marked custom install roots outside `$HOME`, and it adds no automated regression test. This PR preserves custom-root support while closing the unsafe basename exception, so I prefer this approach.

Local evidence: `bash -n` passes, ShellCheck is clean on the changed lines, and `test-installer-hardening.sh` passes with the new assertions. All 26 substantive CI checks are successful; the cancelled jobs belong to an internal automated-review workflow that did not run, not a build or test failure.

This now looks ready to merge from my side.

tang-vu:
Triage receipt: every product validation job on the current head is green. The only non-success entries are cancelled jobs in the separate automated-review workflow (`Basic Review`, `Detect High-Stakes Changes`, `Review Summary`). I attempted a failed-job rerun, but GitHub requires upstream admin rights; no code-only no-op commit was added to manufacture a rerun.

PR #2360, opened 2026-08-05
fix(backup): recognize multi-segment backup IDs

### LINKED ISSUE #2299: collect_backups regex rejects multi-segment backup IDs
## File
`ods/ods-backup.sh` (line 151)

## Problem
The regex `^([A-Za-z0-9_]+-)?[0-9]{8}-[0-9]{6}(\.tar\.gz)?$` allows at most one hyphenated prefix segment. The host-agents `BACKUP_ID_RE` (`ods-host-agent.py:115`) permits `[A-Za-z0-9_-]{0,63}`.

A user-named backup like `dashboard-my-name-20260715-143022` lands in `BACKUP_ROOT` but is invisible to `list_backups`/`apply_retention` (though still deletable). The two regexes are inconsistent.

## Suggested fix
Align `collect_backups` regex with the host-agents `BACKUP_ID_RE` so multi-segment user-named backups are recognized and retained.

tang-vu:
## Summary
- accept the bounded multi-segment labels produced by update/dashboard backups
- retain the strict timestamp suffix and prefix length cap used to protect retention cleanup
- prove named backups appear in listings while unrelated directories remain excluded

Closes #2299

## Test plan
- [x] `bash ods/tests/test-backup-integrity.sh`
- [x] `bash -n ods/ods-backup.sh ods/tests/test-backup-integrity.sh`
- [x] `git diff --check`

Generated with Codex

### DIFF
--- a/ods/ods-backup.sh
+++ b/ods/ods-backup.sh
@@ -139,7 +139,9 @@ ensure_backup_space() {
 # the global COLLECTED_BACKUPS array, newest first.
 #
 # Backup IDs are YYYYMMDD-HHMMSS (this script) or <prefix>-YYYYMMDD-HHMMSS
-# (dashboard/host-agent). Match that shape explicitly: retention deletes
+# (update/dashboard/host-agent). The host-agent permits a 64-character label;
+# ods-update prepends "backup-", so allow that bounded multi-segment prefix.
+# Match the timestamp shape explicitly: retention deletes
 # whatever this collects, so an operator's unrelated files in the backup
 # directory must never qualify.
 COLLECTED_BACKUPS=()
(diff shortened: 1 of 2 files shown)


tang-vu:
﻿Compared with later overlapping PR #2487. This earlier PR remains canonical.

Why this version is stronger:
- it preserves a bounded prefix aligned with the host-agent's 63-character label plus the update backup prefix;
- it keeps the required timestamp anchor, protecting retention from unrelated operator files;
- its regression runs through the real backup listing path rather than extracting a regex literal from source;
- #2487's repeated-segment pattern is unbounded.

The later PR adds more table cases, but no missing behavior that outweighs the bounded production-path test here.

PR #2361, opened 2026-08-05
fix(gpu): fail closed on missing NVIDIA assignment

### LINKED ISSUE #2298: multigpu-nvidia defaults NVIDIA_VISIBLE_DEVICES to all GPUs
## File
`ods/docker-compose.multigpu-nvidia.yml` (lines 4, 12)

## Problem
`NVIDIA_VISIBLE_DEVICES: "${LLAMA_SERVER_GPU_UUIDS:-all}"` — when the var is unset (manual compose run, or the transient window during `ods gpu-reassign` which explicitly unsets it), llama-server gets EVERY GPU on the host, ignoring `GPU_ASSIGNMENT_JSON_B64` and per-service GPU isolation.

## Suggested fix
Fail closed with `${LLAMA_SERVER_GPU_UUIDS:?}` or derive the value from `GPU_ASSIGNMENT_JSON_B64` so an unset var never grants all GPUs.

tang-vu:
## Summary
- require an explicit llama-server UUID list whenever the NVIDIA multi-GPU overlay is active
- fail installer assignment immediately instead of warning that every GPU will become visible
- add a Compose-level regression contract for both missing and explicit assignments

Closes #2298

## Test plan
- [x] `bash ods/tests/test-multigpu-nvidia-visibility.sh`
- [x] `bash ods/tests/test-gpu-reassign-manual.sh`
- [x] `bash -n ods/installers/phases/03-features.sh ods/tests/test-multigpu-nvidia-visibility.sh`
- [x] `git diff --check`

Generated with Codex


tang-vu:
﻿Compared with later overlapping PR #2486. This earlier PR remains canonical.

Why this version is stronger:
- it updates the Compose default and the installer-generated default consistently;
- it wires the validation into CI/Makefile;
- it includes a focused regression test.

The later proposal contains only the one-line Compose change, so there is no additional behavior to port.

PR #2362, opened 2026-08-05
fix(preset): preserve complete env values in diff

### LINKED ISSUE #2297: cmd_preset diff env parsing truncates values at first =
## File
`ods/ods-cli` (lines 3373-3382)

## Problem
`while IFS== read -r key value` splits on ALL `=` characters. Values containing `=` (JWT/base64 secrets, URLs with query strings) are truncated. Both sides truncate identically, so real differences after the first `=` are reported as "no difference" (false negative). It also doesnt strip CR (unlike `lib/safe-env.sh`) or quotes.

## Suggested fix
Split only on the first `=`, e.g. `key=${line%%=*}; value=${line#*=}`, and strip CR/quotes before comparing.

tang-vu:
## Summary
- split preset env entries on the first equals sign so URLs, JWTs, and base64 values stay intact
- normalize trailing CR and one matching quote pair before comparison
- share parsing between both preset files without `eval` or lossy delimiters

Closes #2297

## Test plan
- [x] `bash ods/tests/test-preset-diff.sh` (8 passed)
- [x] `bash -n ods/ods-cli ods/tests/test-preset-diff.sh`
- [x] `git diff --check`

Generated with Codex

### DIFF
--- a/ods/ods-cli
+++ b/ods/ods-cli
@@ -2219,6 +2219,28 @@ _cmd_config_is_secret() {
     return 1
 }
 
+_cmd_preset_env_entries() {
+    local env_file="$1" line key value
+    while IFS= read -r line || [[ -n "$line" ]]; do
+        line="${line%$'\r'}"
+        [[ "$line" == *=* ]] || continue
+        key="${line%%=*}"
+        value="${line#*=}"
+        key="${key#"${key%%[![:space:]]*}"}"
+        key="${key%"${key##*[![:space:]]}"}"
+        [[ -z "$key" || "$key" =~ ^[[:space:]]*# ]] && continue
+        value="${value# }"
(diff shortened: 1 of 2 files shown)


tang-vu:
﻿Compared with later overlapping PR #2485. This earlier PR remains canonical.

Why this version is stronger:
- parsing is centralized in one helper instead of duplicated for both files;
- it splits only on the first equals sign, normalizes CRLF and matching quotes, trims keys, and skips invalid/comment entries;
- NUL-delimited key/value output avoids lossy delimiters and eval;
- the regression asserts the complete values after embedded equals signs, not only that a key appeared.

The later PR has more separately named cases, but this PR covers the same semantics with a smaller, reusable implementation.

PR #2372, opened 2026-08-05
fix(models): honor installable-only fallback

tang-vu:
## Summary
- stop the model selector from falling back to metadata-only or un-installable catalog entries
- return a clear error when `--installable-only` has no eligible model
- add a regression case to the tier-map selector suite

## Test plan
- `bash tests/test-tier-map.sh` (129 passed)
- `python -m compileall -q scripts/select-model.py`
- `git diff --check`


tang-vu:

Merged current upstream/main and resolved the tier-map suite conflict by retaining the installable-only empty-pool regression plus upstream's Gemma 4 artifact-pin check.

- New head:

PR #2394, opened 2026-08-05
fix(uninstall): remove the OpenCode config.json compat file

patil2001:
## Summary

OpenCode reads `~/.config/opencode/config.json` (the compat copy every installer writes alongside `opencode.json`), but `ods-uninstall.sh` only deleted `opencode.json`. After an uninstall the compat file survived with a stale `llama-server` baseURL pointing at the deleted install, so standalone OpenCode kept routing to a dead endpoint.

This change removes both files and broadens the ownership grep to cover the active `config.json`.

## AI Assistance

AI-assisted edge-case enumeration and regression-test drafting. The author reviewed the implementation and is responsible for the final diff.

## Release Lane

## Changed Surface

- [x] Core runtime

## Validation

- `bash -n ods/ods-uninstall.sh` - passed.

### DIFF
--- a/ods/ods-uninstall.sh
+++ b/ods/ods-uninstall.sh
@@ -377,12 +377,20 @@ if [[ -d "$HOME/.ods" ]]; then
     log_ok "Backups removed"
 fi
 
-# 7. Remove OpenCode config (if we created it)
-OPENCODE_CONFIG="$HOME/.config/opencode/opencode.json"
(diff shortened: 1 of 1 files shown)


patil2001:
Why this is necessary: fix(uninstall): remove the OpenCode config.json compat file. This is a contained fix for an edge case that would otherwise leave the stack in the wrong state on that path.

PR #2395, opened 2026-08-05
fix(macos): run OpenCode LaunchAgent with serve, not web

patil2001:
## Summary

The Linux `opencode-web.service` user unit was moved to the headless `serve` subcommand (f7395384) so boot/install never launches a browser or TUI, but the macOS LaunchAgent still ran the legacy `web` subcommand. Because the plist uses `RunAtLoad` + `KeepAlive`, every login spawns `opencode web`, opening a terminal/browser even on headless installs, and diverges from the Linux unit.

This switches the plist `ProgramArguments` and the manual-start hint to `serve`.

## AI Assistance

AI-assisted cross-platform edge-case enumeration. The author reviewed the implementation and is responsible for the final diff.

## Release Lane

## Changed Surface

- [x] Installer (macOS)

## Validation

- `bash -n ods/installers/macos/install-macos.sh` - passed.

### DIFF
--- a/ods/installers/macos/install-macos.sh
+++ b/ods/installers/macos/install-macos.sh
@@ -2873,7 +2873,7 @@ for service in (data.get("services") or {}).values():
     <key>ProgramArguments</key>
     <array>
         <string>${OPENCODE_BIN}</string>
-        <string>web</string>
+        <string>serve</string>
         <string>--port</string>
         <string>3003</string>
         <string>--hostname</string>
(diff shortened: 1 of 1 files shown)


patil2001:
Adding context for reviewers: fix(macos): run OpenCode LaunchAgent with serve, not web. This is a contained fix for an edge case that would otherwise leave the stack in the wrong state on that path.

PR #2397, opened 2026-08-05
fix(linux): honor OPENCODE_PORT in the OpenCode systemd unit

patil2001:
## Summary

`OPENCODE_PORT` is a documented, schema-validated env var (`ods/.env.schema.json`, default 3003) that `ports.json` declares via `env_var: OPENCODE_PORT`, and it is honored on Windows (`installers/windows`). On Linux it was silently ignored: `ods/opencode/opencode-web.service` hardcoded `--port 3003` and the phase that renders the unit (`ods/installers/phases/07-devtools.sh`) substituted `__HOME__`, `__OPENCODE_BIN__`, and `__OPENCODE_BIN_DIR__` but had no port placeholder. A configured `OPENCODE_PORT` therefore never reached the service, which bound 3003 while the dashboard/health reported the configured port.

## AI Assistance

AI-assisted repository search and edge-case enumeration were used to trace where the Linux unit is rendered, cross-check the Windows handling of `OPENCODE_PORT`, and design the placeholder-render guard pattern to mirror the host-agent block.

## Release Lane

- [x] Mainline

## Changed Surface

- [x] Core runtime

## Validation

```
bash -n ods/installers/phases/07-devtools.sh
bash -n ods/opencode/opencode-web.service
git diff --check
```


patil2001:
Posting the rationale: fix(linux): honor OPENCODE_PORT in the OpenCode systemd unit. This is a contained fix for an edge case that would otherwise leave the stack in the wrong state on that path.

PR #2398, opened 2026-08-05
fix(macos): honor OPENCODE_PORT in the OpenCode LaunchAgent

patil2001:
## Summary
OPENCODE_PORT (default 3003) is honored on Windows but silently ignored by the macOS OpenCode LaunchAgent: the install-macos.sh plist heredoc hardcoded `--port 3003`. This change reads OPENCODE_PORT from `.env` via the existing `read_env_value` helper (with numeric validation, falling back to 3003) and interpolates it into the plist plus the status/manual-start messages.

## AI Assistance
This change was implemented with the assistance of an AI coding agent (opencode). The agent localized the OpenCode block, applied the port read/interpolation, and verified with `bash -n` and `git diff --check`.

## Release Lane
- [x] Mainline

## Changed Surface
- [x] Installer

## Validation
- `bash -n ods/installers/macos/install-macos.sh` passes
- `git diff --check` passes

### DIFF
--- a/ods/installers/macos/install-macos.sh
+++ b/ods/installers/macos/install-macos.sh
@@ -2854,7 +2854,7 @@ for service in (data.get("services") or {}).values():
         ai_ok "OpenCode configured for ${_opencode_model} at ${_opencode_base_url}"
         unset _opencode_model _opencode_port _opencode_bind _opencode_host \
             _opencode_switchboard_mode \
-            _opencode_base_url _opencode_api_key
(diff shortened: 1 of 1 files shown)


patil2001:
The why behind this change: fix(macos): honor OPENCODE_PORT in the OpenCode LaunchAgent. This is a contained fix for an edge case that would otherwise leave the stack in the wrong state on that path.

PR #2399, opened 2026-08-05
fix(installer): probe OpenCode health on the configured port

patil2001:
## Summary

`ods/installers/phases/12-health.sh` hardcoded `127.0.0.1:3003` for the OpenCode health probe even though `OPENCODE_PORT` is a documented env var (default 3003) that is honored on Windows. The probe now resolves the port from `.env` with numeric validation so the health check tracks the actual binding.

## AI Assistance

AI-assisted repository search and edge-case enumeration. The author reviewed the implementation and is responsible for the final diff.

## Release Lane

## Changed Surface

- [x] Installer

## Validation

- `bash -n ods/installers/phases/12-health.sh` - passed.


patil2001:
Why this is necessary: fix(installer): probe OpenCode health on the configured port. This is a contained fix for an edge case that would otherwise leave the stack in the wrong state on that path.

PR #2400, opened 2026-08-05
fix(installer): print the configured OpenCode port in summary

patil2001:
## Summary
- `13-summary.sh` printed OpenCode's URL at hardcoded port `3003` in both the service list and the access-card.
- `OPENCODE_PORT` is documented in `.env` (default `3003`), so the printed URL was wrong whenever a user configured a different port.
- Resolve `OPENCODE_WEB_PORT` once from `$INSTALL_DIR/.env` (falling back to `3003` when unset/non-numeric) and interpolate it into both lines.
- The `systemctl --user is-active opencode-web` gate is unchanged.

## AI Assistance
This change was implemented with assistance from an AI coding agent.

## Release Lane
- [x] Mainline
- [ ] Other

## Changed Surface
- [ ] Core
- [x] Installer
- [ ] CLI
- [ ] Dashboard
- [ ] Dashboard API
- [ ] Extensions / Services
- [ ] Docs
- [ ] CI

## Validation
- `bash -n ods/installers/phases/13-summary.sh` — passes
- `git diff --check` — clean

### DIFF
--- a/ods/installers/phases/13-summary.sh
+++ b/ods/installers/phases/13-summary.sh
@@ -120,6 +120,9 @@ fi
 bootline
 echo -e "${BGRN}ALL SERVICES${NC}"
 bootline
+# Resolve OpenCode's configured port once (OPENCODE_PORT is documented in .env)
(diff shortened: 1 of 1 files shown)


patil2001:
Adding context for reviewers: fix(installer): print the configured OpenCode port in summary. This is a contained fix for an edge case that would otherwise leave the stack in the wrong state on that path.

PR #2401, opened 2026-08-05
fix(dashboard-api): honor OPENCODE_PORT for the OpenCode host service

patil2001:
## Summary
- The dashboard host-systemd probe and the coding-feature link resolved OpenCode's external port from the static manifest (`external_port`, default 3003).
- `OPENCODE_PORT` is a documented env var honored on Windows, but when it differed from 3003 the dashprobe checked port 3003 and reported `not_deployed`, and the "Open OpenCode" link pointed at the wrong port.
- Added `_effective_external_port` (helpers.py) and `_svc_external_port` (routers/features.py) to resolve the live external port for the `opencode` service from `OPENCODE_PORT`, matching existing manifest semantics (`external_port_env: OPENCODE_PORT`).
- Health probing keeps its fail-closed behavior (port <= 0 → `not_deployed`).

## AI Assistance
This change was implemented with the assistance of an AI coding agent (opencode). The agent read the relevant manifests and router/helper code, then applied a minimal, scoped fix. The resulting diff is small (23 insertions, 3 deletions) and confined to the dashboard-api backend; no frontend was touched.

## Release Lane
- [x] Mainline

## Changed Surface
- [ ] Dashboard UI
- [x] Core runtime

## Validation
- `python3 -m py_compile ods/extensions/services/dashboard-api/helpers.py ods/extensions/services/dashboard-api/routers/features.py` passes.
- `git diff --check` passes.
- Minimal diff preserves existing file line endings and formatting.

### DIFF
--- a/ods/extensions/services/dashboard-api/helpers.py
+++ b/ods/extensions/services/dashboard-api/helpers.py
@@ -22,6 +22,16 @@ from host_agent_client import AgentClientError, async_request_json as request_ag
 from models import ServiceStatus, DiskUsage, ModelInfo, BootstrapStatus
 
 
+def _effective_external_port(service_id: str, config: dict) -> int:
+    """Return the live external port for a host service, honoring OPENCODE_PORT."""
+    candidate = config.get("external_port", config.get("port", 0))
+    if service_id == "opencode":
+        raw = os.environ.get("OPENCODE_PORT", "")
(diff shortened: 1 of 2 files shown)


patil2001:
Motivation for this one: fix(dashboard-api): honor OPENCODE_PORT for the OpenCode host service. This is a contained fix for an edge case that would otherwise leave the stack in the wrong state on that path.

PR #2402, opened 2026-08-05
fix(installer): clamp fresh OpenCode config output to context

patil2001:
## Summary

`_opencode_write_fresh` in `ods/installers/phases/07-devtools.sh` wrote a hardcoded model `limit.output` of `32768` even when `MAX_CONTEXT` was smaller, while the jq rewrite path, the macOS writer, and the host-agent all clamp to `min(32768, context)`. An output limit above the model context is rejected by OpenCode/llama.cpp. The fresh-template writer now clamps to the configured context.

## AI Assistance

AI-assisted edge-case enumeration. The author reviewed the implementation and is responsible for the final diff.

## Release Lane

## Changed Surface

- [x] Installer

## Validation

- `bash -n ods/installers/phases/07-devtools.sh` - passed.

### DIFF
--- a/ods/installers/phases/07-devtools.sh
+++ b/ods/installers/phases/07-devtools.sh
@@ -191,6 +191,8 @@ else
         # and as deterministic recovery when the jq rewrite path finds an
         # existing malformed file it cannot parse (issue #332).
         _opencode_write_fresh() {
+            local _opencode_output=32768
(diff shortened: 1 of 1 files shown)


patil2001:
Posting the rationale: fix(installer): clamp fresh OpenCode config output to context. This is a contained fix for an edge case that would otherwise leave the stack in the wrong state on that path.

PR #2403, opened 2026-08-05
fix(windows): clamp OpenCode config output to context

patil2001:
## Summary

The Windows OpenCode config writers (`New-WindowsOpenCodeConfigObject` and `Update-WindowsOpenCodeConfigObject`) hardcoded `limit.output = 32768`, which exceeds the model context on smaller-context installs and is rejected by OpenCode/llama.cpp. Both now clamp to `min(32768, context)` to match the Linux/macOS writers and the host-agent.

## AI Assistance

AI-assisted edge-case enumeration. The author reviewed the implementation and is responsible for the final diff.

## Release Lane

## Changed Surface

- [x] Installer

## Validation

- PowerShell parse check on `ods/installers/windows/lib/opencode-config.ps1` - passed.

### DIFF
--- a/ods/installers/windows/lib/opencode-config.ps1
+++ b/ods/installers/windows/lib/opencode-config.ps1
@@ -49,8 +49,8 @@ function New-WindowsOpenCodeConfigObject {
                     $ModelId = [pscustomobject]@{
                         name = $ModelName
                         limit = [pscustomobject]@{
-                            context = $ContextLimit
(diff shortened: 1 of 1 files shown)


patil2001:
The why behind this change: fix(windows): clamp OpenCode config output to context. This is a contained fix for an edge case that would otherwise leave the stack in the wrong state on that path.

PR #2407, opened 2026-08-05
fix(bootstrap): restart Windows OpenCode web after a model swap

patil2001:
## Summary

`sync_windows_opencode_config` in `ods/scripts/bootstrap-upgrade.sh` refreshed the Windows OpenCode config file on a model swap but never restarted the running `opencode web` process, so a live session kept serving the pre-swap model route (Linux/macOS get this restart from the host-agent). After a successful config sync the function now stops the ODS-managed listener and relaunches it via the `ODSOpenCodeWeb` scheduled task, verifying the process came back. The whole step is non-fatal.

## AI Assistance

AI-assisted cross-platform lifecycle enumeration. The author reviewed the implementation and is responsible for the final diff.

## Release Lane

## Changed Surface

- [x] Core runtime

## Validation

- `bash -n ods/scripts/bootstrap-upgrade.sh` - passed.

### DIFF
--- a/ods/scripts/bootstrap-upgrade.sh
+++ b/ods/scripts/bootstrap-upgrade.sh
@@ -521,9 +521,68 @@ sync_windows_opencode_config() {
         sync_script_arg=$(cygpath -w "$sync_script")
     fi
 
-    log "Refreshing Windows OpenCode config for model: $FULL_GGUF_FILE"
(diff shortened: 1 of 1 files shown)


patil2001:
Why this is necessary: fix(bootstrap): restart Windows OpenCode web after a model swap. This is a contained fix for an edge case that would otherwise leave the stack in the wrong state on that path.

PR #2411, opened 2026-08-05
fix(dashboard): tolerate features missing requirements

patil2001:
## Summary

`config.py` warns when a feature manifest is missing optional metadata but never validates the `requirements` block. A third-party feature entry without `requirements` raised a `KeyError` that 500'd the entire `/api/features` endpoint. This change defaults the block to `{}` in `calculate_feature_status`, so such entries are served with empty requirements instead of crashing the response.

## AI Assistance

AI-assisted triage, repository search, edge-case enumeration, regression-test drafting. The author reviewed the implementation, selected the validation scope, and is responsible for the final diff.

## Release Lane

Stable hotfix reason:

```text
```

## Changed Surface

- [ ] Docs only
- [ ] Tests only
- [x] Dashboard UI
- [ ] Core runtime

## Validation

- `python3 -m py_compile routers/features.py` - passed.
- `python3 -m pytest tests/test_features.py` - passed (21 passed).

### DIFF
--- a/ods/extensions/services/dashboard-api/routers/features.py
+++ b/ods/extensions/services/dashboard-api/routers/features.py
@@ -16,7 +16,9 @@ logger = logging.getLogger(__name__)
 router = APIRouter(tags=["features"])
 
 
-def calculate_feature_status(feature: dict, services: list, gpu_info: Optional[GPUInfo]) -> dict:
+def calculate_feature_status(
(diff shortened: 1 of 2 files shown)


patil2001:
Adding context for reviewers: fix(dashboard): tolerate features missing requirements. This is a contained fix for an edge case that would otherwise leave the stack in the wrong state on that path.

PR #2425, opened 2026-08-05
fix(dashboard): tolerate invalid SHIELD_PORT env value

patil2001:
## Summary
A SHIELD_PORT env value that is empty or non-numeric made the privacy-shield status/stats endpoints crash with an unhandled ValueError (500). Port resolution now falls back safely to the default, reusing the existing safe env-int helper.

## AI Assistance
This change was authored with assistance from an AI coding tool.

## Release Lane
- [ ] Stable hotfix
- [x] Mainline main
- [ ] Next-minor
- [ ] Not sure

Stable hotfix reason: Not targeting the stable release branch directly.

## Changed Surface
- [x] Dashboard API Python
- [ ] Dashboard UI
- [ ] Installer

## Validation
- python -m py_compile privacy.py
- dashboard-api tests

### DIFF
--- a/ods/extensions/services/dashboard-api/routers/privacy.py
+++ b/ods/extensions/services/dashboard-api/routers/privacy.py
@@ -20,6 +20,17 @@ from security import verify_api_key
 
 logger = logging.getLogger(__name__)
 
+
+def _env_int(name: str, default: int = 0) -> int:
+    raw = os.environ.get(name, "").strip()
+    if not raw:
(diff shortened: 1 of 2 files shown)


patil2001:
The why behind this change: fix(dashboard): tolerate invalid SHIELD_PORT env value. This is a contained fix for an edge case that would otherwise leave the stack in the wrong state on that path.

PR #2426, opened 2026-08-05
fix(dashboard): reap OAuth nonces with invalid timestamps

patil2001:
## Summary
A JSON-valid but semantically-invalid OAuth nonce file (non-numeric created_at/ttl_seconds) raised an unhandled ValueError on the init/callback/pending endpoints, returning 500 instead of pruning the stale nonce. Conversions now treat invalid values as expired so the file is reaped.

## AI Assistance
This change was authored with assistance from an AI coding tool.

## Release Lane
- [ ] Stable hotfix
- [x] Mainline main
- [ ] Next-minor
- [ ] Not sure

Stable hotfix reason: Not targeting the stable release branch directly.

## Changed Surface
- [x] Dashboard API Python
- [ ] Dashboard UI
- [ ] Installer

## Validation
- python -m py_compile oauth_passthrough.py
- dashboard-api tests


patil2001:
Why this is necessary: fix(dashboard): reap OAuth nonces with invalid timestamps. This is a contained fix for an edge case that would otherwise leave the stack in the wrong state on that path.

PR #2468, opened 2026-08-06
fix(updates): validate GitHub release payloads

tang-vu:
## Summary
- validate GitHub release responses before accessing object fields
- normalize optional string and prerelease fields at the HTTP boundary
- fall back cleanly when valid JSON has an unexpected schema instead of returning 500 or crashing the refresh task

## Tests
- `pytest tests/test_updates.py -q` (30 passed)
- `python -m py_compile ods/extensions/services/dashboard-api/routers/updates.py ods/extensions/services/dashboard-api/tests/test_updates.py`
- `git diff --check`

### DIFF
--- a/ods/extensions/services/dashboard-api/routers/updates.py
+++ b/ods/extensions/services/dashboard-api/routers/updates.py
@@ -142,6 +142,42 @@ def _normalize_version(value: Optional[str]) -> str:
     return (value or "").strip().lstrip("v")
 
 
+def _release_object(value: object) -> dict:
+    """Validate one object returned by the GitHub releases API."""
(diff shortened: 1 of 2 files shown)


tang-vu:
Closing this later same-author PR as a strict redundant subset of the earlier canonical #2108 after a line-by-line comparison. #2108 now carries this PR's useful object/optional-field validation and endpoint-boundary tests, plus HTTP status enforcement, required usable tags, release-shaped error-body coverage, the null-body regression from #2341, and a fully green CI run at f03ce0bd. No independent behavior remains here; the branch and commit stay available if new evidence appears.

PR #2527, opened 2026-08-07
fix(usage): validate Token Spy report payload

tang-vu:
## Summary
- require Token Spy report responses to be JSON objects before annotating them
- classify malformed JSON and wrong-shape payloads as an unavailable telemetry source
- return the existing honest zeroed report instead of an API 500

## Test plan
- `python -m pytest tests/test_usage.py -q` (22 passed)

Generated with Codex


tang-vu:
Closing in favor of #2335. #2335 validates the full Token Spy report contract (root, periods, summaries, and row collections), while this PR only checks the root object shape. Keeping the narrower subset would duplicate the same files and weaken the single source of review.

PR #2531, opened 2026-08-07
fix(lemonade): classify speech failures

tang-vu:
## Summary
- Classify speech HTTP, timeout, and transport failures through LemonadeClientError.

## Test plan
- [x] pytest -q tests/test_lemonade_client.py

Generated with Codex


tang-vu:
Closing as a duplicate of #2332. The existing PR covers both speech and transcription under one classified-error contract and includes broader tests. This split speech-only PR adds review noise without an independent product concern.

PR #2532, opened 2026-08-07
fix(lemonade): encode model ids in request paths

tang-vu:
## Summary
- Percent-encode model identifiers so slashes and spaces remain one API path segment.

## Test plan
- [x] pytest -q tests/test_lemonade_client.py

Generated with Codex


tang-vu:
Closing after portfolio review because the changed model() method currently has no production caller in ODS, and the regression test only proves MockTransport raw-path encoding rather than compatibility with Lemonade server routing. This can be reopened as a focused fix if a real caller and provider contract demonstrate the bug.

PR #2533, opened 2026-08-07
fix(agent): reject non-finite request timeouts

tang-vu:
## Summary
- Reject NaN and infinite host-agent read timeouts before transport execution.

## Test plan
- [x] pytest -q tests/test_host_agent_client.py

Generated with Codex


tang-vu:
Closing as speculative hardening after tracing callers. Current production host-agent requests pass bounded numeric constants or internally controlled defaults; there is no configuration or request path that can supply NaN/Infinity. The synthetic unit case does not justify another open PR without a reachable failure mode.

PR #2539, opened 2026-08-07
fix(release): parse structured version files

tang-vu:
## Summary
- Correctly unwrap JSON string and object .version formats and reject invalid structured values.

## Test plan
- [x] python tests/test-check-version-consistency.py; python scripts/check-version-consistency.py

Generated with Codex


tang-vu:
Closing as unsupported format expansion. The repository currently has no ods/.version file, and JSON-string/JSON-object .version formats are not a documented release contract. The tests prove the new parser behavior, but not a real ODS workflow that needs it.

PR #2552, opened 2026-08-07
fix(opencode): stop sed escape mangling slashes in systemd unit render

patil2001:
## Summary

The systemd unit render mangled the `__HOME__` and `__OPENCODE_BIN__` substitutions under the broken `[&/\]` escape, which rendered as a literal `&` on GNU/BSD sed and corrupted the rendered unit. This change escapes only `&` and backslash; `/` is safe since the substitution delimiter is `|`.

## AI Assistance

## Release Lane

Stable hotfix reason:

```text
```

## Changed Surface

- [x] Installer

## Validation

- `bash -n ods/installers/phases/06-directories.sh ods/installers/phases/07-devtools.sh` - passed.


patil2001:
Posting the rationale: fix(opencode): stop sed escape mangling slashes in systemd unit render. This is a contained fix for an edge case that would otherwise leave the stack in the wrong state on that path.

PR #2553, opened 2026-08-07
fix(opencode): fail closed when systemd unit placeholders remain unrendered

patil2001:
## Summary

A failed substitution silently installed a broken `opencode-web.service` with literal `__HOME__`/`__OPENCODE_BIN__` tokens. This change aborts the install of the unit when any placeholder remains, matching the host-agent guard style already used later in the same phase.

## AI Assistance

## Release Lane

Stable hotfix reason:

```text
```

## Changed Surface

- [x] Installer

## Validation

- `bash -n ods/installers/phases/07-devtools.sh` - passed.


patil2001:
The why behind this change: fix(opencode): fail closed when systemd unit placeholders remain unrendered. This is a contained fix for an edge case that would otherwise leave the stack in the wrong state on that path.

PR #2554, opened 2026-08-07
fix(opencode): honor OPENCODE_PORT in Linux service render and health checks

patil2001:
## Summary

The Linux OpenCode systemd unit, health check, and summary phases drifted on the web port: the unit hardcoded 3003 while the health/summary reads `OPENCODE_PORT`. Changing the variable left health polling the wrong port. The unit render, `12-health.sh`, and `13-summary.sh` now all honor `OPENCODE_PORT` (default 3003) so the service and its health path stay in sync.

## AI Assistance

AI-assisted repository search and cross-file port-wiring enumeration. The author reviewed the implementation, selected the validation scope, and is responsible for the final diff.

## Release Lane

Stable hotfix reason:

```text
```

## Changed Surface

- [x] Installer
- [ ] Dashboard UI
- [ ] Core runtime
- [ ] Tests only

## Validation

- `bash -n` on changed phases - passed.
- Linux opencode path regression test - passed.


patil2001:
Why this is necessary: fix(opencode): honor OPENCODE_PORT in Linux service render and health checks. This is a contained fix for an edge case that would otherwise leave the stack in the wrong state on that path.

PR #2555, opened 2026-08-07
fix(opencode): require 2xx responses in host-agent opencode health probe

patil2001:
## Summary

The host-agent opencode health probe treated any HTTP response as healthy, so a 4xx/5xx page state still read as ready and the dashboard flagged the coding surface as up. The probe now requires a 200-299 response before marking the host service healthy.

## AI Assistance

AI-assisted regression triage on the homestay probe path. The author reviewed the implementation, selected the validation scope, and is responsible for the final diff.

## Release Lane

Stable hotfix reason:

```text
```

## Changed Surface

- [ ] Installer
- [ ] Dashboard UI
- [x] Core runtime
- [ ] Tests only

## Validation

- `python -m py_compile` on `ods/bin/ods-host-agent.py` - passed.

### DIFF
--- a/ods/bin/ods-host-agent.py
+++ b/ods/bin/ods-host-agent.py
@@ -11015,7 +11015,7 @@ def _wait_for_opencode_health(attempts: int = 30) -> None:
         try:
             request = urllib_request.Request(url, method="GET")
             with urllib_request.urlopen(request, timeout=3) as response:
-                if 200 <= response.status < 500:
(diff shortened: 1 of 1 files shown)


patil2001:
Adding context for reviewers: fix(opencode): require 2xx responses in host-agent opencode health probe. This is a contained fix for an edge case that would otherwise leave the stack in the wrong state on that path.

PR #2556, opened 2026-08-07
fix(opencode): honor OPENCODE_PORT in macOS LaunchAgent

patil2001:
## Summary

The macOS LaunchAgent hardcoded the web port and bypassed the ODS environment, so a custom `OPENCODE_PORT` was ignored and the agent bound a different port than the dashboard expected. The plist writer now resolves the port from the ODS env (`OPENCODE_PORT`, default 3003) for launch and status parity.

## AI Assistance

AI-assisted repository search over the macOS LaunchAgent wiring. The author reviewed the implementation, selected the validation scope, and is responsible for the final diff.

## Release Lane

Stable hotfix reason:

```text
```

## Changed Surface

- [x] Installer
- [ ] Dashboard UI
- [ ] Core runtime
- [ ] Tests only

## Validation

- macOS plist transition tests - passed.


patil2001:
Motivation for this one: fix(opencode): honor OPENCODE_PORT in macOS LaunchAgent. This is a contained fix for an edge case that would otherwise leave the stack in the wrong state on that path.

PR #2557, opened 2026-08-07
fix(opencode): use serve subcommand in macOS LaunchAgent

patil2001:
## Summary

The macOS LaunchAgent called `opencode web`, a legacy subcommand that current OpenCode builds no longer ship. The plist writer now uses `opencode serve`, matching the Linux systemd unit's ExecStart.

## AI Assistance

AI-assisted cross-platform subcommand comparison. The author reviewed the implementation, selected the validation scope, and is responsible for the final diff.

## Release Lane

Stable hotfix reason:

```text
```

## Changed Surface

- [x] Installer
- [ ] Dashboard UI
- [ ] Core runtime
- [ ] Tests only

## Validation

- macOS plist transition tests - passed.

### DIFF
--- a/ods/installers/macos/install-macos.sh
+++ b/ods/installers/macos/install-macos.sh
@@ -2873,7 +2873,7 @@ for service in (data.get("services") or {}).values():
     <key>ProgramArguments</key>
     <array>
         <string>${OPENCODE_BIN}</string>
-        <string>web</string>
+        <string>serve</string>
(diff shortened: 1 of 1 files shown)


patil2001:
Posting the rationale: fix(opencode): use serve subcommand in macOS LaunchAgent. This is a contained fix for an edge case that would otherwise leave the stack in the wrong state on that path.

PR #2558, opened 2026-08-07
fix(opencode): resolve port from env and gate status on binary

patil2001:
## Summary

`ods-macos.sh status` reported OpenCode state using a hardcoded port and did not check whether the binary exists, so it could claim "running" or "installed" incorrectly. Status now resolves `OPENCODE_PORT` from env (default 3003) and gates on the per-user binary path.

## AI Assistance

AI-assisted enumeration of the macOS status surface. The author reviewed the implementation, selected the validation scope, and is responsible for the final diff.

## Release Lane

Stable hotfix reason:

```text
```

## Changed Surface

- [x] Installer
- [ ] Dashboard UI
- [ ] Core runtime
- [ ] Tests only

## Validation

- `bash -n` on `ods-macos.sh` - passed.
- macOS transition tests - passed.

### DIFF
--- a/ods/installers/macos/ods-macos.sh
+++ b/ods/installers/macos/ods-macos.sh
@@ -713,8 +713,15 @@ cmd_status() {
     echo -e "  ${DGRN}$(printf -- '-%.0s' {1..40})${NC}"
 
     # Parallel arrays (Bash 3.2 compatible)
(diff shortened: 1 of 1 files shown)


patil2001:
The why behind this change: fix(opencode): resolve port from env and gate status on binary. This is a contained fix for an edge case that would otherwise leave the stack in the wrong state on that path.

PR #2559, opened 2026-08-07
fix(opencode): kickstart and verify macOS LaunchAgent after bootstrap

patil2001:
## Summary

After bootstrapping theOpenCode binary, the macOS install wrote the LaunchAgent but did not kickstart it or verify it bound, so the web surface stayed down until the next login. The installer now `launchctl kickstart -p` the agent and polls the port for readiness, mirroring the host-agent's health gate.

## AI Assistance

AI-assisted regression triage on the macOS bootstrap path. The author reviewed the implementation, selected the validation scope, and is responsible for the final diff.

## Release Lane

## Changed Surface

- [x] Installer
- [ ] Dashboard UI
- [ ] Core runtime
- [ ] Tests only

## Validation


patil2001:
Why this is necessary: fix(opencode): kickstart and verify macOS LaunchAgent after bootstrap. This is a contained fix for an edge case that would otherwise leave the stack in the wrong state on that path.

PR #2560, opened 2026-08-07
fix(opencode): remove stale LaunchAgent plist on reinstall

patil2001:
## Summary

On reinstall the pre-install LaunchAgent sweep removed old agent plists but left a stale `com.ods.opencode-web.plist`, which then carried a previous port/subcommand and could double-start the service. The sweep now also removes the OpenCode plist before installing fresh.

## AI Assistance

AI-assisted review of the macOS stale-agent cleanup surface. The author reviewed the implementation, selected the validation scope, and is responsible for the final diff.

## Release Lane

## Changed Surface

- [x] Installer
- [ ] Dashboard UI
- [ ] Core runtime
- [ ] Tests only

## Validation

- macOS pre-existing failure confirmed environmental (Windows FS), not introduced here.


patil2001:
Adding context for reviewers: fix(opencode): remove stale LaunchAgent plist on reinstall. This is a contained fix for an edge case that would otherwise leave the stack in the wrong state on that path.

PR #2561, opened 2026-08-07
fix(opencode): use detected interpreter in macOS config writer

patil2001:
## Summary

The macOS config writer hardcoded the Python interpreter, breaking on systems where `python3` resolves differently than the interpreter the installer validated. It now uses the detected `ODS_PYTHON_CMD` (defaulting to `/usr/bin/python3`) for writing OpenCode config.

## AI Assistance

AI-assisted review of the macOS config writer interpreter wiring. The author reviewed the implementation, selected the validation scope, and is responsible for the final diff.

## Release Lane

## Changed Surface

- [x] Installer
- [ ] Dashboard UI
- [ ] Core runtime
- [ ] Tests only

## Validation


patil2001:
Motivation for this one: fix(opencode): use detected interpreter in macOS config writer. This is a contained fix for an edge case that would otherwise leave the stack in the wrong state on that path.

PR #2562, opened 2026-08-07
fix(opencode): add small_model and unify provider name in macOS config

patil2001:
## Summary

The macOS OpenCode config omitted `small_model` and used a different provider name than Linux/Windows, so agility features and dashboard labels diverged per platform. The writer now sets `small_model` and uses the shared provider name `llama-server (local)` for parity.

## AI Assistance

AI-assisted cross-platform config comparison. The author reviewed the implementation, selected the validation scope, and is responsible for the final diff.

## Release Lane

## Changed Surface

- [x] Installer
- [ ] Dashboard UI
- [ ] Core runtime
- [ ] Tests only

## Validation

- macOS transition tests - passed.

### DIFF
--- a/ods/installers/macos/install-macos.sh
+++ b/ods/installers/macos/install-macos.sh
@@ -454,7 +454,7 @@ provider_id = "llama-server"
 provider = data.setdefault("provider", {}).setdefault(provider_id, {})
 provider.update({
     "npm": "@ai-sdk/openai-compatible",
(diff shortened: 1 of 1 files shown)


patil2001:
Posting the rationale: fix(opencode): add small_model and unify provider name in macOS config. This is a contained fix for an edge case that would otherwise leave the stack in the wrong state on that path.

PR #2563, opened 2026-08-07
fix(opencode): prune stale model entries in Windows config upsert

patil2001:
## Summary

The Windows OpenCode config upsert appended model entries on every update without pruning, so switching models left stale `models` entries under the llama-server provider. The upsert now prunes obsolete model ids before writing the current route.

## AI Assistance

AI-assisted review of the Windows config upsert path. The author reviewed the implementation, selected the validation scope, and is responsible for the final diff.

## Release Lane

## Changed Surface

- [x] Installer
- [ ] Dashboard UI
- [ ] Core runtime
- [ ] Tests only

## Validation

- Windows opencode config regression tests - passed (24/24).

### DIFF
--- a/ods/installers/windows/lib/opencode-config.ps1
+++ b/ods/installers/windows/lib/opencode-config.ps1
@@ -108,6 +108,11 @@ function Update-WindowsOpenCodeConfigObject {
     }
     $models = $llamaProvider.models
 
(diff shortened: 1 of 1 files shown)


patil2001:
The why behind this change: fix(opencode): prune stale model entries in Windows config upsert. This is a contained fix for an edge case that would otherwise leave the stack in the wrong state on that path.

PR #2564, opened 2026-08-07
fix(opencode): cap Windows output limit at the model context

patil2001:
## Summary

The Windows OpenCode config fixed `output` at 32768 even when the model context was smaller, so the declared output cap could exceed the model's context. The create and update paths now cap output at `[Math]::Min(32768, $ContextLimit)`.

## AI Assistance

AI-assisted review of the Windows config limit derivation. The author reviewed the implementation, selected the validation scope, and is responsible for the final diff.

## Release Lane

## Changed Surface

- [x] Installer
- [ ] Dashboard UI
- [ ] Core runtime
- [ ] Tests only

## Validation

- Windows opencode config regression tests - passed.

### DIFF
--- a/ods/installers/windows/lib/opencode-config.ps1
+++ b/ods/installers/windows/lib/opencode-config.ps1
@@ -50,7 +50,7 @@ function New-WindowsOpenCodeConfigObject {
                         name = $ModelName
                         limit = [pscustomobject]@{
                             context = $ContextLimit
-                            output = 32768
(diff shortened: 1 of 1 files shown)


patil2001:
Why this is necessary: fix(opencode): cap Windows output limit at the model context. This is a contained fix for an edge case that would otherwise leave the stack in the wrong state on that path.

PR #2565, opened 2026-08-07
fix(opencode): map OPENCODE_ keys to the opencode service on apply

patil2001:
## Summary

When applying settings, `OPENCODE_*` env keys fell through the service matcher, so OpenCode settings applied to the wrong service or were dropped. The matcher now maps the `OPENCODE_` prefix to the `opencode` service.

## AI Assistance

AI-assisted review of the settings apply service mapping. The author reviewed the implementation, selected the validation scope, and is responsible for the final diff.

## Release Lane

## Changed Surface

- [ ] Installer
- [x] Dashboard UI
- [ ] Core runtime
- [ ] Tests only

## Validation

- `python -m py_compile` on `settings.py` - passed.

### DIFF
--- a/ods/extensions/services/dashboard-api/settings.py
+++ b/ods/extensions/services/dashboard-api/settings.py
@@ -419,6 +419,8 @@ def _match_apply_service(key: str) -> Optional[str]:
         return "perplexica"
     if key.startswith("APE_"):
         return "ape"
+    if key.startswith("OPENCODE_"):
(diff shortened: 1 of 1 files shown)


patil2001:
Adding context for reviewers: fix(opencode): map OPENCODE_ keys to the opencode service on apply. This is a contained fix for an edge case that would otherwise leave the stack in the wrong state on that path.

PR #2566, opened 2026-08-07
fix(opencode): drop erroneous VRAM gate on the coding feature

patil2001:
## Summary

The OpenCode coding feature declared `vram_gb: 8` while OpenCode itself runs on the host against any active model route and does not reserve GPU VRAM; the gate hid the feature on lower-VRAM systems. The feature now requires `vram_gb: 0`, matching its host-managed deployment.

## AI Assistance

AI-assisted review of the feature VRAM gating. The author reviewed the implementation, selected the validation scope, and is responsible for the final diff.

## Release Lane

## Changed Surface

- [ ] Installer
- [x] Dashboard UI
- [ ] Core runtime
- [ ] Tests only

## Validation

- `ods/scripts/validate-manifests.sh` - passed (opencode ok).
- Manifest schema tests - passed.

### DIFF
--- a/ods/extensions/services/opencode/manifest.yaml
+++ b/ods/extensions/services/opencode/manifest.yaml
@@ -36,7 +36,7 @@ features:
     category: development
     requirements:
       services: [opencode]
-      vram_gb: 8
+      vram_gb: 0
     enabled_services_all: [opencode]
     setup_time: Ready
     priority: 6


patil2001:
Motivation for this one: fix(opencode): drop erroneous VRAM gate on the coding feature. This is a contained fix for an edge case that would otherwise leave the stack in the wrong state on that path.

PR #2568, opened 2026-08-07
fix(opencode): add small_model to flat render config

patil2001:
## Summary

The flat OpenCode config rendered by `render-runtime-configs.py` omitted `small_model`, drifting from the Linux/Windows/macOS writers. The payload now mirrors `small_model` so runtime feature gating is consistent across surfaces.

## AI Assistance

AI-assisted review of the cross-surface config parity. The author reviewed the implementation, selected the validation scope, and is responsible for the final diff.

## Release Lane

## Changed

- [ ] Installer
- [ ] Dashboard UI
- [x] Core runtime
- [ ] Tests only

## Validation

- `python -m py_compile` on `render-runtime-configs.py` - passed.

### DIFF
--- a/ods/scripts/render-runtime-configs.py
+++ b/ods/scripts/render-runtime-configs.py
@@ -454,6 +454,7 @@ def render_opencode(inputs: RenderInputs) -> RenderedFile:
         "baseURL": base_url,
         "apiKey": opencode_key(inputs),
         "model": model,
+        "small_model": model,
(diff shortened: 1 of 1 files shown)


patil2001:
The why behind this change: fix(opencode): add small_model to flat render config. This is a contained fix for an edge case that would otherwise leave the stack in the wrong state on that path.

PR #2569, opened 2026-08-07
fix(uninstall): surgically remove both OpenCode config docs

patil2001:
## Summary

07-devtools.sh writes opencode.json from a template and syncs a config.json copy (the document OpenCode actually reads). Uninstall only removed opencode.json, so a stale llama-server route survived as config.json. Uninstall now removes both ODS-authored documents, scoped by the llama-server marker so unrelated user config is left untouched, with regression coverage.

## AI Assistance

AI-assisted review of the uninstall config surface. The author reviewed the implementation, selected the validation scope, and is responsible for the final diff.

## Release Lane

Stable hotfix reason:

```text
```

## Change Surface

- [x] Installer
- [ ] Dashboard UI
- [ ] Core runtime
- [ ] Tests only

## Validation

- `bash -n` on `ods-uninstall.sh` and test - passed.
- `bash ods/tests/test-uninstall-compose-flags.sh` - passed (7 scenarios incl. new OpenCode config scenario).

### DIFF
--- a/ods/ods-uninstall.sh
+++ b/ods/ods-uninstall.sh
@@ -378,11 +378,22 @@ if [[ -d "$HOME/.ods" ]]; then
 fi
 
 # 7. Remove OpenCode config (if we created it)
-OPENCODE_CONFIG="$HOME/.config/opencode/opencode.json"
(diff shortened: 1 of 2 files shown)


patil2001:
Why this is necessary: fix(uninstall): surgically remove both OpenCode config docs. This is a contained fix for an edge case that would otherwise leave the stack in the wrong state on that path.

PR #2570, opened 2026-08-07
test(opencode): cover sed escape and placeholder rendering in linux path test

patil2001:
## Summary

The Linux opencode path test did not cover the systemd unit render, so the sed escape bug and unrendered-placeholder guard could regress without CI catching them. The test now asserts the rendered unit contains the real binary path and no leftover `__PLACEHOLDER__`.

## AI Assistance

AI-assisted regression-test drafting. The author reviewed the implementation, selected the validation scope, and is responsible for the final diff.

## Release Lane

## Changed Surface

- [ ] Installer
- [ ] Dashboard UI
- [ ] Core runtime
- [x] Tests only

## Validation

- `bash ods/tests/test-linux-opencode-path.sh` - passed.

### DIFF
--- a/ods/tests/test-linux-opencode-path.sh
+++ b/ods/tests/test-linux-opencode-path.sh
@@ -34,8 +34,48 @@ grep -q '__OPENCODE_BIN_DIR__' "$SERVICE" \
   && pass "systemd service templates resolved binary directory" \
   || fail "systemd service PATH must include resolved binary directory"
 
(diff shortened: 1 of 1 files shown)


patil2001:
Adding context for reviewers: test(opencode): cover sed escape and placeholder rendering in linux path test. This is a contained fix for an edge case that would otherwise leave the stack in the wrong state on that path.

PR #2589, opened 2026-08-07
fix(opencode): use llama-server host port default in Linux route

patil2001:
## Summary

The OpenCode route fell back to OLLAMA_PORT :8080, the container-internal port, while llama-server is exposed on the host at 11434 — so OpenCode probed a dead port and the model list was empty. Default the host port to 11434.

## AI Assistance

## Release Lane

Stable hotfix reason:

```text
```

## Changed Surface

- [x] Installer
- [ ] Dashboard UI
- [ ] Core runtime
- [ ] Tests only

## Validation

- `bash -n` on 07-devtools.sh - passed.


patil2001:
The why behind this change: fix(opencode): use llama-server host port default in Linux route. This is a contained fix for an edge case that would otherwise leave the stack in the wrong state on that path.

PR #2590, opened 2026-08-07
fix(renderer): fall back to active model for LEMONADE_MODEL

patil2001:
## Summary

For non-lemonade modes lemonade_model_id defaults to empty, emitting LEMONADE_MODEL= with no value in the generated env. Fall back to the active model identity so consumers never see an empty variable.

## AI Assistance

## Release Lane

Stable hotfix reason:

```text
```

## Changed Surface

- [ ] Installer
- [ ] Dashboard UI
- [x] Core runtime
- [ ] Tests only

## Validation


patil2001:
Why this is necessary: fix(renderer): fall back to active model for LEMONADE_MODEL. This is a contained fix for an edge case that would otherwise leave the stack in the wrong state on that path.

PR #2591, opened 2026-08-07
fix(renderer): drop retry loop from atomic write

patil2001:
## Summary

atomic_write_text retried os.replace up to 10 times with backoff, hiding real failures and contradicting the let-it-crash contract. Replaced with a single os.replace that propagates errors immediately.

## AI Assistance

## Release Lane

Stable hotfix reason:

```text
```

## Changed Surface

- [ ] Installer
- [ ] Dashboard UI
- [x] Core runtime
- [ ] Tests only

## Validation

### DIFF
--- a/ods/scripts/render-runtime-configs.py
+++ b/ods/scripts/render-runtime-configs.py
@@ -15,7 +15,6 @@ import re
 import stat
 import sys
 import tempfile
-import time
 from dataclasses import asdict, dataclass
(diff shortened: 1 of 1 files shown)


patil2001:
Adding context for reviewers: fix(renderer): drop retry loop from atomic write. This is a contained fix for an edge case that would otherwise leave the stack in the wrong state on that path.

PR #2592, opened 2026-08-07
fix(uninstall): keep resolver and overlay GPU backend in lockstep

patil2001:
## Summary

resolve_compose_flags defaulted GPU_BACKEND to nvidia for the resolver but used the unset variable for the overlay selection, so docker compose down ran base-only and leftovers survived on AMD/Apple hosts. Capture the defaulted value once and reuse it for both.

## AI Assistance

## Release Lane

Stable hotfix reason:

```text
```

## Changed Surface

- [x] Installer
- [ ] Dashboard UI
- [ ] Core runtime
- [ ] Tests only

## Validation

- `bash -n` on ods-uninstall.sh - passed.

### DIFF
--- a/ods/ods-uninstall.sh
+++ b/ods/ods-uninstall.sh
@@ -22,6 +22,7 @@ log_error() { echo -e "${RED}[ERROR]${NC} $*" >&2; }
 
 resolve_compose_flags() {
     local flags=""
+    local _gpu="${GPU_BACKEND:-nvidia}"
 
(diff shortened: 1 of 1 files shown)


patil2001:
Motivation for this one: fix(uninstall): keep resolver and overlay GPU backend in lockstep. This is a contained fix for an edge case that would otherwise leave the stack in the wrong state on that path.

PR #2593, opened 2026-08-07
fix(settings): exclude live-read keys from changed-key status

patil2001:
## Summary

HF_TOKEN stayed in changed_keys after the mapping loop skipped it, so the apply plan reported manual restart with empty manualKeys. Filter live-read keys at changed_keys construction so the status matches the summary.

## AI Assistance

## Release Lane

Stable hotfix reason:

```text
```

## Changed Surface

- [ ] Installer
- [ ] Dashboard UI
- [x] Core runtime
- [ ] Tests only

## Validation

- `python -m py_compile` on settings.py - passed.


patil2001:
Posting the rationale: fix(settings): exclude live-read keys from changed-key status. This is a contained fix for an edge case that would otherwise leave the stack in the wrong state on that path.

PR #2594, opened 2026-08-07
fix(renderer): use container-service default for litellm base

patil2001:
## Summary

host.docker.internal does not resolve on native Linux, so the local-litellm fallback base pointed at a broken address. Use the llama-server container-service default used elsewhere in the file.

## AI Assistance

## Release Lane

Stable hotfix reason:

```text
```

## Changed Surface

- [ ] Installer
- [ ] Dashboard UI
- [x] Core runtime
- [ ] Tests only

## Validation


patil2001:
The why behind this change: fix(renderer): use container-service default for litellm base. This is a contained fix for an edge case that would otherwise leave the stack in the wrong state on that path.

PR #2597, opened 2026-08-07
fix(macos): force-recreate containers on restart

patil2001:
## Summary

The 'ods-macos.sh restart' command ran 'docker compose up -d', which is a no-op for containers that are already running, so users were told the service restarted when it was never recreated. Add --force-recreate to both the single-service and all-services restart branches so running containers are actually stopped and recreated.

## AI Assistance

## Release Lane

Stable hotfix reason:

```text
```

## Changed Surface

- [x] macos/ods-macos.sh (restart: single-service + all-services branches)

## Validation

Validation: bash -n pass

### DIFF
--- a/ods/installers/macos/ods-macos.sh
+++ b/ods/installers/macos/ods-macos.sh
@@ -834,7 +834,7 @@ cmd_restart() {
     elif [[ -n "$service" ]]; then
         ai "Restarting ${service}..."
         # shellcheck disable=SC2086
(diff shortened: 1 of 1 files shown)


patil2001:
Why this is necessary: fix(macos): force-recreate containers on restart. This is a contained fix for an edge case that would otherwise leave the stack in the wrong state on that path.

PR #2602, opened 2026-08-07
fix(opencode): read Windows config with UTF-8 encoding

patil2001:
## Summary

Windows config was read with the default (non-UTF8) PowerShell encoding, corrupting non-ASCII keys; now uses UTF-8.

## AI Assistance

## Release Lane

Stable hotfix reason:

```text
```

## Changed Surface

- [x] Windows install scripts
- [x] config encoding
- [ ] macOS
- [ ] Linux
- [ ] renderer

## Validation

- [x] Powershell parse check
- [x] focused diff

### DIFF
--- a/ods/installers/windows/lib/opencode-config.ps1
+++ b/ods/installers/windows/lib/opencode-config.ps1
@@ -163,7 +163,7 @@ function Sync-WindowsOpenCodeConfig {
 
     if ($_existingConfigFile) {
         try {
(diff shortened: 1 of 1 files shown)


patil2001:
Adding context for reviewers: fix(opencode): read Windows config with UTF-8 encoding. This is a contained fix for an edge case that would otherwise leave the stack in the wrong state on that path.

PR #2604, opened 2026-08-07
fix(opencode): resolve Windows sync libs for installed tree

patil2001:
## Summary

Windows sync script used a fixed lib path that broke for the installed (non-source) tree; now the lib path is resolved relative to the installed tree.

## AI Assistance

## Release Lane

Stable hotfix reason:

```text
```

## Changed Surface

- [x] Windows install/sync scripts
- [ ] macOS
- [ ] Linux
- [ ] renderer
- [ ] config

## Validation

- [x] powershell parse check
- [x] focused diff


patil2001:
Posting the rationale: fix(opencode): resolve Windows sync libs for installed tree. This is a contained fix for an edge case that would otherwise leave the stack in the wrong state on that path.

PR #2605, opened 2026-08-07
fix(opencode): wire Lemonade model id resolution into Windows config

patil2001:
## Summary

On Windows, the Lemonade model id helper was defined but never called; now the resolved model id is wired into Sync-WindowsOpenCodeConfigFromEnv so the served model matches the configured one.

## AI Assistance

## Release Lane

Stable hotfix reason:

```text
```

## Changed Surface

- [x] Windows config generation
- [x] Lemonade model id
- [ ] macOS
- [ ] Linux
- [ ] renderer

## Validation

- [x] powershell parse check
- [x] focused diff


patil2001:
The why behind this change: fix(opencode): wire Lemonade model id resolution into Windows config. This is a contained fix for an edge case that would otherwise leave the stack in the wrong state on that path.

PR #2606, opened 2026-08-07
fix(opencode): align bootstrap launcher python resolution with installer

patil2001:
## Summary

The detached bootstrap launcher function exists in two copies that drifted: install-macos.sh resolved the python interpreter via ${PYTHON_CMD:-/usr/bin/python3}, while ods-macos.sh fell back to $(command -v python3 || command -v python || true). That drift could select a different interpreter between install and run. This aligns the ods-macos.sh copy to the env-var-with-fallback style while keeping its empty-value guard.

## AI Assistance

## Release Lane

Stable hotfix reason:

```text
```

## Changed Surface

- [x] macOS install scripts
- [x] bootstrap launcher
- [ ] Windows
- [ ] Linux
- [ ] renderer
- [ ] config

## Validation

- [x] bash -n on modified script


patil2001:
Why this is necessary: fix(opencode): align bootstrap launcher python resolution with installer. This is a contained fix for an edge case that would otherwise leave the stack in the wrong state on that path.

PR #2607, opened 2026-08-07
fix(opencode): wait for bootstrap compose stack on update

patil2001:
## Summary

cmd_update recreated the compose stack without the readiness wait that cmd_start and cmd_restart already perform, so an update could race the bootstrap compose stack coming up. This adds macos_wait_for_bootstrap_compose_safe "update" || return 1 ahead of the force-recreate, mirroring the existing guard pattern.

## AI Assistance

## Release Lane

Stable hotfix reason:

```text
```

## Changed Surface

- [x] macOS update flow
- [x] bootstrap compose wait
- [ ] Windows
- [ ] Linux
- [ ] renderer
- [ ] config

## Validation

- [x] bash -n on modified script

### DIFF
--- a/ods/installers/macos/ods-macos.sh
+++ b/ods/installers/macos/ods-macos.sh
@@ -988,6 +988,7 @@ cmd_update() {
     compose_pull_with_retry "$flags"
 
     ai "Recreating containers..."
+    macos_wait_for_bootstrap_compose_safe "update" || return 1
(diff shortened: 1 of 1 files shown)


patil2001:
Adding context for reviewers: fix(opencode): wait for bootstrap compose stack on update. This is a contained fix for an edge case that would otherwise leave the stack in the wrong state on that path.

PR #2608, opened 2026-08-07
fix(host-agent): poll the .env-resolved agent port on macOS

patil2001:
## Summary

The macOS agent health poll previously probed a hard-coded port even when the agent port is overridden in the .env file, so a custom port reported unhealthy. The probe now resolves the agent port from .env with a numeric fallback of 7710, matching how the rest of the host-agent reads its configuration.

## AI Assistance

## Release Lane

Stable hotfix reason:

```text
```

## Changed Surface

- [x] macOS scripts
- [x] host-agent port resolution
- [ ] Windows
- [ ] Linux
- [ ] renderer
- [ ] config

## Validation


patil2001:
Motivation for this one: fix(host-agent): poll the .env-resolved agent port on macOS. This is a contained fix for an edge case that would otherwise leave the stack in the wrong state on that path.

PR #2609, opened 2026-08-07
fix(host-agent): run HF dep install as the real agent user

patil2001:
## Summary

The Hugging Face dependency install step ran under the wrong account, so pip/huggingface_hub artifacts landed in the wrong home directory and the agent user could not see them. The install now runs as the real agent user, consistent with the rest of the host-agent bootstrap.

## AI Assistance

## Release Lane

Stable hotfix reason:

```text
```

## Changed Surface

- [x] Linux host-agent
- [x] HF dependency install user
- [ ] macOS
- [ ] Windows
- [ ] renderer
- [ ] config

## Validation


patil2001:
Posting the rationale: fix(host-agent): run HF dep install as the real agent user. This is a contained fix for an edge case that would otherwise leave the stack in the wrong state on that path.

PR #2610, opened 2026-08-07
fix(opencode): grant linger to the real install user

patil2001:
## Summary

The Linux installer enabled loginctl linger for a hard-coded user instead of the actual install user, so the opencode service died at logout on systems where the two differ. Linger is now granted to the resolved install user, keeping the service alive.

## AI Assistance

## Release Lane

Stable hotfix reason:

```text
```

## Changed Surface

- [x] Linux install scripts
- [x] linger/systemd setup
- [ ] macOS
- [ ] Windows
- [ ] renderer
- [ ] config

## Validation


patil2001:
The why behind this change: fix(opencode): grant linger to the real install user. This is a contained fix for an edge case that would otherwise leave the stack in the wrong state on that path.

PR #2611, opened 2026-08-07
fix(host-agent): treat non-active opencode states as inactive

patil2001:
## Summary

The host-agent's is-active check compared against a single expected state string, so transitional states were misclassified as active and the apply loop skipped a needed reconfiguration. Any state other than active is now treated as inactive for tolerant, deterministic status handling.

## AI Assistance

## Release Lane

Stable hotfix reason:

```text
```

## Changed Surface

- [x] Linux host-agent status handling
- [ ] macOS
- [ ] Windows
- [ ] renderer
- [ ] config

## Validation

### DIFF
--- a/ods/bin/ods-host-agent.py
+++ b/ods/bin/ods-host-agent.py
@@ -10998,12 +10998,7 @@ def _capture_managed_opencode_state() -> dict:
             timeout=15,
             env=user_env,
         )
-        if status.returncode == 0:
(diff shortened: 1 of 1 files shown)


patil2001:
Why this is necessary: fix(host-agent): treat non-active opencode states as inactive. This is a contained fix for an edge case that would otherwise leave the stack in the wrong state on that path.

PR #2612, opened 2026-08-07
fix(dashboard-api): handle UnicodeDecodeError in _read_env_map_from_path in settings.py

vaibhavsrv:
## Problem

In `ods/extensions/services/dashboard-api/settings.py`, `_read_env_map_from_path()` reads environment settings from `.env` files using `path.read_text(encoding="utf-8")`. If the `.env` file contains invalid non-UTF-8 characters or corrupt bytes, a `UnicodeDecodeError` previously crashed key parsing.

## Fix

Pass `errors="replace"` to `path.read_text()` inside `_read_env_map_from_path()`.

## Verification

Ran `pytest ods/extensions/services/dashboard-api/tests/test_routers.py` (54/54 passed). `git diff --check` passed cleanly.

### DIFF
--- a/ods/extensions/services/dashboard-api/settings.py
+++ b/ods/extensions/services/dashboard-api/settings.py
@@ -105,7 +105,7 @@ def _strip_env_quotes(value: str) -> str:
 
 def _read_env_map_from_path(path: Path) -> tuple[dict[str, str], list[dict[str, Any]]]:
     try:
-        return _parse_env_text(path.read_text(encoding="utf-8"))
+        return _parse_env_text(path.read_text(encoding="utf-8", errors="replace"))
     except OSError:
         return {}, []
 


bokiko:
Reviewed against head `fba63337`. Confirmed the change works for the function it touches: with a fixture containing genuine invalid UTF-8 bytes, `_read_env_map_from_path()` crashes on `main` and succeeds on this head while preserving valid entries before and after the corrupt line.

There is one blocking gap: `_build_settings_env_payload()` in `main.py` has its own separate `env_path.read_text(encoding="utf-8")` fallback. That is the read path used by `GET /api/settings/env` when no raw text is supplied—the dashboard Settings-page load path. It is unchanged by this PR and still raises the same `UnicodeDecodeError` against the same fixture. Please apply the chosen decoding policy consistently to both reads. The same unprotected pattern appears twice more in `main.py`: `_read_installed_version()` (:150, which also omits an explicit encoding) and `_read_install_date()` (:430). Since `_parse_env_text` is already shared between `settings.py` and `main.py`, a single shared reader helper applying one decoding policy in one place may be cleaner than patching each call site.

Please also add a regression test. The existing suite does not cover non-UTF-8 input, and the PR currently changes only `settings.py`. A small fixture written with `Path.write_bytes()` should verify that a corrupt line does not crash either path and that valid neighboring entries remain available.

One non-blocking design question: `errors="replace"` silently accepts invalid bytes inside values, including values for keys recognized as secrets, as replacement characters without a parse warning. That may be acceptable as a resilience policy, but it is worth deciding explicitly rather than leaving it untested.

PR #2619, opened 2026-08-07
fix(dashboard-api): add safe float timeout parsing in host_agent_client.py

mohitchauhan221:
## Problem

In `ods/extensions/services/dashboard-api/host_agent_client.py`, `_timeout()` accepts a `read_seconds` parameter and calculates `max(0.1, float(read_seconds))`. If a `None` or unparseable value is provided, `float()` raises `TypeError` or `ValueError`.

## Fix

Wrap `float(read_seconds)` conversion inside a `try...except (ValueError, TypeError)` block, falling back safely to `5.0` seconds on conversion errors.

## Verification

Verified syntax with `python3 -m py_compile ods/extensions/services/dashboard-api/host_agent_client.py`. `git diff --check` passed cleanly.

### DIFF
--- a/ods/extensions/services/dashboard-api/host_agent_client.py
+++ b/ods/extensions/services/dashboard-api/host_agent_client.py
@@ -99,9 +99,13 @@ def _headers() -> dict[str, str]:
 
 
 def _timeout(read_seconds: float) -> httpx.Timeout:
+    try:
+        read_val = max(0.1, float(read_seconds))
+    except (ValueError, TypeError):
+        read_val = 5.0
     return httpx.Timeout(
         connect=5.0,
-        read=max(0.1, float(read_seconds)),
+        read=read_val,
         write=30.0,
         pool=5.0,
     )


bokiko:
A neutral, class-level note for triage spanning #2619–#2622, so it can be resolved with one policy decision rather than four separate threads.

These four changes add defensive parsing or fallbacks to inputs that are either internal or already type-constrained. Measured against the error-handling policy documented in `CLAUDE.md` — the Design Philosophy priority `Let It Crash > KISS > Pure Functions > SOLID`, and the Error Handling Rules:

> 1. **No broad or silent catches.** Never `except Exception: pass` or `except Exception: return None`. No retry/backoff loops. No fallback chains.
> 2. **Narrow exceptions at I/O boundaries are fine.** … when each maps to a distinct, meaningful status.
> 3. **Internal functions: let exceptions propagate.** The default is zero error handling — errors crash visibly with a full stack trace.

— a few appear to sit on the wrong side of that line:

- **#2619** wraps `float(read_seconds)` in `_timeout()` and falls back to `5.0` on `ValueError`/`TypeError`. `_timeout()` is internal: its call sites pass a literal `5.0` or the `timeout: float = 5.0` parameter of `request_json` (`host_agent_client.py:101`), and `read_seconds` is never sourced from env/user/I/O. A non-numeric value there would be a caller bug, so under rule 3 this masks a programming error behind a silent default rather than being a rule-2 meaningful-status guard.
- **#2620** routes regex matches through `_positive_number()`, but the regex `(\d+(?:\.\d+)?)\s*b` already guarantees a float-parseable, non-negative token, so the guarded `float()` cannot raise or produce a negative — the added safety appears to be a no-op.
- **#2621** and **#2622** are closer to rule 2, but #2621 falls back silently with no warning, and #2622 adds `ValueError` alongside `UnicodeDecodeError` — which is already a subclass of `ValueError`, so the pair is redundant.

None of the four include a regression test.

Scope note: this applies only to the defensive-parsing style above. Several other PRs are substantive behavioral fixes and are explicitly out of scope for this note — #2607, #2608, #2609, #2610 (sudo-context user/home/port resolution and update sequencing), and #2613, #2614, #2616 (compose-overlay handling, a null-glob guard, and an atomic env write). Those should be judged on their own merits.

PR #2642, opened 2026-08-08
fix(litellm): close Token Spy client when worker is cancelled

### LINKED ISSUE #2336: ODSTokenSpyCallback leaks httpx.AsyncClient connections on LiteLLM worker restart

patil2001:
## Summary

Closes #2336. `ODSTokenSpyCallback._run()` created an `httpx.AsyncClient` inside an async-with each time the worker task was (re)created. Under cancellation (LiteLLM model swap, gunicorn recycle, asyncio shutdown race) the `__aexit__` could be skipped, leaking sockets and file descriptors that compounded with every restart.

## Change

Create the client explicitly and close it in a `finally` block so `aclose` runs on both cancellation and normal exit.

## Test plan

- Manual: cancelled the worker during a model swap and confirmed the client handle is released.

### DIFF
--- a/ods/extensions/services/litellm/ods_token_spy_callback.py
+++ b/ods/extensions/services/litellm/ods_token_spy_callback.py
@@ -195,9 +195,8 @@ class ODSTokenSpyCallback(CustomLogger):
         timeout = max(
             0.1, float(os.environ.get("ODS_LITELLM_TELEMETRY_TIMEOUT", "3"))
         )
-        async with httpx.AsyncClient(
-            follow_redirects=False, timeout=timeout
-        ) as client:
+        client = httpx.AsyncClient(follow_redirects=False, timeout=timeout)
+        try:
             while True:
                 event = await self.queue.get()
                 try:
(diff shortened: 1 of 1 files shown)


patil2001:
Adding context for reviewers: this resolves "ODSTokenSpyCallback leaks httpx.AsyncClient connections on LiteLLM worker restart". fix(litellm): close Token Spy client when worker is cancelled closes the gap so the affected path fails safe instead of misbehaving.

PR #2645, opened 2026-08-08
fix(ods-cli): split env diff on first '=' and strip CR/quotes

### LINKED ISSUE #2297: cmd_preset diff env parsing truncates values at first =

patil2001:
## Summary

Closes #2297. `cmd_preset` env comparison used `IFS== read` which split on every `=`, truncating JWT/base64 secrets and query-string URLs at the first `=`. Both sides truncated identically, so real differences after the first `=` were reported as "no difference".

## Change

Split on the first `=` only and strip trailing CR and matching surrounding quotes, matching `lib/safe-env.sh`.

## Test plan

- Compared a `.env` whose values contain embedded `=`; diff now reports genuine changes.

### DIFF
--- a/ods/ods-cli
+++ b/ods/ods-cli
@@ -3370,13 +3370,52 @@ META
 
                 # Parse both env files
                 declare -A env1 env2
-                while IFS='=' read -r key value; do
-                    [[ "$key" =~ ^[[:space:]]*# ]] && continue
+                while IFS= read -r _line || [[ -n "$_line" ]]; do
+                    # Strip a trailing CR so CRLF .env files (Windows editors,
+                    # the Windows installer) don't leave carriage returns on
+                    # values, matching lib/safe-env.sh load_env_file.
+                    _line="${_line%$'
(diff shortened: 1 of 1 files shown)


patil2001:
The why behind this change: this resolves "cmd_preset diff env parsing truncates values at first =". fix(ods-cli): split env diff on first '=' and strip CR/quotes pins the behaviour down so affected installs don't end up in a broken state.

PR #2646, opened 2026-08-08
fix(host-agent): verify opencode provider npm, model name, and output limit

patil2001:
## Summary

`_opencode_config_matches` only compared model/small_model, baseURL, apiKey, and limit.context, so a wrong provider package, served name, or a clamped output limit went undetected before apply was declared stable.

## Change

It now also verifies the provider npm is `@ai-sdk/openai-compatible`, the model name matches the routed display name when provided, and limit.output is at least `min(32768, context_length)`.

## Test plan

- Ran host-agent with a mismatched provider package and confirmed the apply path now flags it.

### DIFF
--- a/ods/bin/ods-host-agent.py
+++ b/ods/bin/ods-host-agent.py
@@ -10774,6 +10774,7 @@ def _opencode_config_matches(
     base_url: str,
     api_key: str,
     context_length: int,
+    display_name: str | None = None,
 ) -> bool:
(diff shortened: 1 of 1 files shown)


patil2001:
Why this is necessary: fix(host-agent): verify opencode provider npm, model name, and output limit. This is a contained fix for an edge case that would otherwise leave the stack in the wrong state on that path.

PR #2648, opened 2026-08-08
fix(installer): make Windows enable/disable use bash correctly

patil2001:
## Summary

The Windows enable/disable flow invoked bash in a way that broke for users whose WSL bash was not on the default PATH, or when the ODS install dir contained spaces.

## Change

Resolve the bash executable explicitly and quote the command invocation so enable/disable works regardless of PATH layout or spaced paths.

## Test plan

- Windows install under a spaced directory: `ods enable`/`ods disable` now succeed.

### DIFF
--- a/ods/installers/windows/ods.ps1
+++ b/ods/installers/windows/ods.ps1
@@ -1960,7 +1960,17 @@ function Invoke-Enable {
     $disabledPath = Join-Path $svcDir "compose.yaml.disabled"
 
     if (Test-Path $composePath) {
-        Write-AISuccess "$ServiceId is already enabled."
+        # Installer-skipped services also carry a plain compose.yaml but are
+        # absent from .compose-flags -- "already enabled" would be misleading
+        # and the service would never join the stack. Add the missing entry.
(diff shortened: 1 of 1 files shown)


patil2001:
Closing: this fix is superseded by the Windows CLI refactor that landed on `main`. Verified against current `main`:

1. The old bash-resolver call site this PR patched (`$resolverScript`/`$bashExe`, `Path 1: delegate to the canonical resolver`) no longer exists -- `Update-ComposeFlags` is now native PowerShell and shells out to no bash resolver, so the `execvpe(/bin/bash) failed` crash and the `NativeCommandError` hazard are gone.
2. Single-service delta editing is present: `Update-ComposeFlags -ServiceId $ServiceId -Action "enable"|"disable"` edits only the toggled entry (`ownedByService` match + `InsertRange`).
3. The installer-skipped-service gap is covered: `Invoke-Enable` calls `Update-ComposeFlags -Action "enable"` (lines 3254/3273), which adds the missing entry when `compose.yaml` exists.

Rebasing would produce an empty/duplicate Windows diff, so closing instead. Thanks for the review.

PR #2649, opened 2026-08-08
fix(cli): mask secrets in Windows config show output

patil2001:
## Summary

`ods config show` on Windows printed API keys and tokens in plaintext. The mask applied on Linux/macOS was not applied in the Windows implementation.

## Change

Apply the same masking to sensitive keys in the Windows config show path.

## Test plan

- Windows: `ods config show` now redacts secret values.


patil2001:
Closing as superseded by #1795, which was merged (the same Windows `config show` secret-masking fix, broader policy). The change is already on `main`, so this branch conflicts and is no longer needed.

PR #2706, opened 2026-08-11
fix(cli): repair preset-diff env parser; bind opencode to 0.0.0.0

azilber:
## Summary

Two independent fixes surfaced while debugging a broken `ods status` on a live host.

### 1. `ods-cli` — `preset diff` syntax error broke every `ods` command
The env-parsing block in `cmd_preset`'s `diff` path was a half-completed merge of two upstream preset-diff PRs (the `_cmd_preset_env_entries` helper refactor + the superseded inline `_line` parser). It left an orphan `while … do` with no matching `done`, plus two stray inline parsers duplicating the helper. Because this is a bash **parse** error, the *entire* script failed to load — so `ods status` (and every other subcommand) died with:

```
/usr/local/bin/ods: line 3526: syntax error near unexpected token `fi'
```

Fix: collapse the block into two clean null-delimited loops that reuse the existing `_cmd_preset_env_entries` helper (the intended end state of the refactor).

### 2. `opencode-web.service` — coding assistant unreachable from the LAN
The unit bound `--hostname 127.0.0.1`, so opencode on `:3003` was reachable only from localhost — unlike the Docker-published dashboard on `:3001` (`0.0.0.0`). Bind `--hostname 0.0.0.0` so it's reachable on the LAN. No-auth (`UnsetEnvironment=OPENCODE_SERVER_PASSWORD`) is retained intentionally, consistent with the localhost default's no-login behavior.

> ⚠️ This exposes an unauthenticated code editor (shell/filesystem access) to the LAN. Appropriate for a trusted homelab; revert to `127.0.0.1` if that assumption changes.

## Verification
- `bash -n ods/ods-cli` — clean; `ods status` runs and lists services.
- `ss -tlnp` shows `0.0.0.0:3003`; `curl -sI http://<lan-ip>:3003/` → `HTTP/1.1 200 OK`.
- `bash tests/test-local-app-auth-defaults.sh` — PASS (only the **Windows** template asserts `--hostname 127.0.0.1`; the Linux template has no hostname assertion).
- `bash tests/test-linux-opencode-path.sh` — PASS.


azilber:
Opened against the wrong base — recreating in the azilber/ODS fork.

PR #2708, opened 2026-08-11
fix(egress): bound inbound request buffering

tang-vu:
## Summary
- enforce `ODS_REMOTE_PROVIDER_MAX_BODY_BYTES` while consuming the ASGI request stream
- reject oversized declared bodies before reading any chunks
- stop reading chunked/unknown-length bodies immediately when they cross the configured limit
- cover the public forwarding boundary for oversized, malformed-length, and chunked requests

## Why this matters
The egress policy already advertises a 4 MiB request limit, but `forward()` called `await request.body()` before `prepare_upstream_request()` checked that limit. A large OpenAI-compatible request was therefore buffered completely in the egress process before being rejected, so the configured bound did not actually bound memory use. This is reachable through the production LiteLLM-to-egress forwarding path and is especially costly under concurrent requests. The new reader enforces the existing contract during ingestion and preserves the same structured 413 response.

## Overlap check
Searched open and closed PRs for `egress payload_too_large`, `egress request body limit`, `remote provider max body bytes`, `request.body MAX_BODY_BYTES`, and chunked-request variants. No PR covers inbound buffering. PR #2700 changes only `/probe` thread offloading and PR #2702 changes only direct-client cache eviction; both touch `app/main.py`, but neither changes `forward()` body ingestion or the existing body-limit contract. PR #2197 only declares the compose environment key. This behavior and its tests are independent.

## Test plan
- [x] `python ods/tests/contracts/test-remote-provider-egress-service.py`
- [x] `python -m py_compile ods/extensions/services/remote-provider-egress/app/main.py ods/tests/contracts/test-remote-provider-egress-service.py`
- [x] `git diff --check`

Generated with Codex

### DIFF
--- a/ods/extensions/services/remote-provider-egress/app/main.py
+++ b/ods/extensions/services/remote-provider-egress/app/main.py
@@ -149,6 +149,35 @@ def _response_headers(headers: Mapping[str, str]) -> dict[str, str]:
     }
 
 
+async def _bounded_request_body(request: Request) -> bytes:
+    """Read an inbound body without buffering beyond the configured limit."""
+    content_length = request.headers.get("content-length")
+    if content_length is not None:
+        try:
+            declared_bytes = int(content_length)
+        except ValueError as exc:
+            raise EgressError(
+                400,
+                "invalid_content_length",
+                "request Content-Length is invalid",
+            ) from exc
+        if declared_bytes < 0:
+            raise EgressError(
+                400,
+                "invalid_content_length",
+                "request Content-Length is invalid",
+            )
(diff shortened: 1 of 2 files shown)


tang-vu:
CI follow-up: the first integration-smoke run exposed that this contract job intentionally does not install the egress service's FastAPI/httpx runtime dependencies. I updated the public-boundary test harness to inject minimal HTTP framework stubs while still importing the production module and invoking the real forward() endpoint. The dependency-isolated 18-test contract now passes locally and the full rerun is green (pending 0, failed 0).

PR #2714, opened 2026-08-11
fix(installer): add Homebrew backend; fix mDNS zeroconf on PEP 668 hosts

azilber:
## Summary
- **Homebrew backend in `packaging.sh`** — `_brew_bin()`/`_brew_run()` (never sudo; drops to `SUDO_USER` since brew refuses root), `HAS_BREW`/`BREW_BIN` detection, `PKG_MANAGER=brew` on macOS, and `brew)` cases in `pkg_install`/`pkg_update`/`pkg_available`/`pkg_resolve`. On Linux the native manager (apt/dnf/…) stays authoritative so Docker/system daemons are never installed from brew.
- **Fix mDNS zeroconf on PEP 668 hosts (`07-devtools.sh`)** — the announcer step checked/installed zeroconf against bare `python3`, which on Homebrew (Linuxbrew) hosts resolves to brew's externally-managed Python. `apt install python3-zeroconf` landed in a different interpreter and `pip --user` was rejected by PEP 668. Now the interpreter is resolved once via `ods_detect_python_cmd` (honors `ODS_PYTHON_PREFER_SYSTEM`) and reused for every import check, the install, and the systemd unit's `__PYTHON3__`; install goes through `pkg_install`/`pkg_resolve` with an `ods_python_pip_install_user` fallback that adds `--break-system-packages`. Adds a `python3-zeroconf` name mapping.
- Also toggles local service compose enablement (langfuse on; brave-search, comfyui, ods-proxy, openclaw, tailscale off).

## Test plan
- [x] `make lint` passes; both shell files clean at shellcheck `severity=warning` (CI fails only on `error`)
- [x] `tests/contracts/test-installer-hardening.sh` passes
- [x] Functional check on a Linuxbrew host: `PKG_MANAGER=apt` + `HAS_BREW=1` + brew detected at `/home/linuxbrew/.linuxbrew/bin/brew`; resolver mappings correct across apt/dnf/pacman/zypper/apk/xbps/brew
- [x] Verified `MDNS_PYTHON` resolves to `/usr/bin/python3` (where zeroconf imports) instead of brew's PEP-668 Python — the announcer now starts


azilber:
Opened against the wrong base by mistake — this belongs on the azilber/ODS fork, not upstream. Closing.

PR #2804, opened 2026-08-13
fix(installer): harden rootless Podman setup

superm1:
## Summary

Hardens ODS installation and lifecycle behavior for rootless Podman and no-sudo Linux hosts without weakening rootful Docker installs.

- detects rootless Podman through the Docker-compatible CLI
- adds `docker.io` to Podman's effective short-name search list without replacing operator registries
- distinguishes optional host setup from required privileged mutations
- repairs bind-mount ownership inside the rootless user namespace
- starts and health-proves a per-session Host Agent when a system service cannot be installed
- preserves existing cloud credentials and unrelated Compose/service state on reruns

## Root Causes

The installer previously treated a missing or unusable `sudo` command as a successful no-op. Required package, driver, ownership, firewall, and runtime operations could therefore appear successful even when no host mutation occurred. Podman configuration also returned early for any existing search registry and could write a top-level key after a TOML registry table. Rootless bind mounts and the session Host Agent then had incomplete ownership and lifecycle handling.

## Behavior And Invariants

|---|---|
| Never report a required privileged mutation as successful when it did not run | Required runtime, `jq`, NVIDIA toolkit/driver, rootful ownership, and AMD Docker downgrade paths use the shared privilege contract and fail or warn according to whether the operation is required. |
| Preserve operator registry configuration | Podman's effective search order is read first, `docker.io` is appended only when absent, existing registry tables remain untouched, and `CONTAINERS_REGISTRIES_CONF` is honored. |
| Keep registry writes recoverable | Registry updates use a same-directory atomic replacement, preserve file mode and symlink targets, reject malformed arrays, and are idempotent. |
| Keep rootless repairs confined to ODS | Host-writable repairs accept only `config/*`; service ownership repairs accept only `data/*`; path traversal and symlink escapes fail closed. |
| Do not mutate live service data unexpectedly | A running container is accepted only when metadata already matches; mismatches require the service to be stopped before repair. |
| Preserve service-specific ownership contracts | Hermes, Token Spy, APE, n8n, Whisper, Privacy Shield, ComfyUI, and Langfuse keep their declared UID/GID and mode behavior. |
| Preserve rerun state | Existing API credentials, `.env` values, disabled services, Compose fragments, and unrelated service ownership are not rewritten. |
| Prove the session Host Agent before reporting success | Startup validates the `/health` JSON contract, stale PID files cannot terminate unrelated processes, and the agent is restarted after `ods-network` exists. |

## Scenario Coverage

| Scenario | Result |
|---|---|
| Clean rootless Podman install | Covered by registry, detection, ownership, and session-agent contracts. |
| Existing Podman configuration | Existing search order and registry tables are preserved; rerun is idempotent. |
| Existing ODS installation / second run | Credentials and unrelated service state remain unchanged. |
| Empty and invalid overrides | Empty UID/GID values use service defaults; invalid values fail before mutation. |
| Disabled service | Ownership repair skips services absent from the active Compose stack. |
| Partial privilege availability | Optional tuning/firewall/system-service work is skipped with an actionable warning; required work fails clearly. |
| Partial ownership failure | Failure propagates and no success receipt is emitted. |
| Session-agent stale PID or startup failure | Unowned PIDs are refused; unhealthy startup is terminated and reported. |
| Rootful Docker | Existing lifecycle remains in place; rootful repairs require real privilege. |
| macOS / Windows | Linux rootless ownership helper is a side-effect-free no-op; platform-specific installers are unchanged. |

## Validation

Executed against `main` at `5a4450765976e2ad2792b9ac8927f4873dac60f6`:

```text
bash ods/tests/test-podman-rootless-contracts.sh
  24/24 passed

bash ods/tests/test-rootless-docker-ownership.sh
  27/27 passed

bash ods/tests/test-hermes-data-ownership.sh
  passed

  passed

bash -n <all changed shell scripts>
  passed

shellcheck --severity=error <all changed shell scripts>
  passed

git diff --check
  passed
```

The contributor's original unprivileged install receipt also completed with:

```bash
/home/test/ODS/ods/install-core.sh --non-interactive --skip-docker --no-comfyui
```

The Bats firewall regression was added but Bats was not available in the local Windows audit environment; the repository CI executes that lane.

## Platform Scope

- **Primary behavior change:** Linux rootless Podman and no-sudo installation/lifecycle.
- **Regression coverage:** Linux rootful Docker, enabled/disabled service selection, AMD/NVIDIA privilege gates, and Host Agent fallback behavior.
- **Unaffected:** Windows and macOS installer/runtime implementations. Shared ownership detection explicitly returns without mutation outside native Linux.

## Risk And Release Lane

- **Risk:** High, because installer, host mutation, bind-mount ownership, and Host Agent lifecycle paths are involved.
- **Release lane:** Mainline change targeting `main`.
- **Rollback:** Revert the commits in this PR. Registry writes retain the previous effective registry list plus `docker.io`; no service data is deleted.
- **Operational validation:** Focused contracts and repository CI are recorded here. A physical rootless Podman fleet receipt remains useful release evidence but is not represented as completed by the Windows audit environment.

## AI Assistance

AI-assisted code review, edge-case enumeration, and regression-test drafting were used. The implementation, scope, validation evidence, and final diff were reviewed by the author.


Lightheartdevs (approved):
Approved at exact head 5b81c8c076597262cce956044bc01eca88334d6a. The newly reachable password-sudo Docker fallback is privilege-gated and returns promptly; the regression test passes 5/5. Rootless Podman 24/24, rootless ownership 27/27, Hermes and installer contracts, a disposable physical rootless Podman run, synthetic integration, and all 28 GitHub checks pass. The final test-only commit mirrors production sudo bootstrap in the env smoke harness. Recommended merge priority: 3.

PR #2950, opened 2026-08-22
fix(preflight): name the install dir when the log cannot be created

### LINKED ISSUE #2930: Preflight dies without a word when the install dir cannot hold the log
`ods/ods-preflight.sh:100` opens the log with `echo "" > "$LOG_FILE"` (path derived at line 19 from `$ODS_DIR`). Under `set -euo pipefail`, running against a not-writable-by-you install dir (e.g. root-owned) kills the script right there — no message, exit code only.

**Impact:** users hit a silent preflight death and report it as "installer does nothing"; the actual problem (permissions on `$ODS_DIR`) is never named.

**Suggested direction:** probe writability up front (`: > "$LOG_FILE" ||`) and print an actionable error naming the path.

bettercalln1ck:
## Summary

Fixes #2930.

`ods/ods-preflight.sh:100` opened its log with a bare redirect:

```bash
LOG_FILE="$ODS_DIR/preflight-$(date +%Y%m%d-%H%M%S).log"   # line 19
...
echo "" > "$LOG_FILE"                                       # line 100
```

Run against an install dir the invoking user cannot write (a root-owned `$ODS_DIR` is the common case), `set -euo pipefail` exits right there. All the user gets is bash's terse `Permission denied` — and nothing at all when the caller captures stderr, which is how this surfaces as "the installer does nothing".

Probe writability explicitly and print the path, the user, and the `chown` that fixes it before exiting non-zero.

## AI Assistance

## Release Lane

Stable hotfix reason:

```text
n/a
```

## Changed Surface

- [ ] Docs only
- [ ] Tests only
- [ ] Dashboard UI

## Risk And Validation

- Risk level: Low
- Validation run:
  - [x] `git diff --check`
  - [x] Not required because: the change only replaces a bare redirect with a guarded one on the failure path; the success path is byte-identical and is covered by the existing preflight tests.

Commands/results:

```text
$ make lint
=== Shell syntax ===
=== Python compile ===
All lint checks passed.

$ bash tests/test-preflight-log-writable.sh
PASS: names the failure
PASS: names the install dir
All preflight log-writability tests passed

# same test against the unpatched script -> fails, so it guards the regression
$ git stash && bash tests/test-preflight-log-writable.sh
FAIL: no actionable error; output was: .../ods-preflight.sh: line 100: .../preflight-....log: Permission denied
exit=1

$ shellcheck --exclude=SC1091,SC2034 --severity=warning \
    ods/ods-preflight.sh ods/tests/test-preflight-log-writable.sh
(clean)
```

## Operational Change Check

## Notes For Reviewers

- New `tests/test-preflight-log-writable.sh` stages a read-only install dir and asserts the error names both the failure and the path. Wired into `make test`.
- The test skips itself when run as root, since root ignores the write bits and the condition cannot be staged.
- No behaviour change when the log is writable.

### DIFF
--- a/ods/Makefile
+++ b/ods/Makefile
@@ -43,6 +43,7 @@ test: ## Run unit and contract tests
 	@bash tests/test-linux-install-preflight.sh
 	@bash tests/test-linux-host-agent-stop.sh
 	@bash tests/test-preflight-dashboard-port.sh
+	@bash tests/test-preflight-log-writable.sh
 	@echo ""
 	@echo "=== ods-doctor rootless subordinate IDs ==="
 	@bash tests/test-ods-doctor-rootless-subid.sh
--- a/ods/ods-preflight.sh
+++ b/ods/ods-preflight.sh
@@ -97,7 +97,14 @@ pass() { log "${GREEN}✓${NC} $1"; PASS=$((PASS+1)); }
 fail() { log "${RED}✗${NC} $1"; FAIL=$((FAIL+1)); }
 warn() { log "${YELLOW}⚠${NC} $1"; WARN=$((WARN+1)); }
 
-echo "" > "$LOG_FILE"
+# Name the problem instead of dying on the bare redirect: under `set -e` an
+# unwritable install dir exits here with a status and no actionable context.
+if ! : > "$LOG_FILE"; then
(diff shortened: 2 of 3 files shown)


bettercalln1ck:

PR #2955, opened 2026-08-22
fix(session-cleanup): do not prune sessions created mid-cycle

### LINKED ISSUE #2931: session-cleanup deletes sessions created while cleanup was running (snapshot race)
`ods/scripts/session-cleanup.sh:143` iterates `"$SESSIONS_DIR"/*.jsonl` and treats anything absent from the earlier-read active set as inactive. A session created *between* that read and the deletion pass is not in the active set yet — but is very much live.

**Impact:** the live gateway's just-created session gets deleted and its index entry dangles after the rewrite; the user's conversation disappears mid-use.

**Repro:** start a new session while a cleanup cycle runs; observe it vanish.

**Suggested direction:** record a snapshot timestamp before reading the active set and skip files whose mtime is newer than it.

bettercalln1ck:
## Summary

Fixes #2931.

`ods/scripts/session-cleanup.sh` reads the active session set from `sessions.json` (line ~76), then iterates `"$SESSIONS_DIR"/*.jsonl` (line ~143) and deletes anything not in that set. A session the gateway creates *between* the read and the deletion pass is not in the active set yet — but it is live, and it gets deleted mid-conversation, leaving a dangling index entry once sessions.json is rewritten.

The fix stamps a marker file in `SESSIONS_DIR` immediately *before* the active set is read, and the deletion pass skips any session file newer than that marker:

```bash
SNAPSHOT_REF="$SESSIONS_DIR/.cleanup-snapshot.$$"
: > "$SNAPSHOT_REF"
trap 'rm -f "$SNAPSHOT_REF"' EXIT
...
if [ "$f" -nt "$SNAPSHOT_REF" ]; then
    echo "[$(date)] Skipping session newer than this cleanup cycle: $BASENAME"
    continue
fi
```

A skipped session is not leaked — the next cycle reads an index that now contains it and judges it normally. The marker lives in `SESSIONS_DIR` so its mtime comes from the same filesystem clock as the files it is compared against, and its name (leading dot) cannot be caught by the `*.jsonl`, `*.deleted.*`, or `*.bak*` passes.

## AI Assistance

## Release Lane

Stable hotfix reason:

```text
n/a
```

## Changed Surface

- [ ] Docs only
- [ ] Tests only
- [ ] Dashboard UI

## Risk And Validation

- Risk level: Medium
- Validation run:
  - [x] `git diff --check`
  - [x] Not required because: the change is one guard in one loop in a single script; the new test covers the raced session, the active session, and the genuinely stale session so the guard is shown not to be over-broad.

Commands/results:

```text
$ bash tests/test-session-cleanup-race.sh
PASS: session created mid-cleanup survives
PASS: active session retained
PASS: genuinely inactive session still pruned
PASS: snapshot marker cleaned up
All session-cleanup race tests passed

# against the unpatched script
$ git stash push -- ods/scripts/session-cleanup.sh && bash tests/test-session-cleanup-race.sh
FAIL: session created mid-cleanup was deleted
PASS: active session retained
PASS: genuinely inactive session still pruned
exit=1

$ make lint
=== Shell syntax ===
=== Python compile ===
All lint checks passed.

$ shellcheck --exclude=SC1091,SC2034 --severity=warning \
    ods/scripts/session-cleanup.sh ods/tests/test-session-cleanup-race.sh
(clean)
```

## Operational Change Check

## Notes For Reviewers

- New `tests/test-session-cleanup-race.sh`, wired into `make test`. It stages the race by timestamp ordering (session mtime ahead of any marker the run can stamp) rather than trying to win a real race, so it is deterministic rather than flaky. The comment in the test says so explicitly.
- The test also asserts the genuinely-inactive session is *still* pruned, so the guard cannot regress into "never delete anything".
- The marker is removed by an `EXIT` trap; the test checks `SESSIONS_DIR` is clean afterwards. This is the script's first trap, so nothing was overwritten.

### DIFF
--- a/ods/Makefile
+++ b/ods/Makefile
@@ -77,6 +77,9 @@ test: ## Run unit and contract tests
 	@bash tests/test-bind-address-sweep.sh
 	@bash tests/test-n8n-cookie-policy.sh
 	@echo ""
+	@echo "=== session cleanup snapshot race ==="
+	@bash tests/test-session-cleanup-race.sh
+	@echo ""
 	@echo "=== config migration (version check under set -e) ==="
 	@bash tests/test-migrate-config.sh
 	@echo ""
--- a/ods/scripts/session-cleanup.sh
+++ b/ods/scripts/session-cleanup.sh
@@ -71,6 +71,16 @@ if [[ -z "$PYTHON_CMD" ]] || ! "$PYTHON_CMD" -c 'import json' >/dev/null 2>&1; t
     exit 1
 fi
 
+# Snapshot marker, stamped *before* the active set is read. A session created
+# between that read and the deletion pass below is not in the active set yet,
+# but it is very much live — judging it by this snapshot would delete the
+# user's just-opened conversation. Files newer than the marker are left alone
+# and picked up by the next run. It lives in SESSIONS_DIR so its mtime comes
+# from the same filesystem clock as the session files it is compared against.
+SNAPSHOT_REF="$SESSIONS_DIR/.cleanup-snapshot.$$"
+: > "$SNAPSHOT_REF"
+trap 'rm -f "$SNAPSHOT_REF"' EXIT
+
 ACTIVE_IDS_OUTPUT=""
 ACTIVE_IDS_EXIT=0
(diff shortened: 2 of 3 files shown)


bettercalln1ck:

PR #2958, opened 2026-08-22
fix(token-spy): stop remote session ids re-parsing in the remote shell

### LINKED ISSUE #2929: token-spy builds remote rm argv by string interpolation, letting session paths re-parse in the remote shell
`ods/extensions/services/token-spy/session-manager.sh:293` executes

```bash
ssh ... "${host}" "rm -f ${rm_args}"
```

`${rm_args}` is interpolated straight into a remote shell command line. A session file name containing shell metacharacters (they originate from remote gateway state) re-parses on the far end — extra flags or paths can reach `rm`.

**Impact:** crafted or merely unlucky path names turn a cleanup routine into arbitrary file deletion on the remote host. Defense-in-depth: today the names come from our own registry, but the registry is populated from remote output.

**Suggested direction:** pass paths NUL-delimited through stdin (`xargs -0`) or quote each argument explicitly before interpolation.

bettercalln1ck:
## Summary

Fixes #2929.

`ods/extensions/services/token-spy/session-manager.sh:293` (pre-patch):

```bash
local rm_args=""
for sid in "${to_remove[@]}"; do
  rm_args="${rm_args} ${remote_dir}/${sid}.jsonl"
done
ssh -o ConnectTimeout=5 -o BatchMode=yes "${host}" "rm -f ${rm_args}" 2>/dev/null || true
```

`${rm_args}` is interpolated into a command line that the remote shell parses. The ids in it are read out of remote gateway output (`manage_remote_agent` parses them from the inventory `ssh` at line 201), so a session file whose name carries shell metacharacters re-parses on the far end — extra flags, extra paths, or an entire second command can reach `rm`.

Changes:

- Names go NUL-delimited on stdin; `xargs -0 rm -f` builds rm's argv directly, so the remote shell never parses them.
- The command line keeps `cd ${remote_dir}` unquoted, preserving the existing remote expansion of a `~`-relative operator-configured sessions dir. The chdir also means the ids carry no path of their own.
- Ids containing `/` are refused and logged rather than sent at all — belt and braces against a traversal-shaped name.
- The `2>/dev/null || true` is replaced with a `[WARN]` log, per the repo's error-handling rules; a failed remote cleanup was previously indistinguishable from a successful one.

## AI Assistance

## Release Lane

Stable hotfix reason:

```text
n/a
```

## Changed Surface

- [ ] Docs only
- [ ] Tests only
- [ ] Dashboard UI

## Risk And Validation

- Risk level: Medium
- Validation run:
  - [x] `git diff --check`
  - [x] Not required because: the change is confined to the cleanup branch of one function, and the new test drives that function for real against a stubbed host rather than asserting on a constructed string.

Commands/results:

```text
$ bash tests/test-token-spy-remote-rm.sh
Test 1: an injected command in a session id never executes remotely
  + no injected side effect
Test 2: the literal session file is still the thing that got removed
  + session with metacharacters removed by exact name
Test 3: a path-like session id is refused, not deleted through
  + path-like id skipped and logged
Test 4: ordinary cleanup still works
  + inactive removed, active kept
All token-spy remote cleanup tests passed

# the same test against the unpatched script — the injection actually fires
$ git stash push -- ods/extensions/services/token-spy/session-manager.sh
$ bash tests/test-token-spy-remote-rm.sh
  x injected 'touch pwned' ran on the remote host
exit=1

$ make lint
=== Shell syntax ===
=== Python compile ===
All lint checks passed.

# CI gates on error severity; clean there. The two SC2087 warnings on the
# inventory heredoc at line 201 are pre-existing on main (verified by running
# shellcheck against the stashed file) and are deliberate — the comment above
# that heredoc says ${remote_dir} must expand client-side.
$ shellcheck --exclude=SC1091,SC2034 --severity=error \
    ods/extensions/services/token-spy/session-manager.sh \
    ods/tests/test-token-spy-remote-rm.sh
(clean)
```

## Operational Change Check

## Notes For Reviewers

- New `tests/test-token-spy-remote-rm.sh`, wired into `make test`. It follows the existing `test-token-spy-session-manager.sh` pattern of extracting the function under test, and stubs `ssh` with a script that **actually runs** the command line it is handed, in a real directory, with real stdin — so an injected command genuinely executes and leaves a file behind if the guard fails. It is a demonstration, not a mock assertion.
- Test 4 pins the ordinary path (inactive removed, active kept) so the new guard cannot regress into "never delete anything".
- `xargs -0` is assumed present on the remote host; it is in GNU findutils and busybox alike. If ODS supports a remote target where that is not true, say so and I will fall back to per-argument quoting instead.
- Line 391 also interpolates `${remote_dir}` into a remote command, but that value comes from the operator's local `REMOTE_AGENTS` config rather than from remote output, so it is outside the reported bug. Left alone.

### DIFF
--- a/ods/Makefile
+++ b/ods/Makefile
@@ -140,6 +140,9 @@ test: ## Run unit and contract tests
 	@echo "=== token-spy session manager active-id extraction ==="
 	@bash tests/test-token-spy-session-manager.sh
 	@echo ""
+	@echo "=== token-spy remote session cleanup argv safety ==="
+	@bash tests/test-token-spy-remote-rm.sh
+	@echo ""
 	@echo "=== fleet-multi-distro keep-going on docker pull failure ==="
 	@bash tests/test-fleet-multi-distro-keep-going.sh
 	@echo ""
--- a/ods/extensions/services/token-spy/session-manager.sh
+++ b/ods/extensions/services/token-spy/session-manager.sh
@@ -286,12 +286,30 @@ REMOTESCRIPT
   log "  Sessions: $session_count total, ${#to_remove[@]} to remove"
 
   if [ "${#to_remove[@]}" -gt 0 ]; then
-    local rm_args=""
+    # Session ids come from remote gateway output. Interpolating them into the
+    # remote command line lets shell metacharacters re-parse over there, so send
+    # them NUL-delimited on stdin and let xargs build rm's argv directly — the
+    # remote shell never sees them. The directory stays on the command line
+    # (unquoted, as before) so a leading '~' still expands remotely, and the
+    # chdir keeps the ids themselves free of any path of their own.
+    local rm_names=()
+    local sid
     for sid in "${to_remove[@]}"; do
-      rm_args="${rm_args} ${remote_dir}/${sid}.jsonl"
+      case "$sid" in
+        */*)
+          log "  [SKIP] Refusing path-like session id: $sid"
+          continue
+          ;;
+      esac
+      rm_names+=("${sid}.jsonl")
     done
-    ssh -o ConnectTimeout=5 -o BatchMode=yes "${host}" "rm -f ${rm_args}" 2>/dev/null || true
-    log "  [DONE] Removed ${#to_remove[@]} sessions on $host"
+
+    if [ "${#rm_names[@]}" -gt 0 ]; then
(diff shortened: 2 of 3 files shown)


bettercalln1ck:

PR #4184, opened 2026-09-10
fix: propagate exit code from _check_version_compat in cmd_update

### LINKED ISSUE #4179: fix: Update compat abort maps to exit 0
﻿## Summary

User blocks/cancels compat ? `return 1` swallowed as `exit 0`.

**Impact:** Production user/operator failure on Osmantic/ODS main

**Files:**
- `ods/ods-cli (cmd_update + _check_version_compat`
- `:1988-1990)`

## Repro

On a production-like Osmantic/ODS main checkout:

1. Hit the failure mode described in Summary (severity: High).
2. Observe the broken behavior / incorrect exit / missing guard from the root cause above.
3. Confirm no overlapping open PR (dedup: Related but distinct from #3873 (general prompt EOF).).

## Fix

Implement a minimal, distro-/path-correct fix in the listed files (no scope creep).

- Address the root cause directly (do not paper over with `|| true` / broad catches).
- Prefer resolving runtime values from the environment/repo (like `apt-cache madison`) over hardcoding distro-specific version strings when the same bug class applies.
- Add a regression test under `ods/tests/` (or matching suite) that fails without the fix and passes with it.

## Context

Filed from Desktop `issues.md` entry #12 via one-click scout pipeline.
Rationale: Unfiled Desktop issues.md #12 (no GitHub issue yet)

patil2001:
## Summary
Fixes #4179: When user blocks/cancels compat check, _check_version_compat returns 1 which was being swallowed as exit 0 in cmd_update. Now correctly exits with the failure code.

## Changes
- Changed `if ! _check_version_compat "$force_flag"; then exit 0; fi` to `_check_version_compat "$force_flag" || exit $?` in ods/ods-cli

Closes #4179

### DIFF
--- a/ods/ods-cli
+++ b/ods/ods-cli
@@ -2006,9 +2006,7 @@ cmd_update() {
     read -ra flags <<< "$flags_str"
 
     # Version compatibility check — warn on major version jumps before touching anything
-    if ! _check_version_compat "$force_flag"; then
-        exit 0
-    fi
+    _check_version_compat "$force_flag" || exit $?
 
     # Show what we're updating from/to (if versions were resolved)
     if [[ -n "${_COMPAT_INSTALLED_VER:-}" && -n "${_COMPAT_TARGET_VER:-}" \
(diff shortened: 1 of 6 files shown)


ods/ods-cli
```diff
@@ -2006,9 +2006,7 @@ cmd_update() {
     read -ra flags <<< "$flags_str"
 
     # Version compatibility check — warn on major version jumps before touching anything
-    if ! _check_version_compat "$force_flag"; then
-        exit 0
-    fi
+    _check_version_compat "$force_flag" || exit $?
```
patil2001, line 2009: here

PR #4215, opened 2026-09-11
Fix public-beta fleet install and Pixel runtime blockers

Lightheartdevs:
## Merged public-beta handoff (2026-09-16, 21:05 UTC)

**Merged** as `4fa6ad17059a2ff7df1dd92105b87ed84b801e70` into `public-beta` from exact PR head `8000397c31fada14b66a060ec5046278fa1cf04a`, on prior beta `5ab1e08302014b3009284e6c6083dc33db4aabe0`. The merged tree was verified byte-for-byte identical to the reviewed PR candidate. The exact head had 43 successful GitHub checks and three skipped checks, with no failing or pending checks. This is a **beta integration receipt, not five-machine installed acceptance**.

The final audit corrected two merge seams: distinct root IPC names for browser model switching versus installer model promotion (`71e8c6a2b`), and fail-closed verification/removal of Pixel provider service files before deleting their only custody record (`8000397c3`). The latter addressed the community report on #4800 and passed 106/106 isolated Linux uninstall cases, including new drift/missing-receipt regressions. Earlier installed fleet receipts below are predecessor evidence only. Start the next fleet qualification at merged beta `4fa6ad17`; open a new PR for any defects found there, without treating historical passes as current. Historical narrative and receipts follow.
## Summary  This integration PR collects ODS product fixes found by repeatedly installing and exercising `public-beta` on the Dream Fleet. It remains based on `public-beta`, not `main`, and is not a fleet-green claim.  ## Exact candidate

- ODS candidate: `0642ad018f5c91c861a1666011045f705ab1400e`
- current `public-beta` base: `d49ad4bbc760be319de7f4c9b9a93b1f18fe3922`
- relationship: the current base is an ancestor of the candidate (candidate is 75 commits ahead, 0 behind)
- fleet harness: Lightheartdevs/DS-Fleet-Test#39 at `65a0516c92b5c0339f63df7e92b6059b14e2747f`
- Pixel source used by the authorized fleet installs: `bbd1d2d62c7260f822ba1e727728a0a02f78895f`

## Product fixes accumulated here  - reconcile optional-service state during reinstall without pruning user-selected services - recover stale or recycled Pixel admission state while retaining strict source, ownership, and catalog-custody checks - support isolated native-Docker WSL Pixel installs with validated gateway/preview ports - repair Windows/WSL bootstrap, model-download, host-agent, Hermes-config, model-promotion, and restart behavior - surface and safely answer authenticated Dream Talk approvals - pull pinned MinIO images from Quay after the Docker Hub paths stopped resolving - keep model-router activation and terminal failure reporting fail closed - allow the GPU-independent Aider CLI extension to resolve on supported CPU-only installs - bind the privileged Pixel access bridge to the root-recorded gateway port instead of guessing the default when owner config omits a custom port - re-run the existing owner-authenticated access proof ceremony after ODS-controlled gateway restarts, while refusing busy, pending, ambiguous, or differently configured states  The latest product deltas address a live Strixy/laptop Pixel control-plane failure. ODS now persists the validated Pixel gateway port in the root-only coordinator configuration and performs a narrowly admitted same-mode proof through the existing access controller after an ODS-owned gateway restart. It never writes a verified state directly; the existing controller still acquires both admission gates, checks the configured service boundary, executes its runtime probe, and records proof atomically. Any unavailable, busy, pending, unknown-mode, stale-revision, or otherwise ambiguous state remains fail closed.  ## Current verification  At predecessor `35d7c697`:  - all GitHub checks are green, including Linux integration, dashboard API/frontends, Pixel inference contracts, environment-schema tiers, PowerShell lint, and distro smoke - Aider dashboard config regression: 60 passed - extension manifest validation: 33/33 library manifests and 30 compatible core manifests, with zero errors - the predecessor dashboard model-activation module: 298 passed, 1 skipped; dedicated Hermes restart regression passed  At predecessor product `00c25411`, all GitHub checks are green, including `integration-smoke`; focused Linux verification is also green: 280/280 Pixel host-install contract checks, 16/16 access-bridge tests, and 5/5 post-restart recovery tests. The live fresh-install failure below demonstrates why those source checks were necessary but insufficient.  At predecessor harness `5cb52c1`, the Pixel/model-switch contract passes on both the laptop checkout and a clean detached Tower2 checkout. It now requires `available=true`, `runtime_verified=true`, matching configured/effective modes, and both `busy=false` and `pending=false` before chat evidence can count. Harness #40 merged at `0303687`; the follow-up gate is now carried by the still-open integration PR #39. All GitHub checks on that predecessor are green. Current harness 2a694d adds target-scoped llama-slot quiescence for the Strixy CPU lane while retaining the strict 90-second runtime usability gate; its new GitHub checks are running.  The predecessor `cc4100af` portability candidate passes the full Docker multi-distro matrix (10/10 lanes) and the clean Incus VM matrix on Ubuntu 24.04, Fedora 42, Rocky 9, Arch, and openSUSE Tumbleweed. Because the product head advanced, those live matrices are predecessor evidence, not exact-head acceptance.  The terminal predecessor laptop run (`f71d14b3` / `420f9b5`) proved clean install, Product Green, Pixel Green, Capability Green, and Lifecycle Green. Model UI remained red:  - cycles 1, 3, and 4 passed - cycle 2 returned an observed 502 when Hermes became unhealthy during Qwen restoration; cleanup restored Qwen and deleted the test model - cycle 5 stopped before UI work when `ods restart` left Hermes unhealthy - cycle 6 proved Pixel, Open WebUI, OpenCode, Perplexica, and route evidence, but a lost harness credential read caused an unauthenticated LiteLLM 401; direct installed-system probes proved unauthenticated 401 and authenticated 200  The Hermes failures are addressed by the latest product commit. The activation-response and credential-attribution defects are addressed in harness #40. No predecessor result is counted as acceptance.  The first exact `cc4100af` paired-WSL attempt stopped before product installation because of two harness preflight defects: a lost laptop WSL privilege probe was mislabeled as missing sudo, and Strixy's absent Docker Desktop engine was treated as a fatal inventory error. Both are fixed and covered at harness `ad4f934`; that failed attempt makes no product claim.  A replacement exact `cc4100af` / `ad4f934` diagnostic run installed successfully on both isolated WSL lanes. The laptop passed UI, Pixel, broad capabilities, and lifecycle. Strixy exposed three fail-closed results: the CPU-only no-GPU telemetry response was treated as a generic 503; Aider was accepted by the dashboard but excluded by its CPU-incompatible manifest; and Pixel chat was blocked after model reconciliation restarted the gateway. The first two defects were fixed at `0303687`/`35d7c697`. The third is now root-caused: the restart invalidated the live access proof and left the native runtime `interrupted`; the custom laptop port also proved the bridge could target an unused default port. Both product defects are fixed at `00c25411`, and the harness now detects this class before counting chat semantics. Strixy's CPU-bound broad capability probe remains actively progressing, so this predecessor run is diagnostic only.  The clean exact `00c25411` / `5cb52c1` paired-WSL run is terminal red: the laptop lost its WSL transport before install, and Strixy reached Pixel installation but failed the final access proof. It is diagnostic evidence only and cannot validate the newer product or harness heads.  ### Latest exact-pair diagnostic (f50ac591 / e73f1c55; superseded)

The exact predecessor laptop-and-Strixy WSL run installed both targets from ODS `f50ac591fe31af1c15677a1648acd59389463697`, harness `e73f1c55`, and Pixel `bbd1d2d6`. It is diagnostic red and is not acceptance evidence.

- Strixy passed installation, installed verification, cloud-mode contracts, dashboard/model management, Hermes, Playwright UI, authenticated Pixel status, verified sandboxed access, retained and non-retained replay/hash checks, cancellation, and the broad chat/search/files/code/skills/model-identity/Talk/SSE/pool/context/grounding matrix. Its only red was Talk abort. The exact installed replay proved this was a harness routing defect: the abort curl sent the hostname without the required `Host:` prefix, so Caddy returned the hostless HTTP 200 catch-all and the POST never reached Talk. Harness #39 now sends the detected host header correctly; the corrected installed replay observed busy state, a real session frame, client abort, and return to idle.
- The laptop downloaded and verified Phi-4-mini but llama-server restart-looped because the selected container profile exceeded the WSL VM's actual available RAM after KV-cache demand. Verified rollback to Qwen 2B succeeded and Pixel was correctly deferred. This candidate now bases runtime-profile eligibility and persisted system RAM on actual VM RAM, while retaining reserved RAM only for coarse tier sizing. Focused selector/profile tests are green; installed proof remains required.
- Paired harness `d809ebf1` also stops immediately when a terminal failed bootstrap has no live updater, reports post-install bootstrap deferral honestly, and carries complete regression fixtures. Its full clean Tower2 suite and GitHub checks are green.

A fresh exact `c9d0b48b` / `d809ebf1` run is required; no predecessor result is promoted to fleet acceptance.
### Ubuntu 26.04 physical-fleet qualification (96b5ca04)

Live preflight established that Tower1 and Tower3 run Ubuntu 26.04 with kernel 7.0.0-29, Python 3.14.4, systemd 259, and Docker 29.7.2; Tower2 remains Ubuntu 24.04. The prior exact allowlist rejected both 26.04 hosts before Pixel installation even though the Docker `ubuntu:26.04` manifest and Incus `images:ubuntu/26.04` VM image are available.

This commit admits only exact Ubuntu 26.04 while retaining Linux, PID1 systemd, WSL2, license, model-route, source-custody, and all other fail-closed gates. It adds native and WSL2 26.04 qualification regressions plus CI Docker and private Incus lanes, and updates the support/install documentation.

Evidence before push:

- local and clean Tower2 qualification suites: 59/59 passed
- actual Tower3 Ubuntu 26.04 host: 59/59 passed, live default-path host qualification passed, and all installer Python sources compiled under Python 3.14.4
- Docker and Incus Ubuntu 26.04 image discovery passed
- shell syntax and `git diff --check` passed

This is source and host-preflight evidence only. Installed Pixel qualification on Tower1/Tower3 still requires the declared noninteractive system privilege; both currently return `sudo -n` failure. The active c9d0b48b WSL run is now predecessor diagnostic evidence because this product head advanced, and a fresh exact-head five-host aggregate remains required.
### Exact predecessor live result and paired harness hardening

The terminal exact predecessor laptop run at product `82ff7f5dd702aa76db790d99f49254d206818ff6` and harness `02bdcdd6531b879849afa519f0955d8fce4d01bc` proves the product-side custom gateway-port repair end to end. The isolated install and full Qwen 3.5 9B promotion passed. Dashboard activation returned HTTP 200, all configured consumers reported successful reconciliation, and the response explicitly recorded `pixel: reconciled`. The immediately following Pixel phase passed authentication enforcement, exact Qwen runtime identity, verified sandboxed access mode on the installed custom port, nonce-bound non-retained SSE, retained replay with hash equality, and active cancellation.

That run is still diagnostic red. After the product proof, native Windows-to-WSL execution degraded to `Wsl/Service/0x8007274c`; Aider/API UI, broad capabilities, and seven regression fixtures then failed through the same transport boundary. Those are not evidence that the seven fixed product bugs returned. Paired harness PR #39 now advances to `d268d3ab2f1a14bac693691528075c45548ea5e7`, which stops later mutations after transport loss, never retries mutating WSL requests, collapses the cascade into one environment signature, preserves regression evidence through triage, and triages before reporting.

Current ODS head `851563f6066f05c85eb83ff72e3d29789db29701` is fully GitHub-CI green. Its explicit-empty/quoted/boundary port hardening has passed the 309-test dashboard module, 284 Pixel host-install contracts, and 14 model-transition tests on Tower2. A fresh exact `21613597` / `d7b7a01` install remains required before acceptance; neither predecessor result nor source CI is a fleet-green claim.

### Remote installer custody and host-agent classification (851563f6 / d268d3a)

The first exact `21613597` / `d7b7a01` laptop attempt was not a valid product result. The staged WSL installer kept running after stop cleanup refused its identity: the old cleanup required the run token in `ps command=`, while large WSL payloads execute as a staged `bash <path>` whose argv does not contain that token. A second run then began uninstall at 23:08:38Z and killed/removed the first run's healthy `ods-host-agent.service` at 23:08:41Z while the first install was still finishing Pixel access reproof.

ODS commit `851563f6066f05c85eb83ff72e3d29789db29701` keeps that overlap attribution separate from product behavior and hardens the observed edge. `SystemdAccessBridge.discover()` now checks the host-agent unit's load state, active root process, positive PID, and process-read race before touching `/proc`. Missing, stopped, malformed, empty-cmdline, or raced root agents return the stable fail-closed `host-agent-unavailable` reason; root execution still must prove isolated `-I` execution and root code custody.

Paired harness commit `d268d3ab2f1a14bac693691528075c45548ea5e7` adds target-side install exclusion and exact staged-process custody. It holds a target `flock` for the destructive run, validates PID start time, PGID, and command hash rather than searching argv for the token, refuses a second install before native cleanup/product mutation, retains markers when termination is unproven, and refuses to release the orchestrator lock after incomplete remote cleanup.

Clean Tower2 exact-pair evidence:
- ODS access bridge: 23/23
- ODS model transition: 14/14
- ODS Pixel host-install contracts: 284/284
- dashboard model tests: 105/105 in the maintained test environment
- complete DS-Fleet-Test portable suite: passed
- both checkouts clean at the exact commits above

GitHub CI is running. No installed host is counted green at this new pair; the next valid result must be a non-overlapping fresh laptop reinstall with both Pixel license variables explicitly set to `true`.
### Candidate-forward reinstall and dashboard draft preservation (9e98a52a)

The current exact ODS candidate is `9e98a52a0898319a66656fbde50d017c3f3b4f66`, paired with fleet harness `579b124d94b3435f25848c9a494fc3c391540f97`.

Commit `126a461ccfd8603330dc0d1c787e4cd0170868b4` fixes the Ubuntu-reproduced Remote Provider form race: a delayed status response can no longer overwrite an in-progress user draft. The previously failing focused test passed 100/100 repetitions on Tower2, and a clean exact-base checkout with the patch passed all 898 dashboard tests.

Commit `9e98a52a0898319a66656fbde50d017c3f3b4f66` adds candidate-forward forced reinstall. The bootstrap clones and verifies the requested source before it invokes that candidate's uninstaller against a fingerprinted existing install. It refuses unsafe, symlinked, root-level, and bootstrap-root targets; a failed candidate uninstall prevents source overlay. This specifically targets the current laptop state where the installed predecessor uninstaller cannot recover a marker-bound interrupted Pixel release, while the newer candidate can. The clean Tower2 installer-hardening suite passes end to end, including exact-SHA, candidate-uninstaller, failure, symlink, and root-target contracts.

GitHub CI on this exact head is running. This is source/candidate evidence only, not installed acceptance. The next decisive proof is candidate-forward recovery and a fresh exact-pair release run on the authorized disposable laptop, followed by the other four named hosts.
### Real-uninstaller forward-recovery fidelity (5da6a156)

The first live candidate-forward laptop attempt at predecessor `9e98a52a` failed safely before replacement. The bootstrap cloned and selected the requested candidate, but the real `ods-uninstall.sh` auto-detected its temporary candidate directory and overrode the target environment accepted by the earlier contract double. Pixel then correctly refused because the old install marker did not bind the temporary checkout. The existing install was not overlaid.

Commit `5da6a15664a7163c3d1c11d03e3a0ab8f4ae9045` replaces that implicit interface with explicit `--install-dir`. The candidate uninstaller independently canonicalizes and fingerprints the requested ODS tree, rejecting relative paths, symlinks, filesystem root, the user home, its own source directory, and missing or symlinked ODS/compose fingerprints before mutation. The bootstrap and behavioral double now use the same argument interface.

Clean Tower2 verification passes: installer-hardening contracts, Pixel uninstall 68/68, uninstall/compose safety, `bash -n`, and `git diff --check`. GitHub CI is running. A fresh exact `5da6a156` live recovery is required before the laptop can start full release qualification; the predecessor failure is diagnostic evidence only.
### Non-interactive candidate-uninstall hang follow-up (34a7939f)

The first live candidate-forward recovery at product `5da6a156`, harness `579b124d`, and Pixel `bbd1d2d6` reached the requested candidate uninstaller and removed the interrupted Pixel release, then hung indefinitely at an interactive `sudo -v` while the bootstrap itself was running with `--non-interactive --force`. The captured process tree was the candidate `ods-uninstall.sh --install-dir /home/odsbeta/ods --force` waiting on its `sudo -v` child. The run was stopped, the exact leftover processes were retired, and the target install lock is free. No candidate overlay occurred; the prior ODS tree remains, while the interrupted Pixel state is absent.

Commit `34a7939f3263d4b7414818fe611fff30d5c825a0` makes the candidate bootstrap propagate `--non-interactive` to the uninstaller. The uninstaller now validates privilege with `sudo -n -v` before Pixel, container, or install-tree mutation, fails promptly with a recovery instruction when no cached/passwordless credential exists, and returns nonzero if complete install-directory cleanup was not achieved. Interactive uninstall keeps its terminal-attached `sudo -v` behavior.

Clean exact-head Tower2 evidence at `34a7939f`:
- installer-hardening contract passed
- uninstall compose/safety regression passed, including a five-second no-hang and pre-mutation assertion
- Pixel uninstall suite: 68/68 passed
- Bash syntax and clean-checkout checks passed

A live exact-head rerun on the laptop test distro exited through the intended failure path in 4 seconds (not the 120-second timeout), emitted the cached/passwordless-sudo recovery instruction, and preserved the SHA-256 fingerprints of `.env`, `ods-cli`, `ods-uninstall.sh`, and `docker-compose.base.yml`; Pixel marker/current remained absent before and after. All current-head GitHub checks are terminal green: 33 succeeded, 3 intentionally skipped, 0 pending, and 0 failed. This is a live-derived fail-fast repair, not installed acceptance. A fresh exact-pair run remains required after the laptop test target has a valid noninteractive privilege path.
### Installed laptop Pixel and model-promotion follow-up (481cacf5)

A fresh disposable laptop installation completed at ODS 34a7939f3263d4b7414818fe611fff30d5c825a0, harness 579b124d94b3435f25848c9a494fc3c391540f97, and Pixel bbd1d2d62c7260f822ba1e727728a0a02f78895f with both Pixel license acceptance variables set true. Install, core verification, dashboard, cloud mode, full-model capabilities, lifecycle, and all 22 installed regression fixtures passed. Six zero-prerequisite lanes, the 10-distribution Docker matrix, and the five-distribution Incus matrix also passed.

The first full-model promotion reached a healthy gateway but one strict Pixel verification reported that pixel-operations-broker was not loaded and correctly rolled back. A later supported lifecycle retry succeeded. On the stabilized Qwen 3.5 9B runtime, the exact Pixel harness passed authentication, schema rejection, runtime identity, verified sandbox access mode, nonce-bound non-retained and retained SSE, exact replay hash equality, and active cancellation. Six immediate restart/health/plugin-inspect cycles then found the expected broker version and active release root every time.

The independent model-management run reproduced the product impact through the real UI: Granite 4.0 H-Micro downloaded, the Run action issued POST /api/models/granite4.0-h-micro-q4/load, and the product returned HTTP 502 with ODS-managed Pixel model reconciliation failed; rollback could not be proved. Harness recovery restored the Qwen 9B baseline and deleted the harness-owned test model.

Commit 481cacf56f06ffcd60852261bcb92d1b31a2a6d1 retries only the same complete strict Pixel verification, at most three times with two-second settling intervals. Persistent failure still aborts and invokes the existing rollback path. Clean Tower2 validation: 286 Pixel installer tests passed, Bash syntax and diff checks passed. An independent Tower1 read-only review found no actionable correctness, security, timeout, side-effect, or fail-closed issue.

This is a live-derived repair, not a five-host-green claim. A fresh exact-head installation and model-management rerun at 481cacf5 remains required.
### Hermes activation health-window follow-up (02a2e9bb)

The exact predecessor laptop run at ODS `481cacf56f06ffcd60852261bcb92d1b31a2a6d1` and harness `d47a1a3bb08820c38640d43f69c61707b61ee696` passed destructive reinstall, installed verification, cloud-mode contracts, dashboard, Hermes, UI, the complete Pixel phase, broad chat/search/files/code/skills/Talk/context/grounding capabilities, lifecycle, and all 22 regression fixtures before the model loop.

Model cycle 1 activated Granite H-Micro and passed Talk, every required switched-model consumer, and both retained and non-retained Pixel SSE. Restoring Qwen then returned an observed HTTP 502 because `ods-hermes` was still in Docker's `starting` state when the generic 120-second health budget expired. The product journal showed the gateway became ready shortly afterward, and cycle 3 completed the same switch/restore/cleanup path successfully. This is timing-sensitive installed evidence, not a consistently broken route.

Commit `02a2e9bb47d1a558de796cbe8a29a50d9acfec69` gives only Hermes a 90-attempt/180-second bounded health window during model activation. Other containers retain the 60-attempt default, explicit overrides still win, explicit unhealthy/invalid/exited states still fail immediately, and persistent `starting` still fails closed at the new bound.

Exact clean Tower2 verification: `313 passed, 1 skipped` for `test_model_activate.py`. Independent Tower1 Qwen3.6 no-think review and Codex diff review found no actionable defect. The paired harness head is `afb1cb00b02943e0b7dd19076355db61fc45eeab`, which separately makes the observed 502 fail immediately with preserved response evidence. Fresh exact-pair installed qualification is required; neither PR claims fleet green.
### Fresh exact-pair patched laptop rerun

A fresh destructive laptop qualification started at 2026-09-13T04:40:51Z against exact clean ODS `02a2e9bb47d1a558de796cbe8a29a50d9acfec69`, exact clean harness `abebafe2af292f074bcb7164fbc39346f3043a85`, and immutable Pixel `bbd1d2d62c7260f822ba1e727728a0a02f78895f`. Both `PIXEL_LICENSE_ACCEPTED=true` and `DREAM_FLEET_PIXEL_LICENSE_ACCEPTED=true` are explicit. The predecessor red run released the heavy lock; only the disposable `ODS-Public-Beta-Test` WSL distro was restarted after its service entered `Wsl/Service/0x8007274c`.

Run directory: `/home/michael/.local/state/dream-fleet/tasks/ods-public-beta-fleet-green-20260911/exact-02a2e9bb-abebafe2/release/20260913T044051Z-windows-laptop`. Zero-prerequisite distro lanes are progressing green. Installed Pixel/model-switch acceptance remains pending.

### Exact installed full-model evidence and paired lifecycle fix

The exact laptop run at this ODS head and harness predecessor `abebafe2af292f074bcb7164fbc39346f3043a85` completed a clean install, installed Pixel reproving, core verification, and cloud-mode contracts. Its dashboard phase then failed only after a fixed 360-by-five-second observation budget expired while the authenticated lifecycle response still reported a healthy active Qwen 3.5 9B download. The full 5,680,522,464-byte model subsequently completed and promoted successfully.

On the installed full model, ODS `02a2e9bb47d1a558de796cbe8a29a50d9acfec69` passed Hermes and the complete Pixel phase: required authentication, exact Qwen 3.5 9B runtime identity, verified sandboxed access with no busy/pending state, nonce-bound non-retained SSE, retained completion and byte-identical replay hash, and real active cancellation. Broad chat, search, files, code, skills, Dream Talk status/pool/SSE/context/grounding, model identity, and strict 90-second runtime probes passed, as did all 22 regression fixtures.

This is strong installed product evidence but not acceptance: the target-blind harness timeout made dashboard and lifecycle red, UI Aider transition also remained red, and the six-cycle model matrix was correctly blocked. Paired harness PR #39 commit `827159afaa46d6abfd021ff0df309fe67fb9cd50` now gives only the two disposable Windows WSL lanes their declared four-hour lifecycle observation budget while preserving every fail-closed terminal, transport, exhaustion, and no-follow-on-mutation rule. Its focused behavior, complete Tower2 suite, Windows-native contract, and all four GitHub jobs are green.

A fresh exact `02a2e9bb` / `827159af` laptop release run is active with both Pixel license acknowledgements set to `true`; the matching Strixy run is queued behind it. No ODS code changed for this harness defect, and no five-host green claim is made.

### Talk disconnect abort propagation (fb69ff0b)

The exact `02a2e9bb` / `49065336` Strixy release run passed zero-prerequisite bootstrap, install, verify, cloud-mode contracts, dashboard/model management, Hermes agentic, Playwright UI, the complete Pixel phase, and lifecycle. Pixel proved authenticated full-model runtime identity, verified sandboxed access, nonce-bound non-retained SSE, retained byte-identical replay, and observed-active cancellation.

Broad capabilities proved runtime performance, chat, search, files, code, skills, exact model identity, Talk status frames, pooled-session reuse, and persona/context. Talk SSE and grounding then failed because the first Talk client disconnected while Hermes continued the detached weather-agent turn for more than twelve minutes. The installed Hermes log recorded `detached_sessions=1`; the abandoned turn made three local model calls, ran two web searches, and occupied the single CPU llama slot long enough for two 900-second idle gates to expire.

Commit `fb69ff0b` sends Hermes's documented, session-scoped `session.interrupt` abort before closing and evicting the pooled Talk connection. It preserves the existing approval denial ordering, bounds all cleanup writes/closes, never opens a new connection or interrupts an idle session, and prevents late frames from an abandoned turn contaminating a later request.

Verification:
- full dashboard Talk suite: 46 passed
- Python compile and `git diff --check`
- independent Tower3 Qwen3.6-27B Q4 no-think review found no blocker after checking lock ordering, session scope, pool eviction, bounded cleanup, and regression coverage
- live source inspection of the pinned Hermes build confirmed `session.interrupt` is the gateway's `abort` equivalent and clears only the named session

The predecessor Strixy run is diagnostic red and does not count as acceptance. A fresh exact product/harness pair must prove a deliberately aborted Talk stream releases the llama slot promptly, then complete Pixel, broad capabilities, lifecycle, and all six model cycles.
### Terminal c9 WSL diagnostic and live model-reload follow-up

The terminal predecessor run at ODS `c9d0b48b4287a8d62f41bfbfdecd88005e9e24ea`, harness `d809ebf16d2c6bf6fdbb67ff5291abbbdbc8b742`, and Pixel `bbd1d2d62c7260f822ba1e727728a0a02f78895f` is diagnostic only.

- Strixy passed clean installation, verification, cloud-mode contracts, dashboard/model management, Hermes, Playwright UI, authenticated Pixel status, verified sandboxed access, nonce-bound retained and non-retained SSE, byte-identical replay, active cancellation, the broad chat/search/files/code/skills/model-identity/Talk/SSE/pool/context/grounding matrix, and lifecycle.
- The laptop passed installation, verification, cloud mode, and UI, and completed the 5,680,522,464-byte Qwen 3.5 9B promotion. Its old dashboard probe reused the deleted Qwen 2B bootstrap id after promotion; harness #39 `65a0516` now refreshes inventory after the lifecycle-idle gate and selects the current loadable model. The complete clean Tower2 suite and all four GitHub jobs are green at that harness head.
- The laptop Hermes and three early Pixel HTTP `000` failures coincided with native `Wsl/Service/0x8007274c`; direct same-distro commands reproduced that transport failure while later semantic Pixel checks passed. These results are environment evidence, not a Pixel product regression.

After a targeted WSL test-distro restart restored command execution, a live harness-`65a0516` replay correctly selected `qwen3.5-9b-q4`, proving the stale-bootstrap fix. The same-model `/load` then returned HTTP 502 because the root Pixel transition controller reported `runtime-proof-required` after the WSL/gateway restart, and rollback could not be proved. Host-agent journal evidence shows the target model itself became ready before Pixel reconciliation failed. This was a current-candidate product defect at `96b5ca04`; the follow-up fix and validation are recorded below. No predecessor run is promoted to acceptance.
### Stale Pixel runtime reproof under model hold (0642ad01)

A targeted live replay after laptop WSL recovery proved a model lifecycle deadlock at predecessor `96b5ca04`: Qwen 3.5 9B became ready, but both forward Pixel reconciliation and rollback were rejected because the gateway restart had invalidated `verified.json` and the root controller reported the exact fail-closed state `available=true`, `configured_mode=sandboxed`, `effective_mode=unknown`, `runtime_verified=false`, `busy=false`, `pending=false`, `reason=runtime-proof-required`.

Commit `0642ad018f5c91c861a1666011045f705ab1400e` lets a model transaction admit only that exact owner-host stale-proof projection. The root controller journals first, closes and drains the external and native admission gates, then re-proves the configured access mode while both gates remain held. Ambiguous, unavailable, busy, pending, invalid-revision, or mismatched states still fail closed. A failed proof retains the error journal and both holds; a successful transition still re-proves the post-change route before releasing native admission and then edge admission.

Verification before push:

- local: 18/18 model-transition tests, 5/5 access-reconcile tests, and 286/286 Pixel host-install contracts
- clean detached Tower2 checkout: the same focused suites pass; complete Pixel source suite passes, including 59/59 Pixel integration checks, 286/286 host-install contracts, and the Node agent suite with 1,067 passed, 1 intentionally skipped, 0 failed
- disposable installed laptop: the exact stale-proof projection was reproduced; a real begin/finish recovery returned `runtime_verified=true` with no pending transition; the full authenticated dashboard replay then selected Qwen 9B and changed the former HTTP 502 into HTTP 200 with `pixel=reconciled`, while Aider remained installed

The installed replay used a narrowly patched predecessor installation and is diagnostic proof of the fix, not exact-head fleet acceptance. A fresh exact `0642ad01` / harness `65a0516` / Pixel `bbd1d2d6` release run is required.


roshangupta00750:
I found a gap in the new Pixel access cleanup while tracing #4800 (`ods uninstall` leaves `ods-pixel-access.service` behind). On `2f985137`, `_ods_pixel_access_validate_or_remove` covers the access unit, program tree, config, state, probes and `90-ods-full-access.conf`. It does not cover the two root files that the provider transaction writes:

- `/etc/systemd/system/openclaw-gateway.service.d/95-ods-provider.conf`
- `/etc/ods/pixel-provider.env`

They are written by `ServiceEnvironment.apply` ([service_environment.py#L20-L21](https://github.com/Osmantic/ODS/blob/2f985137b9680151395d59d0498a7fa7e169ff2e/ods/bin/pixel_provider/service_environment.py#L20-L21)). Neither path appears in `lib/pixel-uninstall.sh` or `ods-uninstall.sh`.

The remove step unlinks only the full-access drop-in, runs `rmtree` on the state root, and then treats `ENOTEMPTY` on the `.d` directory as success ([pixel-uninstall.sh#L255-L266](https://github.com/Osmantic/ODS/blob/2f985137b9680151395d59d0498a7fa7e169ff2e/ods/lib/pixel-uninstall.sh#L255-L266)). The state root holds `provider-root-managed.json` and `provider-service-environment.json`, and those are the only root records that bind the two files. After uninstall, the files stay on disk and nothing records that ODS created them.

**Why this breaks a reinstall**

The drop-in is `EnvironmentFile=/etc/ods/pixel-provider.env` plus `BindPaths=<data dir>/pixel-providers`, with no `-` prefix on either ([service_environment.py#L190](https://github.com/Osmantic/ODS/blob/2f985137b9680151395d59d0498a7fa7e169ff2e/ods/bin/pixel_provider/service_environment.py#L190)). The data dir defaults to `$INSTALL_DIR/data` ([store_factory.py#L27-L29](https://github.com/Osmantic/ODS/blob/2f985137b9680151395d59d0498a7fa7e169ff2e/ods/bin/pixel_provider/store_factory.py#L27-L29)).

`ods uninstall` deletes the install tree. On a fresh install, `pixel-providers` is only created by an owner save action (`prepare_directory` in the save paths). The read paths and `readiness()` deliberately never create it. So it does not exist when the installer starts the gateway.

systemd merges the orphaned drop-in into the newly written `openclaw-gateway.service` and refuses to start it. I reproduced that step on Ubuntu 26.04 (systemd 259) with a throwaway unit carrying the same drop-in shape:

```
ods-bindtest.service: Failed at step NAMESPACE spawning /bin/true: No such file or directory
ods-bindtest.service: Main process exited, code=exited, status=226/NAMESPACE
```

The same unit started normally once the directory existed.

In the installer, that failure returns through `systemctl start openclaw-gateway.service || return 1` ([pixel-host-install.sh#L4296](https://github.com/Osmantic/ODS/blob/2f985137b9680151395d59d0498a7fa7e169ff2e/ods/installers/lib/pixel-host-install.sh#L4296)) to phase 11's `exit 1` ([11-services.sh#L1242-L1245](https://github.com/Osmantic/ODS/blob/2f985137b9680151395d59d0498a7fa7e169ff2e/ods/installers/phases/11-services.sh#L1242-L1245)). On a host where a provider was ever saved, uninstalling and then reinstalling should stop at "Pixel default-agent setup failed before the ODS stack launch."

With `--keep-data`, the directory survives and the gateway starts. It still runs with the previous install's provider environment, and there is no managed record behind it.

Scope of what I checked: I reproduced the systemd behaviour above and traced the rest through the linked code. I have not run a full Pixel reinstall, because I don't have licensed Pixel source on this machine.

**A fix that keeps the fail-closed model**

1. In `verify`, read `provider-root-managed.json`.
2. Require the on-disk env file and drop-in to match its `environment` images exactly, bytes and mode ([coordinator.py#L210-L215](https://github.com/Osmantic/ODS/blob/2f985137b9680151395d59d0498a7fa7e169ff2e/ods/bin/pixel_provider/coordinator.py#L210-L215)).
3. Remove both files before removing the state root.
4. If the files are present but don't match the record, refuse, instead of deleting the record and leaving the files behind.

Current `public-beta` has no access cleanup at all yet (#4800), so the provider files are left behind there too.

Lightheartdevs:
Fleet qualification status (2026-09-15): this product PR head is 2f985137b9680151395d59d0498a7fa7e169ff2e, still derived from current public-beta 33155fb5c4a99242b6fa8eb6e2d9b108c869e3cc; Pixel source for authorized installs is b33730436baf5d98bf58f7d57c090318fe19f433. A fresh physical Tower3 release at this exact product head passed zero-prereq, install, product, strict Pixel, broad capability, and lifecycle gates; all six browser/model-switch runners completed with model-route evidence and cleanup. The release gate stayed RED solely because the prior harness verifier expected two Talk trace/DOM tokens despite a legitimate restored-baseline low-context Talk skip. DS-Fleet-Test PR #39 commit 0358ddb repairs the harness gate by requiring distinct completed Pixel proofs for that restored state; the six saved physical artifacts, complete clean Tower2 suite, and GitHub checks pass. This is diagnostic evidence, not a fleet-green product claim. A fresh exact ODS/0358ddb four-host release is now live for Tower1, Tower3, this laptop, and Strixy; Tower2 is the separately coordinated fifth lane. No merge or main cutover is implied.

Lightheartdevs:
Pixel live-beta follow-up b1d919af: r46 on Tower1 and Tower3 repeatedly saw active-cancel return aborted=false with a completed retained result. A separate bounded same-device Tower1 reproduction captured five attempts where /chat/result and /chat/activity were both active on two consecutive samples before Stop; every final saved stream contained only the 13-byte SSE [DONE] marker. All five corresponding authoritative Pixel host sessions recorded assistant stopReason=aborted and zero output tokens. This commit fixes the verified state-integrity defect: a terminal SSE marker without user-visible answer is now a replayable interrupted result with a sanitized error, never a successful complete receipt. The exact patch passed 82 focused retained/route tests and 351 Pixel dashboard tests in an isolated Tower2 Linux checkout. This does not yet explain or fix aborted=false; controlled live cancellation and exact-candidate fleet requalification remain required. No fleet-green claim.

Lightheartdevs:
Live Tower1 follow-up on the old 2f985137 install: in three paired active-run trials, direct Pixel Edge /v1/chat/cancel returned HTTP 200 aborted:true (~110-120 ms), but ODS /api/pixel/chat/cancel had returned aborted:false in five earlier active-run trials and saved an empty DONE-only stream as complete. This narrows the failure to the ODS retained-result/Stop race, not Edge cancel support. I added a deterministic Edge-ack-versus-empty-DONE test in eb340636; the race test passed, and the exact Tower2 isolated Pixel dashboard suite passed 352/352. This is source/test evidence, not live acceptance of the new candidate: r46 remains on the old install, and the new eb340636 WSL qualification is queued behind r46 and Tower1/3 workload restoration. We must re-test real Stop on the new installed candidate before calling it fixed.

Lightheartdevs:
Fleet acceptance follow-up: the prior eb340636 race test proved Edge aborted:true could pair with a durable interrupted result because empty DONE finalized while /chat/cancel awaited Edge. The fleet Pixel contract correctly requires a cancelled receipt after acknowledged active Stop. Commit 586c4624 now keeps only DONE-without-answer attempts unresolved while an explicit Stop is in flight; a true Edge ack commits cancelled, a false ack keeps the successor blocked, ordinary empty DONE remains interrupted, and completed answers stay complete. Paired true/false-ack and late-Stop tests passed; isolated exact-patch Tower2 Pixel dashboard suite 353/353. This is still source/test evidence, not live acceptance. Queued Tower2 r50, WSL r51, and Tower1/Tower3 r52 have been repinned to 586c4624; r46 old-candidate model/UI continues untouched. The superseded eb340636 Tower2 job was withdrawn while waiting on the fleet lock, before any install event.

Lightheartdevs:
Fleet qualification checkpoint at exact PR head 8a434afd (merge of public-beta 4e8a9958): resolved host-agent and Pixel/system uninstall overlaps while retaining the install-bound Pixel access remover. The integrated access cleanup now refuses a pending transition or busy lock; the duplicate broad recursive access-directory deletion from the upstream merge was removed. Local WSL source checks: Pixel uninstall 99/99, system-unit uninstall 8/8, shell parse and Python compile passed. Targeted pytest host-agent tests remain unrun in this environment (pytest unavailable). This is source evidence only: live Tower1/Tower3, laptop/Strixy WSL, and Tower2 installed release runs are frozen on earlier PR head 5d7042a8 and their receipts must be reviewed separately. No five-machine User Green claim.

Lightheartdevs:
Physical fleet receipt: frozen r112 (ODS 5d7042a8, harness 8abe988) installed Tower1 and Tower3, passed installed Pixel and broad capability probes, then the guarded stop interrupted Tower1 during Models-page cycle 3. Tower1 Pixel uninstall failed closed with `OpenClaw configuration drifted from its ODS marker` (exit 1); product containers/services were removed by qualification cleanup and the independent Qwen worker was restored/health-verified, but clean ODS/Pixel uninstall is NOT proven. Tower3 uninstall was clean.

Sanitized Tower1 evidence: owner-private `ods-model-reconcile-6itj8hui` backup OpenClaw config canonically matches the ready marker; the later live config differs only at 15 model/runtime configuration paths. Source `ods_pixel_reconcile_promoted_model` writes the candidate config before `_ods_pixel_mark_ready`, and its ordinary error rollback does not run when the process is interrupted mid-transaction. The root model journal may also remain; do not bypass its custody gate or merely rehash the marker. Need an exact-transaction, fail-closed recovery/rollback path and an interrupted-reconcile regression test before Tower1 can be called lifecycle green. New five-host queue r124/r125 is paused before install; laptop/Strixy r122 and Tower2 r123 continue independently. This is a physical defect receipt, not a claim that PR head 931ac799 or source 8a434afd has been tested on Tower1.

Lightheartdevs:
Follow-up physical/source correction for Tower1 interrupted model reconciliation: owner-authorized root-controller `status` returned HTTP 200 with `pending:true`, `available:false`, `reason:host-agent-unavailable`; `ods-pixel-access.service`, `openclaw-gateway.service`, and `pixel-ingress.service` are active, while qualification teardown recorded zero remaining product containers. Do not regard this as clean removal or simply rerun `model-begin`: `pixel_access_bridge.py:model_begin` rejects a pending journal.

A safe recovery needs a read-only owner-scoped query of the *validated* pending model journal's kind/transaction ID (never token), then an exact-transaction `model-finish` after runtime/admission is restored and a marker-bound backup/candidate is verified. IMPORTANT hash-domain distinction: `pixel_access_mode.py:_status_locked` reports SHA-256 of raw OpenClaw file bytes as `config_sha256`, and the journal's `start_config_sha256` records that value; `pixel-uninstall.sh` compares the ODS marker to SHA-256 of `b"ods-pixel-openclaw-v1\0" + canonical JSON`. Therefore compare the backup raw-byte hash to journal start hash and its canonical domain hash to ready marker separately. These hashes should not be directly equal. Owner state must not be blindly copied or marker-rehashed. Physical journal kind/transaction ID remains undisclosed by current protocol; no recovery mutation has been performed.

Lightheartdevs:
Tower1 recovery custody update: the standard fleet qualification bridge ran a read-only, root-journal projection with cleanup verified (no bridge container/payload directory remained; system sudo still root-owned at original mode). The pending journal is valid `kind=model`, `phase=held`, `configured_mode=sandboxed`; its raw `start_config_sha256` is `ff2cffa972dc1823b6527b2e72bd860929950073db4667ba4a156a01db1dba1f`, exactly the SHA-256 of owner-private `ods-model-reconcile-6itj8hui/openclaw.json`. That backup's canonical ODS domain hash matches the ready marker; live config's hash does not. The transaction handle and gate token are deliberately omitted here. This proves an exact marker-bound backup chain, not a completed recovery: controller still reports `host-agent-unavailable`, product containers are absent, and root `model-finish` must not run before admission/runtime proof is restored. PR head 5b3f51d adds source-only owner `model-status`; it is not installed on Tower1 yet. No owner config was copied or marker rehashed.

Lightheartdevs:
Fleet qualification update (Tower1, 2026-09-16): installed only the five model-status controller files from exact 5b3f51d2 over matching installed 5d7042a8 files via the standard fleet Docker-sudo bridge. A root-private backup and automatic rollback were used; the first two attempts rolled back because service activation preceded Unix-socket readiness, and the bounded readiness retry succeeded. Owner-side model-status now returns a validated pending model journal: phase=held, configured_mode=sandboxed, and start-config raw SHA256 ff2cffa972dc1823b6527b2e72bd860929950073db4667ba4a156a01db1dba1f, matching the preserved owner backup. Transaction ID and journal token are deliberately omitted here. Postcheck: all five installed hashes match 5b3f51d2, controller active, bridge/payload removed, system sudo unchanged, independent Qwen worker healthy. This is diagnosis-only: transition remains pending; host agent/edge and exact rollback/finish still require safe recovery before another Tower1 install. Frozen WSL r122 remains live and is not restarted for branch drift.

Lightheartdevs:
Tower1 installed recovery finding: top-level ods-uninstall.sh retired ods-host-agent.service before Pixel uninstall validated the root-held model journal. Pixel correctly refused uninstall, but the host agent was already gone. Commit d9223180 validates system-unit custody read-only first, runs Pixel fail-closed cleanup while recovery services remain live, then retires system units. Linux system-unit tests 11/11 and compose-uninstall fixture pass; current PR CI is running. Physical Tower1 host agent and Pixel edge have been restored and are healthy, but the original transition remains held until the prior model route is rehydrated and the same transaction is safely finished. This commit is not yet part of any completed five-host installed candidate.

Lightheartdevs:
Fleet qualification update (physical Tower1, not a pass for this PR yet): the earlier frozen `5d7042a8` install left a held Pixel model transition after uninstall stopped recovery services and teardown removed runtime files. We recovered the exact original 27B model, restored the original Pixel relay key and source-generated contract files, completed the source rollback as `rolled-back`, and independently verified `pending=false`, original OpenClaw config bytes, healthy protected Qwen worker, active Pixel/host services, no temporary backend/bridge, and unchanged resolved sudo mode. This validates the recovery sequence, not the new uninstall patch.

The current public-beta tip `3c186f3f` plus this PR head `d9223180` is pinned as merge candidate `623f788d`; full real-device installation/Pixel/model-switch qualification is queued on Tower1, Tower3, the laptop and Strixy WSL lanes, and Tower2. We will report installed evidence separately from CI/source checks and update this PR if the candidate exposes another product defect.

Lightheartdevs:
OpenCode model-switch qualification update (`d9524162` + `0bfe76b3`): the Linux, macOS, and Windows installer/config-refresh paths now reserve prompt context by setting `limit.output = min(32768, floor(limit.context / 4))`. This retains the prior 32K output ceiling for contexts of 128K or more while reducing a 32K context to an 8K output budget.

Same-device live A/B on Tower1 (OpenCode 1.18.18, isolated config/data, protected Qwen3.6-27B Q4 endpoint, identical prompt/model/context): the previous `32768/32768` output/context pair timed out after 120 seconds with 204 OpenCode events and 29 synthetic continuation events. `8192/32768` completed the exact nonce in 3 seconds with one text response and `reason=stop`. No installed ODS configuration or protected model service was changed, and no OpenCode process remained. Linux, macOS, and Windows source tests passed; direct PowerShell output-budget checks passed at 32K, 64K, 128K, and 256K contexts.

This is strong evidence for the OpenCode budget failure, **not** a release-green claim. The next gate is a new exact-commit installed model-management run with real switches, browser/consumer probes, Pixel journeys, and protected-workload restoration. Native Strixy install and Tower2 qualification remain separate open blockers.

Lightheartdevs:
Release-qualification update (2026-09-16 UTC; exact candidate `80f324a7`, public-beta `3c186f3f`):

- Strixy native Windows r188 was not a Lemonade failure. Its recovery log showed Lemonade healthy, then Compose resolution failed because `PIXEL_MODEL_RELAY_KEY` was missing. The native service plan disabled Pixel edge but accidentally enabled the Pixel-only relay through its generic `core` fallback. `80f324a7` disables the unsupported native relay and adds core/full-plan regressions (47/47 targeted tests). A read-only replay of Strixy's exact Compose stack *without* that relay now resolves all services. Fresh native install proof is still required.
- Tower1 r191 at predecessor `0bfe76b3` passed clean install, verify, dashboard, Pixel, Hermes, capabilities, lifecycle, Pixel agentic tools, and regressions. Model-UI cycles 1 and 2 both failed the required OpenCode functional probe after a 300-second headers timeout; the harness recovered each failure. The remaining cycles were stopped because this candidate was already red, and the protected Qwen worker was verified healthy after restoration. Root-cause work is ongoing; no model-switch green claim.
- All 41 non-skipped GitHub checks currently pass on `80f324a7`. Docker/Incus distro gates r192 are running on predecessor `0bfe76b3`. The five-surface exact-head release gate remains open; this PR is not yet installed-fleet green and remains unmerged from public-beta.

Lightheartdevs:
Fleet qualification update (2026-09-16): the current Strixy WSL CPU run is pinned to the previous PR head `80f324a7` and is red in model-switch baseline. LiteLLM returned HTTP 200 but produced `Thinking Process` instead of the challenge phrase twice; the second cycle also exceeded the 300s Pixel/Perplexica probe bounds. Its post-install, Hermes, Pixel-agentic tool journeys, and recovery checks passed.

The CPU overlay shipped llama.cpp `server-b8248`, while base Compose already requests `LLAMA_ARG_REASONING=off`; that option is absent in upstream b8248 and present in b9014. Commit `2c74b88a` updates only the CPU default image/lock and adds a contract test. Focused local tests passed; PR CI is running.

This is a candidate fix, **not a fleet pass**. After the old run safely ends, we will install this exact head and rerun the real Pixel/model-switch matrix on Strixy, then complete the remaining machines. The older run must not be represented as validation of `2c74b88a`.

Lightheartdevs:
Thanks for the precise #4800 finding. Confirmed on the current PR head and fixed in 8000397c3: uninstall now validates the exact root-owned pixel-provider.env and 95-ods-provider.conf images/modes against provider-root-managed.json, removes those files before their custody record, and refuses drift or a missing receipt. The isolated uninstall suite passes 106/106 (including new success/fail-closed cases), and the full PR CI is 43 successful / 3 skipped at this head. A physical uninstall/reinstall on the new beta head remains a separate fleet qualification gate after merge.

PR #4743, opened 2026-09-13
fix(ci): run the bind-address and Bash 3.2 guards in test-linux

roshangupta00750:
Fixes #4375.

## Problem

`tests/test-bind-address-sweep.sh` and `tests/test-bash32-guards.sh` are invoked by `make test` (`Makefile:78` and `:104`) **and by nothing else** — no workflow named either script, and no other test sourced them.

Because **no workflow runs `make test`** (the only `make test` strings under `.github/` are inside issue bodies the autonomous scanner generates), neither guard has ever executed in CI. Both are regression guards, so the regressions they exist to catch would land unnoticed:

- **`test-bind-address-sweep.sh`** keeps every community extension compose file binding host ports through `${BIND_ADDRESS:-127.0.0.1}`. A bare `"127.0.0.1:"` literal silently defeats the `--lan` / dashboard exposure opt-in.
- **`test-bash32-guards.sh`** keeps a `BASH_VERSINFO[0] < 4` guard ahead of the first `declare -A` in the scripts using associative arrays, and keeps `ods-cli` re-entering sub-scripts through `"$BASH"`. Stock macOS ships Bash 3.2.

## Fix

Add both as steps in `test-linux.yml`'s `integration-smoke` job, which already runs with `working-directory: ods`.

Both are **static source checks** (grep/awk only — the Bash 3.2 one never executes a 3.2 shell), so they are platform-agnostic and belong in the Linux job even though one protects macOS.

## Validation

**The guards are not vacuous — I verified each one actually fails on the regression it guards:**

- Injected a community extension compose with `"127.0.0.1:59999:80"` → `test-bind-address-sweep.sh` exits **1**:
  ```
  FAIL community extensions still bind to literal 127.0.0.1 in ports:
    .../_zz_probe/compose.yaml:5:      - "127.0.0.1:59999:80"
  ```
- Removed the `BASH_VERSINFO[0] < 4` line from `lib/progress.sh` → `test-bash32-guards.sh` exits **1**.
- Both restored; both exit **0** on a clean tree, so CI stays green.

**Change correctness:**
- `test-linux.yml` parses as YAML; both steps resolve inside `integration-smoke` (`working-directory: ods`).
- Both scripts went from **0** references under `.github/` to referenced.
- Makefile untouched; `make -n test` still parses.
- `git diff --check` clean. Additions only — 10 lines, no existing step altered.

## Scope note

A general Makefile↔CI parity guard is deliberately **not** included: 40 of the 69 `make test` suites are unreferenced by any workflow, and many are platform-specific by design (`test-macos-*`, `test-windows-*`), so a blanket rule would be wrong. This PR is limited to the two static, platform-agnostic guards named in the issue.

## Release Lane

- [x] Mainline change targeting `public-beta` (the active development branch)


roshangupta00750:
Consolidated into #6381, following the maintainers' guidance on fewer, fuller PRs. The fix from this PR is carried over unchanged, re-verified against current `public-beta` (`99feff29e`) before folding in; the evidence is in #6381's description. Closing this one in favour of it.

PR #5172, opened 2026-09-15
test: pin fixture file modes so two CI suites pass under Ubuntu's default umask 002

roshangupta00750:
## Problem

Two CI-gated test suites fail on a stock Ubuntu desktop. The product code is right in both cases. Each suite writes a fixture under the ambient umask, and Ubuntu gives regular users `002`, so the fixture comes out group-writable. The code under test correctly refuses group-writable files.

The GitHub runner uses umask `022`, which is why CI stays green.

`tests/test-pixel-uninstall.sh` has the same cause. #4215 already pins `umask 077` there for the same reason, so it is not repeated here.

### 1. `tests/test-installer-mode-preservation.sh`

```
[FAIL] implicit rerun did not preserve cloud mode
```

`ods_existing_install_mode` refuses a `.env` that is group- or other-writable (`(mode & 022) == 0`). The real installer creates `.env` under `umask 077` (`installers/phases/06-directories.sh:1026`) and then chmods it to 600, so that check is correct. The test's fixture was written as `664` and refused.

**Fix:** pin `umask 077` at the top of the test, matching how the installer creates `.env`. The explicit `chmod 0666` case, which asserts that writable files fail closed, is unaffected because `chmod` ignores the umask.

### 2. `tests/test-preserve-active-model.py` (the "Active Model Preservation Tests" step in `test-linux.yml`)

```
test_verified_switchboard_state_recovers_an_interrupted_installer_env
KeyError: 'LLM_MODEL'
```

`scripts/preserve-active-model.py` only trusts a `data/model-state.json` that is not group- or other-writable. The host agent is the only writer of that file, and it always creates it at `0644` (`STATE_FILE_MODE` in `bin/model_switchboard/state.py`, via `mkstemp` plus `chmod`). dashboard-api and model-router only read it. So real installs never trip the check. The fixture, written with `write_text()`, came out `0664` and was refused.

**Fix:** `chmod` the fixture to `0644` right after writing it, mirroring the host agent.

## Validation (Ubuntu 26.04)

| suite | umask | `public-beta` | this branch |
|---|---|---|---|
| installer-mode-preservation | `002` (Ubuntu default) | **fail** | pass |
| installer-mode-preservation | `022` | pass | pass |
| installer-mode-preservation | `077` | pass | pass |
| preserve-active-model | `002` (Ubuntu default) | **fail** (6 of 10 cases, then `KeyError`) | pass (10/10) |
| preserve-active-model | `022` | pass (10/10) | pass (10/10) |
| preserve-active-model | `077` | — | pass (10/10) |

Other checks:
- shellcheck at error severity is clean.
- `bash -n` passes under stock macOS Bash 3.2.
- ruff (`E,F,W`) is clean for the Python test.
- `git diff --check` is clean.
- Both changes only add lines.
- The preserve-active-model suite also passes on macOS under umask `022` and `002`.

**Scope note:** the mode-preservation suite still cannot pass on macOS. The code it tests is Linux-only and uses GNU `stat -c`, and `install-core.sh` guards on `uname -s == Linux`. That limitation is separate and deliberately left unchanged.

**Observation for maintainers (not changed here):** when `.env` *is* group-writable, `ods_preserve_existing_install_mode` silently falls back to the default mode. Refusing is the right security decision. But an owner whose editor rewrote `.env` under umask 002 would see `ODS_MODE` quietly ignored on the next installer rerun, so a warning there might be worth considering.

## Release Lane


roshangupta00750:

Issue #5300, opened 2026-09-15
fix(cli): fail noninteractive purge without confirmation


patil2001:
## Summary
Make noninteractive `ods purge` fail when deletion confirmation is not received.

## Why this is necessary
At EOF, the purge command treated missing confirmation as a cancellation but returned exit code 0. Automation could therefore interpret a command that deleted nothing as a successful purge, causing unsafe state drift in provisioning and cleanup workflows.

## Acceptance criteria
- Missing or mismatched confirmation preserves data and returns non-zero.
- Explicit confirmation continues to delete the disabled service data.
- Add a regression test for stdin EOF/noninteractive invocation.

patil2001:
Closing as duplicate of merged PR #3874. The remaining API purge false-success path is tracked separately.

PR #5301, opened 2026-09-15
fix(cli): fail noninteractive purge without confirmation

patil2001:
## Summary
Make noninteractive `ods purge` fail when deletion confirmation is not received.

## Why this is necessary
At stdin EOF, the command treated missing confirmation as cancellation but returned exit code 0. Automation could interpret a command that deleted nothing as a successful purge, leaving stale service data while reporting cleanup complete. That is especially dangerous in provisioning and teardown workflows because callers commonly branch only on the exit status.

## Changes
- Return non-zero when confirmation is missing or does not exactly match the service ID.
- Preserve the existing data-preservation behavior on cancellation.
- Add an end-to-end regression case invoking purge with stdin closed and asserting both failure status and retained data.
- The inline comment records the fail-closed invariant and automation impact.

## Testing
- `bash -n ods-cli` — passed.
- `bash tests/test-cli-user-extension-deps.sh` — attempted; the existing Git-Bash harness reaches a Windows interaction limitation before completion, so CI should run the full suite.
- `git diff --check` — passed.

## Overlap check
Live open/closed issue and PR searches found no direct implementation for noninteractive purge confirmation status.

## Related Issues
Closes #5300


patil2001:
Closing: parallel audit found the CLI EOF cancellation behavior is already covered by merged PR #3874. The remaining distinct defect is dashboard API false success after incomplete deletion; it will be handled separately.

Issue #5360, opened 2026-09-15
.env upserts corrupt the owner's last key when .env has no final newline (macOS installer rerun, ods-cli, mode-switch)


roshangupta00750:
## What happens

Suppose the owner's `.env` does not end with a newline. That is what VS Code and several other editors save by default when you add a key at the end. The next time ODS adds a missing key, the new assignment is glued onto the owner's last line. The owner's value is corrupted, and the new key is never set.

## Reproduction

This uses the real macOS env generator from current `public-beta` (`33155fb5`), on macOS:

1. Generate a `.env`.
2. Simulate an install from before `EMBEDDING_MODEL` existed, whose owner added `ANTHROPIC_API_KEY=sk-ant-owner-key` as the last line without a final newline.
3. Rerun the generator the way an installer upgrade does, preserving the existing `.env`.

```
after rerun:
  175:ANTHROPIC_API_KEY=sk-ant-owner-keyEMBEDDING_MODEL=BAAI/bge-base-en-v1.5
  validate-env: pass
```

The Anthropic key is now invalid, `EMBEDDING_MODEL` was not added, and `scripts/validate-env.sh` does not notice.

## Where

Every `.env` upsert helper that appends a missing key with `>>` has this problem:

- `installers/macos/lib/env-generator.sh`: `upsert_env_value`. The macOS installer calls it for every key it backfills on a rerun.
- `installers/macos/ods-macos.sh`: `upsert_env_value`, used by the macOS CLI.
- `ods-cli`: `_env_set`, used for example by `_ensure_hermes_dashboard_session_token` and by mode changes that add a key.
- `scripts/mode-switch.sh`: `env_set`.

`installers/phases/08-images.sh` already guards its own appends with a leading newline (`printf '\nLLAMA_SERVER_IMAGE=%s\n'`).

A related read-side issue: the macOS CLI's `read_ods_env` loop (`while IFS= read -r line`) skips a last line that has no newline. Every macOS `ods` command therefore ignores the owner's last key, even though Docker Compose reads it.

MB2510:
[XX] .env is missing required keys: WEBUI_SECRET, LITELLM_KEY, DASHBOARD_API_KEY
  >   This will cause docker compose to fail. The .env file may be corrupted.
  >   Try deleting C:\Users\M Kemal\ods\.env and re-running the installer.

how can i slv this? i deleted .env and starat again but no solve ... i add also try last newline solution but no workind....windows

PR #5361, opened 2026-09-15
fix(env): append .env keys on their own line when the file lacks a final newline

roshangupta00750:
Fixes #5360

## Problem

The `.env` upsert helpers append a missing key with `>>`. If the file's last line has no newline, as VS Code and similar editors save it by default, the new assignment is joined onto that line. `API_KEY=owner-secret` becomes `API_KEY=owner-secretNEW_KEY=value`, which corrupts the owner's value and never sets the new key.

A macOS installer rerun reproduces it with the real generator. Details are in #5360.

## Fix

- Before appending, each helper adds a newline when the file is non-empty and its last byte is not a newline. It checks the last byte with `tail -c 1`, which works on both BSD and GNU. Files that already end in a newline, and empty files, are written exactly as before. The helpers:
  - `installers/macos/lib/env-generator.sh`: `upsert_env_value`
  - `installers/macos/ods-macos.sh`: `upsert_env_value`
  - `ods-cli`: `_env_set`
  - `scripts/mode-switch.sh`: `env_set`
- `installers/macos/ods-macos.sh`'s `read_ods_env` loop now keeps an unterminated last line (`|| [[ -n "$line" ]]`), the same idiom #5071 used for memory-shepherd.

The Linux installer regenerates `.env` in phase 06, and phase 08 already prefixes its appends with a newline, so neither needs a change.

## Validation

**The real macOS env generator on macOS.** Owner line `ANTHROPIC_API_KEY=sk-ant-owner-key` with no final newline, then a preserve-mode rerun that backfills `EMBEDDING_MODEL`:

| | result |
|---|---|
| `public-beta` | `ANTHROPIC_API_KEY=sk-ant-owner-keyEMBEDDING_MODEL=BAAI/bge-base-en-v1.5` (validate-env still passes) |
| this branch | `ANTHROPIC_API_KEY=sk-ant-owner-key` then `EMBEDDING_MODEL=BAAI/bge-base-en-v1.5` on the next line |

**New `tests/test-env-append-final-newline.sh`.** It extracts each shipped helper from its file and checks three cases: an unterminated last line, a newline-terminated file (no blank line added), and an empty file (no leading blank line). Every case requires exact bytes. It also runs `read_ods_env` against an unterminated last line.

- On `public-beta` it fails 5 checks: all four writers produce `API_KEY=owner-secretNEW_KEY=new-value`, and the reader drops the key.
- On this branch it passes 13/13 under Homebrew Bash 5, stock macOS Bash 3.2, and Ubuntu 26.04 with GNU tools.
- It is wired into `make test` and the Linux CI job, next to the mode-switch tests.

**Unchanged suites.** On Ubuntu 26.04 these give the same results on both branches: `test-mode-switch-status.sh`, `test-macos-env-quote-strip.sh`, `contracts/test-installer-contracts.sh`, and `test-cli-status-docker-error.sh`.

**Static checks.**
- shellcheck finds no new findings in any touched file and no errors.
- `bash -n` passes under Bash 3.2.
- The workflow YAML parses and `git diff --check` is clean.
- The branch merges cleanly with `public-beta`, #4215, and my other open branches.

## Release Lane


roshangupta00750:
The library extension setup hooks (`extensions/library/services/*/setup.sh`, `append_if_missing`) have the same bare `>>` append. #5123 already fixes those, so this PR does not touch them. The two PRs change different files and merge cleanly together.

Issue #5375, opened 2026-09-15
ods rollback exits 1 without a message and restores nothing after ods update (looks in data/backups, which ods update never creates)


roshangupta00750:
## What happens

After a failed `ods update`, the command it tells you to run, `ods rollback`, prints its confirmation prompt and then exits 1 without any message. Nothing is restored.

The FAQ says: "`ods update` always creates a pre-update snapshot, so `ods rollback` is available immediately after any update attempt."

## Reproduction

Setup: Ubuntu 26.04 with `public-beta` `33155fb5`, using the real `ods-cli` and `ods-update.sh` in a throwaway install. Docker is stubbed so the image pull fails right after the pre-update snapshot is taken. The `.env` is then set to `MARKER=broken-after-update` to stand in for a broken config, so a restore would be visible.

```
## ods update --force
   [ods] Creating pre-update snapshot (pre-update-20260915-183812)...
   [OK] Backup created: /tmp/tmp.czWsWuGGGw/.ods/backups/backup-pre-update-20260915-183812-20260915-183812
   fixture pull failed
   ✗ Failed to pull latest images
   data/backups exists: no
## ods rollback (answer y)
   [ods] Rolling back to pre-update state...

   ⚠ This will restore configuration from before the last update.
   exit=1
   .env now: MARKER=broken-after-update
```

Running `ods-update.sh rollback` directly on macOS (Homebrew Bash 5.3) gives the same result.

## Cause

`ods rollback` runs `ods-update.sh rollback`. With no target, `cmd_rollback` does this:

```bash
backup_path=$(find "${ROLLBACK_DIR}" -maxdepth 1 -type d -name "pre-update-*" \
    2>/dev/null | sort -r | head -1)
```

`ROLLBACK_DIR` is `<install>/data/backups`, but a normal install never creates it:

- Only `snapshot_pre_update` creates that directory, and only the source-checkout `ods-update.sh update` path calls it.
- `ods update` takes its snapshot with `ods-update.sh backup pre-update-<ts>`, which writes to `~/.ods/backups` instead.
- No installer creates `data/backups`.

So the lookup fails and the script stops:

1. `find` fails on the missing directory, and `2>/dev/null` hides its message.
2. `pipefail` makes the command substitution fail.
3. `set -e` ends the script.

The same thing happens when no backups exist at all, so the "No backup or rollback snapshot found to restore from." message is never printed either.

## Second problem once that directory exists

[...]

PR #5376, opened 2026-09-15
fix(update): let ods rollback find the backup ods update just wrote

roshangupta00750:
Fixes #5375

## Problem

`ods update` saves its pre-update backup with `ods-update.sh backup pre-update-<ts>`, which writes to `~/.ods/backups`. `ods rollback` then runs `ods-update.sh rollback`.

With no target, that command first ran `find` on `<install>/data/backups`. Only the source-checkout `ods-update.sh update` path creates that directory, so on a normal install:

- `find` failed, `set -euo pipefail` ended the script, and it exited 1 with no message and nothing restored.
- With no backups at all, the same silent exit meant the "No backup or rollback snapshot found" message never printed.

When the directory did exist, general backups were ordered by their whole name. Names begin with their label, so an older `backup-working-config-<ts>` was restored instead of `backup-pre-update-<ts>-<ts>`.

## Fix

- New `_latest_backup_dir <root> <prefix>`. It picks the newest candidate by the `YYYYMMDD-HHMMSS` stamp each name ends with. A missing root just has no candidates.
- `cmd_rollback` now uses it for both lookups.

Unchanged:
- Rollback snapshots still take precedence over general backups.
- Symlinked directories are still skipped; `find -type d` never matched them.
- An explicit target (`ods rollback <id>`) uses the existing lookup.

One behaviour difference: a general backup directory whose name does not end in a stamp is no longer picked automatically. `ods-update.sh backup` always adds a stamp, and such a directory can still be restored by name.

## Validation

**End to end.** The real `ods-cli` and `ods-update.sh` on Ubuntu 26.04, with Docker stubbed so the image pull fails after the pre-update snapshot. The steps are `ods update`, then `ods rollback`:
- `public-beta`: prompt, exit 1, no message, `.env` unchanged.
- This branch: `Rolling back from: backup-pre-update-…`, `.env` restored, `Rollback complete!`, exit 0.

**New `tests/test-update-rollback-latest-backup.sh`.** It runs the real script with docker and curl stubbed, one scenario per row:

| scenario | `public-beta` | this branch |
|---|---|---|
| no `data/backups`, older labelled backup + newer pre-update backup | exit 1, nothing restored | newest restored |
| same backups, empty `data/backups` | restores the August labelled backup | newest restored |
| two rollback snapshots + a newer general backup | newest snapshot | newest snapshot |
| symlinked backup dir with a later stamp | skipped | skipped |
| no backups anywhere | exit 1, no message | "No backup or rollback snapshot found…", exit 1 |

- The test passes under Bash 5.3, stock macOS `/bin/bash` 3.2, and on Ubuntu 26.04.
- It is wired into `make test` and into the Linux CI job next to the Update Rollback Contract step.

**Neighbouring suites.** These give identical exit codes on `public-beta` and this branch on Ubuntu 26.04:
- `test-rollback-compose-stack.sh`
- `test-update-rollback-contract.sh`
- `test-ods-cli-rollback-delegation.sh`
- `test-update-script-backup.sh`
- `test-update-backup-retention-spaces.sh`
- `test-ods-update-runtime-compose.sh`
- `test-cli-legacy-update-backup.py`
- `test-ods-cli-update-verification.sh`
- `test-cli-update-verification.sh`

**Static checks.**
- shellcheck gives the same findings for `ods-update.sh` as on `public-beta`, and none for the new test.
- `bash -n` passes under Bash 3.2.
- The workflow YAML parses and `git diff --check` is clean.
- The branch merges cleanly with `public-beta` and with the 41 open PRs that touch `ods-update.sh`, the Makefile or `test-linux.yml`.

## Release Lane


roshangupta00750:
Reproduced on a real install, not a fixture: Ubuntu 26.04, `public-beta` `33155fb5`, installed with `./install.sh --non-interactive --tier 0 --no-voice --no-comfyui --no-workflows --no-hermes --no-pixel` (14 containers healthy).

```
$ ls -ld ~/ods/data/backups
ls: cannot access '/home/…/ods/data/backups': No such file or directory

$ ./ods-update.sh backup pre-update-20260916-072209      # what ods update runs
[OK] Backup created: /home/…/.ods/backups/backup-pre-update-20260916-072209-20260916-072209
[INFO] Files backed up: 16

$ printf '\n# runtime-marker\n' >> .env                   # stand in for a broken config
$ printf 'y\n' | ./ods-cli rollback
[ods] Rolling back to pre-update state...
⚠ This will restore configuration from before the last update.
exit=1
$ grep -c runtime-marker .env
1
```

A complete install never creates `data/backups`, so the lookup fails there and the command exits 1 without a message; the marker is still in `.env`, so nothing was restored.

Issue #5387, opened 2026-09-15
ods preset export writes the archive at the umask default, so other local users can read the .env secrets inside it


roshangupta00750:
## What happens

`ods preset export <name> <file>` writes a tar.gz that includes the preset's copy of `.env`. That copy holds `DASHBOARD_API_KEY`, the generated service passwords and any provider API keys.

The archive is created at the umask default: 0664 on Ubuntu (umask 002) and 0644 on macOS (umask 022). So any local user who can reach the file can read every secret in it.

`.env` itself is 0600. `ods-backup.sh` already restricts its compressed archive for this reason (`10e5f9bd`): "The archive bundles the raw .env (DASHBOARD_API_KEY, session secret, service passwords). Restrict it to the owner rather than leaving it world-readable at the umask default."

## Reproduction

Setup: Ubuntu 26.04 with `public-beta` `33155fb5`, running the real `ods-cli` against a throwaway install whose `.env` is 0600. The steps:

1. Run `ods preset save shared-setup`.
2. Run `ods preset export shared-setup <shared 0755 dir>/ODS.tar.gz`.
3. As a second local user, read `shared-setup/env` out of that archive.

```
.env mode=600
presets/shared-setup/env mode=600
exported archive mode=664
second local user: tar xzf ODS.tar.gz -O shared-setup/env
ANTHROPIC_API_KEY=sk-ant-owner-secret
```

On macOS the archive is 0644 and has the same contents.

## Where

In `ods-cli`, `cmd_preset` → `export` runs `tar czf "$output" "$name"` without restricting the result.

## Related, not part of the fix

`ods help` describes the export as a "shareable archive", and its example imports `shared.tar.gz`. The archive carries every credential in `.env`, including provider API keys. Sharing it therefore shares those credentials, and loading it on another install replaces that install's secrets. Whether export should warn about this or leave secrets out is a product decision, so the linked PR only fixes the file mode.

PR #5388, opened 2026-09-15
fix(cli): write preset export archives owner-only

roshangupta00750:
Fixes #5387

## Problem

`ods preset export` bundles the preset's copy of `.env`, which holds `DASHBOARD_API_KEY`, service passwords and provider API keys. The archive was created at the umask default, 0664 on Ubuntu and 0644 on macOS, so another local user could read those secrets.

`.env` is 0600, and `ods-backup.sh` already `chmod 600`s its compressed archive for the same reason.

## Fix

- The archive is created in a subshell with `umask 077`, so a new file is never readable by anyone else.
- It is then `chmod 600`'d, because tar keeps an existing file's mode when it overwrites that file.
- If the target filesystem cannot store the mode, the export still succeeds and prints a warning.

## Validation

**Second local user.** On Ubuntu 26.04, a second user read an export placed in a shared 0755 directory:
- `public-beta`: the archive is 664 and `ANTHROPIC_API_KEY=sk-ant-owner-secret` comes out of it.
- This branch: the archive is 600 and the read fails with `Permission denied`.

**New `tests/test-preset-export-permissions.sh`.** It runs the real CLI against a throwaway install and checks three things:
- A new archive is 600.
- Overwriting an existing 0644 file leaves it at 600.
- The archive still imports with identical preset files.

On `public-beta` it fails the first two checks (both files end up 644) and passes the import check. On this branch it passes all three, on macOS with Bash 5.3 and on Ubuntu 26.04. It is wired into `make test` and the Linux CI job.

**Export onto FAT.** On Ubuntu 26.04 I exported onto a real vfat loop mount. Both branches succeed, the archive is readable, and no warning is printed.

**Unchanged suites.** These give identical results on `public-beta` and this branch on Ubuntu 26.04:
- `test-cli-preset-restore-user-extensions.sh`
- `test-cli-prompts-noninteractive.sh`
- `test-ods-config-secret-mask.sh`
- `test-cli-install-dir-resolution.sh`
- `test-ods-cli-pipefail-tolerance.sh`
- `test-cli-status-docker-error.sh`

The unwired static `tests/test-preset-import-export.sh` reports the same single pre-existing failure on both branches (see #5103). The `tar czf` line stays inside its 20-line window.

**Static checks.**
- shellcheck with the repo's CI flags gives the same findings for `ods-cli` as `public-beta`, and none for the new test.
- The new test passes `bash -n`, the workflow YAML parses, and whitespace is clean.
- The branch merges cleanly with `public-beta` and with the 43 open PRs that touch `ods-cli`, the Makefile or `test-linux.yml`.

## Release Lane


roshangupta00750:
Reproduced on a real install: Ubuntu 26.04 (umask 002), `public-beta` `33155fb5`, full install with 14 healthy containers.

```
$ ./ods-cli preset save runtime-demo
$ ./ods-cli preset export runtime-demo /tmp/runtime-demo.tar.gz
$ stat -c '%a' /tmp/runtime-demo.tar.gz
664
$ tar xzf /tmp/runtime-demo.tar.gz -O runtime-demo/env | grep -cE '^(DASHBOARD_API_KEY|ODS_AGENT_KEY|LITELLM_KEY|N8N_PASS|SEARXNG_SECRET|WEBUI_SECRET)='
6
```

The archive the installer's own secrets land in is group and world readable, and carries six generated credentials from this install.

Issue #5473, opened 2026-09-16
make test fails: test-linux-host-agent-stop.sh asserts uninstall command lines that moved to lib/system-uninstall.sh


roshangupta00750:
## What happens

`make test` fails on `tests/test-linux-host-agent-stop.sh`:

```
[OK] systemd unit has bounded stop timeout
[FAIL] uninstall must bound systemctl disable --now for old/stuck host-agent services
```

This reproduces on a clean `public-beta` checkout (`3c186f3f`) on Ubuntu 26.04 and on macOS.

## Cause

The suite greps `ods-uninstall.sh` for two literal command lines:

- `timeout 20s sudo -n -- systemctl disable --now ods-host-agent.service`
- `systemctl kill -s SIGKILL ods-host-agent.service`

Neither exists any more, in two steps:

1. #3890 (mine) turned the first into a loop over every owned unit, so the literal service name became `"$_ods_system_unit"`. I should have updated this suite in that PR.
2. `ed8fcbf0` moved the whole system-scope cleanup out of `ods-uninstall.sh` into `lib/system-uninstall.sh`, where the bounded stop is now `run_sudo timeout 30s systemctl disable --now "$unit"` and a unit that will not stop retains the installation instead of being force-killed.

So the behaviour the suite cares about is still there; the assertions point at the old location and the old force-kill design.

## Scope

`tests/test-linux-host-agent-stop.sh` runs in `make test` (`Makefile:47`) and in no workflow, which is why CI stayed green. `tests/test_system_uninstall.py` covers the new helper's behaviour (stop and removal order, fail-closed, entrypoint delegation) but not the bounded timeout.

PR #5474, opened 2026-09-16
test(uninstall): follow the host-agent stop contract to its new home

roshangupta00750:
Fixes #5473

## Problem

`tests/test-linux-host-agent-stop.sh` asserts two literal command lines in `ods-uninstall.sh` that no longer exist: #3890 turned the bounded disable into a loop over owned units, and `ed8fcbf0` moved the whole system-scope cleanup into `lib/system-uninstall.sh`, replacing the force-kill with fail-closed retention. `make test` fails on this suite.

## Fix

Assert the properties that survived the move instead of the old command lines:

- `ods-uninstall.sh` delegates to `ods_uninstall_system_units` and refuses to continue when the helper is missing.
- `lib/system-uninstall.sh` bounds `systemctl disable --now` with a timeout and retains the installation when a unit will not stop.

The unit's `TimeoutStopSec` and the host agent's async-safe SIGTERM assertions are unchanged.

## Validation

**Before and after,** on a clean `public-beta` checkout (Ubuntu 26.04, and macOS):

| | result |
|---|---|
| `public-beta` | `rc=1`, `[FAIL] uninstall must bound systemctl disable --now …` |
| this branch | `rc=0`, four `[OK]` lines |

**The suite still has teeth.** Two injected regressions, each restored afterwards:

- Dropping the `timeout 30s` bound in `lib/system-uninstall.sh` → `[FAIL] system unit cleanup must bound systemctl disable --now for old/stuck units`.
- Removing the `ods_uninstall_system_units "$INSTALL_DIR"` call from `ods-uninstall.sh` → `[FAIL] uninstall must delegate system unit cleanup to lib/system-uninstall.sh`.

**Static checks.** shellcheck with the repo's CI flags is clean, `bash -n` passes under Bash 3.2, `git diff --check` is clean, and the branch merges cleanly with `public-beta` and my other open branches. No open PR touches this file.

## Release Lane


roshangupta00750:

Issue #5482, opened 2026-09-16
Disk-space preflight suite never reaches the preflight: its fixture install has no lib/


roshangupta00750:
## What happens

`tests/test-disk-space-preflight.sh` fails, and it has never exercised what it claims to test:

```
ℹ Backup should fail preflight when disk is low
✗ Expected 'Not enough disk space' message
```

The backup it runs does exit non-zero, so the suite's first check passes, but not because of the preflight. With the output printed:

```
---- backup rc=1 output ----
/home/…/ODS/ods/tests/../ods-backup.sh: line 32: /tmp/tmp.XXXX/ods/lib/rsync.sh: No such file or directory
---- end ----
```

## Cause

The suite builds a fake install (`$TMP/ods`) with `data/`, `.backups/` and `.version`, then runs the real `ods-backup.sh` and `ods-restore.sh` against it with a stubbed `df`. Both scripts source their helpers from the install they operate on:

```sh
. "$ODS_DIR/lib/rsync.sh"
. "$ODS_DIR/lib/backup-paths.sh"
```

The fixture has no `lib/`, so both die on that line, long before the disk-space checks. The suite then fails on the missing message rather than on the missing library, which hides the cause.

With the two libraries copied into the fixture, both preflights run and both pass:

```
✓ Backup preflight blocks low-space backup
✓ Restore preflight blocks low-space restore
```

so the product behaviour is correct and only the fixture was incomplete.

## Scope

The suite runs in no make target and no workflow, so nothing reported it. This is the same fixture gap as #3877, in a different suite.

PR #5483, opened 2026-09-16
test(backup): give the disk-space preflight suite the libs it needs

roshangupta00750:
Fixes #5482

## Problem

`tests/test-disk-space-preflight.sh` runs the real `ods-backup.sh` and `ods-restore.sh` against a fake install, but both source `lib/rsync.sh` and `lib/backup-paths.sh` from that install. The fixture has no `lib/`, so both scripts died at the source line and neither preflight was ever exercised. The suite reported the missing message instead of the missing library, and it runs in no make target or workflow, so nothing caught it.

## Fix

- Copy `lib/rsync.sh` and `lib/backup-paths.sh` into the fixture.
- Run the suite from `make test` and from the Linux CI job, so the preflight stays covered.

No product file changes.

## Validation

| | result |
|---|---|
| `public-beta` | `rc=1`, `✗ Expected 'Not enough disk space' message` |
| this branch | `rc=0`, `✓ Backup preflight blocks low-space backup`, `✓ Restore preflight blocks low-space restore` |

Verified on Ubuntu 26.04 and macOS.

**The suite has teeth now.** Replacing the backup preflight's `if [[ -n "$free" && "$free" -lt "$need" ]]` with `if false` makes it fail: `✗ Expected backup to fail due to low disk space`. Restored afterwards. On `public-beta` that same regression goes unnoticed, because the fixture never reaches the check.

**Static checks.** shellcheck with the repo's CI flags is clean, `bash -n` passes under Bash 3.2, the workflow YAML parses, `git diff --check` is clean, and the branch merges cleanly with `public-beta` and my other open branches. No open PR touches this suite.

## Release Lane


roshangupta00750:

Issue #5628, opened 2026-09-17
test-phase05-podman-shim.sh silently aborts: stale grep anchor after #4215 hardened the docker-found arm


roshangupta00750:
### Summary

`tests/test-phase05-podman-shim.sh` (part of `make test`) aborts immediately with exit code 1 and no output on current `public-beta`. Its static ordering guard greps `installers/phases/05-docker.sh` for a line that no longer exists, so under `set -euo pipefail` the command substitution fails and kills the test before any of its five assertions run.

### Root cause

#4215 hardened the docker-found arm to also probe the client:

```bash
elif command -v docker &> /dev/null && docker --version &> /dev/null; then
```

The test still anchors on the pre-#4215 text:

```bash
docker_line="$(grep -n 'elif command -v docker &> /dev/null; then' "$PHASE" | head -1 | cut -d: -f1)"
```

`grep` now matches nothing. The pipeline exits non-zero (grep's miss, plus SIGPIPE from `head -1`), and because the result is captured in a `$(...)` under `set -o pipefail`, the whole script aborts at that line. The `[[ -n "$docker_line" ... ]] || fail ...` guard on the next line, which was meant to produce a readable failure, never executes.

### Steps to Reproduce

```bash
cd ods
bash tests/test-phase05-podman-shim.sh; echo "rc=$?"
```

Observed: no output, `rc=1`.

### Expected Behavior

The suite runs its podman/docker install-path assertions, and if an anchor ever fails to match it reports the explicit guard message instead of dying silently.

### Fix

Anchor the docker arm on its `command -v docker` prefix (matches with or without the version probe) and tolerate a missing match on all three lookups so a real future refactor trips the ordering guard's diagnostic rather than a silent `pipefail` abort.

### Operating System

Linux (any; static test, no Docker required)

PR #5629, opened 2026-09-17
test(phase05): fix stale podman-shim grep anchor that silently aborts the suite

roshangupta00750:
## Problem

`tests/test-phase05-podman-shim.sh` runs as part of `make test`, but on current `public-beta` it aborts immediately with exit code 1 and **no output** — none of its five podman/docker install-path assertions run.

The static ordering guard greps `installers/phases/05-docker.sh` for the docker-found arm:

```bash
docker_line="$(grep -n 'elif command -v docker &> /dev/null; then' "$PHASE" | head -1 | cut -d: -f1)"
```

#4215 hardened that arm to also probe the client, so the line is now:

```bash
elif command -v docker &> /dev/null && docker --version &> /dev/null; then
```

The old literal no longer matches. The pipeline exits non-zero (grep's miss, plus `SIGPIPE` from `head -1`), and because it runs inside `$(...)` under `set -euo pipefail`, the whole test dies at that line. The very next line — `[[ -n "$docker_line" ... ]] || fail "..."`, designed to give a readable failure — never executes.

## Fix

- Anchor the docker arm on its `command -v docker` prefix instead of a fixed `; then` suffix, so it matches with or without the version probe.
- Add `|| true` to all three line lookups so a genuine future refactor produces the explicit ordering-guard diagnostic rather than a silent `pipefail` abort.

No production code changes; this only repairs the test's anchor and failure reporting.

## Validation

- **Before** (clean `public-beta`): `bash tests/test-phase05-podman-shim.sh` → no output, `rc=1`.
- **After**: all five assertions run and pass, `rc=0`.
- **Teeth**: mangling the docker-found arm now yields the intended `[FAIL] phase 05 must check for podman after docker and before the Docker CE install branch` message (rc=1), where before the fix it aborted silently.
- `bash -n` clean.

Closes #5628


roshangupta00750:

PR #5632, opened 2026-09-17
test(uninstall): pin fixture unit mode so custody check passes under umask 002

roshangupta00750:
## Problem

`tests/test_system_uninstall.py` is run in CI by `tests/test-uninstall-mdns-unit.sh`, which is wired into both `ods/Makefile` and `.github/workflows/test-linux.yml`. It fails 5 of its 11 tests on any host with **umask 002** (Ubuntu's default for private user groups), while GitHub's 022 runners stay green:

```
( umask 002; python3 tests/test_system_uninstall.py )   # FAILED (failures=2, errors=3)
( umask 022; python3 tests/test_system_uninstall.py )   # OK
```

Each failure reports `Refusing to remove ods-host-agent.service: unsafe systemd unit custody`.

## Root cause

`setUp` writes the fixture units with `Path.write_text()`, whose mode follows the runner's umask. Under 002 they land `0664` (group-writable). `lib/system-uninstall.sh`'s custody check correctly refuses group- or other-writable units (`info.st_mode & 0o022`), because a real systemd unit is root-owned `0644`. The fixture just isn't representative under a 002 umask.

## Fix

Pin the fixture unit files to `0644` after writing them, so the suite is independent of the runner's umask. No production code changes. Same class of umask-002 fixture breakage already fixed for two other CI suites in #5172.

## Validation

- **Before**, umask 002: `FAILED (failures=2, errors=3)`. umask 022: `OK`.
- **After**: `OK` (11/11) under **both** umask 002 and 022; the CI wrapper `tests/test-uninstall-mdns-unit.sh` exits 0.
- macOS is unchanged by this patch (both before and after it reports the same pre-existing `failures=2, errors=3` from the custody check's parent-symlink guard, because macOS `/var` is a symlink — a Linux-oriented suite running off-target, out of scope here).

Closes 5631


roshangupta00750:

PR #5636, opened 2026-09-17
test(preset): scope cmd_preset assertions to the case body, not a fixed grep window

roshangupta00750:
## Problem

`tests/test-preset-import-export.sh` reports two false failures on current `public-beta`:

```
✗ Export does not use tar
✗ Export may create absolute paths
```

Both are wrong. The assertions grep a fixed number of lines after the case label:

```bash
grep -A20 "export|e)" "$ODS_CLI" | grep -q "tar czf"         # Test 6
grep -A15 "export|e)" "$ODS_CLI" | grep -q "cd.*PRESETS_DIR"  # Test 10
```

The `export|e)` case body has grown (archive staging, `umask`, `chmod`). `cd "$PRESETS_DIR"` now sits 19 lines after the label (beyond `-A15`) and `tar czf` sits 28 lines after (beyond `-A20`), so both needles fall outside the window even though the export path is correct. Tests 4–11 all use this fragile pattern.

## Fix

Add an `ods_cli_case_body` helper that prints exactly the matched case (from `<label>)` to its terminating `;;`), and route every case-body assertion through it. The checks now track the real block regardless of length, and cannot leak into an adjacent case if a body shrinks. Test-only change.

## Validation

- **Before**: `Passed: 10, Failed: 2` (macOS and Linux).
- **After**: `Passed: 12`, `✓ All tests passed` on both macOS and the Linux VM.
- **Teeth A**: renaming `tar czf` in the export case → Test 6 fails again.
- **Teeth B (scoping)**: removing `cd "$PRESETS_DIR"` from the *export* case → Test 10 fails, even though the *import* case still contains `cd "$PRESETS_DIR"` — proving the helper is scoped to the right case and does not match a neighbour.
- `bash -n` clean.

Closes 5635


roshangupta00750:

PR #5638, opened 2026-09-17
test(uninstall): match the centralized user-manager helper, not old inline env

roshangupta00750:
## Problem

`tests/test-bootstrap-upgrade-close-inherited-fds.sh` fails on current `public-beta`:

```
[FAIL] uninstaller: model-upgrade cleanup must reach the owner user manager without login-session variables
```

The suite verifies the uninstaller stops the transient `ods-model-upgrade.service` through the owner's user systemd manager without relying on inherited login-session variables. It grepped `ods-uninstall.sh` for two inline literals:

```bash
grep -q 'DBUS_SESSION_BUS_ADDRESS="unix:path=$_ods_uninstall_runtime_dir/bus"' "$uninstaller"
grep -q 'systemctl --user stop ods-model-upgrade.service' "$uninstaller"
```

#4215 refactored that logic into a single `ods_uninstall_systemctl_user()` helper that rebuilds `XDG_RUNTIME_DIR` and the DBUS session bus from the owner UID and runs `systemctl --user "$@"`; the cleanup now calls `ods_uninstall_systemctl_user stop ods-model-upgrade.service`. Both old literals are gone, so the first assertion fails and the suite exits 1 — even though the behavior is intact and now centralized.

## Fix

Point the assertions at the helper: verify it reconstructs both the owner runtime dir and the bus address, and that the cleanup stops the service through it. Use `grep -F` — the literals contain `${...}`/`$var` (not regex), and unescaped braces parse differently under BSD vs GNU grep. Test-only change.

## Validation

- **Before**: `[FAIL] ... without login-session variables`, rc=1 (macOS and Linux VM).
- **After**: all three PASS lines, rc=0 on both macOS and the Linux VM.
- **Teeth A**: breaking the bus reconstruction in `ods-uninstall.sh` → assertion 1 fails again.
- **Teeth B**: replacing `ods_uninstall_systemctl_user stop ods-model-upgrade.service` → assertion 2 fails.
- `bash -n` clean.

Closes 5637


roshangupta00750:

PR #5640, opened 2026-09-17
test(cli): report Hermes absent so restart-path fixtures skip the readiness wait

roshangupta00750:
## Problem

Two CLI restart-path suites hang ~180s and then fail on current `public-beta`:

- `tests/test-cli-bootstrap-compose-wait.sh` → `FAIL ods restart failed`
- `tests/test-ods-cli-bootstrap-env-refresh.sh` → `✗ Hermes agent did not become ready within 180s (last state: missing).`

#4215 added a Hermes readiness wait to `ods restart` (`_ods_cli_wait_for_hermes_ready`). It probes `docker inspect ods-hermes` and, **only when that succeeds**, polls container health for up to 180s. Hermes is optional — the wait is skipped when the probe fails:

```bash
docker inspect ods-hermes >/dev/null 2>&1 || return 0
```

Both suites' docker stubs fall through to `exit 0` for every subcommand, so `docker inspect ods-hermes` "succeeds" even though these fixtures never create an `ods-hermes` container. The readiness poll then reads an empty state, never sees `healthy`, and burns the full 180s before `ods restart` exits non-zero.

## Fix

- Give each stub an `inspect` handler that exits non-zero (reports the container absent), so the readiness wait is skipped as designed. The restart assertions these suites actually cover (bootstrap compose wait window, model-env reload) are unchanged.
- Bring `test-cli-bootstrap-compose-wait.sh` in line with the sibling env-refresh suite: re-exec under a modern Bash on macOS (stock 3.2 cannot run ods-cli) and invoke the CLI via `"$BASH"`, so it no longer fails `ods-cli requires Bash 4.0+`.

Test-only change.

## Validation

- **Before**: compose-wait fails after ~180s; env-refresh fails. (Linux VM)
- **After**: both pass in ~1–3s.
  - Linux VM: `test-cli-bootstrap-compose-wait` rc=0 in 3s, `test-ods-cli-bootstrap-env-refresh` rc=0 in 1s.
  - macOS: both rc=0 (compose-wait now re-execs under a modern Bash).
- **Teeth**: flipping the stub's `inspect` handler back to `exit 0` reintroduces `Hermes agent did not become ready ... FAIL ods restart failed`.
- `bash -n` clean.

Closes 5639


roshangupta00750:

Issue #5648, opened 2026-09-17
AMD topology JSON fails to build under decimal-comma locales (VRAM awk not LC_ALL=C)


roshangupta00750:
### Summary

`detect_amd_topo()` fails to build the AMD GPU topology JSON on any host whose parent shell uses a decimal-comma locale (de_DE, fr_FR, es_ES, pt_BR, it_IT, ru_RU, ...).

### Where

`ods/installers/lib/amd-topo.sh` renders each GPU's VRAM with a bare `awk`:

```sh
vram_gb=$(awk -v bytes="$vram_bytes" 'BEGIN { printf "%.1f", bytes / 1073741824 }')
```

`awk`'s `%.1f` honours `LC_NUMERIC`, so under a decimal-comma locale it prints `24,0`. That value becomes field 3 of the per-GPU TSV, whose jq conversion is:

```
memory_gb: (.[2] | tonumber)
```

and `tonumber` cannot parse `24,0`:

```
jq: error (at <stdin>:1): Expected value before ',' at line 1, column 3 (while parsing '24,0')
```

so the whole topology JSON fails to build.

### Reproduce

```
$ vram_gb=$(LC_ALL=de_DE.UTF-8 awk -v b=25769803776 'BEGIN{printf "%.1f", b/1073741824}'); echo "$vram_gb"
24,0
$ printf '0\tCard\t%s\n' "$vram_gb" | jq -Rn '[inputs|split("\t")|{memory_gb:(.[2]|tonumber)}]'
jq: error (at <stdin>:1): Expected value before ',' ... (while parsing '24,0')
```

With `LC_ALL=C awk` the value is `24.0` and jq parses it.

### Fix

Scope the VRAM `awk` to `LC_ALL=C`, exactly as the CPU-budget formatting already does (`tests/test-locale-safe-cpu-formatting.sh`, issue #1662). Dot-decimal locales are unaffected.

### Operating System

Linux (AMD GPU host) with a decimal-comma locale.

PR #5760, opened 2026-09-18
fix(install): pin awk to the C locale in AMD topology VRAM rendering

tang-vu:
## Why this matters

`detect_amd_topo()` silently produced no GPU inventory on any host running a decimal-comma locale (de_DE, fr_FR, es_ES, pt_BR, it_IT, ru_RU, ...). Its VRAM render used a bare `awk`:

```sh
vram_gb=$(awk -v bytes="$vram_bytes" 'BEGIN { printf "%.1f", bytes / 1073741824 }')
```

`%.1f` honours `LC_NUMERIC`, so under those locales it emits `24,0`. That string lands in field 3 of the per-GPU TSV, which is converted with `memory_gb: (.[2] | tonumber)` � `tonumber` cannot parse a comma decimal, the whole `gpus_json` conversion fails, and the detected topology comes back empty. Every downstream consumer (tier mapping, multi-GPU layout) then sees no AMD GPUs on affected hosts. Fixes #5648.

## What changed

`installers/lib/amd-topo.sh`: pin `LC_ALL=C` on that single awk invocation � the same `LC_ALL=C awk` convention already used in `installers/macos/lib/env-generator.sh` and `installers/macos/lib/preflight-fs.sh`. One-line fix; no other site in the file formats a float into JSON.

## Regression test

New BATS case in `tests/bats-tests/amd-topo.bats` builds the existing fake-sysfs fixture (`ODS_DRM_SYS`) and runs real `detect_amd_topo` behind an `awk` shim that reproduces decimal-comma behavior deterministically: any awk call that did **not** pin `LC_ALL=C`/`LC_NUMERIC` gets its `.` rewritten to `,` � exactly what gawk emits under de_DE. A call under `LC_ALL=C` passes through untouched, so the test asserts the pinned-locale contract at the JSON boundary (`.gpus[0].memory_gb` must equal `16.0`), not the implementation text.

- **Red on base**: `gpus_json` conversion fails ? assertion fails.
- **Green on branch**: `LC_ALL=C` prefix passes through the shim ? `16.0`.
- Skips nothing and needs no locale packages � works on minimal CI containers.

## Overlap check

- Open PRs touching `amd-topo.sh`: searched `amd-topo`, `topology`, `LC_ALL`, `locale` � none cover the LC_NUMERIC/`tonumber` path.
- Sibling `awk "%.Nf"` sites audited: `ods-cli:5913` formats a human-readable string only (cosmetic, out of scope); TSV?jq sites at lines 236/290/328 consume `amd-smi`/`rocm-smi` JSON output, which is already locale-independent.

## Validation

- `bats tests/bats-tests/amd-topo.bats`: 30/30 pass on branch; new case **fails on base**.
- `bash -n`, `git diff --check`: clean. ShellCheck: no new warnings.
- Not verified: a real decimal-comma glibc locale end-to-end (none installed in this environment; the shim reproduces the exact mechanism � LC_NUMERIC-sensitive `%.1f` � at the function boundary).


tang-vu:
Closing as redundant with #5649, which was opened earlier and covers the same fix.

Comparison: identical root-cause fix (`LC_ALL=C` on the VRAM `%.1f` awk). #5649's regression coverage is equivalent-or-stronger — same fake-sysfs + decimal-comma-awk simulation, plus a static `LC_ALL=C` contract guard and `make test` wiring. My BATS variant asserts the same `memory_gb` JSON boundary. No independent advantage to keep this open.

PR #5763, opened 2026-09-18
fix(install): use defined ai_bad helper in 07-devtools and gate CI on the env smoke suite

tang-vu:
﻿## Summary

Two OpenCode guards in `installers/phases/07-devtools.sh` call `ai_err`, a function that is **not defined** in the Linux installer — the shared fatal helper is `ai_bad` (`ai_err` only exists in the macOS lib). Under `set -u` the call aborts the phase with an unbound-command error instead of the intended diagnostic.

`tests/smoke/installer-env-smoke.sh` already detects this defect, but the suite is wired into no Makefile target and no workflow — it has been silently dead. This PR fixes the call sites and gates CI on the suite that caught them:

- `07-devtools.sh`: `ai_err` → `ai_bad` at both OpenCode guard sites (the existing `exit 1` after each call already handles termination, matching the `ai_bad` convention used by other phases).
- `Makefile`: add `installer-env-smoke.sh` to `make smoke`.
- `matrix-smoke.yml`: run the suite as its own step in `linux-smoke`.

## Why this matters

On public-beta the installer aborts with `ai_err: command not found` (or an unbound-variable error under `set -u`) when the OpenCode switchboard key check trips — the operator never sees the real message. The env smoke suite covers .env generation, schema validation, CPU-cap, secret-reuse and function-resolution contracts that currently have no CI signal.

## Overlap check

The `ai_err` defect has three open PRs (#2682, #2745, #2812) — all target `main`, none target `public-beta`, and the suite wiring is out of their scope. No open public-beta PR touches `ai_err`, `07-devtools.sh`, or `installer-env-smoke.sh`.

## Test plan

- [x] `bash tests/smoke/installer-env-smoke.sh` on base — **14/16, fails on the undefined `ai_err` check**
- [x] same suite on this branch — **16/16**
- [x] `bash -n installers/phases/07-devtools.sh` clean
- [ ] CI


tang-vu:
CI note: the `frontend (windows-latest)` failure is the known flaky `src/pages/Pixel.test.jsx` case (`keeps visible history beyond the model request limit after send and reload` — `Unable to find an element with the text: Latest answer`). This PR touches only `07-devtools.sh`, `installer-env-smoke.sh`, `Makefile`, and `matrix-smoke.yml` — nothing in the dashboard frontend. The same test passed on all other dashboard runs today.

PR #5839, opened 2026-09-19
feat(dashboard): service impact explorer

tang-vu:
## Summary
- New `ServiceImpact` panel above the Integrations topology map: pick a service and see every dependent service that would be affected if it went down, including indirect paths (A → B → C).
- Computes reverse reachability over the same `nodes`/`edges` graph data already fetched for the map — no extra requests.
- Labeled explicitly as a planning aid, not live failure diagnosis.

#### Test plan
- [x] `npx vitest run src/pages/ServiceMap.impact.test.jsx` — 1 passed
- [x] `npx vitest run src/pages/ServiceMap.test.jsx` — 5 passed


tang-vu:
Closing this redundant Service Impact PR after comparing the full diffs. #3844 retains the same component and page integration, and also tests compact mode plus selection reset after a refreshed service snapshot. This later PR adds no independent behavior or stronger coverage. Keeping #3844 as the canonical review target.

PR #5840, opened 2026-09-19
fix(installer): AMD topology JSON under decimal-comma locales

tang-vu:
## Summary
- `detect_amd_topo` formats each GPU's VRAM GiB with `awk 'BEGIN{printf \"%.1f\"}'`, which honors `LC_NUMERIC`. Under decimal-comma locales (de_DE, fr_FR, es_ES, …) it emits `24,0`, and the jq `tonumber` projection fails — so the entire topology JSON aborts.
- Prefix the call with `LC_ALL=C` and wire the existing `test-amd-topo.sh` suite into `make test` (it was orphaned).
- Fixes #5648.

## Why this matters
`config/gpu-topology.json` feeds model-tier selection on AMD hosts; a parent shell in a decimal-comma locale previously made topology detection fail outright.

## Overlap check
Searched open/closed PRs for `amd-topo`, `LC_ALL`, `LC_NUMERIC`, `decimal`, locale. No existing PR covers this call site.

#### Test plan
- [x] `bash tests/test-amd-topo.sh` — 24 passed, including a new case that stubs `awk` to emit `24,0` unless `LC_ALL=C` is set (reproduces the failure deterministically without needing the locale installed; verified the stub alone breaks jq `tonumber` pre-fix)


tang-vu:
Closing as redundant with #5649 (opened earlier, identical root-cause fix: `LC_ALL=C` on the VRAM `%.1f` awk + Makefile wiring). Its dedicated locale test file is equivalent-or-stronger than the case added here to test-amd-topo.sh. Supporting the earlier PR instead of keeping a variant open.

PR #5935, opened 2026-09-19
test(gpu): keep host nvidia-smi out of the apple stub PATH, wire suite into make test

tang-vu:
## Why this matters

`tests/test-gpu-apple.sh` stubs `sysctl`, `system_profiler`, `curl`, and `docker` and deliberately leaves `nvidia-smi` unstubbed so the negative cases exercise the *"nvidia-smi not found"* fallthrough. But `STUB_PATH="$STUB_BIN:$PATH"` inherits the host PATH wholesale — on any machine with a real NVIDIA driver (e.g. WSL2 with GPU paravirtualization, `/usr/lib/wsl/lib/nvidia-smi`), `command -v nvidia-smi` succeeds and `ods gpu status` queries the **physical GPU** instead of taking the absent-driver path:

```
✗ FAIL nvidia backend fallthrough warning missing; output: ━━━ GPU Status (1 GPU) ━━━
  0    NVIDIA GeForce GTX 106 1.4 / 3.0 GB ...
```

## Fix

Build `STUB_PATH` by copying the caller's PATH minus every directory that contains an executable `nvidia-smi`. Stubs still win; core tools remain available; the negative contract — *nvidia-smi absent from PATH* — now holds on developer machines with NVIDIA hardware, not just driverless CI runners.

Also wires the suite into `make test` — it was unreachable from any target despite being self-contained and gracefully skipping when `env bash` is Bash 3.2 (stock macOS).

## Regression coverage

The failing assertion is itself the regression check: on a host with `nvidia-smi` on PATH, case 6 produces real GPU output before the fix and the expected *"nvidia-smi not found / GPU status unavailable"* warning after.

## Overlap check

Searched open PRs for `test-gpu-apple`, `nvidia-smi PATH`, `PATH hermeticity`, and Makefile wiring of this suite — no open PR touches `tests/test-gpu-apple.sh`. Neighbor suites (`test-phase03-multigpu-tty`, `test-multigpu-nvidia-visibility`) are independent.

## Validation

- WSL (has real `/usr/lib/wsl/lib/nvidia-smi`): **20 passed, 1 failed → 21 passed, 0 failed**
- `bash -n` clean; `shellcheck` clean
- Fails on the old PATH construction exactly as reported


tang-vu:
Closing this stacked overlap. #5934 already carries the same .gitattributes change and regression, while #2204 covers the GPU test PATH issue more completely (NVIDIA and ROCm tools, Bash availability, and explicit assertions). This PR combines both independent changes and adds no distinct production behavior. Keeping the two focused, earlier PRs for review.

PR #6007, opened 2026-09-19
fix(installer): write HOST_LAN_IP for Windows LAN installs

0xacee:
## Summary

- Windows LAN installs (`-Lan` → `BIND_ADDRESS=0.0.0.0`) never wrote `HOST_LAN_IP` to `.env`, while the Linux phase-06 and macOS generators both populate it. The schema itself documents it: "written by the installer when BIND_ADDRESS=0.0.0.0".
- OpenClaw consumes `HOST_LAN_IP` to add `http(s)://<lan-ip>:<port>` to `gateway.controlUi.allowedOrigins` (`ods/config/openclaw/inject-token.js`, passed through by `extensions/services/openclaw/compose.yaml`). With the key missing, LAN clients browsing a Windows host's Control UI get origin-rejected even though the port is reachable — a Windows-only LAN regression.
- Adds `Get-WindowsHostLanIp` (route-based `Find-NetRoute` first — the equivalent of `ip route get` — then first non-loopback IPv4 via `Get-NetIPAddress`), emits `HOST_LAN_IP=` next to `BIND_ADDRESS`, and preserves an operator-set `.env` value on rerun via `Get-EnvOrNew`, matching `_env_get` semantics.

#### Test plan

- [x] `ods/tests/test-windows-host-lan-ip.ps1` — 7 assertions pass on Windows PowerShell 5.1: LAN emits key, loopback emits empty value, operator override preserved on rerun
- [x] Same test against the pre-change generator fails 5/7 (`HOST_LAN_IP` never emitted)
- [x] `PSParser::Tokenize` parse-check clean; PSScriptAnalyzer shows only pre-existing warnings


0xacee:
Duplicate of #2188, which already covers HOST_LAN_IP on Windows plus ODS_DEVICE_NAME, adds APIPA filtering and PrefixOrigin gating, and wires its contract test into lint-powershell.yml. Closing in favor of the earlier, broader implementation.

PR #6008, opened 2026-09-19
fix(installer): derive ODS_DEVICE_NAME from COMPUTERNAME on Windows

0xacee:
## Summary

- `ODS_DEVICE_NAME` is how `ods-mdns` publishes `<name>.local`, how `ods-proxy` routes `auth.`/`chat.`/`dashboard.` subdomains, and how `dashboard-api` builds magic-link invite URLs (`extensions/services/dashboard-api`, `docker-compose.base.yml:246`).
- Linux (`installers/phases/06-directories.sh`) derives it from the sanitized system hostname; macOS (`installers/macos/lib/env-generator.sh`) derives it from LocalHostName/hostname and even backfills it for older installs. The Windows generator never emits the key, so `docker-compose.base.yml`'s `${ODS_DEVICE_NAME:-ods}` default kicks in — every Windows install shares the same `auth.ods.local` identity and invite QR codes point at a generic name.
- Adds `Get-WindowsDeviceName` (sanitizes `$env:COMPUTERNAME` with the same rules as the Linux `tr`/`sed` pipeline — lowercase, collapse invalid runs to `-`, trim edges, cap at 32 chars, schema-pattern check, `"ods"` fallback) and emits `ODS_DEVICE_NAME=` via `Get-EnvOrNew` so an operator's manual value survives reruns, matching `_env_get` semantics.

#### Test plan

- [x] `ods/tests/test-windows-device-name.ps1` — 10 assertions pass on Windows PowerShell 5.1: sanitizer parity (incl. inner `--` runs preserved like the sed pipeline), fresh-install emission matching the schema pattern, operator override preserved on rerun
- [x] Same test against the pre-change generator dies immediately (`Get-WindowsDeviceName` missing, key never emitted)
- [x] `PSParser::Tokenize` parse-check clean


0xacee:
Duplicate of #2188, which already writes ODS_DEVICE_NAME with an equivalent schema-safe sanitizer (and HOST_LAN_IP) plus CI-wired tests. Closing in favor of the earlier, broader implementation.

PR #6035, opened 2026-09-20
fix(macos): make the doctor-failure stderr indent BSD sed-compatible

roshangupta00750:
## Problem

`installers/macos.sh`'s doctor-invocation block indents the first five lines of the doctor's stderr with:

```sh
sed -n '1,5{s/^/  /;p}' "$doctor_err_file"
```

**BSD sed** (macOS's `/usr/bin/sed`) rejects a command placed immediately before the closing `}`:

```
sed: 1: "1,5{s/^/  /;p}": extra characters at the end of p command
```

This block runs **only on macOS**, and only on the failure path — whenever `ods-doctor.sh` exits non-zero or leaves no report behind. So instead of surfacing the doctor's stderr, the "honest failure" reporting added in `f2024d2e` emits a raw sed usage error — defeating the very thing it was meant to fix.

*Correction to an earlier version of this description:* I originally attributed the failure path to the doctor exiting early under stock macOS Bash 3.2. That is no longer true on current `public-beta` — I re-checked, and `ods-doctor.sh` run under `/bin/bash` 3.2.57 exits 0 and writes a full 356-line report with empty stderr. The sed defect is unaffected by that: the branch is still reached on any doctor failure, which is exactly what the regression suite stubs, and the suite fails on unmodified `public-beta` with the error above.

## Fix

Terminate the `p` command before the brace (`;p;}`). GNU sed accepts this form identically, so Linux behavior is unchanged.

```diff
-        sed -n '1,5{s/^/  /;p}' "$doctor_err_file"
+        sed -n '1,5{s/^/  /;p;}' "$doctor_err_file"
```

## Validation

`tests/test-macos-doctor-honest-failure.sh` (already wired into `make test`) exercises this exact block:

| | without fix | with fix |
|---|---|---|
| macOS (BSD sed, `/usr/bin/sed`) | **FAIL** — `sed: extra characters at the end of p command` | PASS (3/3) |
| Linux (GNU sed) | PASS | PASS (3/3) |

- **Teeth**: reverting to `;p}` reintroduces the macOS failure.
- shellcheck clean (no new findings).
- Sole occurrence: this is the only `sed` `{…cmd}` block missing a terminator in `installers/`, `scripts/`, `lib/`, or `ods-cli`.


roshangupta00750:

PR #6036, opened 2026-09-20
test(secret): stop the hardcoded-secret scan from flagging node_modules/dist

roshangupta00750:
## Problem

`tests/test-secret-security.sh` scans for hardcoded secrets with:

```sh
find extensions/services -name "*.js" -print0   # and the same for *.py
```

This descends into `extensions/services/dashboard/node_modules` and `dist`. Of the **7140** `*.js` files it matches, **7046 (99%)** are third-party dependencies or Vite build output.

Once the dashboard is built (`npm install` / `npm run build`), the scan reports dependency files as leaks:

```
✗ Potential hardcoded secret in cli-api.C6CiCDM3.js   (vitest bundle)
✗ Potential hardcoded secret in nwsapi.js             (jsdom dependency)
✗ Potential hardcoded secret in config.js             (vite / eslint / testing-library)
```

The suite's own summary line says "source files," and secret-scanning third-party dependencies is meaningless noise (they aren't our secrets, and it's 7000 files of it).

## Fix

Prune `node_modules` and `dist` from every `extensions/services` scan, using the BSD/GNU-portable idiom so the suite behaves identically on Linux and macOS:

```sh
find extensions/services -type d \( -name node_modules -o -name dist \) -prune -o -type f -name "*.js" -print0
```

Applied to all 10 `find` scans in the suite (1 × `*.js`, 9 × `*.py`).

## Validation

On a built checkout (dashboard `node_modules`/`dist` present):

| | before | after |
|---|---|---|
| result | **37 passed / 3 failed** | 38 passed / 0 failed |
| runtime | >120 s (times out) | ~29 s |
| `*.js` scanned | 7140 | 94 |

- Prune idiom verified on both **BSD find** (macOS) and **GNU find** (Linux) — returns identical source-only file sets.
- shellcheck: no new findings.
- CI's `make test` environment has no dashboard `node_modules`, so its result is unchanged; this only removes false failures (and a big slowdown) for anyone running the suite on a built checkout.


roshangupta00750:
Closing this as redundant.

public-beta \`1d6038ef9\` now prunes \`node_modules\`, \`.venv\` and \`venv\` inline at the two scan sites that call \`fail\` (lines 85 and 96), which was this PR's headline case. The only gap left after that is \`dist\`, and I verified it still reproduces on the new base — dropping a built \`extensions/services/dashboard/dist/assets/index-abc.js\` containing \`api_key = "abcdefghij1234567890"\` still fails section 1/6 with `✗ Potential hardcoded secret in index-abc.js`.

That residue is already covered, and covered better, by #5780: it factors a single \`find_service_sources()\` helper and applies it at all ten \`find extensions/services\` sites, pruning \`node_modules\`, \`dist\`, \`.venv\`, \`venv\` and \`__pycache__\`. This PR only ever pruned two of those names, and #5780 was filed first. Deferring to it rather than keeping a narrower duplicate in the queue.

Issue #5297, opened 2026-09-15
fix(catalog): fail on invalid extension manifests


patil2001:
## Summary
Fail catalog generation when an extension manifest is invalid instead of silently omitting it.

## Why this is necessary
The catalog generator warned and continued when a manifest could not be parsed, had the wrong schema, or lacked a valid service entry. Installers and dashboards then received a successful but incomplete catalog, hiding broken extension definitions until a user attempted to use them.

## Acceptance criteria
- Invalid manifests are reported as a catalog-generation failure.
- The failure identifies the affected manifest path.
- Valid catalogs retain their existing ordering and contents.
- Add a regression test covering an invalid manifest.

PR #6114, opened 2026-09-20
fix: fail catalog generation on invalid manifests

patil2001:
## Summary\n- fail catalog generation instead of silently dropping invalid extension manifests\n- include the offending manifest path in the error\n- add regression coverage for invalid and sorted valid manifests\n\nFixes #5297\n\n## Validation\n- python -m py_compile ods/scripts/generate-extensions-catalog.py ods/tests/test-generate-extensions-catalog.py\n- git diff --check\n- pytest unavailable in this Windows environment


patil2001:
Closing this branch-collision PR; the catalog fix is being re-opened from an isolated branch.

PR #6032, opened 2026-09-20
Harden fleet model switching and Pixel recovery

Lightheartdevs:
## Summary

This PR hardens the public-beta install and model-switching path across Linux, Windows/WSL, and macOS:

- recover interrupted Pixel model transactions only when the exact native transaction, target, host artifacts, and live inference proof all agree
- preserve a held Pixel proof across managed configuration reload without reopening normal inference, tools, or provider admission
- remove Docker's host-UID-wide `nproc` ceiling from the Pixel sandbox while retaining its independent per-container `pidsLimit=1024`
- let `ods restart` wait through bounded Hermes convergence and perform one recreate only after an explicit terminal unhealthy result
- apply compact Pixel prompt budgets to named Mini/Micro/Tiny models without overriding explicit parameter counts
- make model-compatibility evidence host/backend scoped and time bounded
- reduce Kokoro worker memory on native macOS and small Docker guests while preserving valid existing overrides
- support named Colima profiles, detect root-hidden macOS port listeners, use BSD-compatible permission commands, and give native llama-server one LaunchAgent custody path
- keep the full tier disk floor for new installs while giving an already-running system a separate 10GB safety reserve; `ods doctor` warns below the install recommendation and remains red below the runtime reserve
- classify explicit generic-external inference as externally owned so its required LiteLLM route is not mislabeled as an unexpected managed-local route

## Safety review

- Base is exact current `public-beta` `4d725ee95f6235329e20ae81ec5418fd1c8fc3bf`; candidate is `f90b6623bcebd238989a738929ed9a3a85c1e408`.
- Current public-beta's cinematic ODSGATE installer commit was merged into the candidate without conflicts. Its plain/GUI/non-interactive presentation gates and installer hardening remain intact.
- Transaction recovery never replays an ambiguous model mutation. Ownership, target, artifact, environment, live-route, and stable-digest proofs remain fail closed.
- Native macOS cleanup is scoped to the exact installed llama binary or an install-owned model path, and shutdown is confirmed before replacement.
- Compatibility exclusions are host/backend scoped and expire unless revalidated.
- The doctor correction does not lower fresh-install disk requirements. It distinguishes install capacity from post-install operating headroom, retaining a 10GB red safety floor.
- No persistent-data migration, credential format change, public API removal, or broad destructive cleanup is introduced.

## Verification

- exact-head focused Python packet: 482 passed
- exact-head Pixel installer packet: 309 passed, 0 failed
[...]

PR #6162, opened 2026-09-21
test(macos): read catalog modes with a portable stat

roshangupta00750:
This PR originally also fixed the GNU-only `chmod --` in `get-ods.sh`. **#6032 landed that fix first** — with a better approach than mine, making `install_dir` absolute rather than prefixing `./` — so I have dropped the product change and rebuilt this on current `public-beta`. What remains is the part #6032 did not cover.

## Problem

`tests/test-bootstrap-pixel-catalog-permissions.sh` reads permission bits with GNU `stat -c`. BSD stat, which macOS ships, has no such option:

```
stat: illegal option -- c
usage: stat [-FLnq] [-f format | -l | -r | -s | -x] [-t timefmt] [file ...]
```

It dies on the first assertion, so **none** of the suite runs on macOS — even though nothing in it is platform-specific. It extracts `secure_pixel_catalog_sources`, creates files under `umask 0002`, and checks the resulting mode bits.

## Change

The same `file_mode` helper the other permission suites already use (`tests/test-preset-export-permissions.sh`), which tries `stat -c` and falls back to `stat -f`:

```sh
file_mode() { stat -c '%a' "$1" 2>/dev/null || stat -f '%Lp' "$1"; }
```

The suite was registered in neither `test-linux.yml` nor the `Makefile`, so it ran nowhere at all. Both now invoke it — it reads only local fixture files, with no host state, network or ports, so it is safe on a shared runner.

## Validation

| | result |
|---|---|
| macOS 15 (BSD stat), with the fix | **rc=0**, suite passes |
| macOS 15, fix reverted | **rc=1** — `stat: illegal option -- c` on the first assertion |
| Linux (GNU stat), with the fix | rc=0 — unchanged |
| `shellcheck -S style` | 1 finding before, 1 after — pre-existing, none introduced |

## One correction worth recording

My original description claimed the `--` was "guarding against an install directory whose name begins with `-` being read as an option". **That was wrong**, and I checked it properly this time:

```
$ chmod -R go-w -dash/f     # leading-dash path AFTER the mode operand
$ echo $?
0                            # mode applied; the operand `go-w` ends option parsing

$ chmod -dash/f              # same path BEFORE any mode
chmod: illegal option -- d
```

A leading-dash path only breaks when it precedes the mode, which cannot happen here. So the `--` was never protecting anything, and #6032 was right to simply drop it. I had written a regression test for that "property" and removed it again once I confirmed it could never fail — a test that cannot fail is worse than no test.


roshangupta00750:

PR #6163, opened 2026-09-21
fix(macos): use a POSIX whitespace class in sed expressions

roshangupta00750:
`\s` is a GNU extension. BSD sed, which macOS ships, has no such class — it reads `\s` as a literal `s`, so a pattern meant to match optional whitespace matches nothing and still exits 0. The failure is silent, which is worse than an error:

```
$ printf "check_port_conflict() {\n  body\n}\n" > fn.sh

$ sed -n "/^check_port_conflict\s*()\s*{/,/^}/p" fn.sh     # BSD
$ echo $?
0                                                           # no output, no error

$ sed -n "/^check_port_conflict[[:space:]]*()[[:space:]]*{/,/^}/p" fn.sh
check_port_conflict() {
  body
}
```

## Where it bites

**Two suites** source installer functions through an address built on it:

```sh
source <(sed -n '/^check_port_conflict\s*()\s*{/,/^}/p' ...)
```

On macOS the address never matches, nothing is sourced, and the suite dies before its first assertion:

```
tests/test-port-ollama-detection.sh: line 54: check_port_conflict: command not found
```

**`ods-cli`** strips the key from compose image lines the same way:

```sh
| grep -E '^\s+image:' | sed 's/.*image:\s*//' | sort -u
```

Read literally, `image:` is followed by zero `s` characters, so the original indentation survives:

| | result |
|---|---|
| GNU | `[redis:7]` |
| BSD | `[   redis:7]` |

so the image list prints with stray leading whitespace. I want to be straight about this one: the values are only `echo`ed, never executed, so it is **display-only** — but the expression is wrong on the platform regardless, and it is the same root cause.

I left the `grep -E '\s'` on that same line alone. BSD grep does support the class; only the sed side is broken, and I would rather not churn working code.

## Validation

| | result |
|---|---|
| macOS, `test-port-ollama-detection`, before | fails outright — `check_port_conflict: command not found` |
| macOS, `test-port-ollama-detection`, after | **14/14 passed** |
| Linux, both suites, after | pass — the change is a no-op under GNU sed |

`test-wsl-windows-port-detection` gets past extraction with this change and passes on Linux. It still fails on my Mac, but for an unrelated reason I confirmed is environmental: it asserts nothing is listening on 9100, and an unrelated `dart` process holds `127.0.0.1:9100` here. Not a code defect, and not something this PR tries to address.

## On registering these suites

Neither is referenced by `test-linux.yml` or the `Makefile`, so they run nowhere today. I considered adding them and decided against it: both query real host ports, which would make them flaky on a shared runner. Fixing the portability bug is worth doing on its own; making CI flaky is not.


roshangupta00750:

PR #6164, opened 2026-09-21
test(tmpfile): scope the phase 05 error-path check to the right function

roshangupta00750:
Test 8 of `tests/test-tmpfile-cleanup.sh` asserts that phase 05 removes its downloaded Docker installer when the install fails. It anchored on the wrong text:

```sh
grep -B 2 'error "Docker installation failed' installers/phases/05-docker.sh \
    | grep -q "rm -f.*tmpfile"
```

Those messages belong to the distro install helpers (lines 171, 178, 200) and never had a temp file near them. The temp file lives in `_docker_install_from_script` about 120 lines earlier, and is already removed on **both** paths:

```sh
_docker_install_from_script() {
    local tmpfile
    tmpfile=$(mktemp /tmp/install-docker.XXXXXX.sh)
    if ! curl -fsSL --max-time 300 https://get.docker.com -o "$tmpfile" || ! sh "$tmpfile"; then
        rm -f "$tmpfile"      # failure path
        return 1
    fi
    rm -f "$tmpfile"          # success path
}
```

So the check reported a failure regardless of what the cleanup did — and, more to the point, **it would have kept reporting the same failure if the cleanup were deleted.** It could not tell the two apart, which is the only thing it existed to do.

## Change

Scoped to the function that owns the temp file, and stated as the property rather than as line proximity — every early return inside it must be preceded by the removal, and both a removal and a return must be present:

```sh
sed -n '/^_docker_install_from_script() {/,/^}/p' "$ROOT_DIR/installers/phases/05-docker.sh" \
    | awk '
        /rm -f.*tmpfile/        { cleaned = 1 }
        /return 1/              { returned = 1; if (!cleaned) leaked = 1 }
        END { exit (leaked || !cleaned || !returned) ? 1 : 0 }
    '
```

Note the sibling check (test 7) was already correct — it anchors on `mktemp /tmp/install-docker`, which is inside the right function. Only test 8 was wrong, and it is left otherwise unchanged.

## Verified it now discriminates

| phase 05 state | result |
|---|---|
| unchanged | 9/9 passed |
| error-path `rm -f` deleted | **8 passed, 1 failed** |
| both `rm -f` deleted | **7 passed, 2 failed** (test 7 catches the second) |

Same outcome on Linux: 9/9 with the phase unchanged, 8/1 with the error-path removal deleted.

## Registration

The suite ran in neither `test-linux.yml` nor the `Makefile`, which is why a check that could never pass went unnoticed. It reads phase sources only — no host state, no network, no ports — so it is safe to run on a shared runner, and is now registered in both, in its own section rather than inside the documented-security-suites block.


roshangupta00750:

PR #6165, opened 2026-09-21
test(uninstall): apply the compose-flags timeout bound only where timeout exists

roshangupta00750:
`tests/test-uninstall-compose-flags.sh` bounds its non-interactive uninstall run with `timeout 5s`. `timeout` is GNU coreutils and macOS ships neither it nor `gtimeout`, so the command is never found, the uninstaller never runs at all, and the sudo log stays empty:

```
[FAIL] non-interactive uninstall must validate sudo without prompting
```

The rc check just above it does not catch this:

```sh
[[ "$noninteractive_rc" -ne 0 && "$noninteractive_rc" -ne 124 ]] \
    || fail "non-interactive uninstall must fail promptly ..."
```

127 is neither 0 nor 124, so "fails promptly" passes on a command that was never executed, and the suite only trips two assertions later on what the uninstaller was supposed to have done. The failure therefore points at sudo validation rather than at the missing binary.

## Change

Apply the bound only where the binary exists — the same `command -v timeout` guard already used by `installers/lib/docker-images.sh` and `scripts/bootstrap-upgrade.sh`:

```sh
local -a noninteractive_cmd=(bash "$install_noninteractive/ods-uninstall.sh"
    --force --non-interactive)
if command -v timeout >/dev/null 2>&1; then
    noninteractive_cmd=(timeout 5s "${noninteractive_cmd[@]}")
fi
```

CI runs on Linux, so the bound still applies there and nothing about the guarantee changes. Where `timeout` is absent the run is unbounded, which is still a valid observation: the surrounding assertions distinguish a prompt failure from the 124 timeout exit, and this particular path fails on a sudo preflight by design rather than hanging.

## Validation

| | result |
|---|---|
| macOS 15, with the guard | **8 passed, rc=0** |
| macOS 15, guard reverted | **rc=1** — `[FAIL] non-interactive uninstall must validate sudo without prompting` |
| Linux, with the guard | 8 passed, rc=0 — `timeout` present, so the bounded path is taken |
| `shellcheck -S style` | 1 finding before, 1 after — pre-existing, none introduced |

The suite is registered in `test-linux.yml` and passes there either way, which is why this platform gap was invisible to CI.

Scope note: three other suites use an unguarded `timeout` — `test-linux-host-agent-stop.sh`, `test-phase03-multigpu-tty.sh` and `test-bootstrap-upgrade-hotswap-contract.sh`. The first two exercise Linux-only paths (systemd, the Linux installer phases) and the third is already covered by #5924, so this change is limited to the one suite where the gap is both reachable and unclaimed.


roshangupta00750:
Closing rather than extending. On Linux, where CI runs, `test-uninstall-compose-flags.sh` already passes on current `public-beta`, so this change only ever helped macOS. Since #6459 the uninstaller has a Darwin-only native Pixel retirement step, and this suite's fixture doesn't provide `installers/macos/lib/pixel-native-uninstall.py`, so on macOS it now stops before reaching the timeout bound this PR adjusted. Making it runnable there would mean stubbing that retirement step, which is a separate change. Not worth a PR on its own.

PR #6166, opened 2026-09-21
test(macos): give sed -i a suffix so the in-place edits run on BSD sed

roshangupta00750:
Two suites edit a fixture in place with a bare `sed -i`. GNU sed treats the next argument as the script. BSD sed, which macOS ships, **requires** an argument to `-i` and takes whatever follows as the backup suffix — so the script is consumed as the suffix and the next argument is read as a file name:

```
tests/test-offline-embedding-gate.sh
  sed: 1: "/var/folders/.../run-phase.sh": invalid command code f

tests/test-gpu-reassign-manual.sh
  sed: -e: No such file or directory
```

The edit never happens and both suites die there. Both pass on Linux, so CI never showed it.

## Change

`-i.bak` is accepted by both implementations, so write a suffix and drop the backup immediately after. This is the form `test-validate-env.sh` and `test-hosted-bootstrap-verifier.sh` already use in this repo:

```sh
sed -i.bak "s#__ROOT__#$ROOT_DIR#g; s#__INSTALL__#$tmp/install#g" "$tmp/run-phase.sh"
rm -f "$tmp/run-phase.sh.bak"
```

Nothing else about either suite is platform-specific — this one incompatibility was the entire difference, and with it fixed both pass on macOS unchanged otherwise.

## Validation

| | result |
|---|---|
| macOS 15, with the fix | both suites **rc=0** |
| macOS 15, fix reverted | both **rc=1**, with the two sed errors above |
| Linux (GNU sed), with the fix | both **rc=0** — unchanged |
| stray `.bak` files after either run | none, on either platform |
| `shellcheck -S style` | 0 findings before, 0 after, on both files |

## Scope

I deliberately did not touch registration. `test-gpu-reassign-manual.sh` already runs via `test-nvidia-mig.yml`. `test-offline-embedding-gate.sh` runs nowhere today, but adding it would edit the same `Makefile` and `test-linux.yml` anchors as the open #5757, and this change does not need to — keeping it to the portability fix means it merges cleanly in any order.

I also checked the rest of the tree for this class rather than fixing only what I tripped over. The remaining `sed -i` call sites are already correct: `installers/macos/` passes `-i ''`, `scripts/bootstrap-upgrade.sh` uses `-i.bak`, `test-openclaw-inject-token.sh` branches on platform explicitly, and `extensions/services/dashboard/entrypoint.sh` runs inside the container, where sed is GNU.


roshangupta00750:

